PR #217, opened 2011-05-20
[PATCH] msg_t label flags.

ninoles:
This implement the label proposal, without removing the delimiter.  It's also assert/check if the label is missing, breaking backward compatibility.  I also add a check_flags method to msg_t, making the code slightly more readable, especially when checking for multiple flags.

The code is tested through the automated test, which include a new test_queue_tcp.

### DIFF
--- a/src/dist.cpp
+++ b/src/dist.cpp
@@ -105,7 +105,7 @@ void zmq::dist_t::activated (writer_t *pipe_)
 int zmq::dist_t::send (msg_t *msg_, int flags_)
 {
     //  Is this end of a multipart message?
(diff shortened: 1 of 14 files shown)


bluca:
@ninoles could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@ninoles  another ping. Please consider helping us with the relicense. Thank you!

PR #273, opened 2012-03-14
Add WinCE support.

BoresExpress:
- Added two new files: errno.hpp and errno.cpp. They are required to use errno functionality on WM.
- zmq.cpp, msg.h: removed inclusion of errno.h because it is included in zmq.h that is also included by .cpp.
- windows.hpp: process.h is included only for desktop builds.
- thread.cpp: on CE CreateThread is used instead of __beginthreadex
- socket_base.cpp, clock.cpp: on CE include cmnintrin.h instead on intrin.h
- signaler.cpp: on Windows should use special macro around event name (for unicode builds)
- err.hpp: make it include errno.hpp (my file) instead on errno.h when building for CE
- err.cpp: use FormatMessage when building for CE (because CE does not have ANSI API functions)
- zmq.h: do not include errno.h whe building for CE
- libzmq.vcproj: add tro new files

### DIFF
--- a/builds/msvc/errno.cpp
+++ b/builds/msvc/errno.cpp
@@ -1,0 +1,32 @@
+#if defined WINCE
+
+//#include "..\..\include\zmq.h"
+#include "..\..\src\err.hpp"
+
+int errno;
+int _doserrno;
+int _sys_nerr;
+
+char* error_desc_buff = NULL;
+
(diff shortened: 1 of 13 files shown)


ZhouBox:
support wce6?

bluca:
If it doesn't work, please send a pull request to fix it

PR #291, opened 2012-03-25
Fix for issue #307

rcxdude:
dist_t distribute() was skipping over pipes when one failed because the non-working pipe got swapped with a working pipe but the write was never retried on that pipe. This fix retries the write with the same index so the newly swapped pipe is also written to. This should stop apparently dropped PUB/XPUB messages due to closed SUB sockets.


bluca:
@rcxdude could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

PR #456, opened 2012-11-01
Resolve LIBZMQ-464

hurtonm:


ForrestSu:
I reviewed the history issue #448 about _VERBOSE, via git blame.
>Do not filter out duplicate subscriptions on the XSUB side of
XSUB/XPUB, so that ZMQ_XPUB_VERBOSE doesn't get blocked by forwarding
devices (as long as the devices all use ZMQ_XPUB_VERBOSE)

Consider for forwarding devices (eg: sockets proxy).

_So, maybe  unsubscription has the same problem?_

PR #550, opened 2013-04-25
Experimental function zmq_recviov doesnt work correctly in a couple of c...

JonDyte:
...ases

1) VSM - you cannot hand out the 'data' address as it was not allocated on the heap
2) for other messages the 'data' address cannot be handed out either, as it not the address
originally returned by malloc and hence cannot be passed to 'free'.
see msg.cpp
u.lmsg.content = (content_t*) malloc (sizeof (content_t) + size_);
....
u.lmsg.content->data = u.lmsg.content + 1;

So the function is changed to always malloc a data buffer and copy the data into it.
There is a possible optimisation using memmove for the non-VSM case but that is not done yet.

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -478,10 +478,6 @@ int zmq_recv (void *s_, void *buf_, size_t len_, int flags_)
 // The iov_base* buffers of each iovec *a_ filled in by this 
 // function may be freed using free().
 //
-// Implementation note: We assume zmq::msg_t buffer allocated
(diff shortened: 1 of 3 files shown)


bluca:
@JonDyte could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

PR #738, opened 2013-11-06
Fix race condition on shutdown

ricnewton:


tj0120:
Terminating on signal 15
Bad file descriptor (bundled/zeromq/src/epoll.cpp:78)

PR #785, opened 2013-12-11
support signaler_port in config.hpp set to value other than 0 and 5905

pijyoi:
the use case is as follows:
Developer wants to use a different fixed signaler port for different applications.

This patch allows the developer to specify that port in config.hpp. The mutex object name is generated based on the port used, so potentially different groups of applications could use their own port in order to not interfere with one another.

For backwards compatibility, if the fixed port is 5905, then the old implementation is used. (Event and constant object name)

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -69,6 +69,7 @@
 
 #if defined ZMQ_HAVE_WINDOWS
 #include "windows.hpp"
+#include <tchar.h>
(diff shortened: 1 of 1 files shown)


terryrankine:
any chance this could be set at runtime somehow - not a compiled option?
port 5905 may interfere with other things on a uses system (like vnc).

PR #950, opened 2014-03-25
Make a NuGet package

phreed:
See the readme.nuget file, which is part of this commit, for description of nuget.
When I try to add this package to the nuget gallery it fails due to the presence of a package with that name.  The owner is JGOZ.  Also, I am not sure the proper attribution so any pointers would be appreciated.  I put may identifier as the owner and author, I know that is wrong, what is correct?
The build also does not include all pivots, only those that I needed for my purposes.  For a fully configured zmq what would all the pivots be and where would they be found?
Thanks

### DIFF
--- a/builds/nuget/libzmq.autopkg
+++ b/builds/nuget/libzmq.autopkg
@@ -1,10 +1,10 @@
 nuget{
   nuspec{
     id = libzmq;
-    version : 4.1.0;
+    version : 4.1.0-alpha;
     title: ZMQ - Code Connected;
-    authors: {phreed};
-    owners: {phreed};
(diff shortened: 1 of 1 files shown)


bluca:
@phreed could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@phreed  another ping. Please consider helping us with the relicense. Thank you!

PR #1080, opened 2014-06-05
Corrected installer source path for VS pdb debug symbols file.  Removed ...

tberkey:
Based upon the current CMake files, platform.hpp gets created in the binary directory, not the source directory.  This leads to issues when specifying relative paths for the inclusion of headers, which are based upon the source directory.

The Microsoft pdb (debug symbols file) gets created in the bin directory rather than lib.  This leads to a failure at installation time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -766,7 +766,7 @@ if(MSVC)
             ARCHIVE DESTINATION lib
             PUBLIC_HEADER DESTINATION include
             COMPONENT SDK)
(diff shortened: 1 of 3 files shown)


bluca:
@tberkey could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@tberkey another ping. Please consider helping us with the relicense. Thank you!

PR #1248, opened 2014-11-06
Problem: default LINGER value is insane

### LINKED ISSUE #1247: Problem: default LINGER value (block forever) is insane
Solution: use a sane default, e.g. 2 seconds

hintjens:
Solution: use a sane value, e.g. 2 seconds

Fixes #1247


bluca:
This change needs a big fat note on the release notes, as technically it's an API breakage :-)

PR #1281, opened 2014-12-23
Problem: need atomic reference counting in several projects

hintjens:
Solution: as libzmq already provides this across all platforms,
expose an atomic counter API. I've not wrapped atomic pointers,
though someone who needs this may want to do so.


bkmit:
Could you please merge this pool request to zeromq4-1 branch.
This will allow to compile malamute project with it.

hintjens:
@bkmit We do not backport improvements, only bug fixes, to stable releases. This will be available in the 4.2 stable release.

PR #1284, opened 2014-12-31
Problem: incomplete definition of integer types

hintjens:
Firstly, only a few types are defined, leaving it hard for higher layers
to complete the set. Secondly, the code incorrectly tries to use ifndef
to avoid re-defining typedefs, which does not work in C.

Solution: define all types, unconditionally for all MSVC compilers.
Additionally, define ZMQ_DEFINED_STDINT that tells higher layers that we
already defined these integer types.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -71,18 +71,18 @@ extern "C" {
 #endif
 
 /*  Define integer types needed for event interface                          */
+#define ZMQ_DEFINED_STDINT 1
 #if defined ZMQ_HAVE_SOLARIS || defined ZMQ_HAVE_OPENVMS
 #   include <inttypes.h>
-#elif defined _MSC_VER && _MSC_VER < 1600
-#   ifndef int32_t
+#elif defined _MSC_VER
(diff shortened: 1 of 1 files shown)


ricnewton:
Not sure what we are trying to do here but it breaks visual studio builds (see https://libzmq.cloudapp.net/job/libzmq/567/console).

Visual Studios has a stdint.h and has for a while, why are we not just using that?

hintjens:
Trying to get consistency between libzmq and czmq. When stdint.h is
available we should be using it. Since libzmq was/is defining some but not
all int types, I think that's where we should be pulling in stdint.h (not
in CZMQ).

Seems my change wasn't very good. I don't have MSVC to play with; someone
who does could improve /revert the change to zmq.h, perhaps?

ricnewton:
Well just removing the _MSC_VER section works for me, I'll raise a pull request.

I was just wondering why the typedef were introduced in the first place, I'm wondering if it had something to do with MinGW or something.

Issue #1295, opened 2015-01-09
zmq_msg_t is now 48 instead of 32 chars


metadings:
A breaking change to 4.1 is also that the zmq_msg_t has a size of 48 bytes, instead of former 32 bytes.

hintjens:
How does this break code? The previous size wasn't documented anywhere, was
it?

metadings:
It broke my code with SIGSEGV in Marshal.FreeHGlobal...

metadings:
Hello Pieter! I made a completely new clrzmq. Would you like to reference it?

     http://github.com/metadings/clrzmq/
     http://github.com/metadings/clrzmq-test/

xaqq:
@hintjens This probably breaks is user-code wasn't recompiled I suppose.

@metadings Can you confirm that you recompiled your application using the up-to-date `zmq.h` header file (as opposed to just updating the `.so`)

metadings:
Yes, I compiled it against libsodium 1.0.1, libpgm 5.2.122 and libzmq 4.2.0 for x86 and amd64.

hintjens:
We can bring this into the zeromq organization on github. See
http://zeromq.org/docs:organization

Otherwise, you can certainly add it to the .Net bindings page.

metadings:
WOW that is really cool. Sorry I have to tell you :)

rodgert:
I've been meaning to ask this for a while, this seems as good a segue as
any -

Is there any reason to make sizeof(zmq_msg_t) anything smaller than the
size of the target platform's cache line size (e.g. 64 bytes on x86), you
are already paying the fetch cost for that number of bytes whether you use
all of them or not.

Increasing it would have at least two benefits -

* More messages would be 'VSM' messages, so it would reduce allocation
overhead a bit.
* Remove any chance of false sharing
<http://en.wikipedia.org/wiki/False_sharing> of things that are, by design,
pushed by value onto a ypipe_t<msg_t> which is shared between two threads.

The only downside I see is slightly increased memory consumption on memory
constrained applications.

hintjens:
Seems like a good suggestion.

rodgert:
I'll add it to my growing ZeroMQ/Azmq list of things to do :)

hintjens:
:-)

xaqq:
> The only downside I see is slightly increased memory consumption on memory constrained applications.

Would it be possible to make `sizeof(zmq_msg_t)` customizable to work around this issue?
[...]

PR #1298, opened 2015-01-10
Increase size of zmq_msg_t to 64 bytes

rodgert:
Increasing it would have at least two benefits -

* More messages would be 'VSM' messages, so it would reduce allocation
overhead a bit.
* Remove any chance of false sharing of things that are, by design (except
on IBM Power, which has 128 byte cache line), pushed by value onto a
ypipe_t<msg_t> which is shared between two threads.

The only downside I see is slightly increased memory consumption on memory
constrained applications.

- Full discussion of this rationale is part of issue #1295

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -195,7 +195,7 @@ ZMQ_EXPORT int zmq_ctx_destroy (void *context);
 /*  0MQ message definition.                                                   */
 /******************************************************************************/
 
-typedef struct zmq_msg_t {unsigned char _ [48];} zmq_msg_t;
+typedef struct zmq_msg_t {unsigned char _ [64];} zmq_msg_t;
 
 typedef void (zmq_free_fn) (void *data, void *hint);
 
--- a/src/msg.hpp
+++ b/src/msg.hpp
@@ -92,7 +92,7 @@ namespace zmq
 
         //  Size in bytes of the largest message that is still copied around
         //  rather than being reference-counted.
-        enum { msg_t_size = 48 };
+        enum { msg_t_size = 64 };
         enum { max_vsm_size = msg_t_size - (8 + sizeof (metadata_t *) + 3) };
 
         //  Shared message buffer. Message data are either allocated in one


metadings:
Won't you do this change to zeromq4-1 too, to avoid multiple changes since v3?

hintjens:
I can backport this to 4.1, sure. However not to 4.0.x, since it's not
technically a bug fix.

metadings:
As I remember did 4.0 have 32 bytes, as of v4.1 there were 48 bytes and now v4.2 is going to have 64 bytes!

rodgert:
4.1 had to increase size to accommodate a reference to connection metadata
via zmq_msg_gets IIRC (though we don't seem to document any user visible
properties yet).

A related point - It may require an additional change in the ypipe_t
implementation detail to eliminate the potential for false sharing. I am
not ready to make that change yet and it won't be API visible in any event.

metadings:
Can you @rodgert  please change the value in v4.1, you are actually 'in' the library... It's going to be the first change of zmq_msg_t since v3...

rodgert:
Agree, lets back port to 4.1rc. You want to do it or should I?

It would also break the ABI if back ported to 4.0.x

metadings:
No, don't port it back to v4.0, I believe it's good as a first change in v4.1, then v4.2 and so on...

Please can you make the changes? I didn't fork zeromq4-x, zeromq4-1 and libzmq... Hey do you also change the `ToolsVersion` attribute in the files 

    ~/builds/msvc/vs2010/libzmq/libzmq.vcxproj
    ~/builds/msvc/vs2010/inproc_lat/inproc_lat.vcxproj
    ~/builds/msvc/vs2010/inproc_thr/inproc_thr.vcxproj
    ~/builds/msvc/vs2010/local_lat/local_lat.vcxproj
    ~/builds/msvc/vs2010/local_thr/local_thr.vcxproj
    ~/builds/msvc/vs2010/remote_thr/remote_thr.vcxproj
    ~/builds/msvc/vs2010/remote_thr/remote_thr.vcxproj

to `ToolsVersion="4.0"` because that's issuing an error in the build action...

hintjens:
I've backported it to 4.1.x.

PR #1300, opened 2015-01-13
Fix. Use C++03 compatible way to align struct. (Build on MSVC)

moteus:
But is it correct way to align?
`offsetof(zmq::command_t, args) + sizeof(args_t)` is not same as `(sizeof(zmq::object_t*) + sizeof(args_t))`
And still `unused` array also could be aligned.
Also you forget to add `sizeof(type_t)`
On my WinXP and MSVC 2010 `sizeof(zmq::command_t)` with my patch is 72.
May be you need use `pragma pack` or `align` directive?

To make `sizeof(zmq::command_t)==64` on my MSVC I make

``` C
#pragma pack(push, 1)
    struct command_t
    {
        …
        enum { pad_size = 64 - (sizeof(zmq::object_t*) + sizeof(args_t) + sizeof(type_t)) };
        unsigned char unused[ pad_size ];

    };
#pragma pack(pop)
}
static_assert(sizeof(zmq::command_t) == 64, "Invalid align");
```
But I do not know how do this crosspaltrom.

### DIFF
--- a/src/command.hpp
+++ b/src/command.hpp
@@ -59,7 +59,8 @@ namespace zmq
             done
         } type;
 
-        union {
+        union args_t
+        {
 
             //  Sent to I/O thread to let it know that it should
             //  terminate itself.
@@ -147,7 +148,7 @@ namespace zmq
 
         } args;
 
-        enum { pad_size = 64 - (sizeof(destination) + sizeof(args)) };
+        enum { pad_size = 64 - (sizeof(zmq::object_t*) + sizeof(args_t) + sizeof(type_t)) };
         unsigned char unused[ pad_size ];
 
     };


rodgert:
It isn't, I was trying to avoid compiler specific attributes because I
don't have a way of testing with MSVC. I have submitted pull request #1301
which I *think* does the right thing on MSVC (again no way to confirm) and
uses the appropriate __attribute__() for GCC/Clang.

rodgert:
Note, there is still a missing bit to make this work on Windows under MSVC, which doesn't have posix_memalign. yqueue.hpp:119 needs an additional case to call _aligned_malloc. I have no way to confirm the change, so I did not include it.

PR #1310, opened 2015-01-16
Fix typo in zmq_msg_gets documentation

rodgert:


xaqq:
Since this is about type: "auththenticated" has a typo in it too !

rodgert:
Ok I will fix that as well

metadings:
Didn't know about the documentation here... Is it now getnameinfo() or getpeername()?

PR #1314, opened 2015-01-20
Close messages that failed to send

topher200:
pipe_t.write only takes control of the underlying message memory when it
succeeds. When it returns failure, we must close the message ourselves to
clean up that memory.

This patch is sponsored by FarSounder, Inc (farsounder.com)

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -246,12 +246,16 @@ int zmq::router_t::xsend (msg_t *msg_)
         }
 
         bool ok = current_out->write (msg_);
-        if (unlikely (!ok))
+        if (unlikely (!ok)) {
+            // Message failed to send - we must close it ourselves.
+            int rc = msg_->close ();
+            errno_assert (rc == 0);
             current_out = NULL;
-        else
-        if (!more_out) {
-            current_out->flush ();
-            current_out = NULL;
+        } else {
+          if (!more_out) {
+              current_out->flush ();
+              current_out = NULL;
+          }
         }
     }
     else {


hintjens:
If you'd like this to go into 4.1.x, would you like to make the same
patch/pull request against https://github.com/zeromq/zeromq4-1?

Thanks

topher200:
While my fix solves our memory leak issue, I'm concerned about other similar parts of the code. For example, stream.cpp::xsend's code looks almost the same as router.cpp::xsend. Would it suffer from a similar memory leak? Should it get the same fix?

Should msg::close be called after all calls to pipe::write (even successful ones)? It looks like xsub::send_subscription uses the same paradigm as I discovered with my patch (call msg::close only after failed pipe::write). However router::xattach_pipe and dealer::xattach_pipe take the path of msg::close even on pipe::write success. 

Can we kick this problem back up the stack? Should pipe::write call msg::close if it is returning false (I think no)? How about this then - should msg::init call "close" on itself before resetting its metadata? That would mitigate this issue and others like it.

I'm hesitant to make these changes myself because I have only experienced the one memory leak and I'm not super familiar with the libzmq internals. I'd love to get guidance on the best way forward. I think at a minimum stream::xsend should get a similar patch.

hintjens:
You can probably best ask these questions on the zeromq-dev list.

topher200:
Thanks for the advice! Link to question on mailing list here: http://lists.zeromq.org/pipermail/zeromq-dev/2015-January/028169.html

Issue #1144, opened 2014-07-28
[4.0.4] Assertion failed: Successful WSASTARTUP not yet performed (..\..\..\src\signaler.cpp:134)


nexcvon:
Win32, VS2010.

If a global zmq::context_t instance is defined *in a dll*, this assertion will fail when exiting:

	Assertion failed: Successful WSASTARTUP not yet performed (..\..\..\src\signaler.cpp: 134)

call stack:

	libzmq-v100-mt-gd-4_0_4.dll!zmq::zmq_abort(const char * errmsg_)  line 71 + 0x15 bytes	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::signaler_t::send()  line 134 + 0x55 bytes	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::mailbox_t::send(const zmq::command_t & cmd_)  line 56	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::ctx_t::send_command(unsigned int tid_, const zmq::command_t & command_)  line 327	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::object_t::send_stop()  line 183	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::socket_base_t::stop()  line 162	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq::ctx_t::terminate()  line 134 + 0x16 bytes	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq_ctx_term(void * ctx_)  line 171 + 0x8 bytes	C++
 	libzmq-v100-mt-gd-4_0_4.dll!zmq_ctx_destroy(void * ctx_)  line 241 + 0x9 bytes	C++
 	ZeroMQ.dll!zmq::context_t::close()  line 304 + 0xe bytes	C++
 	ZeroMQ.dll!zmq::context_t::~context_t()  line 298	C++
 	ZeroMQ.dll!ZeroMQ::`dynamic atexit destructor for 'g_Context''()  + 0x28 bytes	C++
 	ZeroMQ.dll!_CRT_INIT(void * hDllHandle, unsigned long dwReason, void * lpreserved)  line 415	C
 	ZeroMQ.dll!__DllMainCRTStartup(void * hDllHandle, unsigned long dwReason, void * lpreserved)  line 526 + 0x11 bytes	C
[...]

PR #1341, opened 2015-02-06
Problem: issue #1144, 1377

### LINKED ISSUE #1340: Connect for SUB socket doesn't have immediate effect (test_disconnect_inproc sometimes fails)
I built libzmq (from GitHub master, currently latest commit 8e89d1162808e4432e4d65223599e1f9eee24dbd) on two machines with different hardware, but both running Ubuntu 14.04.1 64-bit. On one, `make check` completes successfully, on the other `test_disconnect_inproc` fails with the following error in test-suite.log:

    FAIL: tests/test_disconnect_inproc
    ==================================
    
    lt-test_disconnect_inproc: tests/test_disconnect_inproc.cpp:131: 
       int main(int, char**): Assertion `publicationsReceived == 3' failed. 

In the main while loop of the test, in the 2nd iteration (iteration == 1), the SUB socket connects to the inproc address of the XPUB socket and then receives messages in the 3rd, 4th and 5th iterations, when it disconnects (iteration == 4).
However, on the machine, on which the test fails, the SUB socket doesn't receive messages in the 3rd iteration, only in the 4th and 5th, thus the assertion at the end of the program that 3 messages have been received doesn't hold.
I ran the test using TCP too, and the result was the same.

If I put e.g. `sleep(5);` after `zmq_connect(subSocket, "inproc://someInProcDescriptor");`, then the test completes successfully.

This seems like a bug in libzmq.

hintjens:
Solution: add settle pause after zmq_connect

Fixes #1340

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -132,11 +132,14 @@ zmq::signaler_t::~signaler_t ()
     const struct linger so_linger = { 1, 0 };
     int rc = setsockopt (w, SOL_SOCKET, SO_LINGER,
         (const char *) &so_linger, sizeof so_linger);
-    wsa_assert (rc != SOCKET_ERROR);
-    rc = closesocket (w);
-    wsa_assert (rc != SOCKET_ERROR);
-    rc = closesocket (r);
-    wsa_assert (rc != SOCKET_ERROR);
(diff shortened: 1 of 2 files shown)


ricnewton:
Anything happening with this pull request?  It can't be automatically merged.

PR #1368, opened 2015-03-15
doc: zmq_msg_init does not set errno

calid:
In fact it always returns zero.

Relevant source: https://github.com/zeromq/libzmq/blob/eb3453372868a583bfff7e83e4dd197806e97ca3/src/msg.cpp#L51

This is also an issue in the stable repos.  If this commit looks good I can submit pull requests for those repos as well.


hintjens:
@calid, nice catch. If you'd like to send the same patches to 4-1 and 4-0 stable that'd be great.

PR #1379, opened 2015-04-21
Problem: integer narrowing issues.

evoskuil:
This resolves all 64 bit integer narrowing warnings and a couple related safety issues.
* An unsafe cast from `strlen` to `unsigned int` is rectified in z85.
* The assumption remains that it is always safe to cast `fd_t` to `int`.
* An assertion guard against `num_methods` overflow has been moved to a parameter constraint. This strengthens the guard in that it will be visible to callers at compile time (via a narrowing warning).
* There remains an unsafe situation where `socks_request_t::hostname` can overflow a buffer, but this is now mitigated by a pair of assertions

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -96,7 +96,7 @@ int zmq::options_t::setsockopt (int option_, const void *optval_,
         case ZMQ_IDENTITY:
             //  Identity is any binary string from 1 to 255 octets
             if (optvallen_ > 0 && optvallen_ < 256) {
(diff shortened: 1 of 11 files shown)


hintjens:
Do you want me to backport this to 4-1 stable?

evoskuil:
Up to you, I don't have a desire either way. I'm not hitting any issues here, this was just good citizenship.

hintjens:
OK, I'll leave it then. Thanks :)

Issue #1382, opened 2015-04-24
Blocking zmq_proxy_steerable with PUSH backend and SUB frontend


rikvdh:
Today I found out that zmq_proxy_steerable blocks after TERMINATE command via the control socket and when no-one is listening to the push socket. When then an incoming connection reads from the PUSH socket, the proxy is stopped.

I understand that the PUSH blocks when no listeners are there, but expected behavior is that the proxy stops when TERMINATE is sent.

```
	zmq::context_t context(1);
	std::string bind, con;

	zmq::socket_t frontend(context, ZMQ_SUB);
	frontend.connect("tcp://127.0.0.1:2345");
	frontend.setsockopt(ZMQ_SUBSCRIBE, NULL, 0);

	zmq::socket_t backend(context, ZMQ_PUSH);
	backend.bind("tcp://127.0.0.1:1234");

	zmq::socket_t control(context, ZMQ_SUB);
	control.connect("tcp://127.0.0.1:12345");
	control.setsockopt(ZMQ_SUBSCRIBE, NULL, 0);

	// Start the proxy
	zmq_proxy_steerable(frontend, backend, NULL, control);
```

hintjens:
Yes, the proxy should stop. Do you want to look at the code and propose a
fix?

rikvdh:
@hintjens I'm already on it. But I'm wondering, what is the best way to fix it.

I've already created a test which reproduces the problem. I found out that is blocks in the send-routine of the forward function in src/proxy.cpp:85. This was expected because the PUSH socket blocks here when the message is not delivered. But in the parent zmq::proxy, a message arrived for the control socket. But this is never reached.

Making the send non-blocking when a control-socket is attached can drop messages which conflicts with the spec.

Another option is to spawn the control handling in a different thread and stop the send (if busy) somehow.

A third option would be to pass along control messages within socket via a new zmq_setsockopt and react to the TERMINATE message.

A last resort would be fixing the documentation, and let the application make sure that the socket is pulled before terminating.

rikvdh:
@hintjens Can you please respond, I can't proceed with fixing this issue without input from your side.

hintjens:
Sorry, I missed your comment.

The correct solution is IMO to only receive from the frontend when the
back​end is able to receive messages (poll out is true). This will take a
little work on the code.

rikvdh:
@hintjens Shall I backport this or do you?

hintjens:
[...]

PR #1388, opened 2015-05-01
Fix for issue #1382

rikvdh:
This is the fix for issue #1382. Once it is merged I shall backport this to 4.1 and 4.0 repositories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ test_linger
 test_security_null
 test_security_plain
 test_proxy
+test_proxy_terminate
 test_abstract_ipc
 test_filter_ipc
 test_connect_delay_tipc
--- a/AUTHORS
+++ b/AUTHORS
@@ -86,6 +86,7 @@ Philip Kovacs
 Pieter Hintjens 
 Piotr Trojanek 
 Richard Newton 
+Rik van der Heijden
 Robert G. Jakabosky 
 Sebastian Otaegui 
 Stefan Radomski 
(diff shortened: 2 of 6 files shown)


hintjens:
Thanks for your fix, especially for adding a test case. Can you please squash these commits into a single commit? That's easier for others who want to work with this patch in the future. See git rebase. I can do the backport to 4.x stable. Also, see C4.1 for the style of commit messages ("Problem: ...").

rikvdh:
Hi @hintjens 
I've squashed it all :)

cachedout:
Hi @rikvdh and @hintjens 

This PR appears to have broken SaltStack's implementation of ZeroMQ whilst running on 4.0.6. :-/

Ref: https://github.com/saltstack/salt/issues/24630

We have determined that reverting the changes in `proxy.cpp` fixes the problem for us.

I do not know enough about ZMQ internals to begin to pretend that I know *why* this is so. However, I can certainly provide additional details as needed.

Since 4.0.6 has started to make its way into official distribution, such as Arch, this is a fairly critical issue for us. I'll continue to update this thread as we continue our investigation.

rikvdh:
Hi @cachedout,

I don't know anything about SaltStack, but there already have been some fixes in proxy.cpp in the 4.x branch after the release. Can you please check whether these fix your problem? If they do, @hintjens can problably make a new release for 4.x.

Does this also happen on 4.1.x?

cachedout:
Hi @rikvdh 

Thanks for your kind reply. Yes, it does look like the HEAD of the master branch on 4.x works correctly. Things also work correctly on the head of the master branch on 4.1.x.

We may just need a new git tag for a 4.x release that distros can pick up and package. That would solve the issue for us. Thanks again!

PR #1417, opened 2015-06-03
Zero copy message receive in v2_decoder

jens-auer:
This patch utilizes the zero-copy constructor of msg_t to construct message directly using the memory from the reception buffer. This trades a small increase in memory consumption for reduced allocations and memcopy operations.

The main idea of the patch is to use a reference-counted buffer allocated for every reception operation and then let messages use this buffer directly as their storage, increasing the reference count. When a msg_t object is destructed, it decreases the reference count and deletes the buffer if it reaches zero. To completely skip malloc calls, msg_t needs to be changed to not allocate a content_t object in msg_t::init_data, but store the members of content_t itself.

In order to not reimplement the decoder_t hierarchy completely, the buffer allocation is implemented as policy class which has to be a base-class of v2_decoder because decoder_t allocates the buffer in its constructor.

The result of these changes can be seen with a profiler. Prior to the change, receiving a message > 64 bytes needs 3 memcopy calls (flags, size and message data) and one malloc call per message. With the patch, this reduces to 2 memcopy calls (flags and size) and one malloc call per reception block of 8192 bytes. In the throughput test receiving 100,000 messages, malloc calls reduced from ~100,000 to 2,500. Throughput values increased by ~40%.

However, since this is a delicate change, I would hope that somebody with a good test library can validate it. I only have basic test setup which sends messages to a reactor and checks the content. Also make check passed all tests.

### DIFF
--- a/src/atomic_counter.hpp
+++ b/src/atomic_counter.hpp
@@ -183,7 +183,7 @@ namespace zmq
 #endif
         }
 
-        inline integer_t get ()
+        inline integer_t get () const
         {
             return value;
         }
--- a/src/decoder.hpp
+++ b/src/decoder.hpp
@@ -42,6 +42,45 @@
 
 namespace zmq
 {
+    // Static buffer policy.
+    class c_single_allocator
+    {
+    public:
+        c_single_allocator(size_t bufsize_):
+                buf((unsigned char*) malloc (bufsize) ),
+                bufsize(bufsize_)
+        {
+            alloc_assert (buf);
+        }
+
+        ~c_single_allocator()
+        {
+            free(buf);
+        }
+
+        unsigned char* allocate()
+        {
+            return buf;
+        }
+
+        void deallocate()
+        {
+
+        }
+
+        size_t size() const
+        {
+            return bufsize;
+        }
+
+    private:
+        unsigned char* buf;
+        size_t bufsize;
+
+        c_single_allocator( c_single_allocator const& );
+        c_single_allocator& operator=(c_single_allocator const&);
+    };
+
     //  Helper base class for decoders that know the amount of data to read
     //  in advance at any moment. Knowing the amount in advance is a property
     //  of the protocol used. 0MQ framing protocol is based size-prefixed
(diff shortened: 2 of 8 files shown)


hintjens:
For what it's worth, I've tested your changes, and CZMQ and Zyre on top of it, and they are happy.

minrk:
I believe this has caused custom free-functions to not be called under at least some circumstances. I'm still investigating the details to come up with either a fix or failing test case.

minrk:
Okay, found it: When an lmsg is copied via `zmq_msg_copy`, it doesn't appear to ever be freed (at least `ffn` is never called). I've got a failing test case; just need to track down the exact cause. @jens-auer any idea what change here would allow creating a copy to break the refcounting or other aspect of closing msg objects?

minrk:
Further narrowing it down: the reference counter is not shared between two copies. The copy appears to get a *copy* of the original's counter, so they both start with a value of 2, but decrementing the first counter does not decrement the second, causing the refcount to stop at 1 instead of 0, so the message is never freed.

My C++ allocation-fu is weak, but fixing this so the actual counter instance is shared after `msg::copy` should do the trick.

jens-auer:
Hi,

Thanks for the investigation. I think i know where the error is and is quite obvious now that the change in msg_t to remove content_t cannot work this way.  I have an idwa for a fix that still eliminates the memoy allocation, but i have to investigate this a little more.

Best wishes,

  Jens

-------- Ursprüngliche Nachricht --------
Von: Min RK <notifications@github.com> 
Datum: 09.06.2015  0:08  (GMT+01:00) 
An: zeromq/libzmq <libzmq@noreply.github.com> 
Cc: Jens Auer <jens.auer-githun@betaversion.net> 
Betreff: Re: [libzmq] Zero copy message receive in v2_decoder (#1417) 
 
Further narrowing it down: the reference counter is not shared between two copies. The copy appears to get a copy of the original's counter, so they both start with a value of 2, but decrementing the first counter does not decrement the second, causing the refcount to stop at 1 instead of 0, so the message is never freed.

My C++ allocation-fu is weak, but fixing this so the actual counter instance is shared after msg::copy should do the trick.

rodgert:
This may over complicate things, but could you introduce a new message
variant 'zcmsg' in the union that has the zero copy properties and fall
back to a traditional lmsg when the client explicitly requests a copy?

jens-auer:
Hi Thomas,

I think i'll do. My idea is to allocate enough space in the receive buffer to store the shared atomic counters, but to this i need a different constructor which does not allocate the atomic counter. Adding a new type fits this approach.
In the mean time, we could revert the change in msg_t and bring content_t back. This fixes the issue but introduces an allocation i would like to save.

Best wishes,
  Jens

-------- Ursprüngliche Nachricht --------
Von: Thomas Rodgers <notifications@github.com> 
Datum: 09.06.2015  13:58  (GMT+01:00) 
An: zeromq/libzmq <libzmq@noreply.github.com> 
Cc: Jens Auer <jens.auer-githun@betaversion.net> 
Betreff: Re: [libzmq] Zero copy message receive in v2_decoder (#1417) 
 
This may over complicate things, but could you introduce a new message
variant 'zcmsg' in the union that has the zero copy properties and fall
back to a traditional lmsg when the client explicitly requests a copy?

jens-auer:
Hi,

I think I have a fix for the problem which keeps the nice zero-copy behavior 
and does not introduce a memory allocation for the refcnt counter. Could you 
send me the test program or the test steps? Is it as simple as
0. Create a global variable bool released = false
1. Create a function release_function which set released=true when it is 
executed
2. Create a msg_t msg1 object with msg_t::init_data(..., release_function, ... 
)
3. Create a second msg_t msg2. 
4. msg2.copy(msg1)
5. msg1.close()
=> released == false
6. msg2.close()
=> released == true

Cheers,
  Jens

minrk:
@jens-auer I added test_msg_ffn in #1435. It's marked as XFAIL now, but if you have a fix you can unmark it as a failure and use it to confirm. The pyzmq refcount lifecycle tests also exercise this functionality.

jens-auer:
Hi,

I've fixed two issues in the zero-copy patch, including the reference-counting 
one. The test_msg_ffn now passes. Could you re-run your tests?

Best wishes,
  Jens

minrk:
Yup, pyzmq refcount tests pass with master now. Thanks!

hintjens:
@minrk, @jens-auer this is such amazing stuff, thanks :-)

PR #1423, opened 2015-06-05
Fail ZMQ_SNDHWM and ZMQ_RCVHWM setsockopt if already connected.

ricnewton:
Not this breaks test_xpub_nodrop, but it looks like the test is incorrect, will raise another pull request if that is so.


src/options.cpp
```diff
@@ -884,3 +885,21 @@ int zmq::options_t::getsockopt (int option_, void *optval_, size_t *optvallen_)
...
     return -1;
 }
+
+bool zmq::options_t::is_valid (int option_) const
+{
+    bool valid = true;
+
+    if (connected) {
+        switch (option_) {
+        case ZMQ_SNDHWM:
```
ricnewton: I just did SNDHWM and RCVHWM for now, presumably there are a load of other ones we could add.

hintjens:
I like this, it removes a large source of error and confusion.

PR #1426, opened 2015-06-05
Allow changing of high water marks after connection is established

ricnewton:


hintjens:
Nice... I was not expecting this. ☺

PR #1448, opened 2015-06-24
Add ZMTP connection heartbeats

jbreams:
This commit adds ZMTP connection heartbeats described in
http://rfc.zeromq.org/spec:37/ZMTP.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -358,7 +358,8 @@ test_apps = \
 	tests/test_server_drop_more \
 	tests/test_client_drop_more \
 	tests/test_thread_safe \
-	tests/test_socketopt_hwm
+	tests/test_socketopt_hwm \
+	tests/test_heartbeats
 
 tests_test_system_SOURCES = tests/test_system.cpp
 tests_test_system_LDADD = src/libzmq.la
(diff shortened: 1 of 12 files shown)


hintjens:
@jbreams this is awesome! :)

Could you add a description of the new socket options to doc/zmq_setsockopt.txt?

Usually the timeout can be fixed to 2 or 3 times the interval, so this option could be hidden from the API unless/until we find that we absolutely need to expose it.

jbreams:
Sure, will do. I also see there's a compile failure in VS2013 in the test I wrote for heartbeats, so I'll fix that as well.

jemc:
Cool!

jbreams:
I've been thinking about a default option for the timeout value - when I was writing this, I thought people would set the interval to be much higher than the timeout. So you send a PING message once every 10 seconds, but you would want a response within 1 second. Setting it to a fixed value 2-3 times the interval would mean that you'd have to send PING messages very frequently if you want to disconnect on high latency. But I guess 2x the interval is a reasonable default because otherwise there's no timeout at all (it will send PINGs without expecting any answer). I'll work on setting this default and updating the docs.

PR #1466, opened 2015-07-05
"zero-copy" raw_decoder

jens-auer:
A memcpy is eliminated when receiving data on a ZMQ_STREAM socket. Instead
of receiving into a static buffer and then copying the data into the
buffer malloced in msg_t::init_size, the raw_decoder allocates the memory
for together with the reference-counter and creates a msg_t object
on top of that memory. This saves the memcpy operation.

For small messages, data is still copied and the receive buffer is reused.

The implementation reuses the allocators from the zero-copy patches. The allocators have been moved to their own header file and raw_decoder uses an allocator member to create the buffers and create a zero-copy message without further malloc calls.


src/decoder_allocators.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <cmath>
+
+#include "msg.hpp"
+
+zmq::shared_message_memory_allocator::shared_message_memory_allocator(size_t bufsize_):
+        buf(NULL),
+        bufsize( 0 ),
+        max_size( bufsize_ ),
+        msg_refcnt( NULL ),
+        maxCounters( std::ceil( (double)max_size / (double)msg_t::max_vsm_size) )
```
tkoeppe: Why a C-style cast in C++?

src/decoder_allocators.cpp
```diff
@@ -0,0 +1,146 @@
...
+        max_size( bufsize_ ),
+        msg_refcnt( NULL ),
+        maxCounters( maxMessages )
+{
+
+}
+
+zmq::shared_message_memory_allocator::~shared_message_memory_allocator()
+{
+    if (buf) {
```
tkoeppe: This check seems unnecessary.

src/decoder_allocators.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    bufsize = max_size;
+    msg_refcnt = reinterpret_cast<zmq::atomic_counter_t*>( buf + sizeof(atomic_counter_t) + max_size );
+    return buf + sizeof( zmq::atomic_counter_t);
+}
+
+void zmq::shared_message_memory_allocator::deallocate()
+{
+    free(buf);
```
tkoeppe: missing `std::`

src/decoder_allocators.hpp
```diff
@@ -0,0 +1,149 @@
...
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef ZEROMQ_DECODER_ALLOCATORS_HPP
+#define ZEROMQ_DECODER_ALLOCATORS_HPP
+
+#include <stdlib.h>
```
tkoeppe: Why deprecated headers here but the correct `<cmath>` elsewhere?

src/decoder_allocators.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include "err.hpp"
+#include "atomic_counter.hpp"
+
+namespace zmq
+{
+    // Static buffer policy.
+    class c_single_allocator
+    {
+    public:
+        c_single_allocator(size_t bufsize_):
```
tkoeppe: Should this constructor be `explicit`?

hintjens: Best way to fix up code is to send pull requests, unless you are unsure of
what to do.

tkoeppe: Well, I'm sure what to do, but it the change by itself wouldn't qualify for a "fix" of a problem of any kind. It's all the little things I would probably just have asked the author to tidy up before merging. There are probably other opportunities for cleanup elsewhere, but then again "don't fix what ain't broken"...

hintjens: Sure, "code is not as clean as it could be" is a valid problem. We do this
all the time. Even a space in the wrong place is worth fixing. We have a
style guide for the code; you may want to check it.
http://zeromq.org/docs:style (it can be improved as we need.)

jens-auer: Hi,

I've created a pull-request for the issues. Thanks for pointing them out. Most 
of them I did just because I tried to adapt to the zeroMQ code which is 
already there, e.g. it includes old-style headers in many places and C-casts 
seem to be the standard.  For my own project, I usually avoid all the old-C-
style stuff and just forbid it, e.g. I would not allow malloc/free or C-style 
casts. 

Best wishes,
  Jens

jens-auer: I looked at the coding standard, and it is no concerned with things like C-
style casts, preferring C++ headers over C headers etc. Maybe it would be good 
to expand it to keep the code consistent (and improve it over time because 
coding style violations could be raised as quality issues).

Cheers,
  Jens

tkoeppe: Thanks! But don't take anything I say too seriously, I'm just an innocent bystander who knows some C++ and mostly takes his cues from the existing code base here. Defer to the maintainers for all questions regarding style.

jens-auer: To be honest, as a pure C++-programmer (I do not write C code), I found the 
code-base in some places rather unusual, e.g. using malloc/free and C-style 
casts abundantly. Sometimes, it looks C++ design which tries to be similar to 
C-code. Although it is very readable and I like the design.

Cheers,
  Jens

hintjens: Ah, we are all maintainers. If the code is readable and works as expected,
that is already a win.

PR #1467, opened 2015-07-05
Addressed comments from T. Köppe

jens-auer:
- Replaced stdlib.h with cstdlib
- Made single-argument constructors explicit

### DIFF
--- a/src/decoder_allocators.cpp
+++ b/src/decoder_allocators.cpp
@@ -38,7 +38,7 @@ zmq::shared_message_memory_allocator::shared_message_memory_allocator(size_t buf
         bufsize( 0 ),
         max_size( bufsize_ ),
         msg_refcnt( NULL ),
-        maxCounters( std::ceil( (double)max_size / (double)msg_t::max_vsm_size) )
+        maxCounters( std::ceil( static_cast<double>(max_size) / static_cast<double>(msg_t::max_vsm_size)) )
 {
 
 }
@@ -55,9 +55,7 @@ zmq::shared_message_memory_allocator::shared_message_memory_allocator(size_t buf
 
 zmq::shared_message_memory_allocator::~shared_message_memory_allocator()
 {
-    if (buf) {
-        deallocate();
-    }
+    deallocate();
 }
 
 unsigned char* zmq::shared_message_memory_allocator::allocate()
@@ -82,7 +80,7 @@ unsigned char* zmq::shared_message_memory_allocator::allocate()
         // allocate memory for reference counters together with reception buffer
         size_t const allocationsize = max_size + sizeof(zmq::atomic_counter_t) + maxCounters * sizeof(zmq::atomic_counter_t);
 
-        buf = (unsigned char *) malloc(allocationsize);
+        buf = static_cast<unsigned char *>( malloc(allocationsize) );
         alloc_assert (buf);
 
         new(buf) atomic_counter_t(1);
@@ -101,7 +99,7 @@ unsigned char* zmq::shared_message_memory_allocator::allocate()
 
 void zmq::shared_message_memory_allocator::deallocate()
 {
-    free(buf);
+    std::free(buf);
     buf = NULL;
     bufsize = 0;
     msg_refcnt = NULL;
@@ -119,7 +117,7 @@ unsigned char* zmq::shared_message_memory_allocator::release()
 
 void zmq::shared_message_memory_allocator::inc_ref()
 {
-    ((zmq::atomic_counter_t*)buf)->add(1);
+    (reinterpret_cast<zmq::atomic_counter_t*>(buf))->add(1);
 }
 
 void zmq::shared_message_memory_allocator::call_dec_ref(void*, void* hint) {
--- a/src/decoder_allocators.hpp
+++ b/src/decoder_allocators.hpp
@@ -30,7 +30,7 @@
 #ifndef ZEROMQ_DECODER_ALLOCATORS_HPP
 #define ZEROMQ_DECODER_ALLOCATORS_HPP
 
-#include <stdlib.h>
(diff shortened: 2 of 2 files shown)


src/decoder_allocators.cpp
```diff
@@ -38,7 +38,7 @@ zmq::shared_message_memory_allocator::shared_message_memory_allocator(size_t buf
         bufsize( 0 ),
         max_size( bufsize_ ),
         msg_refcnt( NULL ),
-        maxCounters( std::ceil( (double)max_size / (double)msg_t::max_vsm_size) )
+        maxCounters( std::ceil( static_cast<double>(max_size) / static_cast<double>(msg_t::max_vsm_size)) )
```
tkoeppe: I've seen other code spell this like `static_cast <double` -- Pieter, what's the style rule on the cast syntax?

src/decoder_allocators.hpp
```diff
@@ -41,16 +41,16 @@ namespace zmq
     class c_single_allocator
     {
     public:
-        c_single_allocator(size_t bufsize_):
+        explicit c_single_allocator(size_t bufsize_):
                 bufsize(bufsize_),
```
tkoeppe: Pieter: What's the style rule on indenting initializer list elements?

src/decoder_allocators.hpp
```diff
@@ -30,7 +30,7 @@
 #ifndef ZEROMQ_DECODER_ALLOCATORS_HPP
 #define ZEROMQ_DECODER_ALLOCATORS_HPP
```
tkoeppe: The include guard is off. Pieter: Might be worth codifying the existing practice in the style guide.

src/decoder_allocators.hpp
```diff
@@ -41,16 +41,16 @@ namespace zmq
     class c_single_allocator
     {
     public:
-        c_single_allocator(size_t bufsize_):
+        explicit c_single_allocator(size_t bufsize_):
                 bufsize(bufsize_),
-                buf((unsigned char*) malloc (bufsize))
+                buf(static_cast<unsigned char*>( malloc (bufsize) ))
```
tkoeppe: I have not seen a verdict on spacing type modifiers, `T* x` vs `T *x`. If we don't get a style rule on this, I recommend going with "majority per file".

src/decoder_allocators.cpp
tkoeppe: The style guide says to stick to 80 cols per line, but it doesn't seem to be enforced. Unless there's a strong sentiment to enforce this, I'd say stick with "keep it reasonable".

tkoeppe: Function calls are misspaced here. It should be `maxCounters (std:ceil (...) )`.

src/decoder_allocators.hpp
```diff
@@ -41,16 +41,16 @@ namespace zmq
...
         {
             alloc_assert (buf);
         }
 
         ~c_single_allocator()
```
tkoeppe: I think function declarations are spaced like function calls: `~c_single_allocator ()`.

src/decoder_allocators.hpp
jens-auer: Well, the usual practice seems to be to use something like 
__ZMQ_STREAM_ENGINE_HPP. If I'm not mistaken, using double underscores is not 
allowed in C++ because it is reserved for internal names.

Anyway, this seems to turn into a discussion of style, and the style guide 
doesn't specify most of the points from the other e-mails. I wouldn't create 
rules for some of the points, e.g. I would not define a rule where to put 
spaces in function calls. I would rather say "use spaces to make it readable". 
If I can't read it, I reformat it when reading the code.

Cheers,
  Jens

tkoeppe: Indeed, the names are reserved. Unfortunately, they are used everywhere else, and this calls for keeping it consistent and fixing consistently everywhere at once, I think.

hintjens:
@tkoeppe the style guide is a living contract. Where it is wrong, please go ahead and fix it. The rule is, the code and style guide should be consistent. Ideally the style is natural and obvious to anyone reading the code.

tkoeppe:
I'm afraid I can't edit that page, and I don't quite have the energy at the moment to go through a lenghty process. I'm sending you one PR that applies my understanding of the style for now, but I won't make any more manual changes.

I recommend two things:

1. Add details to the style guide: a) extend the spacing rules of `if` to include `else`, b) extend the spacing rules for function call subscripting to function definitions and direct-initializers ("constructor calls"), c) document include guards.

2. I expect that contributors will not generally pay a huge amount of attention to detail. It might be helpful to set up a `clang-format` profile for the style guide, then use clang-format to fix everything automatically, and then add a presubmit check that code changes are invariant under clang-format. Formatting is really not something that needs human attention, we have the tools now to make this a completely automatic matter.

tkoeppe:
More rules to specify:

* Policy regarding Newlines at the end of files and around include guards.
* Spacing of casts and of function-like operators (`sizeof`, `typeid`, etc.)
* Not specifying redundant `inline`s.
* Explicitness of one-parameter constructors.

hintjens:
:-) there is no "you", only "we"... if you can register for the wiki and
edit the page, great. Otherwise, your comments will be lost in time, like
tears in rain.

tkoeppe:
I'll see what I can do :-S

hintjens:
:-)

tkoeppe:
OK, done. If you disagree with any of my additions, do change them (but I think having *some* definite statement on each subject is better than having none). Hopefully future cleanup can be affected in drive-by fashion by future contributors.

hintjens:
Thank you :) Drive-by improvements is a great way to work.

PR #1472, opened 2015-07-08
Allow explicitly setting buf sizes to 0

reunanen:
Allow setting buffer sizes to 0 explicitly. At least on my Windows 7 box, the default send buffer size was 8192, and because of the ZeroMQ API design I could not set it to 0, which however appears to be a meaningful value (see e.g. https://support.microsoft.com/en-us/kb/201213), even if not very useful in general (see e.g. https://support.microsoft.com/en-us/kb/214397).


doc/zmq_getsockopt.txt
```diff
@@ -566,14 +564,12 @@ Applicable socket types:: all, when using multicast transports
...
-size for the specified 'socket'. A value of zero means that the OS default is
-in effect. For details refer to your operating system documentation for the
-'SO_SNDBUF' socket option.
+size for the specified 'socket'. For details refer to your operating system
+documentation for the 'SO_SNDBUF' socket option.
 
 [horizontal]
 Option value type:: int
 Option value unit:: bytes
-Default value:: 0
```
reunanen: The default is not 0 on my Windows box, and apparently not in Linux either: http://www.diag.com/reports/TCPWindows.html

PR #1476, opened 2015-07-19
Problem: lack test case for large stream messages

hintjens:
Solution: added to test_stream.cpp

### DIFF
--- a/tests/testutil.hpp
+++ b/tests/testutil.hpp
@@ -32,7 +32,7 @@
 
 #include "../include/zmq.h"
 #include "../src/stdint.hpp"
-#include "../src/platform.hpp"
+#include "platform.hpp"
 
 //  This defines the settle time used in tests; raise this if we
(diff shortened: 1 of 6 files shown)


tests/testutil.hpp
```diff
@@ -32,7 +32,7 @@
 
 #include "../include/zmq.h"
 #include "../src/stdint.hpp"
-#include "platform.hpp"
+#include "../src/platform.hpp"
```
ricnewton: Changing the platform include like this always breaks the cmake build where the plaform.hpp is generated out of source.

Can we change it back?

hintjens: For sure. Sorry, I did not realize. I've broken this multiple times when
building/testing individual cases. It might be worth wrapping this include
in a CMAKE-specific macro so that it's clear. Otherwise, maybe a comment...

hintjens:
@ricnewton I've reverted that breakage to the Cmake build.

PR #1479, opened 2015-07-20
Fixed #1477  corruption in "zero-copy" raw_decoder for payloads large…

### LINKED ISSUE #1477: corruption in "zero-copy" raw_decoder for payloads larger than 8192 bytes
Since commit 3679793, if the TCP server sends a message larger than 8192 bytes, the ZMQ_STREAM client will see zero-ed out bytes at the start of its message.
The number of zero-ed out bytes seems to correspond to "msgsize - 8192".

Some discussion of this was in #1447.

Test cases demonstrating the issue:
C++ version:
https://github.com/pijyoi/test_zmqstream/blob/master/zmqstream_corruption.cpp
Python version:
https://github.com/pijyoi/test_zmqstream/blob/master/tcpserver_zmqclient.py

jens-auer:
- Added test case from pijyoi
- Fixed wrong message::init arguments.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -365,7 +365,7 @@ test_apps = \
 	tests/test_thread_safe \
 	tests/test_socketopt_hwm \
 	tests/test_heartbeats \
-        tests/test_stream_exceeds_buffer
+	tests/test_stream_exceeds_buffer
 
 tests_test_system_SOURCES = tests/test_system.cpp
 tests_test_system_LDADD = src/libzmq.la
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -364,7 +364,8 @@ test_apps = \
 	tests/test_client_drop_more \
 	tests/test_thread_safe \
 	tests/test_socketopt_hwm \
-	tests/test_heartbeats
+	tests/test_heartbeats \
+        tests/test_stream_exceeds_buffer
```
jemc: extra whitespace here?

jemc:
I will happily merge, but wanted to give an opportunity to fix the extra whitespace before merging (for less commit noise).  Otherwise, I'll merge, then followup with a PR to remove the extra leading whitespace.

jens-auer:
Fixed the whitespace error. I used spaces instead of tabs, and my editor did not show whitespaces.

PR #1481, opened 2015-07-22
Fix several missing msg_t::close calls.

brian-peloton:
I traced memory leaks found using AddressSanitizer down to these.


jemc:
Thanks!

PR #1492, opened 2015-07-28
Fix a documented memory leak.

brian-peloton:
Despite the old comments, re-initing the msg_t leaks a refcount to metadata in some situations.


hintjens:
Thanks for this and your other patches. Nic work! If you like, next time,
no need to use branches. You can simply push all your PRs to master. One
merge then takes care of them all.

If there are any fixes you'd like to backport to the current stable
release, you can make the pull requests against
https://github.com/zeromq/zeromq4-1. We like an issue for each fix made to
a stable release, to help users track what changed.

Thanks again :-)

PR #1497, opened 2015-07-30
Check for libtoolize instead of libtool on Linux

bluca:
Same fix has been merged into Zproject in https://github.com/zeromq/zproject/pull/197 and https://github.com/zeromq/zproject/pull/198

The problem is the same:

autogen.sh has a check for the libtool binary as a mean to
check if libtool is available. But distributions like Debian and
Ubuntu are splitting the libtool package, and the libtool binary is now
in a separate package. What autoconf actually need is not the libtool
binary, but libtoolize and other macro files. So check for libtoolize
instead. On the other hand, OSX only ships libtool, not libtoolize,
and uses a pre-generated libtool script to build. So check for
libtoolize first and then for libtool, and fail if neither can be
found.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -20,10 +20,14 @@
 
 # Script to generate all required files from fresh git checkout.
 
-command -v libtool >/dev/null 2>&1
+# Debian and Ubuntu do not shipt libtool anymore, but OSX does not ship libtoolize.
+command -v libtoolize >/dev/null 2>&1
 if  [ $? -ne 0 ]; then
-    echo "autogen.sh: error: could not find libtool.  libtool is required to run autogen.sh." 1>&2
-    exit 1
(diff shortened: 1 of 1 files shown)


bluca:
@jemc - thanks for merging! What is the policy for backporting to stable branches? Should I backport this?

hintjens:
@bluca yes, you can backport to stable release repos. You need an issue for any backport so that users know what changed.

XiaoWanLINJU:
the command line check both libtool and libtoolize, but there is also a error message: autogen.sh: error: could not find libtool.  libtool is required to run autogen.sh. Does it mean both are missing?

bluca:
You need the libtool utility, and depending on the distribution this might be in the package libtool or the package libtool-bin

PR #1511, opened 2015-08-06
Avoid terminating connections prematurely

sorenh:
While sending very large messages (far beyond what fits in a the tcp
buffer, so it takes multiple sendto system calls for it to finish),
zmq_close will close the connection regardless of ZMQ_LINGER.

In case no engine is attached, a pipe->check_read() is needed to look
for the delimiter in the pipe and ultimately trigger the pipe
termination.

However, if there *is* an engine attached, the check_read() looks ahead
and finds the delimiter and terminates the connection even though the
engine might actually still be in the middle of sending a message.

This happens because while the io_thread is still busy sending the data,
the pipe can get terminated and the io thread ends up being terminated.

### DIFF
--- a/src/session_base.cpp
+++ b/src/session_base.cpp
@@ -460,7 +460,8 @@ void zmq::session_base_t::process_term (int linger_)
         //  TODO: Should this go into pipe_t::terminate ?
         //  In case there's no engine and there's only delimiter in the
         //  pipe it wouldn't be ever read. Thus we check for it explicitly.
-        pipe->check_read ();
+        if (!engine)
+            pipe->check_read ();
     }
 
     if (zap_pipe != NULL)


sorenh:
Thanks!

I would have left a comment on https://zeromq.jira.com/browse/LIBZMQ-554 and https://zeromq.jira.com/browse/LIBZMQ-551 asking them to test, but I can't seem to create an account on Jira. It says my domain is not allowed.

xaqq:
AFAIK Jira is not used anymore and all issue tracking happens on Github.

On Thu, Aug 6, 2015 at 8:54 PM, Soren Hansen <notifications@github.com>
wrote:

> Thanks!
>
> I would have left a comment on https://zeromq.jira.com/browse/LIBZMQ-554
> and https://zeromq.jira.com/browse/LIBZMQ-551 asking them to test, but I
> can't seem to create an account on Jira. It says my domain is not allowed.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zeromq/libzmq/pull/1511#issuecomment-128473275>.
>



-- 
Kapp Arnaud - Xaqq

sorenh:
Oh, ok. Thanks!

mikedevx:
Does anyone know the status of this fix, as it relates to a zmq release?  Is it in a release, or not yet in a release?

I see this fix was committed Aug 6, 2015.
The most recent stable release 4.1.4, Dec 18, 2015, does not *appear* to contain this fix.
It appears to not be planned for inclusion in 4.1.5 either?

The company product I work on makes *extensive* use of ZMQ 3 Push Pull, sometimes in a connect-send-disconnect model where because of this bug messages are lost.  We are VERY interested in upgrading to a version of ZMQ that contains this fix.  Our workarounds for this bug are proving to be unreliable.

hintjens:
@mikedevx if it helps, we can push forwards with a 4.2 stable release.

bluca:
@hintjens - should we wait to get ZMQ_REQ_CORRELATE fixed first, given it's marked as a stable API and it's broken at the moment? As far as I understand it's been looked at, see discussion in: https://github.com/zeromq/libzmq/pull/1730

hintjens:
Sure. It's also reasonable to make a release candidate with some pieces still needing work.

mikedevx:
Thank you!
We are trying the 1511 fix as a patch on zmq v 3.2.3, and so far, so good in testing.
I would say, No need to rush 4.2. We’ll evaluate the upgrade path when it is released!

Thank you again!
Michael Devereaux

hintjens:
Good stuff.

bluca:
@hintjens - What about backporting this fix to 4.0 and 4.1? Seems easy enough, I can do that if you want.

hintjens:
@bluca yes, nice. We have an issue, so if you could do that and update the NEWS in each case, that'd be awesome...

bluca:
Sure! Is LIBZMQ-554 the issue? There was no mention of one in this PR

hintjens:
Ah, I was confusing #1511 with an issue. We need an explicit issue then.

bluca:
Will open one.

PR #1520, opened 2015-08-14
remove unnecessary multiple WSAGetLastError() calls

reza-ebrahimi:
remove unnecessary multiple WSAGetLastError() calls

### DIFF
--- a/src/err.cpp
+++ b/src/err.cpp
@@ -89,12 +89,12 @@ void zmq::zmq_abort(const char *errmsg_)
 
 const char *zmq::wsa_error()
 {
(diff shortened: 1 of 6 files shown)


c-rack:
@reza-ebrahimi Thanks!
Just two hints:
1. Variables should be `snake_case`, not `camelCase`
2. Use spaces instead of tabs for indention.

reza-ebrahimi:
@c-rack You'r welcome!

That's Ok, I'll be fix and commit previous variable names in next 10 minutes,

Thanks for hint..

PR #1523, opened 2015-08-14
remove old fashion goto syntax and place ZMQ_DELETE macro for heap object deletion (ticket #1524)

reza-ebrahimi:
remove old fashion goto syntax and replace it with regular internally class method

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -498,6 +498,9 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 /******************************************************************************/
 
 #define ZMQ_UNUSED(object) (void)object
+#define ZMQ_DELETE(p_object) \
+	delete p_object; \
+	p_object = 0;
 
 #undef ZMQ_EXPORT
 
--- a/src/address.cpp
+++ b/src/address.cpp
@@ -49,16 +49,14 @@ zmq::address_t::~address_t ()
 {
     if (protocol == "tcp") {
         if (resolved.tcp_addr) {
-            delete resolved.tcp_addr;
-            resolved.tcp_addr = 0;
+            ZMQ_DELETE(resolved.tcp_addr);
         }
     }
 #if !defined ZMQ_HAVE_WINDOWS && !defined ZMQ_HAVE_OPENVMS
     else
     if (protocol == "ipc") {
         if (resolved.ipc_addr) {
-            delete resolved.ipc_addr;
-            resolved.ipc_addr = 0;
+            ZMQ_DELETE(resolved.ipc_addr);
         }
     }
 #endif
@@ -66,8 +64,7 @@ zmq::address_t::~address_t ()
     else
     if (protocol == "tipc") {
         if (resolved.tipc_addr) {
-            delete resolved.tipc_addr;
-            resolved.tipc_addr = 0;
+            ZMQ_DELETE(resolved.tipc_addr);
         }
     }
 #endif
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -97,15 +97,17 @@ zmq::ctx_t::~ctx_t ()
 
     //  Ask I/O threads to terminate. If stop signal wasn't sent to I/O
     //  thread subsequent invocation of destructor would hang-up.
-    for (io_threads_t::size_type i = 0; i != io_threads.size (); i++)
+    for (io_threads_t::size_type i = 0; i != io_threads.size (); i++) {
         io_threads [i]->stop ();
+    }
 
     //  Wait till I/O threads actually terminate.
-    for (io_threads_t::size_type i = 0; i != io_threads.size (); i++)
-        delete io_threads [i];
+    for (io_threads_t::size_type i = 0; i != io_threads.size (); i++) {
+        ZMQ_DELETE(io_threads [i]);
+    }
 
     //  Deallocate the reaper thread object.
-    delete reaper;
+    ZMQ_DELETE(reaper);
 
     //  Deallocate the array of mailboxes. No special work is
     //  needed as mailboxes themselves were deallocated with their
--- a/src/epoll.cpp
+++ b/src/epoll.cpp
@@ -56,8 +56,9 @@ zmq::epoll_t::~epoll_t ()
     worker.stop ();
 
     close (epoll_fd);
-    for (retired_t::iterator it = retired.begin (); it != retired.end (); ++it)
-        delete *it;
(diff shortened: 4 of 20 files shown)


kevinkreiser:
not sure on the style guidelines but looks like some tabs got in there. just fyi

jemc:
Let me just say I'm against this change - I think the `goto` style for errors is not "old fashioned", and actually has some nice benefits for reducing issues that arise in situations like when people need to refactor this code later, and forget to update one branch.

c-rack:
I agree with @jemc. Pull requests should, by [definition](http://rfc.zeromq.org/spec:22) (excerpt):
1. A patch SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.
2. A patch MUST adhere to the code style guidelines of the project if these are defined.

So from my point of view, this PR does not meet this criteria:
1. What is the problem to be solved? The only answer (which should be given by @reza-ebrahimi) I can imagine is to avoid `goto` completely. If that is the problem, than this PR is incomplete, since there are still a lot of other `goto` statements left in the project. Thus, it's not an accurate answer to the problem (if that is the problem to be solved), because this PR does not remove all `goto`s.
2. Common sense and coding style seems to be to have exactly _one_ exit point per function, which is achieved by using `goto`. Changing this coding style should be discussed in an issue rather than a PR.
3. This PR contains again tabs and not spaces, which is also against coding style guidelines. I already mentioned this in https://github.com/zeromq/libzmq/pull/1520.

@reza-ebrahimi, could you please open an issue which describes the problem you want to solve instead? Meanwhile, I would like to close this PR for the reasons I mentioned above.

@reza-ebrahimi It would be great if you could use git branches for your pull requests, because every pushed commit to your master branch is automatically added to your pull requests :-)

reza-ebrahimi:
hmm, Ok, I would like to say you my plan in ZeroMQ contribution...
I want to improve ZeroMQ to meet the MISRA standard and HIC++ guidelines, these are standards that used in critical systems such as Aerospace or Oil & Gas industry applications, if this is not enough, we can argue about that,

so about the goto statement, you are right, this is my first step to remove it completely from project, this may not solve any problems but code goes to be clean and ..., so this is part of my effort to achieve mentioned standards...
and about my pull requests, that's ok, but before that review my next commit...

reza-ebrahimi:
and about the tabs..., I'll fix all of the incorrect spaces in code.

c-rack:
@reza-ebrahimi OK, not meeting the MISRA standard and HIC++ guidelines (have not heart about both before, links would be great) is a good problem description. Nevertheless, I would suggest to open a regular issue with that problem statement and provide a PR (using a branch) that solves this problem (completely). I cannot see why it should not be merged then.

By the way, thanks for contributing! :+1:

reza-ebrahimi:
@c-rack Ok, but I can't understand, for every commit to master branch (main repo) I should open an issue?

c-rack:
@reza-ebrahimi It is not necessary (but recommended) to open an issue for every change to the source. But it is good practise to open an issue which describes your problem, then create a branch that contains the solution for your problem and then submit a PR as the solution to your problem. If the problem you want to solve is small, it is OK to just send a simple PR (from a branch).

See also:
http://rfc.zeromq.org/spec:22
https://www.youtube.com/watch?v=xFVDNTXIC_Y

hintjens:
It's good to be explicit about the problem you're addressing so that people
don't need to guess. Not meeting industry standards in the codebase is a
valid problem. Let's merge first, clarify second. Makes the conversation
smoother ...

hintjens:
Also, I counsel against using branches, its extra work with little benefit.
Just make sure each PR states the problem clearly. Then then can be
pipelined... Since we aim to always merge anyhow.

c-rack:
@hintjens Thanks for feedback. Let me reopen this PR then.

reza-ebrahimi:
@hintjens Thanks for your hints, all of my efforts focused on clean API and meet industry standards.

reza-ebrahimi:
@c-rack Thanks for reopening this PR, so I merged my new commits with this PR, by merging this commit, issue with ticket #1524 should be closed.

bluca:
@reza-ebrahimi - according to the HIC++ v4.0 standard, using goto statements to jump to a common exit and cleanup point is compliant:

```
bool f3 (int (&array)[10][10])
{
for (int j1 = 0; j1 < 10; ++j1)
{
for (int j2 = 0; j2 < 10; ++j2)
{
if (array[j1][j2] == 0)
{
goto finished;
}
   
// ...
                     
}
}
                 
finished:           // @@+ Compliant +@@
return true;
}
```

http://www.codingstandard.com/rule/6-3-1-ensure-that-the-labels-for-a-jump-statement-or-a-switch-condition-appear-later-in-the-same-or-an-enclosing-block/

The Misra standard does not seem accessible without an account unfortunately, or at least I cannot find it on their website.

reza-ebrahimi:
@bluca Yes, Both HIC++ and MISRA complains with goto statement, in MISRA with some restrictions:

MISRA C++: 2008 - Rule 6-6-1 (Rationale): 
Unconstrained use of goto can lead to programs that are extremely difficult to comprehend, analyse and, for C++, can also lead to the program exhibiting unspecified behaviour.

Since MISRA-C is aimed towards critical systems, MISRA-C:2004 has the approach to ban as many of these unconditional branch features as possible. Therefore, goto, continue and multiple returns were banned. break was only allowed if there was a single break inside the same loop.

so in my opinion the consensus of the world's programming community has roughly ended up something like goto, and our goal to modern C++ in ZeroMQ lead codebase to free from goto statements,

Yes The MISRA standard in not freely available, I have 2008 version, please give me your email address, I'll send it for you.

bluca:
@reza-ebrahimi - But as far as I can understand, both HIC++ and MISRA allow for goto if used reasonably. Here, as far as I can see, it is used in a perfectly reasonable way, that is to have a common cleanup and exit point, which happens to be one of the examples that the HIC++ standard explicitly marks as accepted.

Also, I respectfully disagree with the assertion that there is a consensus in the world's programming community :-) In case you are interested, you can read the opinion of many Linux kernel developers in this thread [1], who _strongly_ feel the opposite.

Finally, there was recently a paper [2] that outlines some empirical research, and shows that there's no actual proof that goto statements are inherently bad and cause bugs.

All rigorously IMHO of course :-)

[1] http://koblents.com/Ches/Links/Month-Mar-2013/20-Using-Goto-in-Linux-Kernel-Code/
[2] https://peerj.com/preprints/826v1.pdf

PR #1525, opened 2015-08-16
problem: user cannot poll on thread safe sockets

somdoron:
modify zmq_poll to support polling on multiple thread safe sockets.

Because thread safe socket doesn't have a fd we cannot poll on it.

To solve we are creating a dedicate fd (call poller, open for other names) per thread polling (same poller can be used for multiple sockets), associate the poller with the socket and then call zmq_poll.

An example can be found in the test, also working on a gist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,9 @@ test_client_drop_more
 test_client_server
 test_server_drop_more
 test_thread_safe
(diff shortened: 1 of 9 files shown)


c-rack:
Thanks!

PR #1529, opened 2015-08-16
Container-based CI build

bluca:
Hello,

Travis is moving away from VM-based builds to container-based builds. Advantages for users are more resources available and less waiting time before a build starts. One of the conditions for eligibility is to avoid using sudo, so these commits remove any use of it and use a local build prefix rather than the default /usr/local/ which requires superuser powers to write to, just like the existing qt-android CI scripts already do.

It can always be reverted by setting "sudo: required" in the travis.yml file, and then a full VM will be used instead.

In case you like this approach, I'll do the work to port it to CZMQ and zproject.

EDIT: also -x is set in ci_build.sh, since it is very useful for debugging purposes to see what command is being run in the Travis log. I can take it out if you'd rather not have it.

Documentation: http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ env:
 - BUILD_TYPE=default
 - BUILD_TYPE=qt-android
 
+sudo: false
+
 before_script:
 # ZMQ stress tests need more open socket (files) than the usual default
 # On OSX, it seems the way to set the max files limit is constantly changing, so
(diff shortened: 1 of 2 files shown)


c-rack:
Good work, thanks!

c-rack:
And yes, porting this to CZMQ and zproject is very welcome!

bluca:
@c-rack - It will take a while unfortunately. CMake is giving me a headache, due to one of its many bugs the pkg-config path cannot be extended via ENV var like you can do with more sane build systems. I'll try to find a workaround, but it might take a while.

Issue #1524, opened 2015-08-14
Problem: heap object deletion without assigning NULL to deleted object


reza-ebrahimi:
Problem: heap object deletion without assigning NULL to deleted object 
Solution: define a macro and use it in project in a unified manner, it helps avoiding double deletion and other similar memory corruption problems,

#define ZMQ_DELETE(p_object) \
	delete p_object; \
	p_object = 0;

I'll be commit the changes to master branch.

hintjens:
Can you use a different naming style for internal macros? We use ZMQ_xxx
for API constants. It's dangerous and confusing to also use this style for
internal macros.

I suggest LIBZMQ_xxx for example.

reza-ebrahimi:
@hintjens Your suggestion was accepted :)) 
but I would prefer to use in lower case zmq_delete, so please tell me which way we should go, my suggestion or your suggestion!

hintjens:
We use uppercase for macros..

reza-ebrahimi:
Ok, I'll change the newly added macro to LIBZMQ_XXX.

PR #1530, opened 2015-08-16
define a macro for heap object deletion in a unified manner (related …

reza-ebrahimi:
define a macro for heap object deletion in a unified manner (related to issue #1524)

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -503,6 +503,9 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 /******************************************************************************/
 
 #define ZMQ_UNUSED(object) (void)object
+#define LIBZMQ_DELETE(p_object) \
+	delete p_object; \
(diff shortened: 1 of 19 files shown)


c-rack:
:+1:

c-rack:
May I suggest to define the macro as:
```c
#define LIBZMQ_DELETE(p_object)  { \
    delete p_object; \
    p_object = 0; \
}
```
to avoid side-effects if someone forgets that this macro consists actually of 2 instructions?

Imagine someone writes a code like this:
```c
if(shall_be_deleted(ptr))
    LIBZMQ_DELETE(ptr);
```
then `ptr` will be `0` even if `shall_be_deleted` returned `false`.

reza-ebrahimi:
:+1:  :+1:  :+1: 

I'll be commit this definition, Thanks a lot...

PR #1531, opened 2015-08-16
apply a change to LIBZMQ_DELETE macro (related to issue #1524)

reza-ebrahimi:
apply a change to LIBZMQ_DELETE macro (related to issue #1524)

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -503,9 +503,10 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 /******************************************************************************/
 
 #define ZMQ_UNUSED(object) (void)object
-#define LIBZMQ_DELETE(p_object) \
+#define LIBZMQ_DELETE(p_object) {\
 	delete p_object; \
-	p_object = 0;
+	p_object = 0; \
+}
 
 #undef ZMQ_EXPORT
 


c-rack:
Thanks! For some reason this PR failed on Travis, but that may be a different problem that could be addressed in another PR. Maybe we need to increase timeouts after switching to container based builds (see https://github.com/zeromq/libzmq/pull/1529).

bluca:
@c-rack - this CI run failed because after a PR is merged the commit is no longer available (Github creates a branch behind the scenes for every PR, and deletes it when it's merged), as you can see in the log it failed because it couldn't pull the git ref:
```
fatal: Couldn't find remote ref refs/pull/1531/merge
The command "eval git fetch origin +refs/pull/1531/merge:" failed. Retrying, 2 of 3.
```

The build after the merge fails because now 3 tests fail (2 segmentation faults and one assertion failed), so I don't think it's related to timers:

https://travis-ci.org/zeromq/libzmq/jobs/75856296

reza-ebrahimi:
@c-rack I have fetched from upstream and pulling my local repo before applying changes, so if it is my mistake please help to resolve it,

c-rack:
@reza-ebrahimi I found the problem and submitted https://github.com/zeromq/libzmq/pull/1533 to fix the tests.

PR #1533, opened 2015-08-17
Problem: tests fail after PR #1531

c-rack:
This was a very subtle bug.

Before the LIBZMQ_DELETE macro, the logic was like:
```c
if(metadata) {
    if(metadata->drop_ref())
        delete metadata;
    metadata = 0;
}
```

which was replaced by:
```c
if(metadata) {
    if(metadata->drop_ref())
        LIBZMQ_DELETE(metadata);
}
```

After the improved version of LIBZMQ_DELETE macro, which should avoid such situations, some tests began to fail because we accidentially removed the `metadata = 0` statement, which was before executed regardless of `metadata->drop_ref()`.

This PR brings back the zero-ing of `metadata`.

### DIFF
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -239,6 +239,7 @@ int zmq::msg_t::close ()
         if (u.base.metadata->drop_ref ()) {
             LIBZMQ_DELETE(u.base.metadata);
         }
+        u.base.metadata = NULL;
     }
 
     //  Make the message invalid.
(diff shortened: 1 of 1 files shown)


reza-ebrahimi:
@c-rack But this is not relates to LIBZMQ_DELETE macro, the previous version same as:

if(metadata) {
    if(metadata->drop_ref())
        delete metadata;
    metadata = 0;
}

and I think improved version fixed the bug in the above code.

reza-ebrahimi:
@c-rack I see the code changed to as below:

```
   if (u.base.metadata != NULL) {
        if (u.base.metadata->drop_ref ()) {
            LIBZMQ_DELETE(u.base.metadata);
        }
        u.base.metadata = NULL;
    }
```
so what happens if drop_ref () returns false, I think there is a memory leakage case, is it Ok?

c-rack:
@reza-ebrahimi No, there is no memory leak, because metadata is [ref-counted](https://en.wikipedia.org/wiki/Reference_counting), if `drop_ref()` returns false, there are still other objects holding a reference to it, and thus we are not allowed to delete it. See:
https://github.com/zeromq/libzmq/blob/master/src/metadata.cpp#L52-L55

reza-ebrahimi:
I got it, :+1:

PR #1535, opened 2015-08-17
problem: trying to support polling on thread safe sockets with zmq_poll failed

somdoron:
solution: revert zmq_poll to the version before the attempt. adding new method zmq_pollfd_poll to poll on thread safe sockets (and regular sockets), migration from the two version should be pretty simple.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -410,8 +410,16 @@ typedef struct zmq_pollitem_t
 #define ZMQ_POLLITEMS_DFLT 16
 
 ZMQ_EXPORT int  zmq_poll (zmq_pollitem_t *items, int nitems, long timeout);
+
+/******************************************************************************/
+/*  Pollfd polling on thread safe socket                                      */
+/******************************************************************************/
+
 ZMQ_EXPORT void *zmq_pollfd_new ();
 ZMQ_EXPORT int  zmq_pollfd_close (void *p);
+ZMQ_EXPORT void zmq_pollfd_recv (void *p);
+ZMQ_EXPORT int  zmq_pollfd_wait (void *p, int timeout_);
+ZMQ_EXPORT int  zmq_pollfd_poll (void *p, zmq_pollitem_t *items, int nitems, long timeout);
 
 #if defined _WIN32
 ZMQ_EXPORT SOCKET zmq_pollfd_fd (void *p);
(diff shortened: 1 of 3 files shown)


bluca:
Hi,

CI build failure on OSX - qt-android is because there's a missing dependency, I'll fix it in another PR. Sorry about that!

Nice to see that OSX CI is running though :-)

paddor:
@somdoron May I ask why it didn't work?

I'm finally implementing my own poller based directly on `zmq_poll`. Of course I'd to use `zmq_poller`, since it also supports SERVER/CLIENT sockets, but it won't allow me to get a list of all readable/writable sockets after one call to `zmq_poller_wait`. Of course I could just recall `zmq_poller_wait` with zero timeout until it somehow reports that there are no new events, but I don't think that would work because ZMQ sockets (option ZMQ_FD) are edge-triggered. (Please tell me if I'm wrong.)

So I'm implementing my poller on `zmq_poll`. After the call to it, I can just iterate through the `zmq_pollitem_t` array and collect all readable and all writable sockets.

I wonder if maybe there is a way I could support SERVER/CLIENT sockets. Any idea?

somdoron:
about FD being edge triggered, it is only if you use the FD directly, if you use zmq_poll and or zmq_poller_wait it will be level triggered. So bottom line you can and should use zmq_poller_wait, to iterate over all ready sockets you can use zero timeout.

Regarding zmq_poll, IMO zmq_poller should deprecate zmq_poll eventually. Everything you can do with zmq_poll you can do with zmq_poller with much simpler API.

By the way, to compensate over zloop not supported I created zmq_timers (which is not expose in czmq yet), so you can use zpoller to poll on sockets and ztimers to manage timers.

paddor:
Oh that's great news. I will use `zmq_poller` at some later point, then. Right now I can't use it because I wanna support ZMQ >= 4.0.

And I think I'll untie CZTop::Loop from zloop then. Maybe even remove it completely. There's no use for it. In Ruby, there are other good timer libraries.

One last question: What if only one part (frame) of a message is available for reading. Will it already report the socket as readable? I guess not because ZMQ guarantees the delivery of complete messages, or nothing at all. But I'm not sure about this case.

somdoron:
Regarding your question you a right, all frames will be ready or nothing.

Regarding zmq_poller, just use zpoller, zpoller use zmq_poller if version >= 4.2

paddor:
Okay, cool.

Sorry, can't use zpoller. I need to be able to poll for writing, too.

PR #1536, opened 2015-08-17
CI run on OSX needs binutils

bluca:
OSX does not ship with binutils, need to install it before the script
run via brew in order to be able to use greadelf, needed by the
qt-android CI script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ env:
 
 sudo: false
 
+before_install:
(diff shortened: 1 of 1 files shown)


bluca:
Should run fine now, let's see :-)

bluca:
All green! Sorry for not catching this earlier.

c-rack:
:+1:

PR #1541, opened 2015-08-20
Problem: Source files contain mixed tabs and spaces.

jemc:
Solution: Convert to spaces and remove trailing whitespace in these files.


c-rack:
:+1:

Issue #1542, opened 2015-08-21
ZeroMQ will crash under Windows if port 5905 is in use by another process


mattbolgercsiro:
The use of port 5906 as the event_signaler_port is hard coded in two locations (signaler.cpp and config.hpp). If this port is already in use (in my case by RealVNC) then ZeroMQ will crash on the first call to WinSock select().

Shouldn't this port being in use be handled more gracefully and be caught at the actual setup of the sockets in signaler_t::make_fdpair()? 

http://lists.zeromq.org/pipermail/zeromq-dev/2015-August/029451.html

pijyoi:
The original implementation of a Windows signaler socket used port 5905 instead of letting the OS choose a free port. This also required a global mutex to be locked to prevent simultaneously running zeromq programs from clashing. 

Unfortunately the mutex was implemented with an Event instead of a Mutex. Changing it to a Mutex would break compatibility with existing programs unless the port was also changed. 

The hardcoded value of 5905 in signaler.cpp is just to check what port was configured in config.hpp. If 5905 it uses back the old Event implementation else it uses a Mutex. 

If 0, it lets the OS choose a free port without requiring use of a global mutex. (This is the best solution imho)

You could submit a PR to change the default value in config.hpp to 0.

hitstergtd:
Close issue as PR was accepted and merged.

bluca:
This was fixed a good while ago, which version are you using?

PR #1543, opened 2015-08-21
Set signaler_port to 0 which allows the OS to find a free port

mattbolgercsiro:
Rather than crashing if port 5905 is in use (https://github.com/zeromq/libzmq/issues/1542)
Added config.hpp to the source list so it shows up in generated projects
Remove CMAKE_BUILD_TYPE setting for generators that don't use it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,7 +326,9 @@ endif()
 #-----------------------------------------------------------------------------
 # default to Release build
 
-if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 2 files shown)


aburgm:
So this was merged in August last year but did not get included in the 4.1.4 version released in December. Is this only scheduled to be released with 4.2.x?

bluca:
If you'd like to have this in 4.1, please do cherry-pick the patch and send a PR to zeromq-4.1. Remember to update the NEWS with the github issue.

PR #1546, opened 2015-08-21
rename ZMQ_UNUSED macro to LIBZMQ_UNUSED

reza-ebrahimi:
rename ZMQ_UNUSED macro to LIBZMQ_UNUSED

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -512,14 +512,17 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 
 /******************************************************************************/
-/*  0MQ General                                                               */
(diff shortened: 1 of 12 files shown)


jemc:
I think `ZMQ_UNUSED` is supposed to be a public macro, so we shouldn't change it as there may be user code relying on it.

jemc:
Actually, I think the real issue here is that `LIBZMQ_DELETE` should not have been `#define`d in a public header `include/zmq.h`, but rather in a private header because it is a private internal-only macro.

jemc:
This relates to https://github.com/zeromq/libzmq/issues/1524#issuecomment-131328646

PR #1569, opened 2015-09-05
Fixed missing subscriptions on XPUB with manual subscriptions

djelenc:
The patch fixes the issue #1568.

PR #1570, opened 2015-09-05
Problem: Tests in Travis are failing randomly

djelenc:
Since [PR 1569](https://github.com/zeromq/libzmq/pull/1569) some tests were failing at random. The reason were illegal deque calls that were causing undefined behavior, for instance calling `front()` or `pop_front()` on an empty deque. Such calls are now safeguarded.


hintjens:
Nice, thanks.

PR #1574, opened 2015-09-08
avoid dereferencing uint32_t on unaligned address

pijyoi:
Unaligned access would cause SIGBUS on architectures that do not allow it.

Furthermore, the (now unneeded) fix for HPUX appears to have a bug in that the defined union may have padding bytes inserted.


malexzx:
it works

c-rack:
Thanks for testing! :+1:

PR #1582, opened 2015-09-17
Fixed the bug which cause the application crashed when use TCP connec…

pengweichu:
Fixed the bug which cause the application crashed when use TCP connec to an invalid address(such as 192.168.1.255).


c-rack:
:+1:

Issue #1554, opened 2015-08-28
Assert fails in signaler.cpp. Unstable issue


Thovart:
**Problem:**
The source of this file is the same as in the latest stable build (4.1.3)
Version I use 4.0.5

_Source file_: zeromq\src\signaler.cpp:

``` c++
int zmq::signaler_t::wait (int timeout_)
{
...............
#ifdef ZMQ_SIGNALER_WAIT_BASED_ON_POLL
    struct pollfd pfd;
    pfd.fd = r;
    pfd.events = POLLIN;
    int rc = poll (&pfd, 1, timeout_);
    if (unlikely (rc < 0)) {
        errno_assert (errno == EINTR);
        return -1;
    }
    else
    if (unlikely (rc == 0)) {
        errno = EAGAIN;
        return -1;
    }
...............
    zmq_assert (rc == 1);
    zmq_assert (pfd.revents & POLLIN);
    return 0;
}
```

Rarely, assert(pfd.revents & POLLIN) fails. In such cases:
pfd.events = 0x0300 (i.e. POLLIN)
pfd.revents = 0x0003 (i.e. POLLERR | POLLHUP)
**rc = 1**

**Solution:**
If we consult Posix documentation http://linux.die.net/man/2/poll 
Or MSDN documentation https://msdn.microsoft.com/en-us/library/windows/desktop/ms741669(v=vs.85).aspx
it is clearly stated, that

> On success, a positive number is returned; this is the number of structures which have nonzero revents fields (in other words, those descriptors with events or errors reported). A value of 0 indicates that the call timed out and no file descriptors were ready. On error, -1 is returned, and errno is set appropriately.

In other words: return value of poll _may_ be >0 and _error_ is reported with 'revents'

_A possible solution is to make additional check after 'poll' call:_

``` c++
    int rc = poll (&pfd, 1, timeout_);
...............
    if (pfd.revents & (POLLERR | POLLHUP))
    {
      errno = EAGAIN;
      return -1;
    }
```

**Please comment on this solution**

P.S. ZMQ related callstacks during crash:

```
Thread 035, 2 frame(s) match
          00 0000000006c6f5d0 000007fefc4f0f75 ntdll!NtWaitForSingleObject+0xa
          01 0000000006c6f650 000007fefc50b9c5 mswsock!_GSHandlerCheck_SEH+0x40ed
          02 0000000006c6f790 000007fefc4f2903 mswsock!MSAFD_WSPPoll+0x335
[...]

PR #1584, opened 2015-09-18
fixed zmq assertion in signaler.cpp under ubuntu

### LINKED ISSUE #1583: zmq assertion in signaler.cpp:292
After updating to the recent version i've began to get this error (the program was run under valgrind, ubuntu 14.04):

> Resource temporarily unavailable (/home/ymax/Projects/tester/zeromq-latest/src/signaler.cpp:292)
> ==14071==    by 0x74B4CC8: raise (raise.c:56)
> ==14071==    by 0x74B80D7: abort (abort.c:89)
> ==14071==    by 0x641801: zmq::zmq_abort(char const*) (err.cpp:84)
> ==14071==    by 0x65029C: zmq::signaler_t::recv() (signaler.cpp:292)
> ==14071==    by 0x642198: zmq::mailbox_t::recv(zmq::command_t*, int) (mailbox.cpp:87)
> ==14071==    by 0x654F99: zmq::socket_base_t::process_commands(int, bool) (socket_base.cpp:1256)
> ==14071==    by 0x65443C: zmq::socket_base_t::send(zmq::msg_t*, int) (socket_base.cpp:985)
> ==14071==    by 0x633763: s_sendmsg(zmq::socket_base_t*, zmq_msg_t*, int) (zmq.cpp:348)
> ==14071==    by 0x633898: zmq_send (zmq.cpp:373)
> ...

Looking into sources I've found the possible cause:
the assertion happens here:
```c
void zmq::signaler_t::recv ()
{
    //  Attempt to read a signal.
#if defined ZMQ_HAVE_EVENTFD
    uint64_t dummy;
    ssize_t sz = read (r, &dummy, sizeof (dummy));
    errno_assert (sz == sizeof (dummy));
```
And the creation point of socket is:
```c
zmq::signaler_t::signaler_t ()
{
    //  Create the socketpair for signaling.
    if (make_fdpair (&r, &w) == 0) {
        unblock_socket (w);
        unblock_socket (r);
    }
```
Here we intentionally put socket into non-blocking state. And in this case read can return -1. From man of eventfd: 
> If the eventfd counter is zero at the time of the call to read(2) then the call either blocks... or fails with the error EAGAIN if the file descriptor has been made nonblocking.

ymax:
Fixes #1583

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -289,18 +289,23 @@ void zmq::signaler_t::recv ()
 #if defined ZMQ_HAVE_EVENTFD
     uint64_t dummy;
     ssize_t sz = read (r, &dummy, sizeof (dummy));
-    errno_assert (sz == sizeof (dummy));
-
-    //  If we accidentally grabbed the next signal(s) along with the current
-    //  one, return it back to the eventfd object.
-    if (unlikely (dummy > 1)) {
-        const uint64_t inc = dummy - 1;
-        ssize_t sz2 = write (w, &inc, sizeof (inc));
-        errno_assert (sz2 == sizeof (inc));
-        return;
+    if (sz == -1) {
+        errno_assert (errno == EAGAIN);
     }
+    else {
+        errno_assert (sz == sizeof (dummy));
 
-    zmq_assert (dummy == 1);
+        //  If we accidentally grabbed the next signal(s) along with the current
+        //  one, return it back to the eventfd object.
+        if (unlikely (dummy > 1)) {
+            const uint64_t inc = dummy - 1;
+            ssize_t sz2 = write (w, &inc, sizeof (inc));
+            errno_assert (sz2 == sizeof (inc));
+            return;
+        }
+
+        zmq_assert (dummy == 1);
+    }
 #else
     unsigned char dummy;
 #if defined ZMQ_HAVE_WINDOWS


c-rack:
Thanks!

pijyoi:
I think an issue is the semantics for code calling this recv() method is that a signal got received upon return.

So if we return on EAGAIN, this contract is broken. 

The issue triggering this assertion may be elsewhere. The assertion is telling us that our assumptions are wrong.

ymax:
pijyoi, you are right. on the other hand as we have wait() and recv() as public methods it is supposed that we can call recv() without wait() and in that case (and taking into account the non-blocking mode of the socket)  it is normal to receive EAGAIN. 
IMHO the best solution may be to add wait() into recv() if the EAGAIN must not happen, or create a method wait_recv(timeout) (and move wait(), recv() to private) so the result will be predictable and wait timeout (EAGAIN) will be expected.

pijyoi:
There are only 3 places in the source that call signaler::recv()
1) mailbox.cpp
2) socket_base.cpp
3) zmq.cpp
The last two cases being for thread_safe sockets only, which are a recent addition.

For the first case, a *successful* call to signaler::wait() is made prior to calling signaler::recv(). So if signaler::recv() fails, that might indicate that there is a race somewhere?

See also #1554, which highlights an issue in signaler::wait()

pijyoi:
@ymax, I have made pull request #1595 that reverts signaler::recv() to its old state and created signaler::recv_failable() that allows EAGAIN and also returns an error code to the caller.

I have only made mailbox.cpp use this new method.

@will133, this pull request also allows the WSAEWOULDBLOCK in the Windows codepath.

Issue #1556, opened 2015-09-01
zeromq-4.1.3 release autoconf fails with defaults if libsodium is not available


doj:
Unpacking the zeromq-4.1.3.tar.gz on a Linux system, which doesn't have libsodium installed, fails a simple autoconf:

````Shell
$ tar xfz zeromq-4.1.3.tar.gz
$ cd zeromq-4.1.3/
$ ./configure
....
checking for sodium... no
configure: error: Package requirements (libsodium) were not met:

No package 'libsodium' found
````

However if I manually disable the libsodium like:
$ ./configure --without-libsodium
the autoconf scripts is running just fine.

Looking at the autoconf help text:
````Shell
$ ./configure --help | grep -A 2 sodium
  --with-libsodium        require libzmq build with libsodium. Requires
                          pkg-config [default=no]
````

So it seems that currently the default autoconf behavior is to require libsodium. I expect that autoconf does not stop if libsodium is not found on the system and it's not explicitely disabled with the --without-libsodium configuration argument.

doj:
fixing this issue is as simple as:

````patch
--- a/configure.ac
+++ b/configure.ac
@@ -418,7 +418,7 @@ have_sodium_library="no"
 AC_ARG_WITH([libsodium], [AS_HELP_STRING([--with-libsodium],
     [require libzmq build with libsodium. Requires pkg-config [default=no]])],
     [require_libsodium_ext=$withval],
-    [require_libsodium_ext=yes])
+    [require_libsodium_ext=no])
 
 # conditionally require libsodium package
 if test "x$require_libsodium_ext" != "xno"; then
````

hintjens:
I think the goal here is to force security into the build by default.

jemc:
I would agree - so what needs fixing is the autoconf help text, not the default behavior.

PR #1602, opened 2015-10-07
Fixes issue #1556

scemama:
Changes:
* libsodium is considered required by default, according to the help message.
* the test for the need of the ssp library for Solaris was moved in the region where we test for libsodium. It seemed to me more clear to test it here than before checking for libsodium.

I don't have a Solaris system, so I couldn't test : I am not sure if it still works on Solaris...


scemama:
Oops... I just realized I misunderstood "security" in the discussion of the issue. I thought it was "security of the build"...

PR #1611, opened 2015-10-18
problem: polling on multiple sockets is complicated

somdoron:
more:
problem2: every binding have to implement zpoller like logic
problem3: user have different API for polling on regular sockets and thread safe sockets.

This pull request port the zpoller from czmq. Also zmq_poller support polling on thread safe sockets (same API for both regular or thread safe sockets).

binding can now call zmq_poller API instead of implement the logic themself.


somdoron:
travis ci is not completing because outage in mac machines:

image

PR #1616, opened 2015-10-21
problem: ZMQ_POLLER doesn't support polling on out

somdoron:
Also fix in part of this PR:
* Problem: eagain is not the corrent error to return when timed out. returning ETIMEDOUT
* Problem: I don't have how to check if the libzmq version support ZMQ_POLLER. added define ZMQ_HAVE_POLLER.

### DIFF
--- a/src/socket_poller.hpp
+++ b/src/socket_poller.hpp
@@ -56,9 +56,9 @@ namespace zmq
             short events;
         } event_t;
 
-        int add_socket (void *socket, void *user_data, short events);
-        int modify_socket (void *socket, short events);
-        int remove_socket (void *socket);
+        int add (void *socket, void *user_data, short events);
+        int modify (void *socket, short events);
+        int remove (void *socket);
 #if defined _WIN32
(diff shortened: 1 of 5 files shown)


src/socket_poller.hpp
```diff
@@ -53,15 +53,19 @@ namespace zmq
...
         } event_t;
 
-        int add_socket (void *socket, void *user_data);
+        int add_socket (void *socket, void *user_data, short events);
+        int modify_socket (void *socket, short events);
         int remove_socket (void *socket);
 #if defined _WIN32
-        int add_fd (SOCKET fd, void *user_data);
+        int add_fd (SOCKET fd, void *user_data, short events);
+        int mofify_fd (SOCKET fd, short events);
```
ricnewton: Spelling mistake here, mofify -> modify

somdoron: thanks. Probably explaining why it is failing on windows.
Fixed in a new commit.

PR #1618, opened 2015-10-22
CMake: use libsodium + add Windows CI

madebr:
CMakeLists.txt did not correctly search for and include libsodium.

The new behavior is as follows.
- Search for libsodium:
  * If found, link libzmq and the test programs with it.
  * If not found, include tweetnacl statically in libzmq.
  * CTest: set TIMEOUT to 10 seconds.

Also added `appveyor.yml` definition for continuous integration tests on Windows.
Output of this PR: https://ci.appveyor.com/project/madebr/libzmq/build/build-16


CMakeLists.txt
```diff
@@ -25,16 +28,12 @@ if(WITH_TWEETNACL)
   set(TWEETNACL_SOURCES
     tweetnacl/src/tweetnacl.c
     )
-  if(WIN32)
-  else()
+
+  if(NOT WIN32)
     list(APPEND TWEETNACL_SOURCES tweetnacl/contrib/randombytes/devurandom.c)
```
ricnewton: So this breaks windows now when not using libsodium.

randombytes is an unresolved external, presumably tweetnacl has never worked on windows but now we are forcing its use.

CMakeLists.txt
```diff
@@ -9,14 +9,17 @@ if(APPLE)
...
 
-if(WITH_TWEETNACL)
-  add_definitions(-DHAVE_TWEETNACL -DHAVE_LIBSODIUM)
+set(USE_TWEETNACL OFF)
+
+if(SODIUM_FOUND)
+  add_definitions(-DHAVE_LIBSODIUM)
+  include_directories(${SODIUM_INCLUDE_DIRS})
+else()
+  set(USE_TWEETNACL ON)
```
ricnewton: So this breaks windows now when not using libsodium.

randombytes is an unresolved external, presumably tweetnacl has never worked on windows but now we are forcing its use.

madebr: Ok, will address this in a next PR.

PR #1620, opened 2015-10-23
CTest: add all sources in tests folder to CTest

madebr:
Changes introduced by this PR:
- CMake: Build with both libsodium and tweetNaCL
- In the tests folder, there were some `test_XXX.cpp` files which were not included in CTest. I have checked which tests could be compiled on Linux and which one could be compiled on Windows. I have added them to the correct category (I think).
  * In addition to this, CMake will test whether there is a test_XXX.cpp present in the tests folder that is not known to CTest. If so, a developper warning will be shown. Hopefully, this feature will avoid having a test not run on the CI.
  * **Main problem is that some of these tests fail.**

The following tests fail on both Linux and Windows:
58 - test_pair_tipc (OTHER_FAULT)
59 - test_reqrep_device_tipc (OTHER_FAULT)
60 - test_reqrep_tipc (OTHER_FAULT)
62 - test_router_mandatory_tipc (OTHER_FAULT)
65 - test_sub_forward_tipc (OTHER_FAULT)

The following tests fail only on Windows:
56 - test_ipc_wildcard (Timeout)
63 - test_srcfd (OTHER_FAULT)
66 - test_xpub_manual (OTHER_FAULT)

The following tests fail only on Linux:
77 - test_connect_delay_tipc (OTHER_FAULT)
78 - test_shutdown_stress_tipc (OTHER_FAULT)
81 - test_term_endpoint_tipc (OTHER_FAULT)

appveyor output:
https://ci.appveyor.com/project/madebr/libzmq/build/build-32

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -53,6 +53,21 @@ set(tests
         test_sockopt_hwm
         test_heartbeats
         test_poller
+        test_atomics
+        test_bind_src_address
+        test_capabilities
+        test_ipc_wildcard
+        test_metadata
+        test_pair_tipc
+        test_reqrep_device_tipc
+        test_reqrep_tipc
+        test_router_handover
+        test_router_mandatory_tipc
+        test_srcfd
+        test_stream_timeout
+        test_sub_forward_tipc
+        test_xpub_manual
+        test_xpub_welcome_msg
 )
 if(NOT WIN32)
   list(APPEND tests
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -53,6 +53,21 @@ set(tests
         test_sockopt_hwm
         test_heartbeats
         test_poller
+        test_atomics
+        test_bind_src_address
+        test_capabilities
+        test_ipc_wildcard
```
ricnewton: We don't support IPC on windows yet, maybe test_ipc_wildcard should be part of the non windows tests?

madebr: That can be arranged.
Please review all the tests in `tests/CMakeLists.txt`.
Might it be possible that a test is only included in the Linux section while the feature-under-test has a Windows implementation?

ricnewton:
Well I'm not sure, I know windows doesn't have IPC, and possibly not TIPC?

PR #1636, opened 2015-11-05
Added solution and project files to build with Microsoft VS2015

roalz:
I added the builds/msvc/vs2015 directory with all the files needed to build with Microsoft Visual Studio 2015.
Also updated the builds/msvc/readme.txt to include VS2015.

Note: I've used the pre-compiled libraries for libsodium to test that build was ok.

P.S. this is my very first Pull Request in GitHub, if I'm doing something wrong please be patient ;-)


jemc:
@roalz - you might also consider doing a PR to [zproject](https://github.com/zeromq/zproject/), which we use to generate the build system files for many other zeromq projects (like czmq and zyre, but no libzmq).  If you add a generator for VS2015, then all those other projects get that benefit "for free".

roalz:
Great! Sorry, I didn't know that.
As I'm currently just using libsodium, libzmq and czmq, I was working on a PR for czmq, but going through zproject sounds like the right way to go.
I'll look at it as soon as I can.
Thank you

PR #1638, opened 2015-11-08
Tcp buffer options to set RECV/SEND buffer

jens-auer:
I've added two new socket options to set the receive/send buffer for TCP sockets. This could be useful to specify a different buffer size from applications to reduce overhead from syscalls, e.g. when sending many large messages.

### DIFF
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -736,6 +736,33 @@ Option value unit:: N/A
 Default value:: not set
 Applicable socket types:: all, when using TCP transport
 
+ZMQ_TCP_RECV_BUFFER: Size of the TCP receive buffer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_RECV_BUFFER' specifies the maximum number of bytes which can
+be received by an individual syscall to receive data from the TCP
+socket. The buffer size is specified as an integer number from 0 (very small)
+to 10 (very large). The default value is 3.
+
+
+[horizontal]
+Option value type:: int
+Option value unit:: N/A
+Default value:: 3
+Applicable socket types:: all, when using TCP transport
+
+ZMQ_TCP_SEND_BUFFER: Size of the TCP receive buffer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_SEND_BUFFER' specifies the maximum number of bytes which can
+be sent by an individual syscall to transmit data to the TCP
+socket. The buffer size is specified as an integer number from 0 (very small)
+to 10 (very large). The default value is 3.
+
+
+[horizontal]
+Option value type:: int
+Option value unit:: N/A
+Default value:: 3
+Applicable socket types:: all, when using TCP transport
 
 RETURN VALUE
 ------------
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -1071,6 +1071,33 @@ Option value unit:: boolean
 Default value:: 1 (true)
 Applicable socket types:: all, when using TCP transports.
 
+ZMQ_TCP_RECV_BUFFER: Size of the TCP receive buffer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_RECV_BUFFER' specifies the maximum number of bytes which can
+be received by an individual syscall to receive data from the TCP
+socket. The buffer size is specified as an integer number from 0 (very small)
+to 10 (very large). The default value is 3.
+
+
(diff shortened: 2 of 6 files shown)


c-rack:
Not 100% sure, but if `ZMQ_TCP_RECV_BUFFER` / `ZMQ_TCP_SEND_BUFFER` are not set via `setsockopt`, the default for `tcp_send_buffer_size` resp. `tcp_recv_buffer_size` will stay at 3 instead of previous default of 8192, right?

May I suggest to use real number of bytes instead of a power of two factor?

Additionally, you set `tcp_send_buffer_size` for both options.
It seems that `tcp_recv_buffer_size` is never modified.

jens-auer:
Hi Constantin,

the default is 3, which corresponds to 2^3 * 1024 bytes = 8192 bytes. When I proposed the options on the mailing list, Pieter said that he preferred not to use the real number of bytes because it would be easy to mess up the behavior. So I came up with the idea to have a scale 1 (slow / small buffer) to 10 (fast / large buffer) instead.

The set case is a copy-paste error and is fixed with the new commit

c-rack:
Hi Jens,

thanks for the feedback! I might have missed that conversation on the ML and while I really respect Pieter's opinion, I disagree that using the real number of bytes isn't a good idea.

So I tried the following test code:
```c
void test_setsockopt()
{
    int rc;
    void *ctx = zmq_ctx_new();

    void *socket = zmq_socket(ctx, ZMQ_PUSH);

    int val = 0;
    size_t placeholder = sizeof(val);

    rc = zmq_getsockopt(socket, ZMQ_TCP_RECV_BUFFER, &val, &placeholder);
    assert(rc == 0);
    printf("val: %d\n", val);
    assert(val == 3);
    
    rc = zmq_setsockopt(socket, ZMQ_TCP_RECV_BUFFER, &val, sizeof(val));
    assert(rc == 0);
    printf("val: %d\n", val);
    assert(val == 3);

    rc = zmq_getsockopt(socket, ZMQ_TCP_RECV_BUFFER, &val, &placeholder);
    assert(rc == 0);
    printf("val: %d\n", val);
    assert(val == 3);
    
    val = 10;
    
    rc = zmq_setsockopt(socket, ZMQ_TCP_RECV_BUFFER, &val, sizeof(val));
    assert(rc == 0);
    printf("val: %d\n", val);
    assert(val == 10);

    rc = zmq_getsockopt(socket, ZMQ_TCP_RECV_BUFFER, &val, &placeholder);
    assert(rc == 0);
    printf("val: %d\n", val);
    assert(val == 10);
    
    zmq_close(socket);
    zmq_ctx_term(ctx);
}
```

and it failed with:

```
FAIL: tests/test_sockopt
Assertion failed: (val == 3), function test_sockopt, file tests/test_sockopt.cpp, line 29.
val: 3
val: 3
val: 8192
FAIL tests/test_socketopt (exit status: 134)
```

So setting the buffer size to 3 and getting it back (8192) return different results.

Also, there is a `return 0;` missing in `zmq::options_t::setsockopt`, so setting the new options will always return a result code of -1 instead of 0.

Anyway, I like your contribution to make the recv/send buffers configurable. So I'll merge your PR and we will address the adjustments/fixes in other PRs.

hintjens:
My idea was to offer just two or three settings (e.g. FAST, CHEAP, UGLY)
and have the code do the right thing automagically. However I've no
experience with the problem to offer more detail. A 1 to 10 setting seems
over-detailed.

We do this in other places, e.g. for HWMs we often switch between the
default, and unlimited. CZMQ even expresses this in its API
(zsock_set_unbounded) which is cleaner than setting the HWM to zero...

Good abstractions are important. I'm not sure they apply in this case, it
was just a crazy idea to explore or discard.

goddanao:
Got this one trying to make libzmq

src/stream_engine.cpp: In member function ‘virtual void zmq::stream_engine_t::out_event()’:
src/stream_engine.cpp:388:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
         while (outsize < options.tcp_send_buffer_size) {
                                  ^

c-rack:
Additionally, I get:
```c
src/options.cpp: In member function ‘int zmq::options_t::setsockopt(int, const void*, size_t)’:
src/options.cpp:288: error: call of overloaded ‘pow(int, int&)’ is ambiguous
```
I will provide a patch shorty.

c-rack:
@goddanao this should be fixed with https://github.com/zeromq/libzmq/pull/1640

c-rack:
After merge of #1640, I am going to submit a follow-up patch that:
1. adds a test case for the new options (as stated above)
2. changes new option's behaviour to represent actual byte buffer size instead of scale factor, since it just introduces another level of complexity where it is not needed.

PR #1643, opened 2015-11-19
Problem: when STREAM socket HWM is reached an assert will happen as metadata is trying to set twice

somdoron:
solution: check if the metadata already set.


hintjens:
Wonderful, thanks :-)

PR #1645, opened 2015-11-21
Problem: router socket can do pipe handover while reading multipart message

somdoron:
Fix pipe terimation in router while reading message


n054:
Tried to run default which errored out unsafe. I pushed the back to safety button came back here. Tried to run set through browse files and stalled at /src/. File passed 1.

PR #1649, opened 2015-11-24
Fix pipe activation race

sheremetyev:
Using ZeroMQ in a system with throughput of several million messages per second I was observing occasional drops of small number of messages, even when HWM is set to a value sufficient to accommodate all possible bursts in the flow.

The problem seems be caused by a race between process_activate_write() and check_hwm() on peers_msgs_read variable in zmq::pipe_t.

This PR adds isolated test and fixes the problem by increasing difference between LWM and HWM from fixed 1024 messages to 1/2 of HWM.


hintjens:
Nice work!

PR #1650, opened 2015-11-25
Solution: print test-suite.log if "make check" fails

c-rack:
Problem: test-suite.log is not accessible on failing CI build

Currently, a test fails on Travis-CI but that can not be reproduced locally.
Without the contents of the test-suite.log, this is difficult to analyze.

Solution: print test-suite.log if "make check" fails

This is based on the following answer on SO:
http://stackoverflow.com/a/32597140

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -21,7 +21,13 @@ if [ $BUILD_TYPE == "default" ]; then
     ( cd libsodium; ./autogen.sh; ./configure --prefix=$BUILD_PREFIX; make check; make install)
 
     #   Build and check this project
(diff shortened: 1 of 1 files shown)


c-rack:
@zeromq/libzmq-maintainers ping :-)

hintjens:
Sorry, would have done it faster but kept seeing this on my phone (and can't merge a failing PR from mobile)

c-rack:
Thanks. My goal was to make build green again from previous failing state.

However, the PR went red (only default build on Mac OS X) without any meaningful output in the test-suite.log and now, after merge, it is green again. I have no idea why but will keep an eye on this.

bluca:
Travis OSX machines seem to be extremely overbooked, so unfortunately tests that require a lot of system resources (file descriptors etc) very often fail :-(

PR #1654, opened 2015-12-08
Add the VMCI transport.

Kentzo:
VMCI transport allows fast communication between the Host
and a virtual machine, between virtual machines on the same host,
and within a virtual machine (like IPC).

It requires VMware to be installed on the host and Guest Additions
to be installed on a guest.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -53,6 +53,7 @@ Gonzalo Diethelm
 Guido Goldstein 
 Ian Barber 
 Ilja Golshtein 
+Ilya Kulakov 
 Ivo Danihelka 
 Jacob Rideout 
 Joe Thornber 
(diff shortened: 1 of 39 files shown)


Kentzo:
@hintjens No comments? :)

hintjens:
Just "awesome!" :-)

kevinkreiser:
Hi all! It seems there is a bit of a problem on ubuntu (12.04 and 14.04 so far I've tested).

When configured like so:

    ./configure --without-libsodium --without-documentation

They both do this:

    In file included from src/socket_base.cpp:73:0:
    src/vmci_address.hpp:39:26: fatal error: vmci_sockets.h: No such file or directory
     #include <vmci_sockets.h>
                              ^
Interestingly configure seems to say that by default its disabled:

      --with-vmci             build libzmq with VMCI transport [default=no]

And indeed I dont see anything in the Makefile directly setting ZMQ_HAVE_VMCI but its generated and a bit hard to follow. Any clue what is going on here? Note that sending `--with-vmci=no` in configure does work though.

This might suggest that there is a problem with the default as well as the detection in `configure.ac` or other autotools stuff.

Kentzo:
I have the same problem with libsodium: unless you specifically sau "no", it will be included. And I used its code as a template.

PR #1681, opened 2015-12-21
Problem: MSVC projects don't include timers.cpp/hpp

hintjens:
Causes link errors during build.

Solution: add these.


hintjens:
@c-rack :-) thanks for the super fast merges....

PR #1683, opened 2015-12-21
Problem: includes timer.cpp/hpp instead of timers.cpp/hpp

hintjens:
Solution: fix it.


hintjens:
@somdoron I realized what you meant, finally... :)

somdoron:
:-)

Issue #1690, opened 2016-01-01
ZMQ_REQ_RELAXED does not work


FredTreg:
The ZMQ_REQ_RELAXED option should allow you to send two messages in a row without a successful receive in between.

This does not seem to work as the simple example below (two "send" in a row) blocks indefinitely.

Note: this does not work even if the server is connected and responding.

Am I mistaken somehow?

I could track down the issue to this piece of code in req.cpp :
``` C++
    if (receiving_reply) {
        if (strict) {
            errno = EFSM;
            return -1;
        }

		if (reply_pipe)
			reply_pipe->terminate(false); <-- Line causing the issue
    ...
```
Indeed, terminating the reply pipe also terminates the send pipe as they are the same (the pipe associated with the socket is bidirectional). Doing a terminate on the pipe sets an internal flag called `out_active` to `false` and the pipe can no longer send messages.

A simple removal of the `terminate`line solves the problem. Removing this line should not be an issue as the incorrect ordering of messages that could be incurred is taken care of by the ZMQ_REQ_CORRELATE option if needed.

Example code demonstrating the issue (without error handling to make it clearer):

``` C++
#include "zmq.h"
#include <stdio.h>

int main (int argc, char *argv [])
{
    void *ctx;
    void *s;
    int on = 1;
    zmq_msg_t msg;

    ctx = zmq_ctx_new ();
    s = zmq_socket (ctx, ZMQ_REQ);
    zmq_setsockopt(s, ZMQ_REQ_RELAXED, &on, sizeof(int));
    zmq_connect (s, "tcp://127.0.0.1:6789");

    zmq_msg_init (&msg);
    zmq_msg_init_size (&msg, 10);
    zmq_msg_send (&msg, s, 0);
    zmq_msg_close (&msg);

    printf ("First message sent\n");

    zmq_msg_init (&msg);
    zmq_msg_init_size (&msg, 10);
    zmq_msg_send (&msg, s, 0); // <--- This line will block indefinitely
    zmq_msg_close (&msg);
   
    printf ("Second message sent\n");

    zmq_close (s);
    zmq_ctx_term (ctx);

    return 0;
}
```

hitstergtd:
@FredTreg,
Thanks for the report. Closing issue since your pull request was merged by @c-rack.

PR #1691, opened 2016-01-01
Fixed issue #1690 (ZMQ_REQ_RELAXED)

FredTreg:
When using ZMQ_REQ_RELAXED and a 'send' is executed after another 'send' the
previous code would terminate the 'reply_pipe' if any.
This is incorrect as terminating the reply pipe also terminates the send pipe
as they are the same (a pipe associated with a socket is bidirectional).
Doing a terminate on the pipe sets an internal flag called out_active to false
and the pipe can no longer send messages.
Removing the 'terminate' solves the problem. Removing this call is not an issue
as the incorrect ordering of messages that could be incurred is taken care of
by the ZMQ_REQ_CORRELATE option if needed.

### DIFF
--- a/src/req.cpp
+++ b/src/req.cpp
@@ -60,8 +60,6 @@ int zmq::req_t::xsend (msg_t *msg_)
             return -1;
         }
 
-        if (reply_pipe)
-            reply_pipe->terminate (false);
         receiving_reply = false;
         message_begins = true;
     }


c-rack:
Thanks. Do you think it would be possible to add a minimal test case based on your code in #1690, too?

FredTreg:
I guess I can. Give me a couple of days though, I am using a Java binding to do my testing so I would need to set the proper C++ environment to create the test case. (and thanks for merging, that was quick :)

PR #1692, opened 2016-01-03
Test for issue #1690 (ZMQ_REQ_RELAXED)

FredTreg:
Current ZMQ_REQ_RELAXED test improvement to check that pipes are not closed
after executing two send() in a row with no recv() in between.


c-rack:
Thanks :+1:

PR #1718, opened 2016-01-20
Last bits of Hurd support

gcsideal:
Work of Svante Signell, quoting him:
"I did manually issue configure, make and make check in the git directory, and they passed fine. It seems like even the git version, 4.2.0?, should build when packaged."

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -702,6 +702,11 @@ if !ON_LINUX
 XFAIL_TESTS += tests/test_abstract_ipc
 endif
 
+if ON_GNU
+XFAIL_TESTS += test_ipc_wildcard \
+		test_term_endpoint
+endif
+
 EXTRA_DIST = \
 	CMakeLists.txt \
 	autogen.sh	\
--- a/configure.ac
+++ b/configure.ac
@@ -146,20 +146,6 @@ CPPFLAGS="-D_REENTRANT -D_THREAD_SAFE $CPPFLAGS"
 # For host type checks
 AC_CANONICAL_HOST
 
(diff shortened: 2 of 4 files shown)


bluca:
:+1:

hintjens:
@gcsideal configure.ac:284 says:

    gnu*)

And I'm fairly sure it should say:

    *gnu*)

Can you check? I'm backporting Hurd support into zproject, so we'll have it in CZMQ, Zyre, and so on.

gnu-srs:
No, for $host_os it should be gnu* and for $host *-gnu* as the last option, see
https://www.gnu.org/software/hurd/hurd/porting/guidelines.html. Otherwise other systems ending with gnu* are included too.

hintjens:
Ah, I thought the asterisk was some kind of delimiter. OK, makes sense,
thanks for that.

hintjens:
I've added GNU/Hurd support to CZMQ though cannot test it, if you get the
chance to do so, that'd be cool:
https://github.com/hintjens/czmq/commit/3104ad1a399cf03a9e7f14ad32f3a6c3eacec7d8

bluca:
@gcsideal could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@gcsideal  another ping. Please consider helping us with the relicense. Thank you!

PR #1719, opened 2016-01-20
truncate rc to INT_MAX for send/recv

### LINKED ISSUE #1697: zmq_send/recv overflow with sizes larger than INT_MAX
Problem:

1. zmq_send/recv, etc. operations return a number of bytes in `rc` or `-1` for errors
2. this value is an `int`
3. the operations support sizes larger than `INT_MAX`

So if you send or receive messages larger than `INT_MAX`, `rc` will overflow and have an invalid negative value.

minrk:
to avoid overflow on large messages

send/recv still succeed for large messages, but `zmq_msg_size` must be used to get the true size.

closes #1697


bluca:
Would it be worth adding a notice to the documentation for this function in the public header?

minrk:
Probably makes sense

PR #1727, opened 2016-01-27
problem: pub/sub pattern in zeromq is complicated and not thread safe

somdoron:
suggestion solution: making new pattern, radio-dish.

radio-dish pattern is thread safe, therefore is single frame protocol.

API is simplified as well, instead of zmq_setsockopt, you can call zmq_join and zmq_leave.
Radio-dish matching is exact matching (vs prefix of pubsub). 
Groups are NULL terminated string. message must start with the group, NULL and than the message.


c-rack:
:+1:

PR #1739, opened 2016-01-30
Problem: Spelling of identifier "interface" matched a macro definition in VS2015

sunkin351:
Solution: spelling correction to "Interface".

Affected files:
udp_address.hpp
udp_address.cpp

### DIFF
--- a/src/udp_address.cpp
+++ b/src/udp_address.cpp
@@ -100,8 +100,8 @@ int zmq::udp_address_t::resolve (const char *name_)
     else
         is_mutlicast = false;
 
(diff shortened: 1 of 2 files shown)


jemc:
@sunkin351 Rather than using an inconsistent capitalization here, what about just choosing a different spelling, like `iface`?

sunkin351:
You can do that it you like, my goal was just to make it functional.

jemc:
#1742 resolves this same problem without using inconsistent capitalization.

hintjens:
We have this problem all the time, and our usual solution is to use 'iface'.

PR #1742, opened 2016-01-30
Problem: `udp_address::interface` conflicts with VS2015 build.

jemc:
Solution: rename to `iface`.
Resolves #1739.


sunkin351:
Mind if I also add the VS2015 project files to the master?

PR #1745, opened 2016-01-30
problem: udp transport doesn't compile on windows

somdoron:


sunkin351:
Its saying that all three checks have failed...?

somdoron:
It is not related to the patch... Currently zeromq master fail frequently.

PR #1757, opened 2016-02-01
Revert "Problem: deprecated use and reference of `zmq_term`"

hintjens:
Reverts zeromq/libzmq#1756

I'm reverting this because for now we try to support old versions of ZeroMQ. We can roll forwards in the next major release of CZMQ. However at present, CZMQ will build on libzmq 2.2 forwards afaik.

### DIFF
--- a/builds/zos/test_fork.cpp
+++ b/builds/zos/test_fork.cpp
@@ -52,7 +52,7 @@ int main (void)
         //  Child process
         //  Immediately close parent sockets and context
         zmq_close (pull);
-        zmq_ctx_term (ctx);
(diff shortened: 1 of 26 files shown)


c-rack:
The `zmq_term` function was untouched and was still available, I only updated the tests, perf, comments and docs to use resp. to reference `zmq_ctx_term` instead. But of course, we can do this later.

hintjens:
Oh. Sorry, I misread it, and thought this commit was against czmq not libzmq.  :-(

Yes, definitely. (How does one revert a revert...?)

PR #1770, opened 2016-02-04
Problem: no support for pre-allocated file descriptors [systemd | upstart socket-based activation]

bluca:
Solution: add a bunch of commits that add it, see that they don't work, then go to a ZMQ hackaton and get help to debug and spot the glaringly obvious and silly mistake that was sitting in front of my face the whole time :-)

Thanks to everyone who helped debug the IPC FD problem last week!

Main use case of this feature is socket-based activation in systemd (and upstart too I believe). With very simple service and socket units it's possible to have a service started on-demand, when packets come in, without any data loss:

test.service:

    [Service]
    ExecStart=/some/binary

test.socket:

    [Socket]
    ListenStream=127.0.0.1:5560
    ReusePort=true
    
    [Install]
    WantedBy=sockets.target

Note that, given the endpoints are known and must match what's in the service, we could in theory be way smarter and use systemd's APIs directly, and figure out if the endpoint being used matches any of the sockets that systemd is managing, and if so use the pre-allocated file descriptor. This would mean that applications using the library do not need to change anything.

But I believe this smartness, if it should happen at all, is better placed to live in higher level bindings. Base library should have just the base feature, I think, and then the higher level one can build upon it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,6 +125,7 @@ test_timers
 test_radio_dish
 test_udp
 test_large_msg
+test_pre_allocated_fd_ipc
 tests/test*.log
 tests/test*.trs
 src/platform.hpp*
--- a/.gitignore
+++ b/.gitignore
@@ -126,6 +126,7 @@ test_radio_dish
 test_udp
 test_large_msg
 test_pre_allocated_fd_ipc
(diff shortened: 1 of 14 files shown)


nickbroon:
Yah! :-)

PR #1774, opened 2016-02-05
Don't call memcpy with 0 size and NULL pointer(s)

brian-peloton:
It's undefined behavior, and ubsan flags it.

### DIFF
--- a/src/dealer.cpp
+++ b/src/dealer.cpp
@@ -70,7 +70,8 @@ int zmq::dealer_t::xsetsockopt (int option_, const void *optval_,
     size_t optvallen_)
 {
     bool is_int = (optvallen_ == sizeof (int));
-    int value = is_int? *((int *) optval_): 0;
+    int value = 0;
+    if (is_int) memcpy(&value, optval_, sizeof (int));
 
     switch (option_) {
         case ZMQ_PROBE_ROUTER:
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -92,7 +92,8 @@ int zmq::options_t::setsockopt (int option_, const void *optval_,
     size_t optvallen_)
 {
     bool is_int = (optvallen_ == sizeof (int));
-    int value = is_int? *((int *) optval_): 0;
+    int value = 0;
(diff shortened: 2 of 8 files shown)


hintjens:
You've not, afaics, identified a real issue. Making the code more complex without solving real problems is not a good idea.

bluca:
If I may add my 2 cents, in my opinion these kind of undefined behaviour issues flagged by static analysis are really worth fixing, even if they make the code a tiny bit more complex.

Just a few months ago I spent a week chasing down a bug in another project. The code would take the wrong turn at a branch. A pointer that was NULL was evaluating as true. You can imagine my utter confusion when running through GDB... I would have pulled out my hair, if I still had any :-)

I don't even remember how, but after much, MUCH pain I finally figure out this snippet was the culprit:

    	char* data = dataLength?(char*)malloc(dataLength):0;
 	memcpy(data, dataTemplateInfo->data, dataLength)
        ...
        if (data)
            ...

Even though dataLength was 0, and data was NULL, "if (data)" was evaluating as true. That's when I learned that memcpy'ing to a NULL pointer is undefined behaviour, and GCC produced that broken conditional jump in assembly as a result when any level of optimization was turned on.

The original code was indeed smaller, neater and better to look at. But also wrong, and the source of a lot of pain :-)

This is the code in question and the fix in case you are curious: https://github.com/bluca/vermont/commit/ea53504746b0b8c983803471329661c678affc30

hintjens:
Fixing undefined behavior is fine. The subject of the commit was clear. The code did not match it, and did not make sense except with extra background explanation. This is not improving the code, it's making it more opaque and fragile.

If we do not allow nulls, then we block them cleanly (with asserts and/or error returns) and we fix the test cases.

If we allow null+0 as a combination then we should IMO state this explicitly and handle it explicitly in the code.

Sorry to be argumentative.

brian-peloton:
OK on fixing it that way. Also, I now realize that me mixing two things into the same commit hasn't been helping since the beginning. Sorry about that...

PR #1777, opened 2016-02-07
Use memcpy instead of assuming option values are aligned

brian-peloton:
Otherwise, it's undefined behavior. ubsan catches alignment issues in
the libzmq test suite without this.


bluca:
@brian-peloton could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

PR #1782, opened 2016-02-08
fallback on tweetnacl if libsodium is not found and not explicitly requested

### LINKED ISSUE #1781: cannot build with tweetnacl on OS X
- tweetnacl config is only supported when using cmake
- the cmake builds don't work on OS X

minrk:
and `--with-tweetnacl` for explicitly requesting build with tweetnacl

closes #1781


hintjens:
Awesome.

PR #1784, opened 2016-02-08
Problem: ZMQ_PRE_ALLOCATED_FD is too long

bluca:
Solution: rename socket option (and variables and files) from
pre_allocated_fd to usefd.

@hintjens - better like this? I've renamed everything to keep it consistent, but I can rename just the user-visible option if you want to keep the churn to a minimum.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,8 +125,8 @@ test_timers
 test_radio_dish
 test_udp
 test_large_msg
-test_pre_allocated_fd_ipc
-test_pre_allocated_fd_tcp
+test_usefd_ipc
+test_usefd_tcp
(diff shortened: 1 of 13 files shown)


c-rack:
I like the shorter name, but would rather name it `ZMQ_USE_FD` and `use_fd` to make it consistent with the other options, where each word is separated. However, let's merge and maybe change afterwards.

bluca:
I don't mind either way. @hintjens any preferences?

hintjens:
@c-rack's suggestion is neat.

bluca:
Ack, will get on to it, thanks for the feedback.

PR #1786, opened 2016-02-09
Cleaning up recent option names

hintjens:
@minrk I've also noticed the large message test freezes my laptop for a few seconds. Since we're now implementing a maximum message size, I've made this configurable via zmq_ctx_set(). This was long overdue I think. Needs a little help to be fully working.

### DIFF
--- a/doc/zmq_ctx_get.txt
+++ b/doc/zmq_ctx_get.txt
@@ -26,19 +26,29 @@ ZMQ_IO_THREADS: Get number of I/O threads
 The 'ZMQ_IO_THREADS' argument returns the size of the 0MQ thread pool
 for this context.
 
+
 ZMQ_MAX_SOCKETS: Get maximum number of sockets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The 'ZMQ_MAX_SOCKETS' argument returns the maximum number of sockets
 allowed for this context.
+
+
+ZMQ_MAX_MSGSZ: Get maximum message size
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_MAX_MSGSZ' argument returns the maximum size of a message
+allowed for this context. Default value is INT_MAX.
+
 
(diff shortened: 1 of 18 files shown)


minrk:
This defines (but does not implement) a "maximum" max msg size as INT_MAX. Shouldn't that be closer to SIZE_MAX, instead?

This seems strange to me, because larger messages do work, providing `zmq_msg` methods are used. I made my patch because messages larger than this were working fine, but reporting a false failure. This would make it impossible to ever work, I think.

hintjens:
OK, so this change is unfinished. The `s_sendmsg` truncates to INT_MAX, so that is the default I used for MAX_MSGSZ. There seems no point defaulting to a larger (>2Gb) size. 

Why do you think it would never work?

minrk:
> Why do you think it would never work?

INT_MAX seems a fine default, but the docs added here state that it's not just the default, it's also the max allowed value:

> The 'ZMQ_MAX_MSGSZ' argument sets the maximum allowed size
> of a message sent in the context.
> 
> Default value:: INT_MAX
> Maximum value:: INT_MAX


I read that as:

- attempting to send a message larger than ZMQ_MAX_MSGSZ will fail
- ZMQ_MAX_MSGSZ may not be larger than INT_MAX

Therefore:

- messages may not ever be larger than INT_MAX

Is that wrong? Are you only meaning to set the truncation for `rc`, but not actually checking and failing in send/recv?

hintjens:
You're right, the code isn't correct. I misread it. So I want to change the problem statement to "unlimited message size is an insecure default." In that case a realistic max size is indeed SIZE_MAX.

I need someone to help me access the parent context from a socket_base and then I can finish this...

Thanks for the explanations.

minrk:
I noticed that there is already a ZMQ_MAXMSGSIZE socket option, which is applied to receiving messages. Perhaps that should be extended to sends, rather than adding a new context option.

hintjens:
Good catch. This is a socket option, whereas I'd started making a context
option.

OK, I'll scratch my patch and look at using the socket ZMQ_MAXMSGSIZE.

PR #1790, opened 2016-02-11
Problem: use of libsodium vs. tweetnacl is confused

hintjens:
It's unclear which we need and in the source code, conditional code
treats tweetnacl as a subclass of libsodium, which is inaccurate.

Solution: redesign the configure/cmake API for this:

* tweetnacl is present by default and cannot be enabled
* libsodium can be enabled using --with-libsodium, which replaces
  the built-in tweetnacl
* CURVE encryption can be disabled entirely using --enable-curve=no

The macros we define in platform.hpp are:

    ZMQ_HAVE_CURVE    1        //  When CURVE is enabled
    HAVE_LIBSODIUM    1        //  When we are using libsodium
    HAVE_TWEETNACL    1        //  When we're using tweetnacl (default)

As of this patch, the default build of libzmq always has CURVE
security, and always uses tweetnacl.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,8 @@ test_udp
 test_large_msg
 test_use_fd_ipc
 test_use_fd_tcp
+test_pre_allocated_fd_ipc
+test_pre_allocated_fd_tcp
 tests/test*.log
 tests/test*.trs
 src/platform.hpp*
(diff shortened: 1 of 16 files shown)


.gitignore
```diff
@@ -127,6 +127,8 @@ test_udp
 test_large_msg
 test_use_fd_ipc
 test_use_fd_tcp
+test_pre_allocated_fd_ipc
+test_pre_allocated_fd_tcp
```
bluca: These 2 don't exist anymore, they were renamed to test_use_fd_tcp :-)

hintjens: Ah, ja... could you send a PR?

PR #1794, opened 2016-02-11
Problem: tweetnacl sources are a mess

hintjens:
- they have no copyright / license statement
- they are in some randomish directory structure
- they are a mix of postable and non-portable files
- they do not conform to conditional compile environment

Overall, it makes it rather more work than needed, in build scripts.

Solution: clean up tweetnacl sauce.

- merged code into single tweetnacl.c and .h
- standard copyright header, DJB to AUTHORS
- moved into src/ along with all other source files
- all system and conditional compilation hidden in these files
- thus, they can be compiled and packaged in all cases
- ZMQ_USE_TWEETNACL is set when we're using built-in tweetnacl
- HAVE_LIBSODIUM is set when we're using external libsodium


bluca:
I'm not a lawyer so naive question: is it possible to change the license from public domain to LGPL3 without the author's permission?

hintjens:
Haha, very fun... oh, you're serious?

The original work and license are still available from the original authors. We, ZeroMQ, when we modify a work in the public domain, we create a combined work. By default that is proprietary, and so previous versions of tweetnacl in our repository were not properly licensed. Or, they fell under the root COPYING.LESSER (which is probably the case).

We can license our combined work under whatever terms each constituent piece allows. Public domain allows *any* license on top.

bluca:
Ah ok, makes sense. As I said, I am not a lawyer :-)

PR #1795, opened 2016-02-11
Problem: CI is borken

bluca:
Solution: unbork it, at least partially. CMake build is still borked, but wanted to submit this while I figure that one out.

Also I'm probably going to add an extra "dimension" to the CI build array, so that we can have tweetnacl and libsodium builds for everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ os:
 env:
 - BUILD_TYPE=default
 - BUILD_TYPE=android
-- BUILD_TYPE=tweetnacl
+- BUILD_TYPE=libsodium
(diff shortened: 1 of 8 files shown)


hintjens:
Yay.

I'd honestly drop CI for libsodium builds since it doesn't add anything of value, and does take time. If we find a use case where we need that, we can add it again.

bluca:
Well the builds are done in parallel (per platform), so it doesn't add to the runtime :-)

Distributions maintainers might decide to use libsodium instead of tweetnacl, until we find out what the plans are for the major ones I would suggest to keep it in the CI, just in case.

PR #1796, opened 2016-02-11
Problem: typo in CMake platform.hpp breaks build

bluca:
Solution: use ZMQ_USE_TWEETNACL as a define as expected instead of
HAVE_TWEETNACL

With this, the build is 100% unborken. Yay! :-)


hintjens:
:-) do you know gyp by any chance?

bluca:
Nope, sorry, never used node.js :-(

PR #1797, opened 2016-02-11
Problem: CI use of sodium or nacl is confusing

bluca:
Solution: establish a matrix of CI options. On one axis we have the
build system (autotools, cmake, android) and on the other axis we
have the encryption options (tweetnacl, libsodium or none).

So this *might* be *slightly* overkill :-) But I'd say let's give it a shot! In most cases, unless Travis is horribly overloaded, builds should start in parallel.

If it turns out it's too slow, we can simply modify .travis.yml and take out some of the ENV lines, for example by doing just the default build on CMake and Android.


jemc:
FWIW - I've started using CircleCI instead of TravisCI these days, because TravisCI gets so darn slow as soon as you start getting a bigger matrix or a longer build - I think they throttle your speed in some way by how much work your builds usually take.

I've found CircleCI to be much faster, maybe because there is less overall load on their infrastucture and they can afford to be more attentive with their task scheduling.

hintjens:
I'll add my 10c here again: I think it's overkill to be testing with
sodium, and without curve. We aren't solving real issues here, and these
config options have no impact on the code we're writing. Also, for upstrem
CI (czmq, etc.) I think we should always build with curve (tweetnacl).
There is no benefit afaics in not doing so. (People who want to strip this
out can do so.)

Remember that every thing you make affects others, and CI is already way
too complex. We need to be simplifying it.

(I'd like e.g. to be building and testing bindings, which is much more
profitable.)


On Fri, Feb 12, 2016 at 12:33 AM, Joe Eli McIlvain <notifications@github.com

bluca:
@jemc - Thanks for the pointer, never heard of that system, I'll have a look!

@hintjens - The one benefit of having a CI run with sodium and without curve, is to make sure that the build system is not broken. It's very easy to break CMake/Autotools by mistake, and without CI it will go unnoticed until the changes piles up and figuring out the problem will be much more work.

What about, as a compromise, to only test the default build with sodium and no curve? That way the number of builds will go from 18 (which is definitely overkill as stated in the original comment :-) ) down to 10.

Regarding upstream, tomorrow I'll push a PR to CZMQ to switch to Tweetnacl (I will not add the other systems there).

hintjens:
Long email coming...

Default build should be out of the box, so curve via tweetnacl. We can then
test various options. If we're serious about coverage we should be testing
all of openpgm, libsodium, gssapi, libnorm, and tipc as these are the
addons libzmq supports. Otherwise I'd say, stick to the default build
(tweetnacl/curve) and let's aim for max coverage of operating systems and
platforms.

Note that for any given stack (e.g. JNI over Zyre) we have a N-dimensional
target matrix: (Linux, OS/X, Windows, Android) x (x86, x64, ARM) x (ABIs).

Right now the trend is towards building the entire matrix and uploading the
results somewhere. This will lead us to thousands of binary packages that
we're constantly rebuilding as masters change.

I think the longer term vision has to be different. We need to be able to
test A and B separately and be 100% sure that A+B will work consistently.
This is in effect what we do with libsodium and tweetnacl. The two have the
same API, and if libsodium passes its test, we are 100% sure (or close)
that it will work identically to tweetnacl in libzmq.

Same for portability issues. If code is 100% portable (e.g. Zyre) then if
CZMQ builds on Android, so will Zyre. No need to CI Zyre over Android.

This cuts the rebuild surface down, so we can reasonably build permanently
on all masters.

Then in the long term (I'm thinking several years away) we can build and
deliver packages on demand, for any given target, with appropriate caching.

This problem is so huge I think it's what a next generation git platform
should provide. Something like npm is doing for node.js.

In the shorter term, I think we could profit by bringing together CI builds
and command line builds. It's fragmented atm.

With the node.js bindings I'm developing the strategy of fetching dependent
projects on demand, otherwise leaving the source code that's already there.
This means you can build zyre alone (automatically fetching czmq, libzmq
from github), or build it with a locally patched czmq, libzmq. It means the
CI and command-line builds work the same way.

Rant over. :)

-Pieter

bluca:
> Default build should be out of the box, so curve via tweetnacl. We can then
> test various options. If we're serious about coverage we should be testing
> all of openpgm, libsodium, gssapi, libnorm, and tipc as these are the
> addons libzmq supports. Otherwise I'd say, stick to the default build
> (tweetnacl/curve) and let's aim for max coverage of operating systems and
> platforms.
> Note that for any given stack (e.g. JNI over Zyre) we have a N-dimensional
> target matrix: (Linux, OS/X, Windows, Android) x (x86, x64, ARM) x (ABIs).

I agree that the default build should be out-of-the-box, and that's the current status.
Unfortunately I don't think any free CI system will out-of-the-box offer anything other than x86, at least that I know of.

We could try to setup cross-compiling, but that's a monumental pain to do. Some coverage is given by Debian, that builds virtually every existing architecture: https://buildd.debian.org/status/package.php?p=zeromq3&suite=unstable

But that of course only covers maintainers upload of releases, so it's not really a development CI.

> Right now the trend is towards building the entire matrix and uploading the
> results somewhere. This will lead us to thousands of binary packages that
> we're constantly rebuilding as masters change.

That is true. But it's also all for free, and does binaries all go to /dev/null once the build is finished, so there's no cost for the thousands of binaries :-)

> I think the longer term vision has to be different. We need to be able to
> test A and B separately and be 100% sure that A+B will work consistently.
> This is in effect what we do with libsodium and tweetnacl. The two have the
> same API, and if libsodium passes its test, we are 100% sure (or close)
> that it will work identically to tweetnacl in libzmq.

The only problem I see with this is that A and B need to be glued together, and that glue can break. I'm talking about the autotools/cmake system. As we've seen times and times again, it's extremely easy to break those.

If we don't notice, the problems pile up. I think there's a lot of value in being able to test via CI that the configuration and build system is not broken. Unfortunately at the moment there is no other way to do that other than just building the thing.

It would be great if we could unit-test autools and cmake, but I've never seen it done and I wouldn't know where to begin :-(

> Same for portability issues. If code is 100% portable (e.g. Zyre) then if
> CZMQ builds on Android, so will Zyre. No need to CI Zyre over Android.

But how do you know that portability of Zyre is not broken, if Zyre is not tested on other platforms? :-)

> This cuts the rebuild surface down, so we can reasonably build permanently
> on all masters.
> 
> Then in the long term (I'm thinking several years away) we can build and
> deliver packages on demand, for any given target, with appropriate caching.
> 
> This problem is so huge I think it's what a next generation git platform
> should provide. Something like npm is doing for node.js.
> 
> In the shorter term, I think we could profit by bringing together CI builds
> and command line builds. It's fragmented atm.
> 
> With the node.js bindings I'm developing the strategy of fetching dependent
> projects on demand, otherwise leaving the source code that's already there.
> This means you can build zyre alone (automatically fetching czmq, libzmq
> from github), or build it with a locally patched czmq, libzmq. It means the
> CI and command-line builds work the same way.

But the CI and the release build systems, even if they can and should share most of the bits and pieces, will never be able to 100% overlap. Indeed I do not believe it would be desirable to do so - they serve different purposes and run in different environment. I do not think it would be an advantage to scale CI testing down to the level of release builds. Coverage and testing would be lost, and the only gain would be having a few 20 lines scripts less to maintain, which in my opinion is not enough to balance the scale.

If anything, I would like to do more in the CI! I am working on making libzmq Valgrind-friendly. At the moment, in the slower CI environment, there's too many race conditions in the tests and tons of false positives. I'm almost through cleaning those.

Another thing I would like to have is static analysis. Haven't started yet, but cppcheck is a good place tool for that. And then code coverage would be useful too.

All rigorously IMHO of course :-)

bluca:
Having said that, opened https://github.com/zeromq/libzmq/pull/1803 to scale a bit down :-)

PR #1805, opened 2016-02-16
Problem: curve keys getsockopt uninitialised read

bluca:
Solution: always initialised zmq::options_t class variables arrays to
avoid reading uninitialised data when CURVE is not yet configured and
a getsockopt ZMQ_CURVE_{SERVER | PUBLIC | SECRET]KEY is issued.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -78,6 +78,9 @@ zmq::options_t::options_t () :
     heartbeat_timeout (-1),
     use_fd (-1)
 {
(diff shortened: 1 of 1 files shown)


hintjens:
Does this risk an access violation? If so we should backport to 4.1.x stable.

bluca:
It's a corner case, given one has to use a curve sockopt when curve is not enabled, but I suppose better safe than sorry. I'll do the backport shortly.

PR #1809, opened 2016-02-17
Number of fixes for gyp builds

hintjens:
Also: raised socket limit to 16K on Windows in CMakeLists.txt

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,17 +101,9 @@ AC_RUN_IFELSE(
 
 AC_MSG_RESULT([$libzmq_tipc_support])
 
-AC_ARG_WITH([relaxed],
-    [AS_HELP_STRING([--with-relaxed],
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -434,7 +422,7 @@ AS_IF([test "x$with_libsodium" = "xyes"], [
 
 AC_ARG_ENABLE([curve],
     AS_HELP_STRING([--disable-curve], [disable CURVE security [default=no]]))
-
+x
```
bluca: What's this X? :-)

hintjens: Huh.

Fixed, commit sent.

hintjens: Oh, and thanks for catching that!

bluca:
:+1: especially on the pedantic warnings, it's good to have that enabled by default.

PR #1811, opened 2016-02-17
problem: dynamic cast is causing issue when compiling for nodejs

somdoron:
solution: remove dynamic cast


hintjens:
Thank you! :+1:

hintjens:
@zoobab can you pull libzmq master and try again on OS/X? It should work now.

PR #1814, opened 2016-02-18
Deprecate zmq_utils.h

minrk:
instead of removing it, which breaks downstream builds.


hintjens:
Thanks for fixing this.

PR #1817, opened 2016-02-19
Problem: tweetnacl gives many warnings on MSVC

hintjens:
Solution: disable the warnings on this file only

We use pragmas wrapped in compiler conditionals. This will need
extending to non-gcc/msvc compilers. We could also fix the warnings
in the code, though I suspect it's not really possible.

### DIFF
--- a/src/tweetnacl.c
+++ b/src/tweetnacl.c
@@ -30,7 +30,13 @@
 #include "platform.hpp"
 #if defined (ZMQ_USE_TWEETNACL)
 
-#pragma GCC diagnostic ignored "-Wsign-compare"
+//  Disable warnings for this source only, rather than for the whole
+//  codebase
+#ifdef __GNUC__
+#   pragma GCC diagnostic ignored "-Wsign-compare"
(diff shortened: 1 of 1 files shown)


bluca:
I would be wary of changing the types in the library to fix the warnings, given how small and efficient that crypto library is (and how critical to security, also), it's probably way too easy for subtle bugs to slip in :-)

What about, instead, bumping the C++ standard up? Those warnings should go away if we move from C++98 to 0x, I think. Are there platforms we support that do not have a C++0x compliant compiler yet?

hintjens:
What's involved in terms of compile switches? We can certainly try this.

bluca:
It's pretty easy in autoconf and cmake, as long as everything still builds. No idea about all the windows stuff though.

I'll have a play during the weekend.

sunkin351:
Windows supports the latest C++ standard, which is C++14, you should have no problem.

bluca:
Opened https://github.com/zeromq/libzmq/pull/1819

PR #1823, opened 2016-02-21
Problem: Several problems found by Coverity Static Analyzer

opedroso:
Solution: The Coverity Static Code Analyzer was used on libzmq code and found
many issues with uninitialized member variables, some redefinition of variables
hidding previous instances of same variable name and a couple of functions
where return values were not checked, even though all other occurrences were
checked (e.g. init_size() return).


c-rack:
Nice, thanks!
The diff showed some changes with the indention (tabs vs spaces), but we can fix that later.

opedroso:
Sorry,

I lost all my changes twice due to my git newbiness, so I must have gotten sloppy 

Second time lost my changes while attempting to build in Ubuntu.

Third time, I checked it after it built in Windows.

Sorry, it won't happen again.
Osiris

bluca:
Don't worry about it!

In git, even if a commit is no longer in any tree, can still be accessed if you know the hash (git log result still in the terminal, for example, or by using commands like git reflog)

PR #1825, opened 2016-02-22
Problem: build borken

### LINKED ISSUE #1824: master - make error,  conversion pointer/non-pointer (g++, Ubuntu 14.04)
latest "master", after success of generated ./configure, nonetheless, make exits with the error:
```
make[1]: Entering directory `/home/user/projects/libzmq'
  CXX      src/src_libzmq_la-address.lo
In file included from src/ctx.hpp:45:0,
                 from src/address.cpp:34:
src/thread.hpp: In constructor ‘zmq::thread_t::thread_t()’:
src/thread.hpp:60:21: error: converting to non-pointer type ‘pthread_t {aka long unsigned int}’ from NULL [-Werror=conversion-null]
    , descriptor(NULL)
                     ^
cc1plus: all warnings being treated as errors
make[1]: *** [src/src_libzmq_la-address.lo] Error 1
make[1]: Leaving directory `/home//user/projectslibzmq'
make: *** [all-recursive] Error 1
```

bluca:
Solution: unbork it

Whitespace fix (tab => 4 spaces), class variable initialization order, assert where it should not.
Fixes #1824 

@opedroso - if Coverity complains about unused return values, we can cast them to (void) to silence it where it makes sense not to check :-)

### DIFF
--- a/src/dish.cpp
+++ b/src/dish.cpp
@@ -333,14 +333,14 @@ int zmq::dish_session_t::pull_msg (msg_t *msg_)
 
         if (msg_->is_join ()) {
             rc = command.init_size (group_length + 5);
-			errno_assert(rc == 0);
-			offset = 5;
+            errno_assert(rc == 0);
+            offset = 5;
             memcpy (command.data (), "\4JOIN", 5);
         }
         else {
             rc = command.init_size (group_length + 6);
-			errno_assert(rc == 0);
-			offset = 6;
+            errno_assert(rc == 0);
+            offset = 6;
             memcpy (command.data (), "\5LEAVE", 6);
         }
 
--- a/src/encoder.hpp
+++ b/src/encoder.hpp
@@ -57,11 +57,11 @@ namespace zmq
     public:
 
         inline encoder_base_t (size_t bufsize_) :
-            bufsize (bufsize_),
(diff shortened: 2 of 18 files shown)


hintjens:
Also I dislike making code obscure and more complex just to satisfy tools
like Coverity. If it catches real issues, fine. If it's to complain about
e.g. unused return values, the correct fix is to add the exception to
Coverity, or ignore these complaints.

bluca:
:+1:

opedroso:
I would say it caught a few calls to init_size() that were not being checked for error which may be a problem in low memory condition. I agree that it is unlikely though. The missing member initializer are not a problem if variable is never touched prior to setting it first time. As it is today, they are not used before being set, but who knows in the future? Might as well initialize them all to a known quantity. Reason I ran Coverity is that all of the code we use in our project has to be Coverity clean, so I needed that. Thought it would make sense to check the results back in.

bluca:
Most of those were valid complaints, as you said. It's just the couple I removed that it's not worth fixing, as Pieter pointed out :-)

BTW on my TODO list there's adding cppcheck to the CI, and also adding libzmq to the free coverity scan. My TODO list is way too long :-)

PR #1826, opened 2016-02-22
Problem: make dist is borken

bluca:
Solution: unbork it


scemama:
@bluca : https://en.wikipedia.org/wiki/Borken,_Hesse

bluca:
@scemama - lol, not that borken :-D

PR #1827, opened 2016-02-22
Problem: make dist does not tar up macros.hpp

bluca:
Solution: add it to Makefile.am file list

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -85,6 +85,7 @@ src_libzmq_la_SOURCES = \
 	src/lb.cpp \
 	src/lb.hpp \
 	src/likely.hpp \
+	src/macros.hpp \
 	src/mailbox.cpp \
 	src/mailbox.hpp \
 	src/mailbox_safe.cpp \
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -28,7 +28,7 @@ if [ $BUILD_TYPE == "default" ]; then
         ./autogen.sh &&
         ./configure "${CONFIG_OPTS[@]}" &&
         make &&
-        ( if make check; then true; else cat test-suite.log; exit 1; fi ) &&
+        make VERBOSE=1 check &&
         make install
     ) || exit 1
 else
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -27,9 +27,8 @@ if [ $BUILD_TYPE == "default" ]; then
     (
         ./autogen.sh &&
         ./configure "${CONFIG_OPTS[@]}" &&
-        make &&
-        make VERBOSE=1 check &&
-        make install
+        export DISTCHECK_CONFIGURE_FLAGS="${CONFIG_OPTS[@]}" &&
+        make VERBOSE=1 distcheck
     ) || exit 1
 else
     cd ./builds/${BUILD_TYPE} && ./ci_build.sh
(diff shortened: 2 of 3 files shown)


bluca:
Note that same fix might have to be applied to all the Windows stuff, but haven't got the first clue how they work :-)

hintjens:
There's no make dist for Windows. I'm thinking of switching to github's
tarball releases for libzmq, simpler and less fragile.

sunkin351:
Well, if you do its a good thing windows archivers have the ability to open *.tar files...

bluca:
@hintjens - that would be easier for us indeed.

I think the only problem is that, as far as I understand, Github simply tars up the repository. On the other hand, make dist also generates everything a user need to be able to install by simply doing:

    ./configure
    make
    make install

If we stop shipping the make dist'ed tarball, users who rely on it will need to have the autotools build  toolchain (or cmake) installed and configured, and generate everything themselves. Which is not right or wrong per se, it's just the convention for Unix stuff and users.

Regardless, I think the lack of this file in automake is a bug anyway :-)

hintjens:
We can simply add the output of autogen.sh to git. I think `configure` is
the only significant difference between git and tarball, and it's always
been confusing to need that extra step when cloning a repo.

bluca:
I've seen projects check in git the output of autoconf and automake in the past, and unfortunately it invariably becomes very messy :-(

The problem is that, depending on OS, distribution and autotools version, you'll get a slightly different result. If the output is checked in, suddenly users and contributors will find big and unexpected diffs in their git tree by simply building, which is quite confusing especially for newcomers.
And then we'll also start getting PRs which change those files, by virtue of commit -a or add -A, which again is confusing for contributors I think. This will create merge/rebased conflicts for contributors WIP branches off master, which again might discourage them, as it will make the process more tedious and definitely less fun. Been there, done that :-)

I understand that keeping things like this in shape is a bit more pain for the mainteiners, but as shown by this and other PRs I'm more than happy to try and help :-)

Finally, once the problem addressed by this commit is fixed, were can switch the CI to do make dist-check, so that we don't have to manually check that it all works, which is boring, but let the CI do it, and we avoid bit rot.

hintjens:
Hang on... Tarballs all contain configure. Git repo doesn't. Are you saying
tarballs are not portable? This is the only file I was thinking of adding.

bluca:
They are portable, but the generated configure script will be different (but compatible and portable) depending on the versions of the toolchain components.

For example, as a quick test I ran autogen.sh on my Debian Jessie system, and committed it in git. That was with autoconf 2.69-8, automake 1.14.1-4 and libtool 2.4.2-1.11.

Then I copied the tree into a Debian Sid chroot with autoconf 2.69-9, automake 1.15-4 and libtool 2.4.6-0.1, and ran autogen.sh again, and there's already a huge diff:

    # git diff --stat  
     configure | 3125 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----------------------------------------
     1 file changed, 1841 insertions(+), 1284 deletions(-)

And that's in the same distribution, just with a slightly more recent version of the autotools (and a nearly identical autoconf I might add)! Across distributions, or even operating systems, things might be even worse :-(

hintjens:
Ah, OK. And no point checking in files that change all the time, it makes a
real mess.. Thanks

bluca:
Indeed :-)

Pushed a couple more commits. One more fix in doc/Makefile.am, and then modified the CI script to run make distcheck (which checks everything, including make dist and the unit tests).

bluca:
@hintjens - is there anything else you'd like me to do for this PR? :-)

hintjens:
Ooops, thanks for reminding me. :)

bluca:
:+1:

PR #1832, opened 2016-03-01
Problem: can't be sure crypto_box always returns 0

### LINKED ISSUE #1831: Problem: Assertion failed: rc == 0 (src/curve_client.cpp:273)
This happens in the client stack if you specify a null (all zeros) server key, and if you are using libsodium stable. Using older versions of libsodium or tweetnacl, this assert does not happen.

The cause is a change in the API where crypto_box () returns -1 on a null server key, whereas it used to ignore this error (as it ignores other invalid server keys).

Solution: do not assert on this error, handle it gracefully.

hintjens:
Libsodium has started returning -1 in some cases.

Solution: allow and handle error returns from these calls.

Fixes #1831

### DIFF
--- a/src/curve_client.cpp
+++ b/src/curve_client.cpp
@@ -202,7 +202,6 @@ int zmq::curve_client_t::decode (msg_t *msg_)
     }
     cn_peer_nonce = nonce;
 
-
     const size_t clen = crypto_box_BOXZEROBYTES + (msg_->size () - 16);
 
     uint8_t *message_plaintext = static_cast <uint8_t *> (malloc (clen));
@@ -270,7 +269,8 @@ int zmq::curve_client_t::produce_hello (msg_t *msg_)
     int rc = crypto_box (hello_box, hello_plaintext,
                          sizeof hello_plaintext,
                          hello_nonce, server_key, cn_secret);
-    zmq_assert (rc == 0);
+    if (rc == -1)
+        return -1;
 
     rc = msg_->init_size (200);
     errno_assert (rc == 0);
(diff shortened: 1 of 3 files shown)


bluca:
Nice catch!

Is it worth backporting to the stable releases?

hintjens:
Yes, I'll do that.

... done for 4.1.x.

bluca:
:+1:

photoszzt:
@hintjens I can build with autogen/configure and pass all the tests with libsodium enable. But 10 tests fail in cmake build. My cmake configuration is: 

cmake .. -DWITH_LIBSODIUM=ON -DSODIUM_INCLUDE_DIRS=$PREFIX/include -DCMAKE_INSTALL_PREFIX=$PREFIX 

The failed tests are: 

89% tests passed, 10 tests failed out of 89

Total Test time (real) =  51.71 sec

The following tests FAILED:
	 36 - test_req_correlate (OTHER_FAULT)
	 37 - test_req_relaxed (OTHER_FAULT)
	 59 - test_pair_tipc (OTHER_FAULT)
	 60 - test_reqrep_device_tipc (OTHER_FAULT)
	 61 - test_reqrep_tipc (OTHER_FAULT)
	 63 - test_router_mandatory_tipc (OTHER_FAULT)
	 66 - test_sub_forward_tipc (OTHER_FAULT)
	 82 - test_connect_delay_tipc (OTHER_FAULT)
	 83 - test_shutdown_stress_tipc (OTHER_FAULT)
	 86 - test_term_endpoint_tipc (OTHER_FAULT)
Errors while running CTest
make: *** [test] Error 8

Failure output: https://gist.github.com/photoszzt/2aa00bbfda6e64979544

hintjens:
Ah, this is an unrelated issue IMO, caused by the CMake file setting a time
limit on each test.

PR #1834, opened 2016-03-02
minor automake fixes plus tweetnacl logic change

garlick:
Happy to rebase this without the tweetnacl logic change, which was not enthusiastically received on mailing list.  Submitting with it included first just to give the idea one more chance.  Hope it doesn't seem like I'm trying to sneak one in through the back door.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -430,7 +430,6 @@ AC_ARG_ENABLE([curve],
 if test "x$enable_curve" == "xno"; then
     curve_library=""
     AC_MSG_NOTICE([CURVE security is disabled])
-    AM_CONDITIONAL(ENABLE_CURVE_KEYGEN, 0)
 
 elif test "x$with_libsodium" == "xyes"; then
     AC_MSG_NOTICE([Using libsodium for CURVE security])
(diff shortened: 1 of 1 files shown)


hintjens:
I can't agree with the problem "Too easy to select tweetnacl"... there is no proven disadvantage to this, and many known costs to not building in security by default. 

I'll merge this if you want yet the first thing I'll do tomorrow is fix the problem "libzmq is insecure by default"... :)

garlick:
I'll force push without the offending commit in a sec but wanted to respond to your comments.

My concern about tweetnacl is not the code (though lack of upstream repo is concerning).
I  provides libzmq RPM packages to several US national labs that run our in-house redhat-based distro.  I worry about
* accidentally packaging libzmq with tweetnacl not libsodium, pushing that out (low probability event)
* we become aware of security issue with tweetnacl (2nd low probabilty event)
* I have to push out emergency updates and get a big WTF from my peers (big ouch)

Roger the cmake concern, was only dimly aware that libzmq had grown an extra build system.

My proposed change does make a hard stop if libsodium is not found, e.g.
```
configure: error: libsodium is not installed.  Install it, then run
    configure again, or configure --with-builtin-tweetnacl or --disable-curve.
```
so this doesn't really break "insecure by default".

garlick:
OK that commit is gone.  Thanks.

hintjens:
Thanks!

PR #1847, opened 2016-03-09
Problem: test_large_msg requires 2GB of free RAM

### LINKED ISSUE #1846: Failing tests. Ubuntu 14
Whilst compiling libzmq on Ubuntu 14.04.4 x64 I noticed 1 FAIL and 2 XFAIL's.

Here's the `test-suite.log` file:

```
====================================
   zeromq 4.2.0: ./test-suite.log
====================================

# TOTAL: 80
# PASS:  77
# SKIP:  0
# XFAIL: 2
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: tests/test_large_msg
==========================

lt-test_large_msg: tests/test_large_msg.cpp:53: int main(): Assertion `rc == 0' failed.

XFAIL: tests/test_req_correlate
===============================

lt-test_req_correlate: tests/testutil.hpp:246: void s_recv_seq(void*, ...): Assertion `rc != -1' failed.

XFAIL: tests/test_req_relaxed
=============================

lt-test_req_relaxed: tests/testutil.hpp:246: void s_recv_seq(void*, ...): Assertion `rc != -1' failed.
```

bluca:
Solution: scale down the test to use only 32 MB. While it will no
longer fulfill the original purpose of checking the truncation of the
return value, it will still check a fairly large message without
causing issues on low power build machines, embedded or CI
environments.
Fixes #1846 

I propose to scale down this test. We keep getting bug reports and issues because it currently needs 2GB of ram, which most VM/build systems/CI systems/embedded systems just don't have.
The downside is that we lose coverage of the INT_MAX return size truncation.

What do you think?


hintjens:
The original test was to check a precise case which IMO we can fix differently, by setting a limit on message sizes that is rather less than MAX_INT. In which case the test should be to set the message limit to various sizes and check that it's applied. 

I had started on something in that direction and got distracted. I suggest this test be removed temporarily until we have limits properly implemented.

bluca:
ACK. Update to remove the test instead.

bluca:
@hintjens - ping! :-) Do you want me to drop this?

hintjens:
Sorry it fell off my radar...

bluca:
No worries! Would you like help for the all alternative solution? What did you have in mind exactly?

hintjens:
Max message size, set per context, then applied on all socket recvs and sends from the API. Defaults to something large yet sane like 32MB.

hitstergtd:
Would it be better to make the code conditional on available RAM instead so at least some coverage is retained instead of totally blitzing the test? I.e. those who can test it where RAM isn't an issue?

bluca:
Sorry, haven't got around to implement @hintjens solution yet. Will try to do it ASAP so we have coverage.

Unfortunately I'm not sure detecting memory can be done reliably everywhere.

sunkin351:
http://stackoverflow.com/questions/2513505/how-to-get-available-memory-c-g

Methods for getting the available system memory in UNIX and Windows can be found on the second answer on the stack overflow thread above. Hope this helps.

bluca:
@sunkin351 - thanks for the pointers! Unfortunately using sysconf won't help when running under Valgrind, which has additional constraints. IMHO losing memory leaks detection would be worse than losing coverage of this corner case of a single 2 GB+ message. What do you think?

@hintjens - I had a brief look, and I remembered there was already a ZMQ_MAXMSGSIZE which affetcs the max size of a RECEIVED message.
What about aliasing that to ZMQ_RCV_MAXMSGSIZE (for backward compat) and adding a ZMQ_SND_MAXMSGSIZE socket option, in line with how we have a SND/RCV_HWN/TIMEO/etc ? And the CTX option you added maybe could be used to set the global default for both.

hitstergtd:
@bluca - I will look into creating cross-platform code for memory detection. At least so long as we are testing our core platforms (Linux, *BSD, Windows) than that should be fine. Ultimately, this is a test, so we are afforded some latitude I would have thought. None-core platforms can be added as and when. Thoughts?

PR #1851, opened 2016-03-13
Add code coverage to build system and CI

bluca:
This PR adds code coverage support via gcov/lcov. A new make target, make check-code-coverage, is added by importing a standard autotools m4 from autotools-archive, and it will create a nice html report locally. We currently sit a 74%, which is not too bad :-)

I propose we also use coveralls.io for a nice and pretty web gui. This allows to track coverage over time, and let us see how we progress. @hintjens - if you agree, I can turn on support on their website. Usage is free for open source projects, like Github and Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ env:
 - BUILD_TYPE=cmake CURVE=tweetnacl
 - BUILD_TYPE=default CURVE=libsodium
 - BUILD_TYPE=default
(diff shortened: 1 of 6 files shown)


hintjens:
Nice stuff. Give it all you got, Scotty!

bluca:
:-) There we go: https://coveralls.io/github/zeromq/libzmq
Badge in the readme.md (and Github home) also works and shows the current coverage %

PR #1852, opened 2016-03-14
Update Unix Domain Socket creation/handling

Suudy:
Updated handling of Unix Domain Sockets, make use of temporary directories, and cleanup afterward.

### DIFF
--- a/src/ipc_listener.cpp
+++ b/src/ipc_listener.cpp
@@ -49,6 +49,7 @@
 #include <sys/socket.h>
 #include <fcntl.h>
 #include <sys/un.h>
+#include <sys/stat.h>
 
 #ifdef ZMQ_HAVE_LOCAL_PEERCRED
 #   include <sys/types.h>
@@ -62,6 +63,58 @@
 #       define ucred sockpeercred
 #   endif
 #endif
+
+const char *zmq::ipc_listener_t::tmp_env_vars[] = {
+  "TMPDIR",
+  "TEMPDIR",
+  "TMP",
+  0  // Sentinel
+};
+
+int zmq::ipc_listener_t::create_wildcard_address(std::string& path_)
+{
+    std::string tmp_path;
+
+    // If TMPDIR, TEMPDIR, or TMP are available and are directories, create
+    // the socket directory there.
+    const char **tmp_env = tmp_env_vars;
+    while ( tmp_path.empty() && *tmp_env != 0 ) {
+        char *tmpdir = getenv(*tmp_env);
+        struct stat statbuf;
+
+        // Confirm it is actually a directory before trying to use
+        if ( tmpdir != 0 && ::stat(tmpdir, &statbuf) == 0 && S_ISDIR(statbuf.st_mode) ) {
+            tmp_path.assign(tmpdir);
+            if ( *(tmp_path.rbegin()) != '/' ) {
+                tmp_path.push_back('/');
+            }
+        }
+
+        // Try the next environment variable
+        ++tmp_env;
+    }
+
+    // Append a directory name
+    tmp_path.append("tmpXXXXXX");
+
+    // We need room for tmp_path + trailing NUL
+    std::vector<char> buffer(tmp_path.length()+1);
+    strcpy(buffer.data(), tmp_path.c_str());
+
+    // Create the directory.  POSIX requires that mkdtemp() creates the
+    // directory with 0700 permissions, meaning the only possible race
+    // with socket creation could be the same user.  However, since
+    // each socket is created in a directory created by mkdtemp(), and
+    // mkdtemp() guarantees a unique directory name, there will be no
+    // collision.
+    if ( mkdtemp(buffer.data()) == 0 ) {
+        return -1;
+    }
+
+    path_.assign(buffer.data());
+
+    return 0;
+}
 
 zmq::ipc_listener_t::ipc_listener_t (io_thread_t *io_thread_,
       socket_base_t *socket_, const options_t &options_) :
@@ -113,7 +166,7 @@ void zmq::ipc_listener_t::in_event ()
     io_thread_t *io_thread = choose_io_thread (options.affinity);
     zmq_assert (io_thread);
 
-    //  Create and launch a session object. 
+    //  Create and launch a session object.
     session_base_t *session = session_base_t::create (io_thread, false, socket,
         options, NULL);
     errno_assert (session);
@@ -148,12 +201,15 @@ int zmq::ipc_listener_t::set_address (const char *addr_)
 
     //  Allow wildcard file
     if (addr [0] == '*') {
-        char buffer [12] = "2134XXXXXX";
-        int fd = mkstemp (buffer);
-        if (fd == -1)
+        std::string tmp_path;
+
+        if ( create_wildcard_address(tmp_path) < 0 ) {
             return -1;
-        addr.assign (buffer);
-        ::close (fd);
+        }
+
+        tmp_socket_dirname.assign(tmp_path);
+
+        addr.assign (tmp_path + "/socket");
     }
 
     //  Get rid of the file associated with the UNIX domain socket that
@@ -169,8 +225,16 @@ int zmq::ipc_listener_t::set_address (const char *addr_)
     //  Initialise the address structure.
     ipc_address_t address;
     int rc = address.resolve (addr.c_str());
-    if (rc != 0)
+    if (rc != 0) {
+        if ( !tmp_socket_dirname.empty() ) {
+            // We need to preserve errno to return to the user
+            int errno_ = errno;
+            ::rmdir(tmp_socket_dirname.c_str ());
+            tmp_socket_dirname.clear();
+            errno = errno_;
+        }
         return -1;
+    }
 
     address.to_string (endpoint);
 
@@ -179,8 +243,16 @@ int zmq::ipc_listener_t::set_address (const char *addr_)
     } else {
         //  Create a listening socket.
         s = open_socket (AF_UNIX, SOCK_STREAM, 0);
-        if (s == -1)
+        if (s == -1) {
+            if ( !tmp_socket_dirname.empty() ) {
+                // We need to preserve errno to return to the user
+                int errno_ = errno;
+                ::rmdir(tmp_socket_dirname.c_str ());
+                tmp_socket_dirname.clear();
+                errno = errno_;
+            }
             return -1;
+        }
 
         //  Bind the socket to the file path.
         rc = bind (s, address.addr (), address.addrlen ());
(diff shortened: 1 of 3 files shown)


bluca:
Hi,

Thanks for the contribution!

Could you please fix the formatting (indentation should be 4 spaces)?

Also one of the test is failing:

lt-test_term_endpoint: ../tests/test_term_endpoint.cpp:148: int main(): Assertion `rc == 0' failed.
/bin/bash: line 5: 35277 Aborted                 (core dumped) ${dir}$tst
FAIL: tests/test_term_endpoint

I suspect it's due to the changes in zmq::ipc_listener_t::set_address from return -1 to goto error;

src/ipc_listener.cpp
```diff
@@ -208,9 +260,13 @@ int zmq::ipc_listener_t::set_address (const char *addr_)
 
 int zmq::ipc_listener_t::close ()
 {
-    zmq_assert (s != retired_fd);
-    int rc = ::close (s);
-    errno_assert (rc == 0);
+    int rc;
+
+    if ( s != -1 ) {
+      zmq_assert (s != retired_fd);
```
bluca: retired_fd is -1, so this check doesn't make much sense. Why the added "if {}" block?

Suudy: On Mon, Mar 14, 2016 at 9:08 AM, Luca Boccassi <notifications@github.com> wrote:
>
> In src/ipc_listener.cpp:
>
> > @@ -208,9 +260,13 @@ int zmq::ipc_listener_t::set_address (const char *addr_)
> >
> >  int zmq::ipc_listener_t::close ()
> >  {
> > -    zmq_assert (s != retired_fd);
> > -    int rc = ::close (s);
> > -    errno_assert (rc == 0);
> > +    int rc;
> > +
> > +    if ( s != -1 ) {
> > +      zmq_assert (s != retired_fd);
>
> retired_fd is -1, so this check doesn't make much sense. Why the added "if {}" block?

I'm trying to handle the case where the directory for the socket
failed to be created (i.e. mkdtemp() failed), then s would remain -1.
This would cause the assertion error "s != retired_fd" and casue
::close to return an error with errno set to EBADF.  Though I suppose
this is also the case if open_socket() failed as well.  So it seems
that in the case of either mkdtemp() or open_socket() failing, we
don't want the assertion or ::close() failure here.

Pete

Suudy:
On Mon, Mar 14, 2016 at 9:04 AM, Luca Boccassi <notifications@github.com> wrote:
> Could you please fix the formatting (indentation should be 4 spaces)?

I've fixed that on my branch.  I'll send a new pull request once I've
cleaned up the test failure and after we've discussed the change in
ipc_listener_t::close().

> Also one of the test is failing:
>
> lt-test_term_endpoint: ../tests/test_term_endpoint.cpp:148: int main():
> Assertion `rc == 0' failed.
> /bin/bash: line 5: 35277 Aborted (core dumped) ${dir}$tst
> FAIL: tests/test_term_endpoint
>
> I suspect it's due to the changes in zmq::ipc_listener_t::set_address from
> return -1 to goto error;

Hmm....  Looking at the test, this shouldn't be the case.  My change
should still return -1, but only after the close.  Let me dig into it.

Pete

bluca:
Thanks for looking into it!

In case you're not too familiar with Github, you can amend your commit and push -f to the same branch, and this PR will get updated automatically.

Suudy:
On Mon, Mar 14, 2016 at 9:04 AM, Luca Boccassi <notifications@github.com> wrote:
> lt-test_term_endpoint: ../tests/test_term_endpoint.cpp:148: int main(): Assertion `rc == 0' failed.
> /bin/bash: line 5: 35277 Aborted (core dumped) ${dir}$tst
> FAIL: tests/test_term_endpoint

This failure is multifaceted.  The root cause of the failure is that
zmq_getsockopt() modifies the optvallen parameter, and it is note
restored before each call.  The second problem is that the variable
passed to the zmq_getsockopt() is declared 'const size_t' even though
it is passed as non-const (which then modifies the value).

I think the reason they were passing before and began failing with my
commit was due to the fact that the AF_UNIX sockets are now longer in
length.  And the previouly modified optvallen was large enough.  My
fix restores buf_size to the actual buffer size before each call to
zmq_getsockopt().

I've fixed test_term_endpoint.  However, I have two XFAIL's that I
can't quite figure out what is going on.  They are in
test_req_correlate and test_req_relaxed.  And looking at the code for
those, they seem completely unrelated to any of the changes I've made.
I reverted all my changes, and these same tests fail.

Suudy:
On Mon, Mar 14, 2016 at 9:58 AM, Luca Boccassi <notifications@github.com> wrote:
> In case you're not too familiar with Github, you can amend your commit and push -f to the same branch, and this PR will get updated automatically.

I'm not git expert, but I did manage two squash my commits into a
single commit and pushed them up to github.  It looks like the pull
request has the proper changes.

Pete

src/ipc_listener.cpp
bluca: Yes, and that's why, I think, the code as-is did a return -1 instead of goto err if open_socket failed.

We need the assert to enforce that this API is not used improperly, IE: called on a socket that's already closed. Do you think it's possible to leave that in with your changes?

Suudy: On Mon, Mar 14, 2016 at 10:51 AM, Luca Boccassi
<notifications@github.com> wrote:
> Yes, and that's why, I think, the code as-is did a return -1 instead of goto err if open_socket failed.
>
> We need the assert to enforce that this API is not used improperly, IE: called on a socket that's already closed. Do you think it's possible to leave that in with your changes?

I may need to rework it a bit.  What I was trying to do was cleanup
the mkdtemp() on an error.  I thought that re-using
ipc_listener_t::close() was the best way to do that.

I can revert ipc_listener_t::close(), and update
ipc_listener_t::set_address() to properly cleanup the directory if
open_socket() fails.  I understand your API checks, and I'll fix
things on my end.

Are the prohibitions on the use of exceptions?  I initially coded it
to throw an exception in create_wildcard_address(), but didn't see
handling of exceptions elsewhere.  If exceptions are ok, I can rework
set_address() to make sure that there are no stray files or
directories upon failure pretty easily.  Otherwise, the code may look
a bit ugly.

Pete

bluca:
On 14 March 2016 at 17:47, Suudy <notifications@github.com> wrote:

> On Mon, Mar 14, 2016 at 9:04 AM, Luca Boccassi <notifications@github.com>
> wrote:
> > lt-test_term_endpoint: ../tests/test_term_endpoint.cpp:148: int main():
> Assertion `rc == 0' failed.
> > /bin/bash: line 5: 35277 Aborted (core dumped) ${dir}$tst
> > FAIL: tests/test_term_endpoint
>
> This failure is multifaceted. The root cause of the failure is that
> zmq_getsockopt() modifies the optvallen parameter, and it is note
> restored before each call. The second problem is that the variable
> passed to the zmq_getsockopt() is declared 'const size_t' even though
> it is passed as non-const (which then modifies the value).
>
> I think the reason they were passing before and began failing with my
> commit was due to the fact that the AF_UNIX sockets are now longer in
> length. And the previouly modified optvallen was large enough. My
> fix restores buf_size to the actual buffer size before each call to
> zmq_getsockopt().
>

It does seem a bit forced, yes. Thanks for taking care of it.


> I've fixed test_term_endpoint. However, I have two XFAIL's that I
> can't quite figure out what is going on. They are in
> test_req_correlate and test_req_relaxed. And looking at the code for
> those, they seem completely unrelated to any of the changes I've made.
> I reverted all my changes, and these same tests fail.


Yes you can ignore those 2: XFAIL in autotools means it's marked as an
expected failure. We marked them as such a while ago, sorry for the
confusion!

src/ipc_listener.cpp
bluca: > I may need to rework it a bit.  What I was trying to do was cleanup
> the mkdtemp() on an error.  I thought that re-using
> ipc_listener_t::close() was the best way to do that.

If you need more time, I can merge as-is for now, and then you can send another PR to fix it when it's ready. Or maybe it's easier for you to work on the same branch?

> I can revert ipc_listener_t::close(), and update
> ipc_listener_t::set_address() to properly cleanup the directory if
> open_socket() fails.  I understand your API checks, and I'll fix
> things on my end.
> 
> Are the prohibitions on the use of exceptions?  I initially coded it
> to throw an exception in create_wildcard_address(), but didn't see
> handling of exceptions elsewhere.  If exceptions are ok, I can rework
> set_address() to make sure that there are no stray files or
> directories upon failure pretty easily.  Otherwise, the code may look
> a bit ugly.

I personally don't have any objections to exceptions, but it's true that they are not used anywhere else. @hintjens any objections/opinions on exceptions?

Suudy: On Mon, Mar 14, 2016 at 11:10 AM, Luca Boccassi
<notifications@github.com> wrote:
> If you need more time, I can merge as-is for now, and then you can send another PR to fix it when it's ready. Or maybe it's easier for you to work on the same branch?

I can wait until I get things into a shape that meets with your
approval.  No point in checking in "sorta" code.

> I personally don't have any objections to exceptions, but it's true that they are not used anywhere else. @hintjens any objections/opinions on exceptions?

Well, if they aren't commonly used, it probably isn't that helpful to
divert from the common style already in use.  I'll rework it without
exceptions.

Pete

Suudy: On Mon, Mar 14, 2016 at 11:10 AM, Luca Boccassi
<notifications@github.com> wrote:
> If you need more time, I can merge as-is for now, and then you can send
> another PR to fix it when it's ready. Or maybe it's easier for you to work
> on the same branch?

Actually, the cleanup was far easier than I thought.  I reverted the
goto's back to returning -1, but also cleaning up any created
directory.  I also reverted the API checks at the beginning of
ipc_listener_t::close() to the prior checks.  I also added code to
preserve errno when trying to cleanup in these cases.

I also have the formatting cleaned up, and tweaked my code to have
opening braces on the same line (to match the rest of the file).

Pete

bluca: Awesome, thank you very much for taking the time to tweak the PR!

bluca:
@Suudy - sorry to ping you again, but the same test is still failing on OSX:

FAIL: tests/test_term_endpoint
Assertion failed: (rc == 0), function main, file ../tests/test_term_endpoint.cpp, line 152.

https://travis-ci.org/zeromq/libzmq/jobs/115954957

Any idea what could be the problem?

bluca:
Never mind, it was the path buffer size, which becomes very long on OSX. Fix via: https://github.com/zeromq/libzmq/pull/1853

bluca:
FYI: this is what the TMP path looks like on OSX:

`ipc:///var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/tmphCFoJH/socket`

Suudy:
Out of curiosity, is TMP, TMPDIR, or TEMPDIR defined in the environment?

bluca:
TMPDIR is set to: /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/
The other 2 are not set.

Suudy:
Ah. I can back out the environment check and use the current working
directory as it did prior if the long names are problematic or undesirable.

bluca:
No need to, I think it's fine as long as we know it can happen and the test has a big enough buffer, which it now has.

Suudy:
It might be better to use FILENAME_MAX (from cstdio) for the buffer
sizing.  I can generate a pull request with that fix for the test.

bluca:
Good idea, that's even better. Thanks!

Suudy:
Done.  Pull request #1857 has been created.

PR #1859, opened 2016-03-19
Fix usage of IPv6 addresses

soulik:
I've noticed recent changes in "select.cpp" file that caused getsockname function to return "Bad address" error due invalid size of the sockaddr structure.

It's better to use sockaddr_storage which uses 128 bytes instead of 16 bytes for sockaddr (IPv4).

Please correct me if the behavior of getsockname function is diffent of other platforms.

Source:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms740496(v=vs.85).aspx

### DIFF
--- a/src/select.cpp
+++ b/src/select.cpp
@@ -435,6 +435,7 @@ bool zmq::select_t::is_retired_fd (const fd_entry_t &entry)
 #if defined ZMQ_HAVE_WINDOWS
 u_short zmq::select_t::get_fd_family (fd_t fd_)
 {
+	// Use sockaddr_storage instead of sockaddr to accomodate differect structure sizes
 	sockaddr_storage addr = { 0 };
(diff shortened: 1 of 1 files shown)


bluca:
I'm no windows expert :-) @opedroso any comment?

soulik:
You don't have to be. :) I've also found another source for this at:
http://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/socket.h.html

I'm quite surprised it worked before.

opedroso:
Sorry, I don't understand the question. Comment on what?

bluca:
Sorry for the confusion! These changes look good and sensible to me but it's Windows specific code, and since I'm not a Windows dev, I was just looking for an opinion more qualified than mine :-)

Issue #1695, opened 2016-01-05
ZMQ_REQ_CORRELATE may not work when sent messages are not received immediately


FredTreg:
## Problem
Considering a REQ socket with both the `ZMQ_REQ_RELAXED` and `ZMQ_REQ_CORRELATE` options enabled, if two messages are sent in a row, then the next `recv()` done on the socket will fail to receive the answer corresponding to the second message (as would otherwise be expected) if the server connected to the REQ socket **is bound only after the two messages are sent**. Answer from the first message is received instead.

The example code pasted at the end of this issue reproduces this behavior.

Notes:

1. This issue is linked to issue #1690 and its fix. Without the fix the issue cannot happen but the `ZMQ_REQ_RELAXED` option cannot be used either.

2. The problem exposed is only one example of a more general issue regarding the option `ZMQ_REQ_CORRELATE`. The incorrect correlation issue may arise with other setups such as when using an `inproc://` ROUTER server bound before the messages are sent *but slow* to respond to them.

## Analysis
By checking the message parts sent, I could track the issue down to this piece of code in the `req.cpp` file:

``` C++
int zmq::req_t::xsend (msg_t *msg_)
{
    ... 

        if (request_id_frames_enabled) {
            request_id++;

            msg_t id;
            //  --> Next line causes the issue: request_id may change in the future
            //  --> as it is not duplicated prior to being sent to the pipe.
            int rc = id.init_data (&request_id, sizeof (request_id), NULL, NULL);
            errno_assert (rc == 0);
            id.set_flags (msg_t::more);

            rc = dealer_t::sendpipe (&id, &reply_pipe);

    ...
```
The `request_id` in the code above is a member variable and is not copied in memory to a new variable when fed to the `init_data` method. So multiple sent messages share the same variable.

As a consequence, when issuing two `send()` in a row, the first message may not have been sent yet down the wire when the second message is sent to the pipe. The `request_id` value of the first message is then overridden by the `request_id` value of the second message before being wired. **Sent messages no longer have unique ids and can no longer be correlated**.

## Solutions
1. A first solution would be to be able to clear the send pipe of any unwired messages before sending the next message down the pipe. I am unfamiliar with the internals of zmq and could not find an obvious API that could do the job. For example the method `terminate()` that was used prior to the fixing of defect #1690 was too much as it would close the send pipe forever.

[...]

PR #1862, opened 2016-03-20
Fixed issue #1695 (ZMQ_REQ_CORRELATE)

FredTreg:
**Problem**: when using ZMQ_REQ_RELAXED + ZMQ_REQ_CORRELATE and two 'send' are executed in a row and no server is available at the time of the sends, then the internal request_id used to identify messages gets corrupted and the two messages end up with the same request_id. The correlation no longer works in that case and you may end up with the wrong message.

**Solution**: make a copy of the `request_id` instance member of the `req_t` class before sending it down the pipe.

Issues #1690 and #1695 are linked: I also updated documentation to reflect the changes (there is no longer a disconnection of the peer when two sends in a row are sent from a relaxed REQ socket).

It seems I am the first to use `malloc()` on the ZMQ project: I hope this is ok! (I could not see a way around it).

Thanks!

### DIFF
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -667,11 +667,9 @@ ZMQ_REQ_RELAXED: relax strict alternation between request and reply
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 By default, a REQ socket does not allow initiating a new request with
 _zmq_send(3)_ until the reply to the previous one has been received.
-When set to 1, sending another message is allowed and has the effect of
-disconnecting the underlying connection to the peer from which the reply was
-expected, triggering a reconnection attempt on transports that support it.
-The request-reply state machine is reset and a new request is sent to the
-next available peer.
+When set to 1, sending another message is allowed and previous replies will
+be discarded if any. The request-reply state machine is reset and a new
+request is sent to the next available peer.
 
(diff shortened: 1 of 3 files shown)


hintjens:
Why did you close the pull request?

FredTreg:
Hi Pieter,

I saw that tests were failing on your automated build system, specifically
test_req_relaxed.cpp and test_req_correlate.cpp which are the two options I
am trying to fix.

As I had only tested my patch against the official 4.1.4, I closed my pull
request. I then tested my patch with the HEAD of the master branch and
indeed the two tests are failing. I am not sure it is related to my patch,
but I would like to investigate more before submitting my patch again.

Sorry if it caught you in the middle of the code review.

Thanks,

2016-03-20 21:42 GMT+01:00 Pieter Hintjens <notifications@github.com>:

bluca:
Hi Fred,

Those two tests are marked as expected failures, as the feature is currently not working.
See https://github.com/zeromq/libzmq/pull/1730 for more details. Brownie points if you can get that issue fixed too :-) But don't hold back these changes because of that!

FredTreg:
Ok, thanks for pointing this out. I'll submit my changes then and look into #1730.

On a personal note, I started investigating ZMQ_REQ_CORRELATE and ZMQ_REQ_RELAXED because I wanted to use these options in my workflow. But since I looked into it more deeply I came to believe that these options are of very little use : in most scenario if you need to make two 'send' in a row on a REQ socket, closing it and reopen it before the second send will do the job : it is simple and effective. So maybe the suggestion in #1730 to drop these options in a future release of libzmq makes sense. Just a thought...

PR #1869, opened 2016-03-29
zeromq-4.1.4/src/tcp_address.cpp : fix relaxed use of sizeof() without parenthesis

jimklimov:
Still compiles on Linux. Newly compiles on Solaris (specifically OpenIndiana userland). Both with gcc ;)


bluca:
Thanks!

For your next contributions, could you please write a commit message with a style as explained in http://zeromq.org/docs:contributing ? (title: Problem: ABC  body: Solution: XYZ

Issue #1867, opened 2016-03-26
change to zmq_utils.h breaks MSVC pyzmq (easy fix)


gnieboer:
The new zmq_utils.hpp using the #warning syntax to emit the deprecation warning, however msvc uses a different syntax.  Recommend the following:

```
#ifndef _MSC_VER
#warning(zmq_utils.h is deprecated.All its functionality is provided by zmq.h.)
#else
#pragma message("Warning: zmq_utils.h is deprecated. All its functionality is provided by zmq.h.")
#endif
```

c-rack:
Thanks. Would you mind to submit a pull request with this change?

gnieboer:
See pull request #1871

bluca:
Closing as this is now fixed. Thanks for sending the PR!

PR #1871, opened 2016-03-29
Update zmq_utils.h

gnieboer:
Previous change broke MSVC b/c MSVC doesn't support #warning


gnieboer:
Reference issue #1867

Issue #1866, opened 2016-03-24
Make failed with SunOS 5.10 sparc


riboseyim:
**SunOS 5.10 Generic_148888-01 sun4v sparc SUNW**
**version: zmq-4.1.2**

`# ./configure
checking for a BSD-compatible install... config/install-sh -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... config/install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... nawk
checking whether make sets $(MAKE)... yes
checking how to create a ustar tar archive... plaintar
checking whether make supports nested variables... yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking for gcc option to accept ISO C99... -std=gnu99
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking dependency style of g++... gcc3
checking whether gcc -std=gnu99 and cc understand -c and -o together... yes
checking for a sed that does not truncate output... /usr/bin/sed
checking for gawk... (cached) nawk
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for xmlto... no
checking for asciidoc... no
checking build system type... sparc-sun-solaris2.10
checking host system type... sparc-sun-solaris2.10
checking how to print strings... printf
checking for a sed that does not truncate output... (cached) /usr/bin/sed
checking for grep that handles long lines and -e... /usr/xpg4/bin/grep
checking for egrep... /usr/xpg4/bin/grep -E
checking for fgrep... /usr/xpg4/bin/grep -F
checking for ld used by gcc -std=gnu99... /usr/ccs/bin/ld
checking if the linker (/usr/ccs/bin/ld) is GNU ld... no
checking for BSD- or MS-compatible name lister (nm)... /usr/ccs/bin/nm -p
checking the name lister (/usr/ccs/bin/nm -p) interface... BSD nm
checking whether ln -s works... yes
checking the maximum length of command line arguments... 786240
checking whether the shell understands some XSI constructs... yes
checking whether the shell understands "+="... yes
[...]

PR #1892, opened 2016-04-12
Problem: build broken on Solaris 10

bluca:
Solution: see commits

Note that this is a partial solution for #1866, there is one more fix to be able to build 4.2 on Solaris, but I'm not sure whether we should add it.

The problem is that we have started using C++11 features - which is great IMHO, but this also means dropping support for older platforms where GCC is just too old, like Solaris 10.
In my opinion at this point it would be acceptable to declare that 4.1 is the last release that supports compilers without C++11.

@hintjens (and any other contributor of course) what do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,10 @@ set (CMAKE_REQUIRED_INCLUDES sys/time.h)
 check_function_exists (gethrtime HAVE_GETHRTIME)
 set (CMAKE_REQUIRED_INCLUDES)
 
+set (CMAKE_REQUIRED_INCLUDES stdlib.h)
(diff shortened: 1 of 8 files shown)


bluca:
@hintjens @c-rack - ping! Any comment on the C++11 issue?

bluca:
Also I backported (or otherwise fixed) 4.1 and 4.0 as well:

https://github.com/zeromq/zeromq4-x/pull/153
https://github.com/zeromq/zeromq4-1/pull/117

c-rack:
@bluca for me, it is ok to declare 4.1 as last release with C++11 support.

PR #1895, opened 2016-04-14
Fix typos similar to #1894

hitstergtd:
Similar to pull request #1894

### DIFF
--- a/src/array.hpp
+++ b/src/array.hpp
@@ -57,7 +57,7 @@ namespace zmq
         {
         }
 
-        //  The destructor doesn't have to be virtual. It is mad virtual
+        //  The destructor doesn't have to be virtual. It is made virtual
         //  just to keep ICC and code checking tools from complaining.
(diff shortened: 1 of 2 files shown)


c-rack:
@hitstergtd Thanks for the typo fixes!

If you look at our commit history, you will find that we use a problem/solution pattern for our commits.
It would be great if you could follow this style for your upcoming pull requests as well ;-)
I'd also recommend to learn more about our [C4.1 process](http://rfc.zeromq.org/spec:22).

hitstergtd:
@c-rack, no problem. Apologies for not following process.

In terms of the problem/solution pattern, are you referring to the format of the Pull Request message?

hitstergtd:
OK, I think I understand now. I will follow the format PH used in #1884, henceforth, for future Pulls.

hitstergtd:
@c-rack Thanks for the merges!

PR #1896, opened 2016-04-14
Problem: Running Linux-only tests results in 10% of tests failing on non-Linux systems (OSX, et. al.)

hitstergtd:
**Problem:**
* Running Linux-only tests on non-Linux systems results in ~10% of the tests (9 out of 88) failing.
* Building and running tests for Linux-only features on non-Linux systems, e.g. Mac OS X, is pointless.
* Linux-only features:
  * Abstract Namespace support for AF_UNIX sockets
  * AF_TIPC sockets

**Solution:**
* Move tests for Linux-only features under a platform conditional thereby eliminating unnecessary builds and fixing "make test" on Mac OS X and possibly other non-Linux systems.
* Test success rate jumps from 90% to 100% on Mac OS X after this change.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -63,14 +63,9 @@ set(tests
         test_capabilities
         test_ipc_wildcard
         test_metadata
-        test_pair_tipc
-        test_reqrep_device_tipc
-        test_reqrep_tipc
         test_router_handover
-        test_router_mandatory_tipc
         test_srcfd
         test_stream_timeout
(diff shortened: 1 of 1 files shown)


bluca:
For the next PR, could you please use the same Problem: .. Solution: .. format in the git commit message? Thanks!

hitstergtd:
@bluca, hmm, seems a lot of the rules have changed. I never had to do that before.

Anyway, I just did a git log and noticed my transgression; apologies! :)

Thanks for the merge!

PR #1897, opened 2016-04-15
Problem: errno not set if UDP is NOT used with ZMQ_RADIO or ZMQ_DISH

hitstergtd:
**Solution:** set `errno` to `ENOCOMPATPROTO` in `check_protocol()`


hitstergtd:
@jemc Thanks! 👍

hintjens:
Nice :)

hitstergtd:
@hintjens Thanks. :)

PR #1898, opened 2016-04-15
Problem: check_protocol() logic duplicated twice and needs cleanup

hitstergtd:
**Problem:**
* Conditional logic in `check_protocol()` that checks if a protocol is supported,
is duplicated twice. Moreover, the first set of checks to ascertain if a
protocol is supported is done regardless of whether the particular protocol
will be built into the library or not.

**Solution:**
* Simplify/collapse all supported protocol checks into one in `check_protocol()`
* Enclose `pgm/epgm/norm` socket+protocol match checks with requisite macros

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -263,66 +263,41 @@ int zmq::socket_base_t::check_protocol (const std::string &protocol_)
 {
     //  First check out whether the protocol is something we are aware of.
     if (protocol_ != "inproc"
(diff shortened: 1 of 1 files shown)


hitstergtd:
**Please note:**
* I would appreciate it if this pull is reviewed by another set of eyes before it is merged.
* I have tested it on Mac OS X and Ubuntu Linux.
* I have ran "make test" and the various performance tests; everything seems fine.

c-rack:
LGTM

hitstergtd:
@c-rack, thanks for the review and merge! 👍

PR #1900, opened 2016-04-18
Problem: test_udp does not release resources properly

hitstergtd:
**Problem:**
* `tests/test_udp.cpp` does not release resources properly under certain error conditions

**Solution:**
- call `zmq_msg_close(3)` if there is an error
- call `free(3)` to release resources if there is an error ([CWE-404](http://cwe.mitre.org/data/definitions/404.html))


hitstergtd:
@bluca Thanks! 👍

PR #1902, opened 2016-04-18
Problem: can't unbind with bound addr with IPv6

bluca:
Solution: un-break un-binding a socket with the same endpoint string that was used to bind it when IPv6 is enabled. The breakage was caused by using the actual last_endpoint as a key instead of the string passed by the user, which is necessary in the wildcard case. But it broke the normal case with IPv6, since even if the address is an IPv4 one it will be saved in IPv6 format, so it doesn't match anymore. This is the culprit: https://github.com/zeromq/libzmq/commit/09e7416ee9588f131493e29491b46cc09a9560b5

My solution is a liiitle bit of a hack - basically I try to use the used-provided endpoint string and resolve it, just like it's used in connect/bind case, and see if that succeeds before giving up. At least it should be future-proof, and if more changes are done to the connect/bind they should be picked up in the unbind/disconnect. And the existing API for wildcard sockets is maintained, nothing is broken.

Also added a bunch of tests for various combinations of TCP sockets endpoints. Note that Travis on Precise (legacy) or Trusty does not support IPv6, so I added a test run on the container-based environment which is the only one that has a loopback interface with ::1 bound on it, so IPv6 test cases can run there. Note that I can't test this on Windows - I think it should work as I've used the same options as in existing part of the library.

Fixes https://github.com/zeromq/zeromq4-1/issues/114

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,11 @@ matrix:
        os: osx
     -  env: BUILD_TYPE=valgrind CURVE=tweetnacl
        os: osx
+  include:
+    -  env: BUILD_TYPE=default CURVE=tweetnacl IPv6=ON
+       os: linux
+       dist: precise
+       sudo: false
 
 sudo: required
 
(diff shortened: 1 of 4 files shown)


bluca:
Backported to 4.1: https://github.com/zeromq/zeromq4-1/pull/118

PR #1906, opened 2016-04-19
Problem: Linker search path warnings on OS X

hitstergtd:
**Problem:**
A per-test Linker search path was added in commit a911fa4 to `tests/CMakeLists.txt` as part of fixing Windows builds. Whilst this is silently ignored by `ld(1)` on Linux, it does not settle well with OS X. Spurious warnings are generated about missing directories, which unnecessarily convolutes the build logs.

**Solution:**
Make per-Test `LINK_DIRECTORIES()` call to be conditional for non-Apple platforms.


hitstergtd:
@bluca 👍

hitstergtd:
I just reviewed the Travis build logs and realised that this was a quick-fire pull request; this issue can happen whenever using the LLVM suite (Clang and friends).

I will send another pull shortly.

PR #1909, opened 2016-04-21
problem: push-pull socket types are not thread safe

somdoron:
solution: add new socket types, scatter and gather which are thread safe.

@hintjens thought it might make you smile a little.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,6 +126,8 @@ test_poller
 test_timers
 test_radio_dish
 test_udp
+test_scatter_gather
+test_socketopt_hwm
 test_use_fd_ipc
 test_use_fd_tcp
 tests/test*.log
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -490,7 +490,9 @@ set (cxx-sources
         radio.cpp
         dish.cpp
         udp_engine.cpp
-        udp_address.cpp)
+        udp_address.cpp
+        scatter.cpp
+        gather.cpp)
 
 set (rc-sources version.rc)
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,6 +57,8 @@ src_libzmq_la_SOURCES = \
 	src/fd.hpp \
 	src/fq.cpp \
 	src/fq.hpp \
+	src/gather.cpp \
+	src/gather.hpp \
 	src/gssapi_mechanism_base.cpp \
 	src/gssapi_mechanism_base.hpp \
 	src/gssapi_client.cpp \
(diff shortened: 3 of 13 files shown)


hitstergtd:
@somdoron - just wondering - what specifically makes them thread-safe?

hitstergtd:
@somdoron 
I see `test_large_msg` was added to `.gitignore` but don't see that file anywhere?

somdoron:
First is rejecting multipart messages. Then the implementation (mutex) is already in socket_base, so the the last parameter to the base ctor enable the thread safety

hitstergtd:
@somdoron - Thanks!

hitstergtd:
@somdoron - did you need to add `test_large_msg` to the pull?

somdoron:
Regarding the gitignore, I added all the stuff that shouldn't be commited as and git status as show as new files. Bottom line when adding new test the executable should be added to ignore, someone didn't do it so I just added it

somdoron:
Or maybe it deleted test that only exist in my machine?

somdoron:
You are right, it only on my machine, i will fix that

hitstergtd:
Thanks! I just re-checked my tree just to be sure as well.

hintjens:
cool cool cool!

hitstergtd:
@somdoron - I will pull it once you fix the `test_large_msg` addition.

somdoron:
@hitstergtd fixed

hintjens:
@somdoron I'm so impressed :)

somdoron:
Thanks :)

Only the pair socket type is now missing... You had a plan to use client-server instead, do you we need the thread safe pair?

hintjens:
Well, we can at least emulate pair-to-pair with client-to-server. This suggests we do not need a threadsafe pair equivalent. We can also do more interesting things in the only use case I know for pair, which is actors. I.e. start up internal services that are accessible from any thread in the process. Like a logging actor, or a permissions actor.

somdoron:
If we don't want req-res we can just use same client for threads accessing the actor. 

if we need req-res we have few options:
1. If we want req-res each thread or user need it is own client socket, we can create those on the spot, thread safe socket is less expensive (no file descriptor to create, only mutex), but still can be expensive, we can some kind of pool. 

2. Any user of the actor must create a reference, so the api can be:

```C
zactor_ref_t* zactor_ref_new (zactor_t *actor);
zactor_ref_t  zactor_ref_send (zactor_ref_t *self, zframe_t **frame_p);
zframe_t* zactor_ref_recv (zactor_ref_t *self);
```

3. Add zmq_send_recv that is atomic and send and then wait for message. Only issue with this is that it will block all other operations until the response is received.

hintjens:
synchronous req rep should be called the "strange cousin" pattern. let's
never talk about him again.

PR #1911, opened 2016-04-21
problem: udp is failing on windows

somdoron:
Fix UDP failing on windows

### DIFF
--- a/src/select.cpp
+++ b/src/select.cpp
@@ -439,14 +439,25 @@ u_short zmq::select_t::get_fd_family (fd_t fd_)
 	sockaddr_storage addr = { 0 };
 	int addr_size = sizeof addr;
 
-    int rc = getsockname (fd_, (sockaddr *)&addr, &addr_size);
+    int type;
+    int type_length = sizeof(int);
+    
+    int rc = getsockopt(fd_, SOL_SOCKET, SO_TYPE, (char*) &type, &type_length);
 
-    //  AF_INET and AF_INET6 can be mixed in select
-    //  TODO: If proven otherwise, should simply return addr.sa_family
-    if (rc != SOCKET_ERROR)
-        return addr.ss_family == AF_INET6 ? AF_INET : addr.ss_family;
-    else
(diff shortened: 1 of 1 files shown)


hitstergtd:
@c-rack, @bluca -
When pulling PRs, should we not wait until at least Travis CI passes just to be sure? Just need to be sure so that I know next time when I pull anything into the repo.

hitstergtd:
Just checked. It works well. Thanks @somdoron.

bluca:
@hitstergtd - this topic has been debated a great deal :-)

The C4.1 process ( http://rfc.zeromq.org/spec:22 ) says to merge rapidly a patch if it looks correct. That has been the policy for these projects - in case a failure/regression in the tests is introduced, it can be spotted quickly (Travis saves history, so it's trivial to see which PR introduced a failure) and then be fixed either by the submitter, or if they don't respond, by another contributor or maintainer. If something is badly broken, and it was not apparent immediately, it is also allowed to just revert the offending commits and ask the submitters to try again.

This has worked quite well IMHO. I think the holes in the system are the Windows builds - since the tests have always been failing on Jenkins and Appveyor, it's extremely hard to tell if a PR breaks them or not, so we are missing coverage there. Hence I think fixing those should be a priority, if possible.

Some contributors prefer to be sure they are not breaking something before a PR is merged - that is fine, and they can be directed at instructions on how to enable appveyor/travis builds on their own forks. It's trivial (and free) to do so. I myself do this - I always wait for the CI to run on my forks before I submit a PR. But IMHO it should not be a hard rule.

hitstergtd:
@bluca - Hah, seems like it, given your detailed answer. :)

OK. So for my own purpose I will be enabling the Travis and AppVeyor builds on my fork just as I had anticipated and also the way you recommend; but it is my opinion that such a practice cannot be assumed of all contributors who submit PRs. Having CI builds pass on the main repo would be considered essential in most other projects I have worked on, as you never know if someone misses something even rather trivial, which my next msg/pull will portray. :)

Issue #1903, opened 2016-04-19
CI and AppVeyor Build Issues


hitstergtd:
The CI builds seem to be failing and the shields in README are marked as such, so I have put this up in case anyone gets to it before I do, or if anyone would like to help.

**Issues:**
- AppVeyor builds failing at the test stage (about 5 of the tests fail)
- AppVeyor build configurations need to be reviewed
- Travis CI builds are failing

These are good starter problems for anyone who wants to get involved in the project.

bluca:
The coverage job in Travis failed last time, but probably nothing to worry about, I've kicked it again

hitstergtd:
OK, thanks @bluca. It is just that just about every pull I sent had indeterminate CI build failures and just kept hating to see the red shields on the GitHub page. :)

bluca:
Yeah worked now, I think the coveralls service was just unable to fetch from github for whatever reason

bluca:
Yes that's a very valid concern (Travis shield is green again now), and if the Windows build on Appveyor worked it would be great for us without Windows boxes :-)

hitstergtd:
Excellent.

* Yeah just noticed the Travis shield green status. Thanks!
* One down, X different AppVeyor builds to go!
* I have a Windows box and need to set it up for various `libzmq` and `czmq` build configurations; although just for VS2015 and mingw-w64 builds.

bluca:
There are, however, tests intermittently failing on Travis, like test_shutdown_stress or test_proxy. Might be good to look into that to make them more resilient.

hitstergtd:
Agreed.

hitstergtd:
* Perhaps we should not run test_shutdown_stress with Travis?
* Maybe `ZMQ_DISABLE_TEST_TIMEOUT` should be defined per `tests/testutil.hpp` and also added into `tests/CMakeLists.txt`?

bluca:
- If it's not run then it's as like it wasn't there, and we lose coverage. Better to have it flacky on some platforms temporarily, and fix it as soon as possible
- IMHO that would just make things worse, as tests would timeout instead. If there are race conditions or quiescence periods are not long enough those should be fixed instead

hitstergtd:
OK. Agreed.

hitstergtd:
@bluca 
I was just looking through the AppVeyor documentation. Have you looked at?:

<http://www.appveyor.com/docs/how-to/rdp-to-build-worker>

hitstergtd:
[...]

PR #1912, opened 2016-04-21
Problem: ZMQ_SRCFD test does not work on Windows

hitstergtd:
**Solution:**
* Add Winsock specific assertions, since [getpeername()](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738533(v=vs.85).aspx) will return `SOCKET_ERROR` (-1) and `WSAGetLastError()` will return last error as `WSAENOTSOCK`.

**Notes:**
* The issue has been around since the initial commit of `tests/test_srcfd.cpp`
* Please wait until AppVeyor build is complete before merging this into base

**Issues affected:**
* #1903


hitstergtd:
@somdoron Thanks!

PR #1914, opened 2016-04-22
Problem: LD search path warnings if using Clang

hitstergtd:
Solution: PR #1906 did not solve the problem entirely; subsequent Travis CI builds indicated that the issue happens due to Clang/LLVM, naturally, so make sure to fix the issue by detecting if CMake is using Clang for building the tests.


hitstergtd:
Doron, thanks!

PR #1918, opened 2016-04-23
Problem: XPUB test broken on Windows since #1569

hitstergtd:
**Solution:**
- Adjust `test_subscriptions()` to support different protocol types
- Run TCP and IPC tests everywhere but on Windows and OpenVMS


hitstergtd:
@somdoron - thanks! 👍

bluca:
The assertion is the other way around, opened https://github.com/zeromq/libzmq/pull/1919 to fix :-)

PR #1919, opened 2016-04-23
Problem: assertion in test_xpub_manual is swapped

bluca:
Solution: check that both pointers passed as arguments to
test_missing_subscription are non-NULL, instead of the opposite.


hitstergtd:
I just noticed this blunder; thanks! :-)

bluca:
No problem, it happens :-)

PR #1920, opened 2016-04-23
problem: Tests are failing on windows

somdoron:
please let the CI run, I want to check if this fixed the problem

### DIFF
--- a/tests/test_term_endpoint.cpp
+++ b/tests/test_term_endpoint.cpp
@@ -129,9 +129,9 @@ int main (void)
     assert (push);
     rc = zmq_bind (push, ep_wc_tcp);
     assert (rc == 0);
+    pull = zmq_socket(ctx, ZMQ_PULL);
+    assert(pull);
 #if !defined ZMQ_HAVE_WINDOWS && !defined ZMQ_HAVE_OPENVMS
-    pull = zmq_socket (ctx, ZMQ_PULL);
-    assert (pull);
     rc = zmq_bind (pull, ep_wc_ipc);
     assert (rc == 0);
 #endif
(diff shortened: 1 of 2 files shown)


hitstergtd:
@somdoron,
This will "fix it", in the sense that now you are using a non-wildcard address. The question is why aren't wildcard addresses working at all. I believe your fix is masking that issue.

somdoron:
not really, I still using wildcard addresses.

The problem is that on windows you cannot connect to 0.0.0.0 (I'm not sure what the behavior on linux). So I'm still using the wildcard on the bind but for the connect I'm using localhost. The test is for unbind wildcard address, which I'm still testing.

somdoron:
not really, I still using wildcard addresses.

The problem is that on windows you cannot connect to 0.0.0.0 (I'm not sure what the behavior on linux). So I'm still using the wildcard on the bind but for the connect I'm using localhost. The test is for unbind wildcard address, which I'm still testing.

somdoron:
@hitstergtd the bottom line is that we are still checking unbind of wild card, I just changed the connected address to solve specific problem on windows, the binding is still on wildcard addresses.

bluca:
On Linux (and I believe *nix in general) connecting to localhost.

But more importantly: shouldn't ZMQ_LAST_ENDPOINT return an actual address, rather than INADDR_ANY?

somdoron:
@bluca I don't think so, when you are binding on * you actually bind all interfaces, so ZMQ_LAST_ENDPOINT return INADDR_ANY, which seems right.

somdoron:
So the different behavior is, on linux connecting to 0.0.0.0 connect to localhost, on windows it just fail. So I just changed it so on windows it will connect to 127.0.0.1.

hitstergtd:
@somdoron - you are absolutely spot on.

bluca:
Ah yes, it makes sense thinking about it.

hitstergtd:
Next time @bluca, we will just wildcard all issues to @somdoron (just kidding). :-)
@somdoron Thanks again!

somdoron:
This can be merged. Tests are fixed, now the jenkins failing on those:
48 - test_thread_safe (Timeout)
	 61 - test_xpub_manual (Failed)
	 63 - test_timers (Failed)

somdoron:
@hitstergtd you welcome :-)

PR #1927, opened 2016-04-25
Problem: tune_tcp_keepalives signature more than 80 chars on a line

hitstergtd:
**Solution:**
Fix it.


hitstergtd:
@somdoron - thanks!

PR #1929, opened 2016-04-26
Problem: test_pub_invert_matching missing in autotools build

hitstergtd:
**Solution:**
* Add it to `Makefile.am`

**Note:**
* Performed autotools build and test to verify inclusion of `test_pub_invert_matching`
* Number of tests is now equal on both Autotools and CMake build infrastructure


hitstergtd:
@somdoron thanks!

PR #1932, opened 2016-04-29
Problem: libzmq lacks a good doxygen config

hitstergtd:
**Solution:**
- Add Doxygen.cfg file to simplify viewing of internals/externals for library users, other parties, etc.
- Add README.doxygen.md as an Overview for Doxygen-generated documentation


hitstergtd:
@somdoron, thanks Doron!

PR #1934, opened 2016-04-29
problem: no documentation for Radio-dish and UDP

somdoron:
also solving: no way to specify bind address when using Radio with UDP

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -21,7 +21,7 @@ MAN3 = zmq_bind.3 zmq_unbind.3 zmq_connect.3 zmq_disconnect.3 zmq_close.3 \
     zmq_atomic_counter_value.3 zmq_atomic_counter_destroy.3
 
 MAN7 = zmq.7 zmq_tcp.7 zmq_pgm.7 zmq_inproc.7 zmq_ipc.7 \
-    zmq_null.7 zmq_plain.7 zmq_curve.7 zmq_tipc.7 zmq_vmci.7
+    zmq_null.7 zmq_plain.7 zmq_curve.7 zmq_tipc.7 zmq_vmci.7 zmq_udp.7
 
 MAN_DOC =
 
--- a/doc/zmq.txt
+++ b/doc/zmq.txt
@@ -165,6 +165,9 @@ Local in-process (inter-thread) communication transport::
 Virtual Machine Communications Interface (VMC) transport::
     linkzmq:zmq_vmci[7]
 
+Unreliable unicast and multicast using UDP::
+    linkzmq:zmq_udp[7]
+
 
 Proxies
 ~~~~~~~
--- a/doc/zmq_socket.txt
+++ b/doc/zmq_socket.txt
@@ -47,9 +47,18 @@ incoming connections *from multiple endpoints* bound to the socket using
 _zmq_bind()_, thus allowing many-to-many relationships.
 
 .Thread safety
-0MQ 'sockets' are _not_ thread safe. Applications MUST NOT use a socket
+0MQ has both thread safe socket type and _not_ thread safe socket types.
+Applications MUST NOT use a _not_ thread safe socket
 from multiple threads except after migrating a socket from one thread to
 another with a "full fence" memory barrier.
+
+Following are the thread safe sockets:
+* ZMQ_CLIENT
+* ZMQ_SERVER
+* ZMQ_DISH
+* ZMQ_RADIO
+* ZMQ_SCATTER
+* ZMQ_GATHER
 
 .Socket types
 The following sections present the socket types defined by 0MQ, grouped by the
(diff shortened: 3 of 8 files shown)


jemc:
@somdoron - Regarding the UDP transport, I want to be able to write an [interoperable implementation in another language](https://github.com/jemc/pony-zmq).  Is there anything that needs to be updated in the ZMTP spec for this?

somdoron:
UDP is not using ZMTP, just plain UDP. check udp_engine class. If I remember correctly the first byte is the length of the group, following by the group following by the message. That all. As radio/dish doesn't supporr multipart so the wiring. Also there is no handshake.

jemc:
Thanks for the info.  It might be nice to formalize this spec at some point, but that info gives me a good place to start.

somdoron:
I will, I just documented the transport and the radio-dish pattern. Also
need to create rfc for radio dish and zeromq over udp.

sunkin351:
Well, it might be worth while to implement some way to mimic multipart messages for UDP. Or do you plan to let the application developers deal with this?

jemc:
@sunkin351- unfortunately, none of the new (threadsafe) socket types support multipart.

This is a consequence of upholding the current libzmq API, and also retaining zero-copy.  multi-part could be implemented in a zero-copy and threadsafe way, but would require a change of API.

sunkin351:
How long has it been since you guys last had a change in API? And what kind of change would it take to make multi-part threadsafe?

sunkin351:
(sorry, I cant edit while I am on my phone.)

Also, just to be clear, what does one mean here by threadsafe?

jemc:
> what kind of change would it take to make multi-part threadsafe?

All parts would need to be sent/received in a single API call, instead of in one API call per part in the multipart message.  Just how this would look (in a way that keeps zero-copy in mind) has never really been fleshed out by anyone.

> what does one mean here by threadsafe?

The threadsafe socket types use a mutex to synchronize access, so they can be shared among threads, each thread doing its own API calls.  The only thing that doesn't work in this paradigm is operations that need to span multiple sequential API calls, hence no more multipart-as-we-know-it.

sunkin351:
I see. If such an API change were to occur, I would fully support it. If I may, I really like the approach that zmqpp takes on multipart messages, having a message object that is able to hold however many messages you need to send before you send them, and sending them all in one go.

jemc:
I would support it too (well, depending on the details of how it was implemented).  I think multipart is an important part of the ZeroMQ value proposition, and I'd hate to see us move into a future without it.

Someone needs to take the time to design and implement such an API.

sunkin351:
I could try my hand at API design, though I don't know how good I would be at plain C.

somdoron:
I'm not a big fan of multipart, I think those kind of things need to be handled by the application, the problem is that multipart add a lot of complication to libzmq, for example implementing the UDP with multipart support probably won't take one day as it did, a lot of earlier tries to implement UDP failed, because of the complication of implemented a transport in zeromq. Implementing a transport that only support single-part open a window to a lot of other transport, for example as Web Socket is datagram protocol as well (like UDP), adding Web Socket transport to libzmq will be very simple (I'm planning to do that).

Also avoiding multipart yield better performance even for non thread-safe zeromq sockets.

Regarding multiple part, as I said I think it shouldn't be solved in the socket level, but we can try and solve that in the zeromq library. Actually czmq has bsend and brecv multipart like behavior.

With czmq we can also encode zmsg into byte array and send it over thread safe sockets.

We also have zproto (bsend/brecv compatible with zproto) which we can use to do multipart like.  

What we are missing is a way to do zero copy, we can create new message class (name suggestions?) in libzmq, which we can also send over a socket that support zproto protocol and can be zero-copied.

hitstergtd:
WebSocket support, in `libzmq`... well. 👍 :)

sunkin351:
To get this straight, you want it to be implemented in such a way so as to compile a multipart message into a single message, and then send it. Correct or no? If yes, then we think alike. :P

somdoron:
@sunkin351 yes :-)

sunkin351:
Do you think we can do that with the current API? Somehow storing the data provided to each call to the sendmore() function until the final call to send()?

somdoron:
I won't be thread safe that way, we need to accept array, but that way it won't be zero-copy and yield not good performance...

sunkin351:
Are contexts thread-safe?

somdoron:
take a look at bsend/brecv, now we can somehow use that to support zero copy, so we can the the bsend logic and move it to new class, like:

```C
void* zmq_bmsg_new (char* picture);
int zmq_bmsg_send (void* self_, void* socket_);
int zmq_bmsg_recv (void* self_, void* socket_);
int zmq_bmsg_write(void* self_, ...);
int zmq_bmsg_read(void* self_, ...);
int zmq_bmsg_pop(void* self_, char* picture, ...)
```

So send, recv, write and read act normal as we expect, but pop is the magic that doing the zero-copied. We can pop some parts of the message and then send the rest of the message without decode it.

[1] https://github.com/zeromq/czmq/blob/master/doc/zsock.txt#L205

somdoron:
> Are contexts thread-safe?

any API with multiple calls is not thread safe.

sunkin351:
I am really not liking how limited C is. And we cant expose any of the underlying C++ implementation. Passing a vector to libzmq would be very useful.

scemama:
@sunkin351 Please let the API of libzmq be compatible with standard C. Using STL structures in the API would certainly be good for C++ codes, but using libzmq in other languages will become **extremely painful**.

hitstergtd:
Just spitballing here - would it not be possible to support thread-safe multi-part messages with the `zmq_sendiov(...)` and `zmq_recviov(...)` functions?

sunkin351:
@scemama I am just ranting on about how plain C doesn't have any kind of vector support.

somdoron:
@hitstergtd we can (someone also suggested that on the list), however there are some problems with that:

1. Are we going to encode into one message or real zeromq multipart? If the second that is going to make it harder to add new transports, as we now need to support zeromq parts on the transport. One of the reasons I like single-part is that it making much more easier to add transports
2. If we encode it into one message someone can get confuse when trying to receive it with zmq_msg_recv.
3. We need different implementation for those as now, because current implementation is not thread safe, but the single call API make it possible.
3. The main issue is, that if someone is using zmq_recviov and zmq_sendiov he have to be consistent.

somdoron:
I like the czmq picture solution, I thought about the API little more:

```C
void* zmq_bmsg_new ();
int zmq_bmsg_send (void* self_, void* socket_);
int zmq_bmsg_recv (void* self_, void* socket_);
int zmq_bmsg_recv_no_wait (void* self_, void* socket_);
int zmq_bmsg_write(void* self_, const char *picture, ...);
int zmq_bmsg_read(void* self_, const char *picture, ...);
int zmq_bmsg_pop(void* self_, const char* picture, ...)
```

Usage:

```C
//  Sending
int rc = zmq_bmsg_write (bmsg, "44s", 1,2, "Hello");
assert (rc == 0);
rc = zmq_bmsg_send(bmsg, socket);
assert (rc == 0);

//  Receiving
int n1,n2;
char *str;
rc = zmq_bmsg_recv(bmsg, socket);
assert (rc == 0);
rc = zmq_bmsg_read(bmsg, "44s", &n1, &n2, &str);
assert (rc == 0);

//  Zero-copy
rc = zmq_bmsg_recv(bmsg, socket);
assert (rc == 0);
rc = zmq_bmsg_pop(bmsg, "4", &n1); // will pop the first integer from the message, leaving the rest
assert (rc == 0);
zmq_bmsg_send(bmsg, other_socket);
```

hitstergtd:
@somdoron 
Understood. Just to clarify, I was thinking along the following lines but I could be trivialising the matter:

1. **Thread-safe route** - encode and send/recv provided `iovec` as a single-part message
2. **Non-thread safe route** - if `ZMQ_SNDMORE` or `ZMQ_RCVMORE` is given in `flags`

The above would be applicable to all the send and receive capabilities in the API, but I am not sure if that would induce API breakage which is a no-no.

I also concur that multi-part sends should be the responsibility of the API user as it removes unnecessary complexity from the library and allows simpler and cleaner implementations.

hitstergtd:
Anyway, I am not expressing preference as I am sure you know what feels right, but just expressing thoughts on how we could salvage the existing API to improved and thread-safe semantics.

jemc:
I like @somdoron's "picture message" solution, as it actually gives applications more support than current multipart (encoding integers and such instead of just strings), while also avoiding the need for an array/vector/multipart struct.

The big sticking point for me would be that we need to document the "picture message" encoding scheme and keep it stable so that interoperable implementations can share in the elegant solution.

hitstergtd:
@jemc - I don't believe I ever condoned exposing C++ stuff to the API; rather, I concur with you. :-)

sunkin351:
Does having all threadsafe sockets mean that you can have one context for the entire application of a multi-threaded environment?

somdoron:
@sunkin351 context was already thread safe, also context is kind of deprecated if you are using CZMQ. Thread safe sockets can be shared between sockets, although it is not always recommended.

sunkin351:
I see, that could be where I could improve my applications, because I have been using a different context for each thread. whoops. XD

PR #1936, opened 2016-04-30
[WIP] Enable UDP transport support for CLIENT and SERVER protocols

Bitiquinho:
I'm trying to add UDP transport support for CLIENT/SERVER connections, based on the previous RADIO/DISH work by @somdoron (thanks a lot for it), as it would be useful for me (should scratch my own itch, right ?).

So far, I have only changed what seems obvious, but there is some specific code on radio.hpp and radio.cpp to handle UDP pipes that I don't quite understand. 

Could you point me what still needs to be changed for this to work ?

Sorry for possibly weird English and thanks in advice.

### DIFF
--- a/src/session_base.cpp
+++ b/src/session_base.cpp
@@ -572,7 +572,7 @@ void zmq::session_base_t::start_connecting (bool wait_)
 #endif
 
 if (addr->protocol == "udp") {
-    zmq_assert (options.type == ZMQ_DISH || options.type == ZMQ_RADIO);
+    zmq_assert (options.type == ZMQ_DISH || options.type == ZMQ_RADIO || options.type == ZMQ_CLIENT || options.type == ZMQ_SERVER);
 
     udp_engine_t* engine = new (std::nothrow) udp_engine_t ();
     alloc_assert (engine);
@@ -586,6 +586,10 @@ if (addr->protocol == "udp") {
     }
     else if (options.type == ZMQ_DISH) {
         send = false;
+        recv = true;
+    }
+    else if (options.type == ZMQ_CLIENT || options.type == ZMQ_SERVER) {
+        send = true;
         recv = true;
     }
 
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -308,7 +308,9 @@ int zmq::socket_base_t::check_protocol (const std::string &protocol_)
 #endif
 
     if (protocol_ == "udp" && (options.type != ZMQ_DISH &&
-                               options.type != ZMQ_RADIO)) {
+                               options.type != ZMQ_RADIO &&
+                               options.type != ZMQ_CLIENT &&
+                               options.type != ZMQ_SERVER)) {
         errno = ENOCOMPATPROTO;
         return -1;
     }
@@ -882,7 +884,7 @@ int zmq::socket_base_t::connect (const char *addr_)
 if (protocol  == "udp") {
     paddr->resolved.udp_addr = new (std::nothrow) udp_address_t ();
     alloc_assert (paddr->resolved.udp_addr);
-    rc = paddr->resolved.udp_addr->resolve (address.c_str(), options.type == ZMQ_DISH);
+    rc = paddr->resolved.udp_addr->resolve (address.c_str(), (options.type == ZMQ_DISH || options.type == ZMQ_CLIENT || options.type == ZMQ_SERVER));
     if (rc != 0) {
         LIBZMQ_DELETE(paddr);
         EXIT_MUTEX ();


bluca:
@Bitiquinho - would you like to keep this PR open to get indications? If so I won't merge, and I suggest to add [WIP] to the title so it doesn't get merged :-)

somdoron:
I think it is close to impossible to do that (or the complexity is very high), but we can try.
By the way it will be easy to support scatter-gather, but client-server it two way communication.

UDP is not reliable, that is nice for pubsub,  because you don't really care if the other side gets it or not. But for client-server that is another story, for example, if you are sending a message from client to server and the server is just down you have no way to know and then if you wait for a reply you might never get it. If you will try to develop reliability over it you probably just implementing TCP.(or UDT or SCTP) 

The problem is that UDP is not connected, there is no notion of session like in TCP and in client-server, as we need session for client-server we have to implement this our own, in client-server the session id is probably the IP of the remote peer, so we need to create a session for each remote peer (when we bind). I might even use different code base for client-server as it is not very close to radio/dish. So in the udp_engine, when the endpoint is the bound one, for every new IP address received you need to create new session (how do you kill sessions? probably inactivity timeout, as there is no notion of session closed like in TCP) and session should create new udp_engine for this specific IP address and port.

Bottom line this will be a lot of work, it is possible, but the price is high (this is way a lot of failed before), my approach to solve this was to only support pubsub (radio/dish) which is much simpler. 

If you want we can support gather-scatter which will much more simple, then you can use pair of scatter-gather for two way communication.

By the way, why do you need UDP with client-server? what is the use case?

Bitiquinho:
@bluca Thanks for the tip. Added it.

Bitiquinho:
@somdoron Thanks for your detailed answer. I really didn't know it was a much more complex issue. I only saw the similarities between both protocols (thread-safe, no snd_more/rcv_more) and judged it would be relatively unharmful to add it, as the underlying infrastructure was already set up. My mistake.

I actually have two use cases: 

- Low-latency robot control over network: I have a real-time system as a controller for robotic actuators, and it runs a server through which remote clients can access the joint sensors measurements (position, velocity, torque) or define joint setpoints that enter robot control. As the update messages are sent continuously in both ways, there is no problem in dropping some packets. So I'm going for UDP (I have custom code by now) for lower latency and transparency.

- Common client-server game architecture: I'm building a centralized (with master server) multiplayer game system. Update messages (containing objects position, velocity, etc.) are update continuously as well. So my needs are pretty much the same as the first case.

Bitiquinho:
Couldn't just be assumed that, by choosing UDP, the user is willing to do unreliable messaging ?

somdoron:
In which scenario can you use multicast?

One of way to simplify the UDP transport is to avoid handshake, because of that if you are doing unicast UDP the sender must connect to the receiver, so in case of a server publishing messages the server has to connect to all clients, therefore know the IP address and port of each dish, and the dish has to bind the port. This can be complicated in WAN environments when you sometime need to punching hole, which the current implementation doesn't support.

So if you are over WAN the current UDP transport might not work for you. 

For the first scenario if you are using it on LAN or WIFI you can use Radio-Dish with multicast, you can have to groups, one is for commands on one for updates, so only the server listen for the command group.

For the second scenario, you can use UDP because it from the outside world to the server, so you can have UDP unicast, the client is the radio and the server is the dish, the radio connects to the dish.

somdoron:
So you can use UDP with radio-dish for low latency and client-server for commands.

somdoron:
> Couldn't just be assumed that, by choosing UDP, the user is willing to do unreliable messaging ?

You still have the session problem, so server has routing id notion, when using UDP all the incoming messages seems to come from same routing id and you have no way to reply. Implementing the session above UDP, so each IP & Port uniqueness will look like different routing id is very complicated.

My guess is that previous attempt failed on trying to do handshake and implementing session like behavior. Avoiding those two is what make this possible.

Bitiquinho:
This really clarifies things. Thanks a lot for all this information.

Multicast would be nice in both scenarios, as the game case uses a [semi-]authoritative server that enforce consistency of objects position across all clients.

But I still need to receive on both servers joint setpoints/player inputs from all clients on similar rate that the servers send joint measures/object positions.

Could it be that the best way to do what I am trying to achieve is to combine a radio-dish[es] (multicast messages from server to clients) with a scatter[ers]-gather (input messages from clients to server) ?

somdoron:
That make sense.

Also remember that Unicast is just another case of multicast, you can do it all with multicast without even gather-scatter, like so:

1. Multicast network, some nodes are radio, some dishes and some both.
2. Use two groups, one for pushing updates to server and one by the server pushing commands, it can be Players and Server
3. Server publish messages on the Server group and listening on the Player group
4. Players listen on the Server group and push changes to the Player group

somdoron:
@Bitiquinho can I close the PR?

Bitiquinho:
@somdoron Yes, please. Thanks a lot for all tour help. I'll try what you recommended.

PR #1937, opened 2016-04-30
Problem: multiple issues with zmq_sendiov/recviov

hitstergtd:
**Solution:**
- Update `zmq_sendiov()` signature to track number of messages that were
successfully sent; this makes it consistent with `zmq_recviov()`

- Add relevant notes on top of zmq_sendiov() to reflect updated behaviour

- Add check for the `count` parameter in `zmq_sendiov()` and `zmq_recviov()`

- Add test for `zmq_sendiov()` in `tests/test_iov.cpp`

- Cleanup `tests/test_iov.cpp` for style, consistency, clarity and simplicity

**Notes:**
* Given that these methods are not really "official" in terms of API, at least according to `zmq.h`, I was not worried about breakage; but if that is an issue then we can hold off the merge and discuss it further

* In the future need to add tests for the following combinations: send+recviov, sendiov+recv

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -452,6 +452,11 @@ int zmq_sendiov (void *s_, iovec *a_, size_t count_, int flags_)
         errno = ENOTSOCK;
         return -1;
     }
+    if (unlikely (count_ <= 0 || !a_)) {
+        errno = EINVAL;
+        return -1;
+    }
+
     int rc = 0;
     zmq_msg_t msg;
     zmq::socket_base_t *s = (zmq::socket_base_t *) s_;
@@ -555,6 +560,11 @@ int zmq_recviov (void *s_, iovec *a_, size_t *count_, int flags_)
         errno = ENOTSOCK;
         return -1;
     }
+    if (unlikely (!count_ || *count_ <= 0 || !a_)) {
+        errno = EINVAL;
+        return -1;
+    }
+
     zmq::socket_base_t *s = (zmq::socket_base_t *) s_;
 
     size_t count = *count_;
--- a/tests/test_iov.cpp
+++ b/tests/test_iov.cpp
@@ -39,47 +39,88 @@ struct iovec {
 };
 #endif
 
-void do_check(void* sb, void* sc, unsigned int msgsz)
+void do_check(void* sb, void* sc, size_t msg_size)
 {
-    setup_test_environment();
-    int rc;
-    int sum =0;
-    for (int i = 0; i < 10; i++)
+    assert (sb && sc && msg_size > 0);
+
+    int rc = 0;
+    const char msg_val = '1';
+    const int num_messages = 10;
+    size_t send_count, recv_count;
+
+    send_count = recv_count = num_messages;
+
+    char *ref_msg = (char *) malloc (msg_size);
+    assert (ref_msg);
+    memset (ref_msg, msg_val, msg_size);
+
+    // zmq_sendiov(3) as a single multi-part send
+    struct iovec send_iov[num_messages];
+    char *buf = (char *) malloc (msg_size * num_messages);
+
+    for (int i = 0; i < num_messages; i++)
     {
-        zmq_msg_t msg;
-        zmq_msg_init_size(&msg, msgsz);
-        void * data = zmq_msg_data(&msg);
-        memcpy(data,&i, sizeof(int));
-        rc = zmq_msg_send(&msg,sc,i==9 ? 0 :ZMQ_SNDMORE);
-        assert (rc == (int)msgsz);
-        zmq_msg_close(&msg);
-        sum += i;
+        send_iov[i].iov_base = &buf[i * msg_size];
+        send_iov[i].iov_len = msg_size;
+        memcpy (send_iov[i].iov_base, ref_msg, msg_size);
+        assert (memcmp (ref_msg, send_iov[i].iov_base, msg_size) == 0);
     }
 
-    struct iovec ibuffer[32] ;
-    memset(&ibuffer[0], 0, sizeof(ibuffer));
-    
-    size_t count = 10;
-    rc = zmq_recviov(sb,&ibuffer[0],&count,0);
-    assert (rc == 10);
+    // Test errors - zmq_recviov - null socket
+    rc = zmq_sendiov (NULL, send_iov, send_count, ZMQ_SNDMORE);
+    assert (rc == -1 && errno == ENOTSOCK);
+    // Test errors - zmq_recviov - invalid send count
+    rc = zmq_sendiov (sc, send_iov, 0, 0);
+    assert (rc == -1 && errno == EINVAL);
+    // Test errors - zmq_recviov - null iovec
+    rc = zmq_sendiov (sc, NULL, send_count, 0);
+    assert (rc == -1 && errno == EINVAL);
 
-    int rsum=0;
-    for(;count;--count)
+    // Test success
+    rc = zmq_sendiov (sc, send_iov, send_count, ZMQ_SNDMORE);
+    // The zmq_sendiov(3) API method does not follow the same semantics as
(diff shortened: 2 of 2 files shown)


hitstergtd:
Some thoughts on these methods:
* I am not quite sure how to deal with `zmq_sendiov(..., 0)` and `zmq_recviov(..., 0)`, as the present behaviour does not seem intuitive. Maybe I am wrong, thoughts are welcome.

* The use of `struct iovec` in the public API should be changed to our own `struct zmq_iovec`; initially it would look like `struct iovec`, but if it needs to change in the future, it becomes easier for us.

* Rename these methods to `zmq_msg_send_multiple()` and `zmq_msg_recv_multiple()`, as they seem more intuitive then the current names.

hitstergtd:
I will add manual pages for these methods soon.

include/zmq.h
```diff
@@ -550,7 +550,7 @@ ZMQ_EXPORT void zmq_atomic_counter_destroy (void **counter_p);
 
 struct iovec;
 
-ZMQ_EXPORT int zmq_sendiov (void *s, struct iovec *iov, size_t count, int flags);
+ZMQ_EXPORT int zmq_sendiov (void *s, struct iovec *iov, size_t *count, int flags);
```
somdoron: this is breaking the API, are you sure we want to do this?

somdoron:
@hitstergtd I'm a little worried with breakage of the API...

bluca:
I agree, we should try to avoid breaking APIs without ample warning (deprecation and so on) unless absolutely necessary (ie: everything's horribly broken). What about adding a new API call that has the behavior you want?

hitstergtd:
@somdoron, @bluca,
I agree, it breaks the API, but as far as the code history and the comments are indicating, these functions don't seem to be part of the official API; they do not even have a man page yet.

The good thing is, I am only breaking zmq_sendiov; as I am not sure that it was even right to begin with; if I am correct it returned and arbitrary number of bytes as its return value which doesn't even make sense in an iovec context. The change I made actually brings it in line with recviov.

Anyway, I am not precious about it, so if you guys think from your experience that this will lead to a lot of breakage then I have no problem creating a new set of methods called *zmq_msg_send/recv_multiple*, because those names would actually make more sense.

bluca:
The problem is that anything that is in the public header is fair game, despite what a comment might say. For distribution packagers, if a public function signature changes, then it's an API breakage, which means the new version must de-facto create a separate new library etc etc.

I realize this can sometimes (often!) be a pain for library developers, but on the other hand if a library breaks API compatibility too often then users get annoyed and switch to more stable alternatives.
We do have a deprecation process, that takes intentionally a long time, to make transition as smooth as possible.

hitstergtd:
Thanks @bluca. That seems like a serious PITA. Anyway, I agree with you for anything that makes sense except that the zmq_sendiov semantics just seem broken. Nevertheless, I am re-working the Pull now to use a `zmq_msg_send/recv_multiple` form, and deprecating the old methods.

bluca:
Thanks!

hitstergtd:
No problem at all, I was on the fence too. I also want to add some more GitHub labels for Pull Requests.

hitstergtd:
@bluca, @somdoron 
Since we are on the topic of sends and receives:

* What about the truncation of messages greater than `MAXMSGZ` in size?
* Instead of truncating, should we not return meaningful errors, such as `EMSGSIZE`, `ERANGE`?
* I am referring to the semantics of `s_sendmsg` and `s_recvmsg` (see: `src/zmq.cpp`)

hitstergtd:
OK. For the latest commit in the Pull:

**Solution:**
- Add check for the `count` parameter in `zmq_sendiov()` and `zmq_recviov()`
- Use and add test for `zmq_sendiov()` in `tests/test_iov.cpp`
- Add error state tests for `zmq_sendiov()` in `tests/test_iov.cpp`
- Add error state tests for `zmq_recviov()` in `tests/test_iov.cpp`
- Cleanup `tests/test_iov.cpp` for style, consistency and clarity
- Generally improve test coverage for both API methods

**Hat-tip:**
@somdoron, @bluca

bluca:
Nice, especially the additional tests, thanks!
Minor nit, but would you mind please rebasing -i and dropping the first 2 commits, rather than having a revert?

bluca:
Regarding the max size, I'd have to check, but I think I remember this behaviour of truncating is coded in the ZMTP RFC. Is that the case?

hitstergtd:
Thanks @bluca. Ok point taken, I will clean it up and send it across later today!

hitstergtd:
OK, done. Cleaned up all the old commits and forced pushed a fresh commit. :)

hitstergtd:
With regard to ZMTP, I checked the RFC, and I could be wrong, but it seems that message data handling have been deliberately left out of the semantics:

(RFC ZMTP 3.1)[http://rfc.zeromq.org/spec:37]:
> **Messages**
> Messages carry application data and are not generally created, modified, or filtered by the ZMTP implementation except in some cases. Messages consist of one or more frames and an implementation SHALL always send and deliver messages atomically, that is, all the frames of a message, or none of them.

The wording is the same for ZMTP 3.0 and 2.0 has no mention at all on send/receive level semantics. The emphasis is mostly on framing, atomicity and socket combinations.

hitstergtd:
Thanks for the merge!

bluca:
Ok, thanks for checking. The reason I thought it was, I think, it's because on the receive side it's enforced straight in the protocol decoder.

hitstergtd:
No worries. Ah ok, I didn't know that. Moreover, whilst the protocol is quite descriptive, I do not believe all the semantics are fully captured. So if truncation semantics are relied upon, even though I would not have thought so, then we should attempt to leave them the way it is for the sake of eliminated API breakage.

I will see what @somdoron also has to say.

hitstergtd:
@bluca, @somdoron -
Here is the note in `zmq.h`; I will send PR to change these to indicate that they are deprecation candidates.

```cpp
/******************************************************************************/
/*  These functions are not documented by man pages -- use at your own risk.  */
/*  If you need these to be part of the formal ZMQ API, then (a) write a man  */
/*  page, and (b) write a test case in tests.                                 */
/******************************************************************************/

struct iovec;

ZMQ_EXPORT int zmq_sendiov (void *s, struct iovec *iov, size_t count, int flags);
ZMQ_EXPORT int zmq_recviov (void *s, struct iovec *iov, size_t *count, int flags);
```

somdoron:
@hitstergtd the code doesn't really truncate the msg, it just change the returned size in case it is larger than maximum positive integer.

hitstergtd:
Agreed. I should have been clear about that. I just saw Pieter's comments on top of s_sendmsg about returning `options->get(MAXMSGSZ)` instead of INT_MAX, and went down this spiral!

hitstergtd:
It would be good to add into s_sendmsg and s_recvmsg logic to stop people sending messages bigger than the set maximum (MAXMSGSZ) and that's one of the things I am working towards.

somdoron:
i'm not sure zmq.cpp is the correct place, it should be internally inside the library.
I'm pretty sure it is being checked somewhere.

Anyway max msg size can be larger than INT_MAX, so the code is still correct

somdoron:
also from the docs MAXMSGSZ is only to filter inbound message.

somdoron:
https://github.com/zeromq/libzmq/blob/de56bc272995c6584d68d78996aa2d92d5b9133b/doc/zmq_getsockopt.txt#L376

bluca:
Yes that's what I meant when I wrote it's done in the decoder:

https://github.com/zeromq/libzmq/blob/master/src/v2_decoder.cpp#L101

```
        if (unlikely (msg_size > static_cast <uint64_t> (maxmsgsize))) {
            errno = EMSGSIZE;
            return -1;
        }
```

hitstergtd:
OK. Thanks guys!

PR #1945, opened 2016-05-02
problem: no way to disable draft when compiling

### LINKED ISSUE #1942: problem: no way to disable draft when compiling
As part of preparing for version 4.2 we need a way to compile zeromq without new draft addition to the API.

Following are new and draft addition to libzmq:

* zmq_poller CLASS
* zmq_timers CLASS
* zmq_join and zmq_leave methods

We also need to disable the new socket types: gather, scatter, client, server, dish, radio. 
We can define them in zmq.h only if draft macro is defined, if draft macro is not defined we can include them in the precompiled.hpp so internal code will still compile.

Lastly we need a way to define the macro from autotools, cmake and visual studio. I don't have much experience with autotools and cmake and appreciate help with this.

bluca:
Solution: add DRAFT option to CMake and Autotools (see commits).
Fixes #1942 

Need help from the Windows gurus for VS201X and others :-)
Changed half the test runs to build with drafts, left some without to make sure we don't break anything.

Also tested building CZMQ against both with and without drafts, all looks good. Next, we need a way in zproject to support enable/disable of drafts in dependencies for CI jobs.

Besides zmq_pollers*/timers*/client/server/radio/dish/scatter/gather, other APIs that should be under DRAFT?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,6 +130,7 @@ test_scatter_gather
 test_socketopt_hwm
 test_use_fd_ipc
 test_use_fd_tcp
+test_pub_invert_matching
 tests/test*.log
 tests/test*.trs
 src/platform.hpp*
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,19 +9,19 @@ os:
 dist: trusty
 
 env:
-- BUILD_TYPE=default CURVE=tweetnacl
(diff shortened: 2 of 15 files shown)


somdoron:
very nice!

I forgot to mention the new zmq_msg methods (group and routing id methods).
I will sent a pull request to add those and give VS a try.

PR #1946, opened 2016-05-03
problem: no way to enable draft on windows

somdoron:
+ problem: zmq_msg new methods are not marked as draft

### DIFF
--- a/src/precompiled.hpp
+++ b/src/precompiled.hpp
@@ -160,9 +160,15 @@
 #define ZMQ_GATHER 16
 #define ZMQ_SCATTER 17
 
-/*  DRAFT Socket events.                                                      */
+/*  DRAFT Socket methods.                                                     */
 int zmq_join (void *s, const char *group);
 int zmq_leave (void *s, const char *group);
+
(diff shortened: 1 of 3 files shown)


src/precompiled.hpp
```diff
@@ -160,10 +160,16 @@
...
 #define ZMQ_SCATTER 17
 
-/*  DRAFT Socket events.                                                      */
+/*  DRAFT Socket methods.                                                     */
 int zmq_join (void *s, const char *group);
 int zmq_leave (void *s, const char *group);
 
+/*  DRAFT Msg methods.                                                        */
+ZMQ_EXPORT int zmq_msg_set_routing_id(zmq_msg_t *msg, uint32_t routing_id);
+ZMQ_EXPORT uint32_t zmq_msg_routing_id(zmq_msg_t *msg);
```
bluca: Should be without the ZMQ_EXPORT since, well, they're not for export :-)

somdoron: fixed

somdoron:
@bluca fixed and rebased.

PR #1948, opened 2016-05-03
Problem: Not all GCOV extensions in .gitignore

hitstergtd:
Solution:
Fix it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ aclocal.m4
 autom4te.cache
 *.o
 *.gcno
+*.gcda
+*.gcov
 *.ncb
 *.lo
 *.loT


hitstergtd:
Thanks!

hitstergtd:
How is it that Coveralls gets all excited about non source code commits?

bluca:
It's very pedantic! :-)
Not sure if it can be fixed, couldn't see any settings. It will run on any build since it's on the CI, and just report what it finds.

I already tried to set the threshold for reporting but that only affects the test failing/passing, since there's obviously some rounding errors that make it report changes where there are actually none.

hitstergtd:
Clearly. Thanks for the explanation. I am sure you were wondering if I had better PRs to spend time on instead of worrying about Coveralls' rounding errors and configurative inadequacies. :-)

PR #1953, opened 2016-05-04
problem: src is broken and unneeded as we have metadata

### LINKED ISSUE #1950: Problem: ZMQ_SRCFD is broken from the beginning and takes precious space in zmq_msg structure
the implementation from following commit only works if socket is bound and have only one connected peer.

When more than one peer the return FD is actually the FD of the last connected peer.

https://github.com/zeromq/libzmq/commit/f30fb8501ef845eae9a44c032813e5c45e189ac4

my suggest solution is remove it, to unbreak the API we can just return -1 when some try to get the SRCFD.

@sterad are you okay with us removing this? are you using it?

@hintjens do you think it is ok to remove the implementation and just return -1 and set errno to not break the API?

somdoron:
PR solve the broken SRCFD by using metadata to store the file descriptor.
However as we now have metadata which include the peer address we don't need the SRCFD, therefore marked SRCFD as deprecated. If someone need more metadata I suggest adding it instead of using the SRCFD. 

Fixes #1950

### DIFF
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -87,7 +87,6 @@ int zmq::msg_t::init ()
     u.vsm.size = 0;
     u.vsm.group[0] = '\0';
     u.vsm.routing_id = 0;
-    u.vsm.fd = retired_fd;
     return 0;
 }
 
@@ -100,7 +99,6 @@ int zmq::msg_t::init_size (size_t size_)
         u.vsm.size = (unsigned char) size_;
         u.vsm.group[0] = '\0';
         u.vsm.routing_id = 0;
-        u.vsm.fd = retired_fd;
     }
     else {
         u.lmsg.metadata = NULL;
@@ -108,7 +106,6 @@ int zmq::msg_t::init_size (size_t size_)
         u.lmsg.flags = 0;
         u.lmsg.group[0] = '\0';
         u.lmsg.routing_id = 0;
-        u.lmsg.fd = retired_fd;
         u.lmsg.content = NULL;
         if (sizeof (content_t) + size_ > size_)
             u.lmsg.content = (content_t*) malloc (sizeof (content_t) + size_);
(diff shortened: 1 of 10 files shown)


src/msg.cpp
```diff
@@ -86,8 +86,7 @@ int zmq::msg_t::init ()
     u.vsm.flags = 0;
     u.vsm.size = 0;
     u.vsm.group[0] = '\0';
-    u.vsm.routing_id = 0;
-    u.vsm.fd = retired_fd;
+    u.vsm.routing_id = 0;
```
bluca: there's trailing whitespace here and in other places below - want to fix before merge? or in a new PR?

somdoron: I will fix it, this is what happen when I'm using visual studio :-(

src/msg.cpp
```diff
@@ -137,8 +134,7 @@ int zmq::msg_t::init_external_storage(content_t* content_, void* data_, size_t s
     u.zclmsg.flags = 0;
     u.zclmsg.group[0] = '\0';
     u.zclmsg.routing_id = 0;
-    u.zclmsg.fd = retired_fd;
-
+
```
bluca: Missed one :-)

src/msg.cpp
bluca: Ok I'll wait then, thanks!

somdoron:
@bluca should be good now

bluca:
Mmh OSX (or rather Clang) is not happy:

```
../src/stream_engine.cpp:990:62: error: cannot take the address of an rvalue of type 'std::__1::basic_ostringstream<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
    std::string fd_string = static_cast<std::ostringstream*>(&(std::ostringstream() << (int)s))->str();
```

Do you have Clang available to give it a look?

somdoron:
I have osx vm, checking...

somdoron:
fixed #1954

PR #1955, opened 2016-05-04
Problem: MAINTAINERS file has no purpose anymore

hitstergtd:
**Solution:**
Remove until such time that a new and appropriate file can replace it should it be deemed necessary.


bluca:
You need to remove it from the list in Makefile.am too, otherwise the build breaks

hitstergtd:
Ouch. The 4th wasn't with me @bluca. Let me fix it.

hitstergtd:
Done.

hitstergtd:
Thanks!

PR #1956, opened 2016-05-04
Problem: Not shamelessly plugged into AUTHORS

hitstergtd:
Solution:
Plug myself in, should the commit and BDFL overlords approve.

Special-copy-to:
@hintjens, @somdoron, @bluca.


bluca:
This is part of the C4.1 process, so no problem :-)

PR #1958, opened 2016-05-04
Problem: zmq_sendiov/zmq_recviov not Deprecated

hitstergtd:
**Solution:**
As preparation for 4.2 release, move the zmq_sendiov and zmq_recviov API methods under the Deprecated Methods section.

**Note:**
The actual methods have NOT been deprecated yet, functionally speaking however it is good to let API users know early. Moreover, these methods were not ever considered stable, at least according to `src/zmq.h`, and have no associated man pages.


hitstergtd:
Thanks!

PR #1960, opened 2016-05-05
Problem: AUTHORS file is not sorted

hitstergtd:
**Solution:**
Sort the list and strip the trailing whitespace.


hitstergtd:
Thanks!

Issue #1930, opened 2016-04-27
Assertion failed: nbytes == sizeof dummy (src/signaler.cpp:202)


paddor:
I'm getting the following error most every time (not always) when running CZTop's test suite on Rubinius at process exit:

```
Assertion failed: nbytes == sizeof dummy (src/signaler.cpp:202)
```

Then the process hangs and hogs the CPU. Any ideas?

somdoron:
What socket type are you using ? Which os? any chance you are using the
socket from multiple threads?

paddor:
Both OSX and Linux. See [here](https://travis-ci.org/paddor/cztop/jobs/126188627). I'm not using the same socket from different threads.

somdoron:
can you provide stack trace?

paddor:
It seems to happen because of the poller specs (CZTop::Poller is based purely on `zmq_poller_*()`). I start the poller specs alone using: `/Users/paddor/.rubies/rbx-brew/bin/rbx -I/Users/paddor/.gem/rbx/2.2.2/gems/rspec-support-3.4.1/lib:/Users/paddor/.gem/rbx/2.2.2/gems/rspec-core-3.4.2/lib /Users/paddor/.gem/rbx/2.2.2/gems/rspec-core-3.4.2/exe/rspec spec/cztop/poller_spec.rb` and it happens sometimes.

The sockets used in the poller specs are mainly PUSH and PULL, but also some DEALER. And one spec uses SERVER/CLIENT. See [here](https://github.com/paddor/cztop/blob/master/spec/cztop/poller_spec.rb).

I'm working on getting a stack trace. Just recompiling ZMQ and CZMQ from master, then running `rbx -d ...` to get more debug information.

paddor:
Sorry, can't get a stack trace from Rubinius, because when this error happens, it just hangs and doesn't react to anything. No stack trace when I `kill -9` it.

somdoron:
that going to be hard...

can you try to reproduce with branch? I'm trying to figure out the errno

https://github.com/somdoron/libzmq/tree/Signaler

hitstergtd:
@paddor / @somdoron,
Perhaps a partial dtrace (OS X) or strace (Linux) can help with this issue?

somdoron:
@paddor few questions just to make sure:

1. Are all sockets disposed before the process exits?
2. The poller get disposed before the process exits?

somdoron:
3. Are you closing the socket before removing it from the poller?

[...]

PR #1961, opened 2016-05-05
problem: closed poller still associated with socket

somdoron:
If thread socket was closed before poller the socket hold a reference to closed signaler.

solution: when closing a socket remove reference to existing pollers.

solves #1930

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -1281,6 +1281,12 @@ int zmq::socket_base_t::recv (msg_t *msg_, int flags_)
 
 int zmq::socket_base_t::close ()
 {
+    ENTER_MUTEX ();
+    
+    //  Remove all existing signalers for thread safe sockets
+    if (thread_safe)
+        ((mailbox_safe_t*)mailbox)->clear_signalers();
+
     //  Mark the socket as dead
     tag = 0xdeadbeef;
 
@@ -1289,6 +1295,7 @@ int zmq::socket_base_t::close ()
     //  process.
     send_reap (this);
 
+    EXIT_MUTEX ();
     return 0;
 }
 
--- a/src/socket_poller.cpp
+++ b/src/socket_poller.cpp
@@ -186,19 +186,14 @@ int zmq::socket_poller_t::remove (socket_base_t *socket_)
         return -1;
     }
 
+    items.erase(it);
+    need_rebuild = true;
+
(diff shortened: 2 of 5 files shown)


src/socket_poller.cpp
```diff
@@ -186,19 +186,14 @@ int zmq::socket_poller_t::remove (socket_base_t *socket_)
...
-
-    if (thread_safe) {
-        if (socket_->remove_signaler (&signaler) == -1)
-            return -1;
-    }
-
-    items.erase (it);
-    need_rebuild = true;
+    if (socket_->getsockopt (ZMQ_THREAD_SAFE, &thread_safe, &thread_safe_size) == 0 && thread_safe)
+        socket_->remove_signaler (&signaler);
```
bluca: Before these 2 calls were checked for error and -1 returned, is that no longer necessary now?

somdoron: The change is that we allow to remove a dead socket.
So we allow the -1 from both and just ignore it.

somdoron: or socket is in the poller set but not associated anymore because socket was closed

src/socket_base.cpp
```diff
@@ -1289,6 +1295,7 @@ int zmq::socket_base_t::close ()
     //  process.
     send_reap (this);
 
+    EXIT_MUTEX ();
```
bluca: Does the EXIT_MUTEX need to be after the send_reap?

The problem is that it causes read-after-free, as shown by Valgrind tests now failing. I guess the mutex object is freed before it returns.

Moving the mutex release before the send_reap makes the tests happy. Is it the right thing to do?

somdoron: in general the EXIT_MUTEX should be before returning from the method.

read-after-free can only happen in non-thread-safe socket and then it doesn't matter as the mutex is not being used. however I understand we need to solve it, I think adding the following to the socket_base destructor can solve the problem:

```C
sync.lock ();
sync.unlock ();
```

Another solution is to actually remove the mutex from the close method, leaving it unsafe by design (which kind of unsafe).

Also we can just move the EXIT MUTEX before the send_reap, I don't see much harm in that.

bluca: Adding sync.lock/unlock to the destructor doesn't seem to make a difference. I'll send a PR to move the unlock earlier.

PR #1962, opened 2016-05-05
problem: timers test is flapping on windows

somdoron:
please let it run


somdoron:
test passed on jenkins windows

can be merged

PR #1970, opened 2016-05-08
Problem: Not all ancillary API methods tested

hitstergtd:
**Solution:**
- Add file for testing ancillary API methods and any misc internal machinery
- Add tests for zmq_version(3) and zmq_strerror(3)
- Add test file into gitignore, Autotools and CMake build files
- Increase test coverage

**Note:**
MSVC solution files not updated.


hitstergtd:
Thanks!

PR #1971, opened 2016-05-09
Problem: Deploying release artifacts to github is a manual process

sappo:
Solution: Use travis to deploy these artifacts automatically.

The deployment is triggered by tagging on the zeromq/libzmq repository.
Of the many builds travis is checking only the default one with
libsodium and drafts disabled is used for deployment.

For now the results of `make distcheck` are deployed as well as their
md5 and sha1 hash sums. Further changes may upload a generated
Changelog as well.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -x
+set -e
 
 if [ $BUILD_TYPE == "default" ]; then
     mkdir tmp
@@ -29,7 +30,7 @@ if [ $BUILD_TYPE == "default" ]; then
         CONFIG_OPTS+=("--enable-drafts=yes")
     fi
 
-    #   Build and check this project
+    # Build and check this project
     (
         ./autogen.sh &&
         ./configure "${CONFIG_OPTS[@]}" &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,15 @@ env:
 
 matrix:
   exclude:
-    -  env: BUILD_TYPE=coverage CURVE=tweetnacl DRAFT=enabled
-       os: osx
-    -  env: BUILD_TYPE=valgrind CURVE=tweetnacl DRAFT=enabled
-       os: osx
+  - env: BUILD_TYPE=coverage CURVE=tweetnacl DRAFT=enabled
+    os: osx
+  - env: BUILD_TYPE=valgrind CURVE=tweetnacl DRAFT=enabled
+    os: osx
   include:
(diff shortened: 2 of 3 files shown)


ci_build.sh
```diff
@@ -29,13 +30,12 @@ if [ $BUILD_TYPE == "default" ]; then
...
-        ./configure "${CONFIG_OPTS[@]}" &&
-        export DISTCHECK_CONFIGURE_FLAGS="${CONFIG_OPTS[@]}" &&
-        make VERBOSE=1 distcheck
-    ) || exit 1
+    # Build and check this project
+    ./autogen.sh
+    ./configure "${CONFIG_OPTS[@]}"
+    make -j5
+    make check
+    make install
```
bluca: Is this change necessary? Dist-check does the above and more (most importantly it checks that make dist works), so won't we lose coverage?

sappo: I moved the dist-check to ci_deploy.sh but if it's worth running it to check the build I happily revert it.

bluca: I think so, otherwise we won't notice if make dist is broken until we try to do a release, which is not nice :-)

sappo:
@bluca seems like by pushing the update I deleted our conversation. The original ci_build behavior is restored now.

bluca:
Yes the comments on outdated diffs are hidden. At least now you can click on it and they will reappear, it used to be that they disappeared forever.

Anyway, this is great stuff, thanks!

sappo:
I'll create a temporary tag now, to check if everything works as expected.

bluca:
It worked! Awesome :-)

sappo:
I'd approximately 60 travis builds on my fork to get it to to that :-D

hitstergtd:
@sappo, thanks! looks good. Also, it would be good to post snapshot releases of the merge builds, so that in the future if we need to narrow down a commit that caused weird performance issues, it would just be a matter of downloading a snapshot release and then running the appropriate scripts. This was suggested in one of the issues recently.

bluca:
git is better for that kind of work via git bisect, doing it via manually downloading hundreds of tarballs would be the stuff of nightmares!

hitstergtd:
It might help you narrow down to a set of commits, but you still have to build them. Having those commits prebuilt makes things quicker to compare.

bluca:
They are not prebuilt. This is not shipping binaries - that would be useless (and wrong) on Linux. It's shipping distributable tarballs.

hitstergtd:
Ah, ok.

PR #1975, opened 2016-05-10
problem:Socket try to access closed mutex

somdoron:


somdoron:
solves #1930

bluca:
@somdoron - seems some tests are borken: https://travis-ci.org/zeromq/libzmq/builds/129103961
any idea what the problem might be?

somdoron:
@bluca yes, patch is wrong,sent a revert patch

PR #1977, opened 2016-05-10
Problem: sockfd_ parameter throws unused warning

hitstergtd:
**Solution:**
Mark it with `LIBZMQ_UNUSED` macro as per our convention.


hitstergtd:
Thanks!

PR #1982, opened 2016-05-11
tipc: add support for address domain suffix

Hugne:
The TIPC protocol bindings in ZeroMQ defaults to a lookup domain
of 1.0.0 to prevent 'closest first' search, and instead always
do round robin if several sockets in the network or node have
the same name published. In retrospect, this might have been a
bad idea because it won't work on standalone configurations.
We solve this by allowing an optional domain suffix to be provided
in the address, and 0.0.0 should be used in that case, or if the
TIPC address range in the cluster configuration is defined to some
other value. Domain suffixes are only relevant for connecting
addresses.

Signed-off-by: Erik Hugne <erik.hugne@gmail.com>

### DIFF
--- a/src/tipc_address.cpp
+++ b/src/tipc_address.cpp
@@ -59,8 +59,16 @@ int zmq::tipc_address_t::resolve (const char *name)
     unsigned int type = 0;
     unsigned int lower = 0;
     unsigned int upper = 0;
-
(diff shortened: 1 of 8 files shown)


hitstergtd:
@Hugne thanks so much for this. At least this way we can ensure out of the box use for this protocol, and also makes it easy to add examples in the documentation.

PR #1983, opened 2016-05-12
Problem: scheduling parameters unused on OSX

hitstergtd:
**Solution:**
* Mark them with LIBZMQ_UNUSED macro as per convention; although in future the appropriate pthread code should be updated to support thread scheduling priorities.

* Removed some stray trailing whitespace as well.

### DIFF
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -28,6 +28,7 @@
 */
 
 #include "precompiled.hpp"
+#include "macros.hpp"
 #include "thread.hpp"
 #include "err.hpp"
 #include "platform.hpp"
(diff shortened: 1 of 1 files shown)


src/thread.cpp
```diff
@@ -116,6 +119,9 @@ void zmq::thread_t::stop ()
 
 void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_)
 {
+    LIBZMQ_UNUSED (priority_);
```
bluca: Shouldn't the go in an \#else block below? Otherwise it's confusing

hitstergtd:
@bluca 
Good idea. Done. Thanks.

PR #1984, opened 2016-05-12
Problem: No coverage for ctx termination errors

hitstergtd:
**Solution:**
- Add error state coverage for `zmq_ctx_term()`, `zmq_term()` and `zmq_ctx_shutdown()`; `zmq_ctx_destroy()` is already covered since it only calls `zmq_ctx_term()`
- Add coverage for `zmq_term()`


hitstergtd:
Thanks!

PR #1987, opened 2016-05-13
Problem: test send count too high for AppVeyor CI environment

hitstergtd:
**Solution:**
Reduce send count in CLIENT/SERVER thread-safe socket test, so that it does not timeout on AppVeyor CI environment and older Windows boxes. The former is slow, uni-processor and resource constrained.


hitstergtd:
Thanks!

Issue #1985, opened 2016-05-12
Problem: MSVC build configuration inflexible and broken.


evoskuil:
This  relates to [another issue](https://github.com/zeromq/libzmq/issues/1963) but is more extensive.

There are several problems in the MSVC builds presently:

`configure.bat` is defining `HAVE_LIBSODIUM` which is no longer used in the sources, it should define `ZMQ_USE_LIBSODIUM`.

`configure.bat` effectively *requires* use of libsodium if it is found. For this reason the GUI option for selecting use of libsodium was removed. So presently the only way to disable libsodium is to remove the sibling directory or edit the `platform.hpp`

`configure.bat` provides a command line switch in autoconfigure style for enabling or disabling `ZMQ_BUILD_DRAFT_API` and therefore there is also no GUI option provided for it.

The implementation of `configure.bat` does not support similar options for enable/disable of:
* `ZMQ_USE_TWEETNACL`
* `ZMQ_USE_LIBSODIUM`
* `ZMQ_HAVE_OPENPGM`
* `HAVE_LIBGSSAPI_KRB5`
* `ZMQ_STATIC`
* `DLL_EXPORT`

These options (with the exception of the recently-added `ZMQ_USE_TWEETNACL`) are configured via the GUI (and `ZMQ_STATIC` and `DLL_EXPORT` are also controllable via the build command line).

Finally, `ZMQ_HAVE_CURVE` should be set any time that `ZMQ_USE_TWEETNACL` or `ZMQ_USE_LIBSODIUM` is set, however there is no way for `configure.bat` to perform this configuration without knowledge of the intended NaCl configuration.

To resolve these problems `configure.bat` needs to either go away or become much more robust. If it is to become more robust, with a full set of configuration options, the GUI options will be overridden and would therefore need to be removed. Otherwise it could simply go away and `platform.hpp` can become static for Windows (as was previously the case).

The change to a dynamic `platform.hpp` was made to enable build of ZMQ *without* libsodium. The original configuration had hard-coded the enable of libsodium into the `libzmq.props` file. With this removed the option is preserved in the GUI.

The autoconfigure technique doesn't apply well in the case of MSVC projects. One of ht benefits of the solutions is the ability for the GUI to expose options. These options can be saved via the GUI or applied at the build command line. The extra step of running a `configure.bat` eliminates this benefit and due to inflexibility will require building up the same options on its command line - which is itself a UI.

[...]

PR #1988, opened 2016-05-13
Problem: msvc builds are misconfigured.

evoskuil:
See https://github.com/zeromq/libzmq/issues/1985


evoskuil:
I can't tell because of the bogus Travis failures on the last run, or the one stress test failure on the previous run, whether the build actually succeeded or not - but it seems that it has.

bluca:
Kicked the osx one, all good: https://travis-ci.org/zeromq/libzmq

hitstergtd:
@evoskuil, this is really good, thanks! It might be an idea to remove the vs2008 msvc build as discussed, before shipping version v4.2. Or do you want me to send the PR?

sunkin351:
It would be a good idea to consult the community before you stop support for vs2008.

rgbkrk:
:smile:

evoskuil:
Thanks @bluca.

@hitstergtd wrt 2008 it would be nice to get it out, but I agree with @sunkin351 that it would be a good idea to poll the community first.

PR #1991, opened 2016-05-14
Problem: zmq_poller_destroy can segfault

hitstergtd:
**Solution:**
- Add checks for `**poller_p_` parameter to ensure that the function does not segfault when either it or the value within it are NULL
- Add tests for the above and increase error state coverage

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -1110,8 +1110,9 @@ void *zmq_poller_new (void)
 
 int zmq_poller_destroy (void **poller_p_)
 {
-    void *poller = *poller_p_;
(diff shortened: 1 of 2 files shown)


sunkin351:
I wasn't aware that there was a more advanced API for the zmq socket poller. Is there documentation I could look at? Last I checked, zmq_poll() was all there was.

hitstergtd:
@sunkin351 
It is fairly recent, and it's a port from CZMQ to support convenience cases.

For your reference:
* Initial PR is #1611
* CZMQ documentation - <http://czmq.zeromq.org/manual:zpoller>

Hope that helps.

hitstergtd:
I am not sure why the Coverage build is dumping core! :-(

Issue #1989, opened 2016-05-13
Problem: Draft API is declared in precompiled header.


evoskuil:
https://github.com/zeromq/libzmq/blob/master/src/precompiled.hpp

somdoron:
What is the problem? It is also defined in zmq.h

evoskuil:
It seems to me that it would either be an unnecessary duplication of code or a confusing inconsistency. No other part of the API is declared in the pch.

bluca:
That's done by design, just like in zproject/czmq. Else, the build will break when DRAFT is disabled.

evoskuil:
@bluca I don't follow, what precludes these declarations from being placed in the headers of the files for which they are defined and conditionally excluded in those locations?

bluca:
Because then they would be duplicated X times instead of one, and some would most likely be left behind. Much clearer if they are in single well-known place. Also it's to follow on what zproject/czmq do as a model - public header + private header, ifdef + ifndef

evoskuil:
Repetition is resolved by use of one or more headers to contain the declarations. Even if there is no existing appropriate header in which to place the API, it would still seem preferable to create a place for it as opposed to overloading the pch for this purpose. We could easily resolve this by creating a `zmq_draft.h` for internal use, and including it in the pch - just as is done with the `zmq.h`

bluca:
Sure, no problem in having a different internal header. I picked that one because it was already there and included everywhere, so it resulted in less churn.

evoskuil:
Sure, that's what I figured. If it's a temporary thing I wouldn't bother. But it seems like this would be a perpetual process.

somdoron:
draft.h files is good idea. However I still think it is a good idea to make it explicit in the zmq.h file and not by including a file.

bluca:
@somdoron - sorry, what do you mean?

evoskuil:
@somdoron I'm sort of ambivalent about that choice. There is extra work/risk in manually maintaining duplicate copies of the same API, but test coverage and CI should mitigate most of the risk. But a second (optional) public header seems inelegant.

hitstergtd:
I suggest that the Draft and Deprecated API stays in `zmq.h`, clearly sectioned away, so that it is abundantly crystal to past and future library users.

[...]

PR #1992, opened 2016-05-14
Problem: name confl from windows.h, draft API decl in pch.

evoskuil:
See https://github.com/zeromq/libzmq/issues/1989


src/proxy.cpp
```diff
@@ -46,8 +46,9 @@
 #include "socket_base.hpp"
 #include "err.hpp"
 
+// TODO: determine if this is an issue, since zmq.h is being loaded from pch.
 // zmq.h must be included *after* poll.h for AIX to build properly
```
evoskuil: I wanted to highlight this comment. The previous implementation did not conform to the requirement indicated by the comment. I don't have the ability to test this, but this pull request does not change the actual behavior.

src/zmq.cpp
```diff
@@ -40,16 +40,14 @@
 #include <poll.h>
 #endif
 
+// TODO: determine if this is an issue, since zmq.h is being loaded from pch.
 // zmq.h must be included *after* poll.h for AIX to build properly
```
evoskuil: I wanted to highlight this comment (also). The previous implementation did not conform to the requirement indicated by the comment. I don't have the ability to test this, but this pull request does not change the actual behavior.

bluca:
Note that the precompiled headers support is only available on Windows.
This means that on *nix,  by removing the include of platform.hpp, we are at the mercy of the preprocessor - if it happens to parse first a file that includes the platform.hpp before it includes this headers, all the constants below will be available, else they won't.
So we should either leave platform.hpp included where it is, or manually include precompiled.hpp everywhere you are removing the include of platform.hpp.

evoskuil:
> Note that the precompiled headers support is only available on Windows.

@bluca The only aspect of the implementation that pertains to precompilation is these instructions within the msvc `libzmq.props` and `libzmq.vcxproj` files:

```xml
<ClCompile>
    <PrecompiledHeader>Use</PrecompiledHeader>
    <PrecompiledHeaderFile>precompiled.hpp</PrecompiledHeaderFile>
    ...
</ClCompile>
```

```xml
<ClCompile Include="..\..\..\..\src\precompiled.cpp">
    <PrecompiledHeader>Create</PrecompiledHeader>
</ClCompile>
...
<ClCompile Include="..\..\..\..\src\tweetnacl.c">
    <PrecompiledHeader>NotUsing</PrecompiledHeader>
</ClCompile>
```

So it doesn't matter whether another compiler supports or is configured to apply precompiled headers.

> This means that on *nix, by removing the include of platform.hpp, we are at the mercy of the preprocessor...

The include of the `platform.hpp` isn't removed, it's moved to the `precompiled.hpp`. What I didn't notice is that `precompiled.cpp` is not included in the *nix makefile. I did not expect this and will correct it.

There is no harm if for some reason precompilation is not going to be used with the other compilers. And if the name "precompiled" is disfavored because precompilation isn't going to be configured for other compilers, then the name of the file can certainly be changed.

The preceding design produces namespace conflicts arising from the fact that it is impossible to see or control when the `windows.hpp` will be included. As a result a slight change to the project can cause it to be included after some other header that is intended to be affected by the configuration in `windows.hpp`. It is essentially a requirement that in every file headers are included in the following order:

* `precompiled.hpp`
* `platform.hpp`
* `windows.hpp` (as applicable).
* ...

If we don't do the above, we are at the mercy of the compiler's include ordering - which created a number of errors in the Windows build when I added the `zmq_draft.hpp` - which should not have been a problem.

> or manually include precompiled.hpp everywhere you are removing the include of platform.hpp.

I understood it to be a requirement of the preceding design that `precompiled.hpp` must be the first include of every cpp file, so I did not consider this a change.

bluca:
If it's fixed in autotools/cmake then it's fine by me as a solution - given the platform.hpp is a result of the build system generating a configuration, it's ok IMHO to have it included by the same build system that creates it.

bluca:
@evoskuil - don't forget CMake :-)

evoskuil:
The above commit will resolve any potential problem of `platform.hpp` not having precedence in any build, though I'm anticipating an empty compilation unit warning.

evoskuil:
@bluca Okay, I didn't realize that CMake had it's own file listing - thanks.

evoskuil:
@bluca Turns out that `precompiled.cpp` is already included by CMake, so it was only missing in the autotools builds.

bluca:
Ah sorry should have checked, thanks for looking into it!

evoskuil:
There are no empty compilation unit warnings, so I guess those are already suppressed.

bluca:
Perfect. Looks good to me - do you have any more pending changes? Else I'll go ahead and merge

evoskuil:
@bluca I've got nothing queued up, all done here.

evoskuil:
@bluca Thanks for the help!

PR #2003, opened 2016-05-19
Problem: no function to derive public key from curve secret.

evoskuil:
The lack of a public function to derive the public key from a curve secret key requires that both keys be preserved together indefinitely. This is a "truth in two places" problem when it comes to configuration, especially if one is not using "certificate" files. Making matters worse, there is no straightforward way to validate that a given public key correctly corresponds to a given secret key. This change allows a secret key to be retained alone and for an application to derive the corresponding public key as needed.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -465,9 +465,13 @@ ZMQ_EXPORT char *zmq_z85_encode (char *dest, const uint8_t *data, size_t size);
 /*  Decode data with Z85 encoding. Returns decoded data                       */
 ZMQ_EXPORT uint8_t *zmq_z85_decode (uint8_t *dest, const char *string);
 
-/*  Generate z85-encoded public and private keypair with libsodium.           */
+/*  Generate z85-encoded public and private keypair with tweetnacl/libsodium. */
 /*  Returns 0 on success.                                                     */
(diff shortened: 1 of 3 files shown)


bluca:
Cool! Would it be possible to have a unit test for this?

evoskuil:
Absolutely, I looked for tests against the other related methods but didn't find them. Can you direct me to the proper location?

bluca:
Uh might very well be that there are none :-)
If you have time and are willing would be nice to create a new test module for these features.

evoskuil:
Sure, will do. `zmq_curve_keypair` can't be tested for correctness, since there's no way to inject the RNG. But `zmq_curve_public`, `zmq_z85_encode` and `zmq_z85_decode` can be fully covered.

These functions are used within `test_security_curve.cpp` although I would't consider that a unit test. How about I create `test_utils.cpp` for this?

bluca:
Sounds like a plan. Thanks!

evoskuil:
I found that most of the other functions in `zmq_utils` are covered elsewhere, so I created `test_base85` and `test_sodium`.

evoskuil:
https://github.com/zeromq/libzmq/pull/2004

PR #2004, opened 2016-05-19
Problem: no unit tests for base85 or public curve functions.

evoskuil:


tests/test_base85.cpp
```diff
@@ -0,0 +1,92 @@
+/*
+    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca: copyright range should be just 2016 since it's a new file

tests/test_sodium.cpp
```diff
@@ -0,0 +1,98 @@
+/*
+    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca: ditto

bluca:
Thanks! Looks good, but let's get the copyright correct the first time to be safe :-)

tests/test_sodium.cpp
```diff
@@ -0,0 +1,99 @@
...
+    const int rc = zmq_curve_public (out_public, secret_key);
+
+#if defined (ZMQ_HAVE_CURVE)
+    assert (rc == 0);
+    assert (zmq_errno () == 0);
+    assert (streq (out_public, public_key));
+#else
+    assert (rc == -1);
+    assert (zmq_errno () == ENOTSUP);
+    (void) public_key
```
bluca: missing ";"

bluca:
If it's not too much trouble, would you mind please squashing the fixup commits into the first one? This helps to keep git bisect working, which is an invaluable tool

bluca:
Thanks! Good to go or do you have more commits coming?

evoskuil:
No more as long as it clears the gate. Thanks for your help.

PR #2013, opened 2016-05-27
Windows Subsystem for Linux compatibility fix

aseering:
On systems where getifaddrs() exists but isn't implemented, behave as if 'resolve_nic_name()' were entirely unsupported.

This is a fix for jupyter/notebook#1331 and a workaround for Microsoft/BashOnWindows#185 .  It's not exactly a beautiful solution, and it's probable that the Bash-on-Windows folks will have fixed the root cause of this problem before they next pull in the latest build of ZeroMQ.  I certainly wouldn't object if you declined this PR on those grounds.  But I wrote the fix, and it works, so I figure I might as well share it.


rgbkrk:
Thank you @aseering

PR #2015, opened 2016-05-30
[RFC] Problem: there is no fine grained way to setup unix sockets in libzmq

vyskocilm:
Solution: use systemd socket activation to solve the problem at least on
Linux with systemd.

The support is limited to ipc transport atm.

Inspiration comes from zeromq-devel Setting privileges on a UNIX socket

Tested on malamute broker (ROUTER socket + ipc transport)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -298,6 +298,11 @@ src_libzmq_la_CPPFLAGS += ${pgm_CFLAGS}
 src_libzmq_la_LIBADD += ${pgm_LIBS}
 endif
 
+if HAVE_SYSTEMD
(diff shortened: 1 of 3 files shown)


bluca:
Hi,

As I wrote on the mailer, we already have this functionality.

The low level library has the ZMQ_USE_FD socket option to pass a file
descriptor, and the high level CZMQ has ZSYS_AUTO_USE_FD env var or
zsys_set_auto_use_fd(1) function to let zmq automatically match
endpoints to the corresponding sockets managed by systemd. As long as
the metadata matches (eg: file path for IPC, address + port for TCP), it
will just work out of the box.

vyskocilm:
Thanks for the feedback, I am closing the pr then.

PR #2018, opened 2016-06-02
Fix compilation with mingw64 using autotools

diorcety:
Here also the change for mingw64
- missing to set the _WIN32_WINNT version in external headers
- Explicit null cast
- On my machine WSA_WAIT_FAILED is unsigned int which is an issue
- Missing socket type in fd.hpp


bluca:
@diorcety - thanks.
For you next contributions, could you please follow the guidelines (http://zeromq.org/docs:contributing) and use the following format for the commit message:

```
Problem: x is broken

Solution: fix x by doing y and z
```

And keeping one problem and solution per commit? Thanks!

sappo:
@diorcety this change broke the visual studio build. See here https://ci.appveyor.com/project/zeromq/libzmq/build/build-641

diorcety:
Really strange the file exists ... Maybe repleace <windows.hpp> by "windows.hpp"?

bluca:
@diorcety could be. If you want to test it, in case you are not familiar with Appveyor, you can set it up for your fork to test the build. Just go to https://ci.appveyor.com/projects/new login with your Github account, and add libzmq in that project page, and then when you push it will build it.

bluca:
@diorcety - any news? We will have to revert this commit if there's no fix, to un-break the VS build

bluca:
Fixed by: https://github.com/zeromq/libzmq/pull/2032

PR #2024, opened 2016-06-04
allow to build sparc (v8,leon)

wbx-github:
Only set sparcv9 optimization for sparc64 systems.
This allows to run for example application using zeromq
on sparc32 systems.

Signed-off-by: Waldemar Brodkorb <wbx@uclibc-ng.org>


bluca:
Thanks!

PR #2025, opened 2016-06-04
Problem: libzmq README has no API description

sappo:
Solution: Use gitdown, mkman from zproject and CLASS documentation to build a README with an API


sappo:
I'm gonna fix the broken build and open another PR

PR #2030, opened 2016-06-08
Problem: std:vector.data breaks compat with C++98

bluca:
Solution: malloc and free a temporaty buffer instead

@somdoron - as we discussed on Saturday, this fixes the build with compilers that don't support C++11, for example on Solaris.


ricnewton:
Isn't replacing buffer.data() with &buffer[0] the C++98 standard way of doing this?

bluca:
@ricnewton - might be, I'll try in a little while, I'll close meanwhile

bluca:
https://github.com/zeromq/libzmq/pull/2031

PR #2031, opened 2016-06-09
Problem: std:vector.data breaks compat with C++98

bluca:
Solution: use buffer address instead

@somdoron - as we discussed on Saturday, this fixes the build with compilers that don't support C++11, for example on Solaris.


bluca:
Ping :-)

c-rack:
Pong :-)

somdoron:
Thanks

PR #2034, opened 2016-06-11
Poll() implementation for Windows Vista+

michicc:
Provide poll() for Windows as well. This is a build option that defaults to off as the resulting binary will only run on Windows Vista or newer.

This is not tested with alternative Winsock service providers like VMCI, but the documentation for WSAPoll does not mention limitations.

On my local machine, throughput improves by ~10 % (20 simultaneous remote_thr workes to one local_thr, 10 byte messages), while latency improves by ~30 % (measured with remote/local_lat).


sunkin351:
Dang, will update my binaries with this as soon as possible, how exactly do i use this?

michicc:
It's a project option, just like the other existing options where you select which libraries to use and so on.

PR #2036, opened 2016-06-13
Flush stderr buffer before calling zmq_abort in assert macros

aburgm:
On Windows, the written message does not seem to be guaranteed to be
written to stderr, in particular when stderr is redirected to a file. I
suppose this is because RaiseException terminates the process in a way
that does not give the CRT a chance to flush stdio buffers (or if it
does, there might be a problem when more than one CRT instance is linked
into the program and they overwrite each other's exception handler). Either
way, just make sure the assertion message ends up written to stderr to
ease diagnostics.

### DIFF
--- a/src/err.hpp
+++ b/src/err.hpp
@@ -78,6 +78,7 @@ namespace zmq
             if (errstr != NULL) {\
                 fprintf (stderr, "Assertion failed: %s (%s:%d)\n", errstr, \
                     __FILE__, __LINE__);\
+                fflush (stderr);\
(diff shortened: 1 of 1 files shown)


aburgm:
Are these test failures really caused by this commit? Is there more log output for them to get an insight into what exactly caused them to fail?

bluca:
Those are pre-existing if you check older builds, don't worry.

```Problem: X is broken

Solution: do Y and Z to fix X```

aburgm:
Will do. Thanks!

PR #2039, opened 2016-06-16
Select fix

dbc60:
`family_entry.fd_entries` is a `std::vector`. The code was taking a reference to the `i`th element, which wouldn't be so bad, except that sometimes the vector is full. A few lines down `fd_entry.events->in_event ();` can add an element to `fd_entries`. If it's full, then the whole vector is re-allocated. If the new vector isn't on the same address as the original our reference refers to invalid memory.

I changed the code so `fd_entry` is a copy of the `i`th element and it will be valid throughout the `for` loop.

The .gitignore update is just a bonus.


dbc60:
Sorry, wrong repo

PR #2040, opened 2016-06-16
Select fix

dbc60:
family_entry.fd_entries is a std::vector. The code was taking a reference to the ith element, which wouldn't be so bad, except that sometimes the vector is full. A few lines down fd_entry.events->in_event (); can add an element to fd_entries. If it's full, then the whole vector is re-allocated. If the new vector isn't on the same address as the original our reference refers to invalid memory.

I changed the code so fd_entry is a copy of the ith element and it will be valid throughout the for loop.


bluca:
Could you please clean up the history, by rebasing and dropping the 2 commits and their reverts?

dbc60:
I'm going to submit a new PR that has clean history.

bluca:
Thanks! But in general, you can force push to the same branch and same PR, no need to open a new one.

```
git fetch upstream
git rebase -i upstream/master
git push -f
```

PR #2042, opened 2016-06-17
Send manual unsubscriptions on pipe termination in XPUB

sheremetyev:
Currently unsubscriptions sent by XPUB in manual mode on pipe termination are in terms of topics, which is different from explicit unsubscriptions sent by client (some custom format). This change stores manual subscriptions in XPUB and sends them on pipe termination to make behaviour consistent.

### DIFF
--- a/src/xpub.cpp
+++ b/src/xpub.cpp
@@ -92,6 +92,12 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
         if (size > 0 && (*data == 0 || *data == 1)) {
             if (manual)
             {
+                // Store manual subscription to use on termination
+                if (*data == 0)
+                    manual_subscriptions.rm(data + 1, size - 1, pipe_);
+                else
+                    manual_subscriptions.add(data + 1, size - 1, pipe_);
+
                 pending_pipes.push_back(pipe_);
                 pending_data.push_back(blob_t(data, size));
                 pending_metadata.push_back(sub.metadata());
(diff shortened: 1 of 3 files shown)


bluca:
Hi, does this change existing behaviour? Eg, do applications need to be rewritten or changed?

src/xpub.cpp
```diff
@@ -191,10 +197,19 @@ int zmq::xpub_t::xsetsockopt (int option_, const void *optval_,
...
+        //  Remove the pipe from the trie and send corresponding manual
+        //  unsubscriptions upstream.
+        manual_subscriptions.rm (pipe_, send_unsubscription, this, false);
+    }
+    else
+    {
+        //  Remove the pipe from the trie. If there are topics that nobody
+        //  is interested in anymore, send corresponding unsubscriptions
+        //  upstream.
+        subscriptions.rm (pipe_, send_unsubscription, this, !verbose_unsubs);
```
somdoron: I think you still want to remove the pipe subscription even if it is manual.  So maybe just remove the else.

src/xpub.hpp
```diff
@@ -79,6 +79,9 @@ namespace zmq
         //  List of all subscriptions mapped to corresponding pipes.
         mtrie_t subscriptions;
 
+        //  List of manual subscriptions mapped to corresponding pipes.
+        mtrie_t manual_subscriptions;
```
somdoron: You actually don't need a trie,  you can do it with multimap.

somdoron:
I think you have a bug where you dont remove subscriptions of terminated pipe.

somdoron:
@bluca,  it might casue issues,  but it is a bug. Maybe we can make option switch for the fix.  My opinion is that it would not cause real issues because there no much sense in handling those kind of unsubscriptions messages.  So when using manual mode you usually ignore all unsubscriptions. What do you think?

bluca:
@somdoron I agree, it makes sense to fix this. The fact that the existing tests work means nothing should be affected, API-wise. Was just checking :-)

PR #2043, opened 2016-06-17
gyp: creating gyp file to work with nsolid

thlorenz:
- separating out gypi file to contain sources
- including debug and release conditions
- test with ninja generator 'gyp --depth=. -f ninja && ninja -C out/Default' (need to have _gyp_ in your path)


bluca:
Hi, I'm not familiar with gyp but there are already some in the builds/gyp directory, would these be different?

rgbkrk:
Whoa, glad to see contributions towards the node bindings as well as support from NodeSource.

PR #2046, opened 2016-06-20
Problem: MSVC warns about event monitoring functions converting SOCKET to int

pijyoi:
When compiling for x64 using MSVC, the event monitoring functions generate several warnings. To squash these warnings, use zmq::fd_t instead of int for the function prototypes expecting a socket descriptor.

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -1714,7 +1714,7 @@ int zmq::socket_base_t::monitor (const char *addr_, int events_)
     return rc;
 }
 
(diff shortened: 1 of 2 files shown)


bluca:
Thanks!

```
Problem: X is broken

Solution: do Y and Z to fix X
```

pijyoi:
Right. I thought doing that in the pull request message would be sufficient
but apparently that doesn't get merged.

bluca:
Yes, the PR comment is "lost" to the git repository, it only exists on Github.

PR #2071, opened 2016-07-31
Convert manual (locking and unlocking) mutexes to scoped mutexes for the case of unlocking mutex even if the protected operation throws an exception

reza-ebrahimi:
Convert manual (locking and unlocking) mutexes to scoped mutexes for the case of unlocking mutex even if the protected operation throws an exception


bluca:
Thanks!

```
Problem: X is broken

Solution: do Y and Z to fix X
```

Also for indentation we use only 4 spaces, no tabs.

reza-ebrahimi:
Thanks for your hint, I'll be fixing indentations.

PR #2073, opened 2016-07-31
Problem: no OBS _service and various packaging problems

bluca:
See individual commits.

These changes allow to build the latest HEAD of libzmq for Debian 7 and 8, Ubuntu 14.10, 15.10 and 16.04, Fedora 23 and 24, and OpenSUSE Tumbleweed and Leap 42.1:

https://build.opensuse.org/package/show/home:bluca/libzmq


bluca:
@sappo - I setup a token and created the project: https://build.opensuse.org/project/show/home:zeromq:libzmq
Once we have the official project we should build libzmq and CZMQ in the same project, otherwise if the ABI changes they won't be usable together

sappo:
Does obs resolve dependencies within the same repository. For example there's no official libzmq package for cent os but we still can build czmq with the package from obs.

bluca:
Yes, it's automatic if they are part of the same project, or the meta can reference other projects, but then it's confusing for users as they have to add all of them on their machines individually.
I was thinking we could have multiple subproject in home:zeromq, but not divided per library but per topic, for example a DRAFT one and a non-DRAFT one.

sappo:
Neat!

PR #2074, opened 2016-08-01
Problem: OBS compress not available on CentOS

bluca:
Solution: run the obs-service-compress at service time rather than
buildtime


bluca:
@sappo - CentOS 7 and RHEL 7 builds added: https://build.opensuse.org/package/show/home:zeromq:libzmq/libzmq

PR #2084, opened 2016-08-16
Fix compatibility with newer CMake.

Kentzo:
See CMake bug for details.

https://gitlab.kitware.com/cmake/cmake/issues/16201


bluca:
Thanks!

```
Problem: X is broken

Solution: do Y and Z to fix X
```

PR #2088, opened 2016-08-23
is_ipv6_available: Create a fake zmq context on windows

t-b:
This is required as zmq_ctx_new calls WSAStartup. Without that the IPV6
socket creation always fails.

This now makes test_unbind_wildcard failing on my machine.

### DIFF
--- a/tests/testutil.hpp
+++ b/tests/testutil.hpp
@@ -339,6 +339,10 @@ is_ipv6_available(void)
     inet_pton (AF_INET6, "::1", &(test_addr.sin6_addr));
 
 #ifdef ZMQ_HAVE_WINDOWS
+    // initialize network stack
+    void *ctx = zmq_ctx_new();
+    assert(ctx);
+
     SOCKET fd = socket (AF_INET6, SOCK_STREAM, IPPROTO_IP);
     if (fd == INVALID_SOCKET)
         ipv6 = 0;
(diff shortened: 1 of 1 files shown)


bluca:
Thanks!

```
Problem: X is broken

Solution: do Y and Z to fix X
```

bluca:
Also, I think it would be better to call the function after the context is initialized in the tests.
We don't want to create multiple contexts, I think. I'll send a PR.

t-b:
@bluca Thanks for the quick reply. I'll read the contributing docs before my  next PR.

I did not move the `is_ipv6_available` call below the context intialization, as that is way too fragile for my taste. If you want to do it differently, move the is_ipv6_available call into the setup_test_environment and just call WSAStartup.

bluca:
If the test is failing on Windows, and if you can, please do send a PR to fix it. Unfortunately I do not have a Windows machine available.

From the appveyor log it's failing to bind to tcp://127.0.0.1:* when ipv6 is enabled globally, which will result in a v4-mapped-in-v6 type of address being used. I really have no idea how Windows handles that.

For the test, I agree that it's more fragile from a code point of view, but having multiple contexts hidden in the background is worse.
At the moment only that tests uses the function. I have it on my list to add more ipv6 address, and when I do, I'll refactor it, but for now I'd rather avoid the churn to refactor all the existing test modules.

PR #2094, opened 2016-08-28
Problem: cannot build packages with DRAFT APIs

bluca:
Solution: see commits

Both a manual build and an OBS build solution are added.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -10,6 +10,27 @@ ifeq ($(DEB_BUILD_ARCH_OS), kfreebsd)
 	DO_TEST = no
 endif
 
+DRAFTS=no
+
+# OBS build: add
+#   Macros:
+#   %_with_drafts 1
+# at the BOTTOM of the OBS prjconf
+OBS_BUILD_CFG=/.build/build.dist
+ifeq ("$(wildcard $(OBS_BUILD_CFG))","")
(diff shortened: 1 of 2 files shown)


bluca:
@sappo - I switched the trigger in libzmq and czmq to kick home:zeromq:git-stable, and linkpac'ed the packages there to the older locations at home:zeromq:czmq and home:zeromq:libzmq so that they get triggered too at the same time

sappo:
@bluca - this is great news but what exactly do you mean by linkpac'ed?

Am 28.08.2016 20:59 schrieb "Luca Boccassi" <notifications@github.com>:
>
> @sappo - I switched the trigger in libzmq and czmq to kick
home:zeromq:git-stable, and linkpac'ed the packages there to the older
locations at home:zeromq:czmq and home:zeromq:libzmq so that they get
triggered too at the same time

bluca:
A package can be linked to another in a different project, and they will share the service file. When the original is triggered, the linked one is too. But the linked is built in the context of its own project, so with its dependencies, config, build targets, etc.

I also have aggregated the build dependencies like asciidoc and libsodium. Aggregation is one step further, and it means that the binaries built in the original project are copied verbatim to the aggregated one, they are not rebuilt. Since we don't needed to rebuild those, it makes it a bit faster and avoids double builds.

PR #2099, opened 2016-09-04
Problem: socket_base uses macros for mutex lock and unlock

a4z:
Solution: add a lock guard that takes optional a mutex and use it

### DIFF
--- a/src/mutex.hpp
+++ b/src/mutex.hpp
@@ -181,6 +181,33 @@ namespace zmq
         scoped_lock_t (const scoped_lock_t&);
         const scoped_lock_t &operator = (const scoped_lock_t&);
     };
+
+
+    struct scoped_optional_lock_t
+    {
+        scoped_optional_lock_t (mutex_t* mutex_)
+            : mutex (mutex_)
+        {
+            if(mutex != 0)
+                mutex->lock ();
+        }
+
+        ~scoped_optional_lock_t ()
+        {
+            if(mutex != 0)
+                mutex->unlock ();
+        }
+
+    private:
+
+        mutex_t* mutex;
+
+        // Disable copy construction and assignment.
+        scoped_optional_lock_t (const scoped_lock_t&);
+        const scoped_optional_lock_t &operator = (const scoped_lock_t&);
+    };
+
+
 }
 
 #endif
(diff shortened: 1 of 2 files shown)


bluca:
Thanks!

Unfortunately it looks like this change breaks existing tests:

FAIL: tests/test_use_fd_ipc
===========================
Device or resource busy (../src/mutex.hpp:113)

Could you please have a look?

src/mutex.hpp
```diff
@@ -181,6 +181,33 @@ namespace zmq
...
         const scoped_lock_t &operator = (const scoped_lock_t&);
     };
+
+
+    struct scoped_optional_lock_t
+    {
+        scoped_optional_lock_t (mutex_t* mutex_)
+            : mutex (mutex_)
+        {
+            if(mutex != 0)
```
bluca: Why 0 and not NULL, given it's a pointer? Would be clearer for readers

a4z:
yes, of course, if NULL this is the convention, i will change it 
thanks for the hint.

ad failing test, this test does not exist on my current mingw64 environment, but in an hour or 2 I will sit on a Linux machine and can have a look
but mutex.hpp:113 is interesting, this should not be through code I have changed but I will see

bluca:
Thanks!

The IPC test is only for *nix so that's why you don't see it, but there are more as you can see on the Travis log.

Notice how it's the tests that check the thread-safe sockets that fail (note that you need DRAFT APIs to be enabled to run those):

```
FAIL: tests/test_use_fd_ipc
===========================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_use_fd_tcp
===========================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_poller
=======================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_client_server
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_thread_safe
============================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_radio_dish
===========================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_udp
====================
Device or resource busy (../src/mutex.hpp:113)
FAIL: tests/test_scatter_gather
===============================
Device or resource busy (../src/mutex.hpp:113)
```

a4z:
fixed in the next commit

bluca:
Thanks!

PR #2112, opened 2016-09-20
Pollset

hnwyllmm:
add a new poller named pollset which will get benefit of performance in AIX platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ if (WITH_MILITANT)
 endif()
 
 set (POLLER "" CACHE STRING "Choose polling system. valid values are
-                            kqueue, epoll, devpoll, poll or select [default=autodetect]")
+                            kqueue, pollset, epoll, devpoll, poll or select [default=autodetect]")
 
 include (CheckFunctionExists)
 include (CheckTypeSize)
@@ -88,6 +88,15 @@ if (POLLER STREQUAL "")
     set (CMAKE_REQUIRED_INCLUDES)
     if (HAVE_KQUEUE)
         set (POLLER "kqueue")
+    endif()
+endif ()
+
+if (POLLER STREQUAL "")
+    set (CMAKE_REQUIRED_INCLUDES sys/pollset.h)
+    check_function_exists (pollset_create HAVE_POLLSET)
+    set (CMAKE_REQUIRED_INCLUDES)
+    if (HAVE_POLLSET)
+        set (POLLER "pollset")
     endif()
 endif ()
 
@@ -136,6 +145,7 @@ if (POLLER STREQUAL "")
 endif ()
 
 if (POLLER STREQUAL "kqueue"
+ OR POLLER STREQUAL "pollset"
  OR POLLER STREQUAL "epoll"
  OR POLLER STREQUAL "devpoll"
  OR POLLER STREQUAL "poll"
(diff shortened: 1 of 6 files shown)


somdoron:
@hnwyllmm you broke the build, can you move the new cpp and hpp files to the src directory?

CMakeLists.txt
```diff
@@ -92,6 +92,15 @@ if (POLLER STREQUAL "")
 endif ()
 
 if (POLLER STREQUAL "")
+    set (CMAKE_REQUIRED_INCLUDES sys/pollset.h)
```
bluca: Please don't change the default, move this snippet below

bluca:
Thanks for sending the pr!

Apart from the comments from Doron and myself, a couple suggestions. If you have multiple commits, use a different commit message for each one. Also each commit should work by itself (by which I mean not break the build and the tests) if possible.

Also, for your next contributions, could you please follow the contributing guidelines for the git commit message?

```
Problem: X is broken

Solution: do Y and Z to fix X
```

CMakeLists.txt
hnwyllmm: I will move it just before `poll` just like below, does it ok for that? (I am so sorry for my stupid commit)
`if (POLLER STREQUAL "")
    set (CMAKE_REQUIRED_INCLUDES sys/devpoll.h)
    check_type_size ("struct pollfd" DEVPOLL)
    set (CMAKE_REQUIRED_INCLUDES)
    if (HAVE_DEVPOLL)
        set (POLLER "devpoll")
    endif ()
endif ()

if (POLLER STREQUAL "")
    set (CMAKE_REQUIRED_INCLUDES sys/pollset.h)
    check_function_exists (pollset_create HAVE_POLLSET)
    set (CMAKE_REQUIRED_INCLUDES)
    if (HAVE_POLLSET)
        set (POLLER "pollset")
    endif()
endif ()

if (POLLER STREQUAL "")
    set (CMAKE_REQUIRED_INCLUDES poll.h)
    check_function_exists (poll HAVE_POLL)
    set (CMAKE_REQUIRED_INCLUDES)
    if (HAVE_POLL)
        set (POLLER "poll")
    endif ()
endif ()`

bluca: Yes that would be perfect, thank you!

CMakeLists.txt
```diff
@@ -468,6 +478,7 @@ set (cxx-sources
         plain_server.cpp
         poll.cpp
         poller_base.cpp
+		pollset.cpp
```
bluca: In the next commit could you also please fix the indentation? It's spaces only, no tabs

acinclude.m4
```diff
@@ -901,7 +917,7 @@ AC_DEFUN([LIBZMQ_CHECK_POLLER], [{
     fi
     if test "$pollers" == "auto"; then
         # We search for pollers in this order
-        pollers="kqueue epoll devpoll poll select"
+        pollers="kqueue pollset epoll devpoll poll select"
```
bluca: Please also move this after devpoll, we do not want to change defaults

CMakeLists.txt
hnwyllmm: Done, please view #2120 , Thanks

acinclude.m4
hnwyllmm: Done. Please see #2120

Issue #25, opened 2010-05-13
Add support for function callbacks to REQ/REP and PUB/SUB sockets


chuckremes:
Per a discussion in the comments here: http://www.zeromq.org/blog:multithreaded-server

Expand the 0mq socket api to allow for attaching callbacks to REQ sockets on send. The callback should be invoked when its matching response arrives.

Similarly, allow for attaching callbacks to SUB sockets. All incoming messages should be passed to the callback for processing.

While these features can currently be built on top of the existing api via zmq_poll, it would be cleaner to have it available as part of the supported api.

PR #2113, opened 2016-09-20
remove pollset.hpp/.cpp to src directory

hnwyllmm:
remove pollset.hpp/.cpp files to src directory to fix the break build.


utam0k:
So it is better to quickly merge it.

hnwyllmm:
Who can tell me why this is failed:
rc = zmq_bind (sb, "tcp://127.0.0.1:*");
assert (rc == 0);

build console shows:
`25: Test command: C:\projects\build_libzmq\bin\Release\test_unbind_wildcard.exe
25: Test timeout computed to be: 10
25: Assertion failed: rc == 0, file C:\projects\libzmq\tests\test_unbind_wildcard.cpp, line 151
25/70 Test #25: test_unbind_wildcard .............***Exception: Other  7.24 sec
test 26
      Start 26: test_ctx_options`

It works fine in my Linux os.

bluca:
Don't worry about that one, is a pre-existing failure.

Thanks for the followup, please also send a PR to move the CMake snippet.

PR #2120, opened 2016-09-21
Pollset-change the position of `pollset`

hnwyllmm:
1. move the 'pollset' to the position between 'devpoll' and 'poll'
2. replace the tabs with spaces


bluca:
Thanks!

PR #2122, opened 2016-09-23
Problem: Pub socket crashes on client disconnect. 

gena-moscow:
family_entry.fd_entries [i] is modified in zmq::select_t::rm_fd()

Solution: do not copy family_entry.fd_entries [i]


bluca:
@gena-moscow - this was set -explicitly- to a copy by https://github.com/zeromq/libzmq/pull/2041 by @dbc60 
Douglas, what was the use case there? This change will move it back to a reference, will it cause you issues again?

gena-moscow:
As I understand we can't use reference here because vector can be reallocated as Douglas said.

PR #2127, opened 2016-09-26
	fix bugs of the pollset

hnwyllmm:
1. extend 'fd_table' when fd_ is greater or equal than the size of 'fd_table';
2. delete specific fd from pollset before reset pollin or pollout according the description of AIX document

### DIFF
--- a/src/pollset.cpp
+++ b/src/pollset.cpp
@@ -252,4 +252,3 @@ void zmq::pollset_t::worker_routine (void *arg_)
 }
 
 #endif
--
--- a/src/pollset.cpp
+++ b/src/pollset.cpp
@@ -135,7 +135,7 @@ void zmq::pollset_t::reset_pollin (handle_t handle_)
     pc.events = 0;
 
     pc.cmd = PS_DELETE;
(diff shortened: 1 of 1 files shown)


bluca:
This breaks the build:

```
../src/pollset.cpp:255:1: error: expected unqualified-id before ‘-’ token
```

Want to fix it before I merge?

src/pollset.cpp
```diff
@@ -250,3 +252,4 @@ void zmq::pollset_t::worker_routine (void *arg_)
 }
 
 #endif
+-
```
bluca: This extra - is breaking the build

hnwyllmm: Yes, surely. Sorry for the low level fault

bluca:
Thanks. For your next contributions, please:

- if you have a first commit that's broken and you fix it in the second commit in the same PR, please rebase and squash. Having broken commits in the tree breaks git bisect.
- Please follow the contributions guidelines for the commit message format:

```
Problem: X is broken

Solution: do Y and Z to fix X
```

PR #2128, opened 2016-09-27
Problem: zmq_poller only signals one event

### LINKED ISSUE #2108: Problem: zmq_poller only signals one event
zmq_poller only returns a single event, even if several events are available to process. This precludes the ability of the caller to decide which events should get processed in which order, and allows starvation of sockets if one socket is particularly busy.

Currently, the only way to do traditional waiting on a collection of sockets is zmq_poll, but zmq_poll doesn't support the threadsafe sockets (I would also consider this a bug).

minrk:
Solution: zmq_poller_wait_all signals all events

allows signaling multiple events with one call to zmq_poller_wait_all rather than emitting only one event.

this prepares for zmq_poll being based on zmq_poller, which requires events for all sockets rather than just one.

closes #2108

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -577,6 +577,7 @@ ZMQ_EXPORT int  zmq_poller_add (void *poller, void *socket, void *user_data, sho
 ZMQ_EXPORT int  zmq_poller_modify (void *poller, void *socket, short events);
 ZMQ_EXPORT int  zmq_poller_remove (void *poller, void *socket);
 ZMQ_EXPORT int  zmq_poller_wait (void *poller, zmq_poller_event_t *event, long timeout);
+ZMQ_EXPORT int  zmq_poller_wait_all (void *poller, zmq_poller_event_t *events, long timeout);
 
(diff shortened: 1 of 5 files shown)


somdoron:
cool

somdoron:
I have one questions, don't you prefer that the size of the events array will be static, so you won't allocate array from the heap everytime?

So you can dequeue events only as the size of the array. I think this will yield better performance and won't loose much.

Now with the single wait implementation we allocate an array on every call.

minrk:
I opened #2129 to remove the `new` invocations. For some reason, I didn't think you could do that with size from variables. My C++ isn't so strong.

somdoron:
I actually thought the same way as you, we can do that? 
nice...

somdoron:
Anyway, I still think we need the size parameter to the method, what if the size changed from the size() call to wait call() ?

We just need to limit the max number of items that can be be fetched, and return the number of actually fetched items.

minrk:
#2130 adds n_events arg to wait_all.

PR #2130, opened 2016-09-27
add n_events argument to zmq_poller_wait_all

minrk:
avoids unnecessary heap allocations, races on the number of items

follow-up to #2128

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -577,7 +577,7 @@ ZMQ_EXPORT int  zmq_poller_add (void *poller, void *socket, void *user_data, sho
 ZMQ_EXPORT int  zmq_poller_modify (void *poller, void *socket, short events);
 ZMQ_EXPORT int  zmq_poller_remove (void *poller, void *socket);
 ZMQ_EXPORT int  zmq_poller_wait (void *poller, zmq_poller_event_t *event, long timeout);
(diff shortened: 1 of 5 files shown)


minrk:
Looks like [I was wrong](https://travis-ci.org/zeromq/libzmq/jobs/163085798#L1660) about `x = array[arg]`. While this works for me (`Apple LLVM version 8.0.0 (clang-800.0.38)`), Travis fails with:

    ../src/zmq.cpp:1249:47: error: ISO C++ forbids variable length array ‘evts’ [-Werror=vla]

somdoron:
You can use the array you get as parameter, we just need to make sure the structure are the same, in the future as well.

minrk:
#2131 passes it through. I checked that the structs are the same. I can't think of a reason why they would diverge. That said, I also can't think of a reason why there are two names for this struct.

somdoron:
One is a public name (exported type) and one internal. We should have only one.
I was not sure how to do it at the time, we don't want the socket_poller.cpp to include zmq.h and zmq.h to include socket_poller.hpp. 

Can we do this in zmq.h?

```
class socket_poller_t;
struct socket_poller_t::event_t;

typedef struct socket_poller_t::event_t  zmq_poller_event_t;
```

We can take the event_t struct out of the class if it help

somdoron:
Or we can just define it in the zmq.h and include zmq.h in socket_poller.cpp, I cannot think of something simpler...

PR #2131, opened 2016-09-27
pass through poller events

minrk:
instead of allocating a new, identical array and copying the data.

This is only safe while zmq_poller_event_t and zmq::socket_poller_t::event_t are the same struct, which they presumably will remain.


bluca:
Since today the CZMQ build has been timing out during tests:

https://travis-ci.org/zeromq/czmq/jobs/162307658

The only other PR was for the pollset, which is not used by default. Any idea if these changes could have caused it?

PR #2140, opened 2016-09-28
Support all sockets in zmq_poll

### LINKED ISSUE #2109: Problem: zmq_poll only supports some sockets
Users shouldn't have to know whether a socket is threadsafe in order to tell which poll APIs can be used.

Implementing zmq_poll on top of zmq_poller (when available) should solve the issue, and provide a nicer transition path for people already using zmq_poll in their applications.

The only impediment to that is that zmq_poller only returns one event, but this can be fixed as well (#2108).

minrk:
zmq_poll calls zmq_poller if available, since only zmq_poller supports new threadsafe socket types.

When zmq_poller can be assumed to be present, the duplicate poll implementation can be removed.

closes #2109

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -742,10 +742,60 @@ const char *zmq_msg_gets (zmq_msg_t *msg_, const char *property_)
 
 // Polling.
 
+#if defined ZMQ_HAVE_POLLER
+inline int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
+{
+    // implement zmq_poll on top of zmq_poller
+    int rc;
+    zmq_poller_event_t events[nitems_];
+    void *poller = zmq_poller_new ();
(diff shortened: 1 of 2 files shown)


somdoron:
very cool

bluca:
@minrk - Windows build seems to be borken unfortunately:

```
..\src\zmq.cpp(750): error C2057: expected constant expression [C:\Jenkins\workspace\libzmq\build\libzmq-static.vcxproj]
..\src\zmq.cpp(750): error C2466: cannot allocate an array of constant size 0 [C:\Jenkins\workspace\libzmq\build\libzmq-static.vcxproj]
..\src\zmq.cpp(750): error C2133: 'events' : unknown size [C:\Jenkins\workspace\libzmq\build\libzmq-static.vcxproj]
```

https://libzmq.cloudapp.net/job/libzmq/1265/console

PR #2141, opened 2016-09-28
heap-allocate events in zmq_poller_poll

minrk:
variable-size array allocation is not allowed by some compilers

This should fix the build failures due to #2140, but there are still some proxy tests failing.

One is due to zmq_poller not allowing sockets to be registered more than once, which zmq_poll allows and zmq_proxy currently requires.

I'm not sure what the other failure is caused by.

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -747,10 +747,12 @@ inline int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
 {
     // implement zmq_poll on top of zmq_poller
     int rc;
-    zmq_poller_event_t events[nitems_];
+    zmq_poller_event_t *events;
+    events = new zmq_poller_event_t[nitems_];
+    alloc_assert(events);
     void *poller = zmq_poller_new ();
     alloc_assert(poller);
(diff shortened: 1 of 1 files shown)


hnwyllmm:
Would you like to try to use 'alloca' routine to allocate memory, which will be faster than 'new' or 'malloc' that it allocate memory from stack. And it be supported in the most of OS platform ( at least Linux, HPUX, SunOS and AIX).

bluca:
alloca is a bit dodgy due to stack overflow concerns, also don't know what's the support on windows which is the original issue

hnwyllmm:
Hmm, it is a boring problem. But why not use the way in the routine 'zmq_poll':
`pollfd spollfds[ZMQ_POLLITEMS_DFLT];
    pollfd *pollfds = spollfds;

    if (nitems_ > ZMQ_POLLITEMS_DFLT) {
        pollfds = (pollfd*) malloc (nitems_ * sizeof (pollfd));
        alloc_assert (pollfds);
    }`

I think performance is very import that ZMQ will be used in online services. Please consider about it, Thanks very much.

PR #2145, opened 2016-09-29
allow duplicate entries in zmq_poller_poll

minrk:
This should restore full compatibility with earlier zmq_poll behavior.

It complicates things a little bit, as collisions must be detected and when collisions are found:

- event masks must be merged
- use modify instead of add
- pollitems, events arrays are no longer co-ordered

Reverts the recent zmq_proxy patch to workaround the lack of repeat-item support in zmq_poll that is now fixed.

If someone wants to optimize the collision detection / event merge using a map, that would probably be appropriate, but is a bit outside my expertise.

### DIFF
--- a/src/proxy.cpp
+++ b/src/proxy.cpp
@@ -130,15 +130,6 @@ int zmq::proxy (
         { backend_, 0, ZMQ_POLLOUT, 0 }
     };
 
-    int control_idx = 2;
-    if (frontend_ == backend_) {
-        // when frontend & backend are the same,
-        // avoid duplicate poll entries
-        qt_poll_items -= 1;
-        items[1] = items[2];
-        control_idx = 1;
-    }
-
     //  Proxy can be in these three states
     enum {
         active,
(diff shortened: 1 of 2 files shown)


bluca:
@minrk @somdoron I see random failures in test_poller when DRAFT APIs are enabled, both on Travis and on the Suse's package builds for x86 and amd64 packages. Not always, maybe 1 out of 10. Any idea what could be happening? Race condition? Timing issue?

somdoron:
@bluca which test is failing?

We might want to deprecate the zmq_poll as creating zmq_poller on each call is expensive. zmq_poller should be the new suggest way to poll on sockets.
Also we might want to not call the zmq_poller and use old zmq_poll implementation if ZMQ_DRAFT is not enabled.

bluca:
@somdoron - it's test_poller: https://github.com/zeromq/libzmq/blob/master/tests/test_poller.cpp
Unfortunately it never happened locally, always see it on the CI/build system, so I don't have more details. But it's always with the DRAFT API enabled builds.

For the rest, yes it sounds like a plan.

somdoron:
You have a link to failing test?

bluca:
Not at the moment, will post one next time I see a failure

bluca:
Here's one, during a Fedora x86_64 build:

https://build.opensuse.org/package/live_build_log/home:zeromq:git-draft/libzmq/Fedora_24/x86_64

PR #2147, opened 2016-10-01
Fix zmq poller poll return

goodfella:
First, thank you for all the great work with libzmq.  This library is great.  When I upgraded to the current master I first found that it didn't build, hence the commit to include unistd.h.  Then I discovered that polling was not working, hence the second commit.

### DIFF
--- a/src/stream_engine.cpp
+++ b/src/stream_engine.cpp
@@ -31,6 +31,7 @@
 #include "macros.hpp"
 
 #include <string.h>
+#include <unistd.h>
 #include <new>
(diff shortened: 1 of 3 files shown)


bluca:
Thanks! For your next contributions, could you please follow the contributing guidelines for the git commit message?

```
Problem: X is broken

Solution: do Y and Z to fix X
```

goodfella:
You're welcome Luca.  I'll insure to follow the contribution guidelines
next time.

Issue #2151, opened 2016-10-04
Problem: 2 connects with same sourceip:port to different destip:port fail


Prachi2807:
_send.cpp_

***********************************************************************************
void *ctx=zmq_ctx_new();
**void *sock1=zmq_socket(ctx,ZMQ_DEALER);
int rc=zmq_connect(sock1,"tcp://127.0.0.1:5500;127.0.0.1:5558");**

**void *sock2=zmq_socket(ctx,ZMQ_DEALER);
rc=zmq_connect(sock2,"tcp://127.0.0.1:5500;127.0.0.1:6611");**

***********************************************************************************
is this possible??

somdoron:
Did you try? I think it depends on the OS

Prachi2807:
I have tried it . but the first one is only getting connected.
After i stop the server on 5558 , then only it will go and connect to the 6611.

whereas in TCP implementation , it both the connections are connected at the same time.

Prachi2807:
is anyone looking into it ??
I need the solution to it,

bluca:
What's the OS?

Prachi2807:
OS - Linux Version 2.6.32-504.30.3.el6.x86_64 x86_64

Prachi2807:
Hey , have you tried?

bluca:
No haven't had time, might have some during the weekend

bluca:
Found the issue, on connect the SO_REUSEADDR flag is not set on the socket if a source address is passed before binding, so the second connect would never complete as the bind will always return EADDRINUSE.

Will send a fix shortly. Thanks for the report!

Prachi2807:
Thanks for the fix :+1:

PR #2171, opened 2016-10-23
Problem: 2 connects with same sourceip:port to different destip:port fail

bluca:
Solution: during a connect with a TCP endpoint if a source address is
passed set the SO_REUSEADDR flag on the socket before the bind system
call.
Add unit test to cover this case for both IPv4 and IPv6.

Fixes https://github.com/zeromq/libzmq/issues/2151

@somdoron - I believe using SO_REUSEADDR on Windows here is fine since it's a client socket that connects immediately after binding, so the complete 5-tuple should prevent hijacking. What do you think?

### DIFF
--- a/src/tcp_connecter.cpp
+++ b/src/tcp_connecter.cpp
@@ -314,6 +314,18 @@ int zmq::tcp_connecter_t::open ()
 
     // Set a source address for conversations
     if (tcp_addr->has_src_addr ()) {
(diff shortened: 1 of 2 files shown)


somdoron:
Sounds fine,  but I think you compleye the 5 tuple with the connect,  anyway I'm not familiar with this attack or vulnerablity

PR #2173, opened 2016-10-27
Problem: build API incompatible with 4.1

bluca:
Solution: keep the new --without-docs option, but also keep the old
--without-documentation with an added deprecation warning.
We can then remove it in the next major release, to leave enough time
for users and maintainers to change it without disruptions.

Let's be nice to maintainers and users with build scripts :-)


c-rack:
👍

PR #2178, opened 2016-10-28
[DO NOT MERGE YET - JENKINS TEST] Problem: MS VC++ build broken

bluca:
Solution: use __alignof instead of sizeof inside align intrinsic as
expressions are not allowed

sorry for the breakage... vc++ quirks


bluca:
grr not yet - help from someone with windows and vc++ would be greately appreciated

PR #2179, opened 2016-10-28
Problem: MS VC++ build broken

bluca:
Solution: use __alignof instead of sizeof inside align intrinsic as
expressions are not allowed

Third try - can't use an intrinsic with another intrinsic in vc++, because reasons.
So detect arch and hard code pointer size depending on it. Suggestions welcome.


bluca:
@rodgert @somdoron - now Jenkins is happy too. I could have just ignored VC++ but then I thought of the ARM-based surface and the Windows IOT port for Raspberry PI 3. Trying to be nice for those poor souls!

somdoron:
@bluca Can I merge this? I have an access to windows machine, do you want me to test something?

bluca:
@somdoron yes please I think it's ready now, both Jenkins and appveyor are green.
If you have a setup with a ms compiler and want to double check I would also appreciate, to be extra sure, thanks :-)

PR #2181, opened 2016-10-31
Remove an unnecessary while (true)

jolting:
This block of code will either return -1 or fall through.
The while (true) does nothing.
The braces limit the scope of int rc.


jolting:
Theres some more that I just saw. It seems like while (true) is used everywhere.

PR #2187, opened 2016-11-02
 Add WinXP compatibility

Dmitriy-GH:
In 4.2.0 used Windows Condition variable. Is supported from Windows Vista and later.

I added emulation with std::condition_variable in class condition_variable_t

This code no effect for Windows Vista and later. It used only if compile with 
#define ZMQ_HAVE_WINDOWS_TARGET_XP

[test_multithread.cpp](https://github.com/Dmitriy-GH/libzmq/blob/master/builds/msvc/vs2015_xp/test_zmq/test_multithread.cpp) is a multithread test ZMQ_CLIENT/ZMQ_SERVER sockets. 5 clients thread with 100 sockets per each permanently send messages to server.
On i7 with 4 core working stable without deadlock and any problem. Already sent 70 000 000 messages and it is working on.

PS Sorry for my English. I used autotranslator.


somdoron:
Very nice

PR #2203, opened 2016-11-08
Problem: 4.2.0 is out, time to bump version for development

bluca:
Solution: update NEWS, include/zmq.h and packaging


metadings:
Dear bluca, could we find a way to update the (precompiled, binary) Downloads section on zeromq.org? I'd like to do this without installers, just the precompiled binaries for x86 and x64, for Linux on GCC48 and for Windows on MSVC2010 or MSVC2012.

I hate to tell users we are always using an outdated version. A version which maybe crippled due to compilation failures - for example, the last time I made zeromq on Windows, I didn't see the x64 version was compiled without libsodium. Now I can't make this again - because I don't own the Windows computer, I just have a Linux computer.

The current Windows version is v4.0.4 - this is from 2014 - with the old `zmq_msg_t` of 32 instead of 64 bytes and without all the goodies from newer versions.

bluca:
Hi, unfortunately I don't have a Windows dev machine either, so we'd need someone to do this work. If someones volunteers then of course we can upload the binaries to the page.

evoskuil:
I maintain a nuget package for VC2013, currently 4.2. Libsodium is disabled in favor of tweetnacl. You can just extract the binaries. 

https://www.nuget.org/packages/libzmq_vc120/

lgeiger:
In [zeromq/zeromq.js](https://github.com/zeromq/zeromq.js) we also use prebuilt static libs for VS2015. You can find them at [nteract/libzmq-win](https://github.com/nteract/libzmq-win).

bluca:
@lgeiger @evoskuil I think anyone who's registered can edit the windows page: http://zeromq.org/distro:microsoft-windows

Given I'm not sure about the exact build environment for those files, could you please add both of your builds? So that if they are different they may be useful for more users

lgeiger:
I have a VM with VS2015. I can build libs for [all supported configs](https://github.com/zeromq/libzmq/blob/master/builds/msvc/build/buildbase.bat) and upload them in two weeks when I have more time.

bluca:
Thank you!

Issue #2033, opened 2016-06-09
Linker fails on master: undefined reference to `dladdr'


ishbir:
I ran `./autogen.sh && ./configure && make`. Problem is fixed by using `LIBS=-ldl ./configure && make`. I suspect something is up with autoconf and it doesn't add `dl` as a dependency.

O.S.: Arch Linux
GCC Version: 6.1.1

bluca:
Version of autoconf, automake, libtool et al?

ishbir:
autoconf (GNU Autoconf) 2.69
automake (GNU automake) 1.15
libtool (GNU libtool) 2.4.6
GNU ld (GNU Binutils) 2.26.0.20160501

bluca:
I tried to reproduce on Debian sid, same version of the whole toolchain including the compiler, but can't.
Is it possible that something is up with your environment on Arch?

bluca:
```
# ldd src/.libs/libzmq.so.5.0.0 
	linux-vdso.so.1 (0x00007fffd0e2c000)
	/usr/lib/cowdancer/libcowdancer.so (0x00007f658ff88000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f658fd7a000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f658fb5d000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f658f7dd000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f658f4de000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f658f13a000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f658ef24000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f658ed1f000)
	/lib64/ld-linux-x86-64.so.2 (0x000055c2f9413000)
```

libdl is linked without any workaround.

ishbir:
I haven't changed the environment much. `printenv` gives me:
```
XDG_VTNR=1
SSH_AGENT_PID=774
XDG_SESSION_ID=c1
ANDROID_HOME=/opt/android-sdk
GLADE_PIXMAP_PATH=:
TERM=xterm
SHELL=/bin/bash
[...]

PR #2206, opened 2016-11-10
Problem: linker fails looking for dladdr

bluca:
Solution: search and add it via AC_CHECK_LIB when building with
libunwind, as the backtrace function uses dladdr. This problem
only appears on some distributions and with some compiler/toolchain
versions.

Fixes https://github.com/zeromq/libzmq/issues/2033


bluca:
Anyone for a merge? :-)

c-rack:
Sure :-)

bluca:
Thanks!

Issue #2208, opened 2016-11-14
setsockopt ZMQ_TOS doesn't work for IPv6


saifhhasan:
To set `Type Of Service` for IP layer packets ZMQ provides `ZMQ_TOS` socket option. However this only works for v4 sockets. Considering things are moving to IPv6 heavily (especially within enterprise networks), ZMQ should support setting `traffic class` for v6 based on `ZMQ_TOS`.

There is a subtle difference between v4 and v6 in terms of the positioning of field but TOS has same meaning in both v4 and v6. Linux provides following APIs for v4/v6 to set TOS field value.
```
// For v4
setsockopt(fd, IPPROTO_IP, IP_TOS, tos, sizeof(tos));

// For v6
setsockopt(fd, IPPROTO_IPV6, IPV6_TCLASS, tos, sizeof(tos));
```

Ask for this issue is to extend `ZMQ_TOS` to support ipv6 `traffic class` as well.

bluca:
Do you have a use case or a way to test that it would work properly if implemented?

saifhhasan:
Yes. Here at Facebook we build `Link State Routing Protocol` on the top of `ZeroMQ` transport (IPv6 only communication). It is very critical for us to distinguish control plane traffic from ordinary traffic to not hammer communications between routing daemon even under heavy network load.

We have internally carried out the patch and it is working well. Trying to send it upstream. Happy to take back your feedback and iterate as per the requirements.

OpenR: https://code.facebook.com/posts/1142111519143652/introducing-open-r-a-new-modular-routing-platform/

bluca:
Great! If you have already done the work, then please do send a PR and we'll merge it. Your diff looks good, the only thing you'll want to make sure is that if IPv6 is not enabled it doesn't cause a failure. Thanks!

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/2209

PR #2209, opened 2016-11-14
Make ZMQ_TOS work with IPv6 sockets

saifhhasan:
Summary:
To set `Type Of Service` for IP layer packets ZMQ provides `ZMQ_TOS` socket
option. However this only works for v4 sockets. Considering things are moving to
IPv6 heavily (especially within enterprise networks), ZMQ should support setting
`traffic class` for v6 based on `ZMQ_TOS`.

There is a subtle difference between v4 and v6 in terms of the positioning of
field but TOS has same meaning in both v4 and v6. Linux provides following APIs
for v4/v6 to set TOS field value.
```
// For v4
setsockopt(fd, IPPROTO_IP, IP_TOS, tos, sizeof(tos));

// For v6
setsockopt(fd, IPPROTO_IPV6, IPV6_TCLASS, tos, sizeof(tos));
```

Test Plan:
Make sure Cmake works fine and all tests passes.
Imported this change to `OpenR` project and tested on our local testbed.
Captured some packets exchanged between PUB/SUB and ROUTER/ROUTER sockets
pairs. Verify that `TCLASS` value is set as per expectation.

Tasks: #2208

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -172,4 +172,17 @@ void zmq::set_ip_type_of_service (fd_t s_, int iptos)
 #else
     errno_assert (rc == 0);
 #endif
+
+    rc = setsockopt(
+        s_,
+        IPPROTO_IPV6,
+        IPV6_TCLASS,
+        reinterpret_cast<const char*>(&iptos),
+        sizeof(iptos));
+
+#ifdef ZMQ_HAVE_WINDOWS
+    wsa_assert (rc != SOCKET_ERROR);
+#else
+    errno_assert (rc == 0);
+#endif
 }


bluca:
Thanks! Will it handle gracefully the case where IPv6 is not enabled? If not we can fix it up.

bluca:
Also,  for your next contributions, could you please follow the contributing guidelines for the git commit message?

```
Problem: X is broken

Solution: do Y and Z to fix X
```

saifhhasan:
Thanks @bluca for quick response. Amazed by the speed of zeromq community :) I will follow the guide for future contributions (somehow missed that part).

I have not tested with IPv4 only system yet. Not sure what will be the behaviour (it might need masking off in case if IPv6 is not enabled).

bluca:
Welcome to the community :-) What you've just seen in action is our C4.1 process: https://rfc.zeromq.org/spec:42/C4/ - IE. merge first, ask questions later :-)

I'll give it a go for the v4 case and fixup if there's a fallout, no worries.

bluca:
ENOPROTOOPT is returned if IPv6 is not available, so it's a simple fix: https://github.com/zeromq/libzmq/pull/2210

PR #2210, opened 2016-11-14
Problem: IPV6_TCLASS setsockopt fails on IPv4 socket

bluca:
Solution: if setsockopt errors out and errno is set to ENOPROTOOPT
ignore it and carry on.


saifhhasan:
Thanks for following up to support on v4 only platform. #moving_fast

Issue #2169, opened 2016-10-22
More explicit docs for setsockopt() with strings regarding NUL-termination


rotty:
As far as I can glance from the implementation, `setsockopt()` does not require NUL-termination of string options (e.g. `ZMQ_PLAIN_PASSWORD`). This is quite nice behavior for rust language bindings, as rust strings are not NUL-terminated, but a sequence bytes of UTF-8 encoded string, along with the length in bytes.

However, I could not find documentation of NUL-termination being optional, so I feel a bit uncomfortable relying on that in the bindings. Even if I strongly assume this is more-or-less intended, and will not change easily, I'd like to see this documented explicitly.

Adding something along the following lines near the beginning of `zmq_setsockopt(3)` might suffice:

> When the option value type is listed as "character string", an optionally NUL-terminated string is expected, unless more specific requirements are given for that option.

bluca:
The setsockopt itself mimicks the BSD socket API so as you say it does not require null termination, as it takes an explicit size.

However we would need to check each and every internal client of the get/setsockopt, to make sure that when a string is passed, nowhere in the code there are assumption about this.

If it is all safe, then by all means please send a PR to update the docs with your suggestion and we'll merge it.

rotty:
I'm not sure why internal clients need to be checked, can you elaborate? I'm not talking about preserving embedded NUL-bytes across a `setsockopt()`, `getsockopt()` sequence, but only that it is safe to pass a non-NUL terminated string without embedded NUL bytes to setsockopt(). Checking the implementation of  `zmq_setsockopt()` alone would suffice to ensure that property, I think.

If we want to guarantee preserving embedded NUL bytes, more checking indeed needs to be done. I am not sure, however, if *that* guarantee should be part of the API contract. If there is consensus this is an additionally desired property, I can try to extend the scope of the resulting PR to include it, however.

bluca:
Sorry, I replied from the phone so I wasn't clear and I didn't look at the code.

What I was concerned about was possible use of char pointers, but I've now seen that all the strings are saved in std:string types and the content is used via c_str() in other modules, so we are completely fine.

Personally I'm OK with adding this to the docs if it helps users or writers of bindings, but in the future we'll need to take care in using the same mechanism if more string parameters are added. I don't think it will be a problem, so please do send the correction to the doc. Thanks!

rotty:
[...]

PR #2216, opened 2016-11-16
Document terminating NUL behavior in API more explicitly

rotty:
This addresses #2169.


rotty:
Note that this is somewhat spiritually at odds with API change made to `ZMQ_CURVE_PUBLICKEY`:

> NOTE: an option value size of 40 is supported for backwards compatibility,
though is deprecated.

Although the actual wording introduced does not conflict, as the curve options are listed with type "Z85 text string", not "character string".

Issue #1599, opened 2015-10-03
size of zmq_msg_t is not known to FFI wrappers


Asmod4n:
When writing a FFI wrapper for libzmq you have to guess the size of `zmq_msg_t`

What could help against this is to define a public function to return `sizeof(zmq_msg_t)`, where would be a good place to expose it?

hintjens:
Why can't you just use 'sizeof (zmq_msg_t)' in your code? The type is
defined in zmq.h as a structure.

Asmod4n:
FFI doesn't have access to typedefs made in C, only to functions. The libuv project had a similar issue, so they made functions to return the size of each type they export publicly.

hintjens:
You could try using the zmq_has method.

Asmod4n:
@hintjens zmq_has returns the size of the opaque zmq_msg_t struct?

something like this is needed for FFI wrappers of libzmq:
```c
size_t
zmq_msg_t_size(void)
{
    return sizeof(zmq_msg_t);
}
```

Asmod4n:
FFI doesn't have access to header files, nor can it interpret typedefs, or tell you the size of a opaque struct.

hintjens:
I meant, you could extend zmq_has to return this value. Actually a better
solution: add it to zmq_ctx_get () e.g.

    size = zmq_ctx_get (void *context, ZMQ_MSG_SIZE)

Though it suggests this is variable per context, which is not the case,
it's at least clean for the caller.

bjourne:
Pretty please add this function. I, and the other developers who are working on the Factor language, has spent a huge amount of time (weeks!) debugging an issue which was caused by libzmq increasing the size of the `zmq_msg_t` structure. Our code assumed the old format where the struct was 40 bytes so new versions of libzmq, where the struct is 64 bytes, would cause very hard to debug memory overwrite problems.

Afaict, the libzmq developers idea is that `zmq_msg_t` should be opaque. Users of the library shouldn't need to know details of it so that the format can be updated in the future. That is good engineering practice, but the struct isn't completely opaque because you need to know the size of it to allocate correctly sized buffers. One way to solve it is to add the suggested function `zmq_msg_t_size` which returns the size. Another is for libzmq itself to handle allocation. So for example instead of:

```c
zmq_msg_t *msg = (zmq_msg_t *)malloc(64);
zmq_msg_init_size(msg, 1024);
...
```
You would have functions that return pointers to newly allocated struct instances, like this:
```c
[...]

PR #2220, opened 2016-11-20
Problem: size of zmq_msg_t is not known to FFI wrappers

bluca:
Solution: add a ZMQ_MSG_T_SIZE context read-only option so that
wrappers can call zmq_ctx_get (ctx, ZMQ_MSG_T_SIZE) to get the
size at runtime.

Fixes https://github.com/zeromq/libzmq/issues/1599

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 0MQ version 4.2.1 stable, released on 201x/xx/xx
 =============================================
 
+* New Context options:
+  - ZMQ_MSG_T_SIZE
+
 
 0MQ version 4.2.0 stable, released on 2016/11/04
 =============================================
--- a/doc/zmq_ctx_get.txt
+++ b/doc/zmq_ctx_get.txt
@@ -57,6 +57,13 @@ zero if the "block forever on context termination" gambit was disabled by
 setting ZMQ_BLOCKY to false on all new contexts.
 
 
+ZMQ_MSG_T_SIZE: Get the zmq_msg_t size at runtime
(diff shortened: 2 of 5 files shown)


bluca:
@Asmod4n @bjourne please let me know if this solves your problem.

bjourne:
In the long run it will, thanks! Because we want to support older zeromq versions too, we can't rely on the feature yet.

bluca:
You can check the version:

2.x: 32 + sizeof (void *)
3.x: 32 bytes
4.1.0: 48 bytes
4.1.1 and up: 64 bytes

Don't both with 4.1.0, and I wouldn't spend time on 2.x either, but up to you.

Asmod4n:
@bluca this kind of works, but FFI has no access to the preprocessor, so you just have to use the number of it and its ok.

bluca:
Ah I see. Once a new option is released we commit to never change the number, so it should be all right.

bluca:
@bjourne @Asmod4n I'm not 100% sure of the name or the format of this option (ctx option vs zmq_has option vs its own method) so I would like to move this behind DRAFT flag to do a bugfix release while we think about it. Would this be OK for you? Are you already using this in your bindings?

Note that I'm fine with this being available in some form, it's just the form that I'd like more time to think about before releasing it

Asmod4n:
@bluca looks good, i wouldn't mind a method either.
FFI can check if a method exists in a library at runtime but it cannot check if a preprocessor #define exists.

PR #2233, opened 2016-12-01
bug fix: AIX pollset error

hnwyllmm:
buf fix:  AIX only, pollset 'rm_fd' set fd_table to null first then set pe->fd to retired_fd


bluca:
Thanks! Next time, instead of having a second commit that fixes a problem introduced by the first, please rebase and squash it

PR #2239, opened 2016-12-05
avoid crashing in the multi-thread operation for std::vector

loachfish:
1. stl container is not thread safety
2. rm_fd() and loop() end to clear the retired event source will in multi-thread operation
3. may be crashed in reaper thread to delete the items in the std::vector as the source is nullptr


bluca:
@loachfish thanks.

In which situation does the problem appear? Do you have a test case that reproduces the failure?

I didn't think the pollers would get ran from multiple threads

PR #2250, opened 2016-12-10
Problem: check for item.fd causes zloop_poller regression

bluca:
Solution: fix the check for the socket.
This regression happens when using zloop with zmq_pollitem_it with
only file descriptors registered through zloop_poller.

@minrk narrowed a test failure I got in our application at work to commit f302d8a7b6cc74f6e8308e1d61ffb4804961c346 and this change fixes it. As the commit says we are simply registering some file descriptors and using them in a CZMQ zloop, but the events stopped working with that commit.

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -834,7 +834,7 @@ inline int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
         for (int j = j_start; j < found_events; ++j) {
             if (
                 (items_[i].socket && items_[i].socket == events[j].socket) ||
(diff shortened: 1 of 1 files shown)


bluca:
Same regression also affected zbeacon, after this fix the issue mentioned at https://github.com/zeromq/zyre/issues/496 no longer happens for me.

minrk:
Interesting. This *seems* wrong to me (if either item is a zmq socket, neither FD should be considered), but if it fixes something, A+!

bluca:
It did seem wrong to me as well, and haven't fully wrapped my head around it yet. As soon as I have time i want to add a test case to cover this and hopefully look into it again

PR #2255, opened 2016-12-15
Problem: no mixed FD and zmq socket zmq_poll test

bluca:
Solution: add one

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,6 +138,7 @@ test_dgram
 test_base85
 test_bind_after_connect_tcp
 test_sodium
+test_zmq_poll_fd
 tests/test*.log
 tests/test*.trs
 src/platform.hpp*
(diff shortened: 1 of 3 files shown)


bluca:
@minrk - without the change of https://github.com/zeromq/libzmq/pull/2250 this test case fails, so hopefully it can help give some insight in what was going wrong

minrk:
Thanks! I see what was wrong now, and #2250 is almost right. I think I can come up with a test that will fail with master, and open a PR with the fix.

bluca:
Perfect, thanks!

minrk:
#2257 should be the 'right' fix for #2250, though once I realized the typo that prompted #2250, it's a safety check against internal code in libzmq, not against user code. That makes it much less likely to happen, and can't be triggered by a test case.

Issue #2254, opened 2016-12-14
Problem: zmq 4.1.6 cannot pub msg to a zmq 2.x


laplaceyang:
In some of my code, my server use the zmq 4.1.6 with PUB mode, and another old server with SUB mode build with zmq 2.2.0.

Then I cannot receive anything in the sub server.

Here is the example:
[https://gist.github.com/laplaceyang/8cf1c004832f4ef618c69ceb868ef33b](https://gist.github.com/laplaceyang/8cf1c004832f4ef618c69ceb868ef33b)
build the PUB server with zmq 4.1.6 and SUB server with zmq 2.x

but if i build the server with zmq 3.x, it works.

It seems that change in stream_engine.cpp lead to this problem

---
In zmq 3.x, we have function

![image](https://cloud.githubusercontent.com/assets/7036040/21185741/d6852c80-c24c-11e6-8041-b17e27e56090.png)

we  inject a phantom subscription message and process identity msg in the same function

but in zmq 4.x, we separate it to two function

identity process func
![image](https://cloud.githubusercontent.com/assets/7036040/21185863/5538de00-c24d-11e6-869f-bab393c48e3e.png)
inject phantom subscription message function
![image](https://cloud.githubusercontent.com/assets/7036040/21185885/6f9451b2-c24d-11e6-8d85-8f1dfe18d444.png)

This change lead that,we should trigger two socket in event to finish subscribe.

but I dump the tcp package,
The SUB server just send one package to PUB.

![image](https://cloud.githubusercontent.com/assets/7036040/21186221/c0222874-c24e-11e6-9a42-92c5950fd68e.png)
After this no more package exchanges

laplaceyang:
well , I create a pull request just now. pr[https://github.com/zeromq/libzmq/pull/2256](https://github.com/zeromq/libzmq/pull/2256)

PR #2256, opened 2016-12-15
fix bug zmq4.x PUB msg to ZMTP1.0 SUB server

laplaceyang:
This pr fix issues[https://github.com/zeromq/libzmq/issues/2254](https://github.com/zeromq/libzmq/issues/2254) which zmq4.x pub msg to sub server run with ZMTP 1.0 protocol, such as zmq 2.x

The problem was included in pr[https://github.com/zeromq/libzmq/pull/542](https://github.com/zeromq/libzmq/pull/542)

### DIFF
--- a/src/stream_engine.cpp
+++ b/src/stream_engine.cpp
@@ -752,10 +752,19 @@ int zmq::stream_engine_t::process_identity_msg (msg_t *msg_)
         errno_assert (rc == 0);
     }
 
-    if (subscription_required)
-        process_msg = &stream_engine_t::write_subscription_msg;
-    else
-        process_msg = &stream_engine_t::push_msg_to_session;
+    if (subscription_required) {
+        msg_t subscription;
+
+        //  Inject the subscription message, so that also
+        //  ZMQ 2.x peers receive published messages.
+        int rc = subscription.init_size (1);
+        errno_assert (rc == 0);
+        *(unsigned char*) subscription.data () = 1;
+        rc = session->push_msg (&subscription);
+        errno_assert (rc == 0);
+    }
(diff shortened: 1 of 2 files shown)


bluca:
I'm not sure I follow this change. At a first sight, it seems that identical code is being inlined. Could you please explain?

laplaceyang:
@bluca If we send PUB msg (4.x) to SUB server (2.x). 
At first situation, we send handshake to SUB. And the SUB send identity to us at the same time.

Then in SUB end, the process_msg point is setted to `stream_engine_t::process_identity_msg`, because we detect that peer is a ZMTP1.0 Server.

Then we are out of handshake function, and go back to` zmq::stream_engine_t::in_event () `

and execute `rc = (this->*process_msg) (decoder->msg ()); `
we execute the `process_identity_msg` actually.
At the end of `process_identity_msg`, we set process_msg point to `write_subscription_ms`g by old code.
To wait for next in_event.

But unfortunately， the old zmq 2.x server will never send msg to us again, so no in_event will be tiggered, and we will not inject fake subscribe msg to session.

**What's wrong is that, we now need two socket in event to finish subscribe in old master code of PUB end, but the zmq2.x actually send only one msg.**

In new code, I remove the `write_subscription_msg` and move the code in process_identity_msg function, so that we can finish subscribe by one in_event. which is the ZMTP 1.0 behavior.

And also, it's what we do in zmq 3.x. So the zmq 3.x behave right.

There is also sample code in my issues[https://github.com/zeromq/libzmq/issues/2254](https://github.com/zeromq/libzmq/issues/2254)

PR #2257, opened 2016-12-15
zmq_poll: only compare FD when neither item is a zmq socket

minrk:
The bug fixed by #2250 was that `j` is an index into `events`, not `items_`. Removing the comparison altogether in #2250 ought to be safe because unlike `items_`, `events` comes internally from libzmq. Still, I'm more comfortable with only comparing FD if neither side is a zmq socket event.

Still passes the test case in #2255

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -834,7 +834,7 @@ inline int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
         for (int j = j_start; j < found_events; ++j) {
             if (
                 (items_[i].socket && items_[i].socket == events[j].socket) ||
-                (!items_[i].socket && items_[i].fd == events[j].fd)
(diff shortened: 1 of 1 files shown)


bekritsk:
I had discovered an issue related to this that when polling for stdin, zmq_poll would return a positive integer to the  zmq_poll(), but did not have the .revents field filled properly (https://lists.zeromq.org/pipermail/zeromq-dev/2016-December/031242.html). 
This fix seems to solve that problem, as well.

bluca:
Thanks for reporting that. Could you please reply this to your mailing list post as well, so that people searching for that problem in the future will see the resolution as well?

bekritsk:
I had discovered an issue related to this that when polling for stdin,
zmq_poll would return a positive integer to the zmq_poll(), but did not
have the .revents field filled properly (
https://lists.zeromq.org/pipermail/zeromq-dev/2016-December/031242.html).
This fix seems to solve that problem, as well.

PR #2259, opened 2016-12-16
Fixed handling of WSAEWOULDBLOCK to be generic

Cziken:
I don't know what was the intention of this early if statement but
now this is properly evaluated in wsa_error_no function if this is
performance issue I suggest moving evaluating this error code to the
beginning of wsa_error_no.


Cziken:
OK, I missed something. Can You tell me how to run ut at local machine? :) I will be glad!

PR #2262, opened 2016-12-16
Problem: Duplicated code

Cziken:
Solution: set_curve_key method in options_t struct with little
improvement in switch-case block

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -90,6 +90,35 @@ zmq::options_t::options_t () :
 #endif
 }
 
+int zmq::options_t::set_curve_key(uint8_t * destination,  const void * optval_, size_t optvallen_)
+{
+    char * optval = (char *) optval_;
+    switch (optvallen_) {
+
+        case CURVE_KEYSIZE:
+            memcpy (destination, optval_, optvallen_);
+            mechanism = ZMQ_CURVE;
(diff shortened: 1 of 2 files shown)


bluca:
This change actually breaks CURVE, so I'm reverting it.

How did you test it?

Cziken:
I've run tests from ci_build.sh, all of them passed so this is strange. I will take a look on it at monday :) sorry, I'm total noob in this project but I will try to track it down and fix it. Do you have crash or somewhere is Null instead of actual value?

bluca:
you can check the test runs on linux osx and windows:

https://travis-ci.org/zeromq/libzmq/builds/184582112
https://ci.appveyor.com/project/zeromq/libzmq/build/build-924
https://libzmq.cloudapp.net/job/libzmq/1348/console

also a simple ```make check``` will show that it fails

Cziken:
Maybe we can add this info about make check to docs? I'll fix this at Monday and open another pull req. Sorry for inconvenience and have a nice weekend! :)

bluca:
It's already there https://github.com/zeromq/libzmq/blob/master/INSTALL#L69

Cziken:
I was always too eager :( ok, thanks for the tip :)

PR #2271, opened 2016-12-22
Added some improvements 

Cziken:

### DIFF
--- a/src/mailbox_safe.cpp
+++ b/src/mailbox_safe.cpp
@@ -26,6 +26,7 @@
     You should have received a copy of the GNU Lesser General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
+#include <algorithm>
 
 #include "precompiled.hpp"
 #include "mailbox_safe.hpp"
@@ -59,16 +60,8 @@ void zmq::mailbox_safe_t::add_signaler (signaler_t* signaler)
 
 void zmq::mailbox_safe_t::remove_signaler (signaler_t* signaler)
 {
-    std::vector<signaler_t*>::iterator it = signalers.begin();
-
     // TODO: make a copy of array and signal outside the lock
-    for (; it != signalers.end(); ++it){
-        if (*it == signaler)
-           break;
-    }
-
-    if (it != signalers.end())
-        signalers.erase(it);
+    signalers.erase(std::remove(signalers.begin(), signalers.end(), signaler), signalers.end());
 }
 
 void zmq::mailbox_safe_t::clear_signalers ()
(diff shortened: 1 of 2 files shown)


src/radio.cpp
```diff
@@ -100,18 +100,18 @@ void zmq::radio_t::xwrite_activated (pipe_t *pipe_)
 
 void zmq::radio_t::xpipe_terminated (pipe_t *pipe_)
 {
-    for (subscriptions_t::iterator it = subscriptions.begin (); it != subscriptions.end (); ) {
```
bluca: There's a reason I did this change, perhaps there should be a comment.
Iterators are invalidated when deleted, so this will cause a read of freed memory and potentially a crash.

See https://github.com/zeromq/libzmq/pull/1771

src/mailbox_safe.cpp
```diff
@@ -59,16 +60,8 @@ void zmq::mailbox_safe_t::add_signaler (signaler_t* signaler)
...
-
     // TODO: make a copy of array and signal outside the lock
-    for (; it != signalers.end(); ++it){
-        if (*it == signaler)
-           break;
-    }
-
-    if (it != signalers.end())
-        signalers.erase(it);
+    signalers.erase(std::remove(signalers.begin(), signalers.end(), signaler), signalers.end());
```
bluca: This will break on Solaris, AIX and the like as it needs newer C++ standard library

src/radio.cpp
bluca: https://travis-ci.org/zeromq/libzmq/jobs/186068184

bluca:
Declining as it breaks backward compatibility and re-introduces a read-after-free regression

src/radio.cpp
Cziken: Maybe better way to express your intentios will be
it = subscriptions.erase(it);
This will clearly stand a point about what You are doing there :)

src/mailbox_safe.cpp
Cziken: Due to #include<algorithm> ?

src/radio.cpp
bluca: That doesn't work either unfortunately

Cziken: So... Standard lies? ;) 

Erase returns iterator following the last removed element. If the iterator pos refers to the last element, the end() iterator is returned.

bluca: Standard implementation unfortunately have bugs or subtle differences on decades-old platforms and that caused a regression. Unfortunately it was almost a year ago so I can't remember exactly anymore, but I think it was on Solaris 10

Cziken: OK, this project is fancy due to lot of dependencies from legacy compilers. But still I will try to contribute some code to it

bluca: You are very welcome to contribute, and we are thankful for that.
Sorry to be strict, but we have many users on legacy platforms, and everytime something breaks I have to scramble and spend time to fix it up again. So if I see something that I know will be broken because we've already seen it, I can't accept it.

PR #2272, opened 2016-12-22
Problem: fix for read-after-free not documented

bluca:
Solution: add a comment as it's easy to miss and very tempting to
"optimize"

### DIFF
--- a/src/radio.cpp
+++ b/src/radio.cpp
@@ -100,6 +100,8 @@ void zmq::radio_t::xwrite_activated (pipe_t *pipe_)
 
 void zmq::radio_t::xpipe_terminated (pipe_t *pipe_)
 {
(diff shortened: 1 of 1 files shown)


src/radio.cpp
```diff
@@ -100,6 +100,8 @@ void zmq::radio_t::xwrite_activated (pipe_t *pipe_)
 
 void zmq::radio_t::xpipe_terminated (pipe_t *pipe_)
 {
+    //  NOTE: erase invalidates an iterator, and that's why it's not incrementing in post-loop
+    //  read-after-free catched by Valgrind, see https://github.com/zeromq/libzmq/pull/1771
```
gasche: catched -> caught

bluca:
@c-rack ping :-)

Issue #2227, opened 2016-11-27
troubleshooting curve connections


amirtuval:
Hi,

I have successfully been able to use curve to encrypt the communication between client and server.
It works great.

My main concern is that when the curve handshake fails for some reason, it is very hard to troubleshoot.
For example, if the client has the wrong server public key, the client will timeout with an EAGAIN error - exactly the same behavior as if the server is down or if there's no network, so it is very hard to figure out the root cause is the wrong certificate.

The only thing I've been able to see, is that the server writes "CURVE I: cannot open client HELLO -- wrong server key?" to stdout, but that seems more of a debug message than anything else, and would probably be removed in future versions. (The code says //  Temporary support for security debugging - curve_server.cpp:314).
Also, this is running on both windows and linux. On windows it's running as a windows service, so writing to stdout is not a good solution.

What I'd like to have is some programatic interface that tells me handshake failed, so I can decide what to do (like write to application log file, prompt the user etc.).
This can be in the form of an error or an exception, or in the form of an event, like a zmonitor "Handshake Failed" event.

Is there something like this that is already implemented that I missed? Any advice on this would be appreciated.

Thanks
Amir

bluca:
I don't think there is such a feature yet, but if you could send a PR to implement it it would be most welcome.

In general we don't use exceptions, so a zmonitor event as you suggested would probably be the best solution.

vtellier:
Hi,

I've been exactly into the same use case, (see [stackoverflow](http://stackoverflow.com/questions/41380623/zmq-encryption-how-to-know-when-the-handshake-failed/41381693)).

I tried to enter a bit (just a bit) into the code of libzmq and it looks like the ERROR command of [CurveZMQ protocol](https://rfc.zeromq.org/spec:26/CURVEZMQ/) is already fully implemented.
So actually what is missing is to broadcast the event to the user? I'm asking because this feature is more than a nice to have for me, so I would be happy to try to implement it.

When thinking about it, we have the [zmq_socket_monitoring](http://api.zeromq.org/4-2:zmq-socket-monitor) which is really network related.
Would it not make more sense to add another layer of monitoring just above, which would provide information only related the "internal ZMQ layers" (can't find a better name):
Beyond the encryption handshakes, it might also provide information depending on the configuration of the ZMQ socket. For instance, a ROUTER could notify about a routing fail due to wrong identity, or when it drops messages due to HWM reached.
This is just an idea of course, so far I just miss the handshakes state.

[...]

PR #2280, opened 2016-12-30
Fixed issue #2227

vtellier:
# Problem
See #2227 
ZMQ doesn't provide status about the encryption handshake. When it fails, the client tries to reconnect indefinitely.

# Solution
Added two new monitoring events:
 - **ZMQ_EVENT_HANDSHAKE_SUCCEED** is raised once the encryption handshake succeed
 - **ZMQ_EVENT_HANDSHAKE_FAILED** is raised when it failed
Both events are raised on server and client side.

Also, when the curve engine find out the client uses a different (server) public key, it sends back a hard error to the client so it doesn't tries to reconnect infinitely.
That error message contains an error code abitrarily set to 100:

```
// src/curve_server.cpp
status_code = "100";
```

This can (or should) be modified, as the error code is not read so far, I had to set a value.

_**Nb:** I compiled and tested my code under **Visual Studio 2015** (non XP), I did **not** tested with any language binding or other zmq projects._

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -417,6 +417,8 @@ ZMQ_EXPORT const char *zmq_msg_gets (zmq_msg_t *msg, const char *property);
 #define ZMQ_EVENT_CLOSE_FAILED      0x0100
 #define ZMQ_EVENT_DISCONNECTED      0x0200
 #define ZMQ_EVENT_MONITOR_STOPPED   0x0400
+#define ZMQ_EVENT_HANDSHAKE_FAILED  0x0800
+#define ZMQ_EVENT_HANDSHAKE_SUCCEED 0x1000
 #define ZMQ_EVENT_ALL               0xFFFF
 
 ZMQ_EXPORT void *zmq_socket (void *, int type);
--- a/src/curve_server.cpp
+++ b/src/curve_server.cpp
@@ -310,13 +310,14 @@ int zmq::curve_server_t::process_hello (msg_t *msg_)
                               sizeof hello_box,
                               hello_nonce, cn_client, secret_key);
     if (rc != 0) {
-        //  Temporary support for security debugging
-        puts ("CURVE I: cannot open client HELLO -- wrong server key?");
-        errno = EPROTO;
-        return -1;
+        // Hard error, the client knows a wrong server public key, it shall not try to reconnect using the same.
+        status_code = "100";
+        state = send_error;
+        rc = 0;
     }
+    else
+        state = send_welcome;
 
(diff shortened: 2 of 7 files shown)


include/zmq.h
```diff
@@ -417,6 +417,8 @@ ZMQ_EXPORT const char *zmq_msg_gets (zmq_msg_t *msg, const char *property);
 #define ZMQ_EVENT_CLOSE_FAILED      0x0100
 #define ZMQ_EVENT_DISCONNECTED      0x0200
 #define ZMQ_EVENT_MONITOR_STOPPED   0x0400
+#define ZMQ_EVENT_HANDSHAKE_FAILED  0x0800
```
bluca: Please move these in the DRAFT section at the bottom of the file (and copy them in src/zmq_draft.hpp too)

src/curve_server.cpp
```diff
@@ -310,13 +310,14 @@ int zmq::curve_server_t::process_hello (msg_t *msg_)
                               sizeof hello_box,
                               hello_nonce, cn_client, secret_key);
     if (rc != 0) {
-        //  Temporary support for security debugging
-        puts ("CURVE I: cannot open client HELLO -- wrong server key?");
-        errno = EPROTO;
-        return -1;
+        // Hard error, the client knows a wrong server public key, it shall not try to reconnect using the same.
```
bluca: Please revert this change - as mentioned in https://github.com/zeromq/libzmq/issues/2227#issuecomment-269786977 this is not backward compatible

src/socket_base.cpp
```diff
@@ -1678,6 +1678,16 @@ void zmq::socket_base_t::event_disconnected (const std::string &addr_, zmq::fd_t
     event(addr_, fd_, ZMQ_EVENT_DISCONNECTED);
 }
 
+void zmq::socket_base_t::event_handshake_failed(const std::string &addr_, int err_)
+{
+	event(addr_, err_, ZMQ_EVENT_HANDSHAKE_FAILED);
```
bluca: The formatting in the library is spaces-only, 4 spaces indentation, please fix this and other tabs in this diff

bluca:
The test_monitor unit test is broken by this change, could you please have a look and fix it?

See https://travis-ci.org/zeromq/libzmq/builds/187723849

bluca:
@vtellier ping - are you looking into this?

vtellier:
Yes! Sorry I was away, back online in few minutes and fix it

bluca:
Thank you!

PR #2281, opened 2016-12-30
Code formatting + reverted hard error handshake fail

vtellier:
## Regarding the regression introduced in the PR #2280 :

 - Moved new events in draft section + added to zmq_draft.h
 - Removed the remainning tabs
 - Reverted the _hard_ error (_back_ to soft error) in `curve_server.cpp`

## The feature doesn't work anymore:
 - The ZMQ_EVENT_HANDSHAKE_FAILED is raised only on the server's side.
 - The ZMQ_EVENT_HANDSHAKE_SUCCEED still works on both sides

## About the build fails:
 - I was not able to replay the tests on my computer so far.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -417,8 +417,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (zmq_msg_t *msg, const char *property);
 #define ZMQ_EVENT_CLOSE_FAILED      0x0100
 #define ZMQ_EVENT_DISCONNECTED      0x0200
 #define ZMQ_EVENT_MONITOR_STOPPED   0x0400
-#define ZMQ_EVENT_HANDSHAKE_FAILED  0x0800
-#define ZMQ_EVENT_HANDSHAKE_SUCCEED 0x1000
 #define ZMQ_EVENT_ALL               0xFFFF
 
 ZMQ_EXPORT void *zmq_socket (void *, int type);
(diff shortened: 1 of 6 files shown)


vtellier:
The test fails because the last monitoring event (**ZMQ_EVENT_MONITOR_STOPPED**) is not catched by the handler.

```
void zmq::socket_base_t::stop_monitor (bool send_monitor_stopped_event_)
{
    // this is a private method which is only called from
    // contexts where the mutex has been locked before

    if (monitor_socket) {
        if ((monitor_events & ZMQ_EVENT_MONITOR_STOPPED) && send_monitor_stopped_event_)
            monitor_event(ZMQ_EVENT_MONITOR_STOPPED, 0, ""); // This call is performed, but the event doesn't reaches destination.
        zmq_close (monitor_socket);
        monitor_socket = NULL;
        monitor_events = 0;
    }
}
```

I tried to comment all the calls to the new events, but this is not enough to solve the problem.

I have a train to take right now, so I have to stop for today, I'll be back on this issue ASAP.

bluca:
Thanks! I'll have time to have a look at the test in a couple of hours, no worries

bluca:
PR fix coming up soon

bluca:
@vtellier see https://github.com/zeromq/libzmq/pull/2282

Thanks for your contributions. Do you think is there anything useful we could add  to the event value?

vtellier:
@bluca You're welcome, I'm glad to contribute :)

About the event value, at the beginning I was thinking that a more detailled information about the reason of the failure could be interesting. However I do not need it, because I think that most handshakes fails are due to keys discrepency, either the server changed its key pair, either the client has a wrong one, in both cases it's the same.

For the moment I don't see anything interesting event value, so I set it 0. this can still be changed of course.
Also, I don't feel my knowledges in ZMQ good enough to define it right now, but _in the future_ if more details are needed it will be possible to define zero as _"no detail"_ while other values can be fixed.

bluca:
That's fine. It's still in DRAFT state so we can change the values as much as required until it's declared stable, so if you find anything useful while testing it just send a PR.

PR #2282, opened 2016-12-30
Problems: test_monitor failing, DRAFT events leaking when draft APIs are disabled

bluca:
Solutions: see commits

The basic issue was that the catch-all flag, ZMQ_EVENT_ALL, was enabling the new events. This is kindof expected given the name. But to further clarify, I've added a note to the docs. In my opinion using such a flag implies that new events will be enabled as they are added. It wouldn't make any sense to have a ZMQ_EVENT_ALL flag that doesn't trigger all events (stable ones, that is).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,7 @@
 
 * New DRAFT (see NEWS for 4.2.0) Context options:
   - ZMQ_MSG_T_SIZE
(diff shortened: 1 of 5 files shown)


bluca:
Anyone for a merge? @c-rack @sappo

PR #2283, opened 2016-12-31
Fixed issue #2227 second part

vtellier:
**Problem:** the feature was signaling the handshake fail only on the server side (see #2281 )

**Solution:**
 - removed the previously added encryption_error => _less changes less bug_
 - handshake fail is now signaled when an error happen while the
   mechanism is still hanshaking

> @bluca  **Be careful of this before merge!**
> In `stream_engine.cpp` I check `if(mechanism->status() == mechanism_t::handshaking)`
> I am not 100% sure that this does not leads to crashes (if mechanism becomes null for some reason) in race conditions especially.

It was the case, see commit e6b057f05cf5a861ffc652805b025482554a5d34

### DIFF
--- a/src/session_base.cpp
+++ b/src/session_base.cpp
@@ -421,8 +421,7 @@ void zmq::session_base_t::engine_error (
     if (pipe)
         clean_pipes ();
 
-    zmq_assert (reason == stream_engine_t::encryption_error
-             || reason == stream_engine_t::connection_error
+    zmq_assert (reason == stream_engine_t::connection_error
              || reason == stream_engine_t::timeout_error
              || reason == stream_engine_t::protocol_error);
 
@@ -434,7 +433,6 @@ void zmq::session_base_t::engine_error (
             else
                 terminate ();
             break;
-        case stream_engine_t::encryption_error:
         case stream_engine_t::protocol_error:
             terminate ();
             break;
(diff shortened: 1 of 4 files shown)


vtellier:
Some tests still fail, I can have a look at it probably tomorrow.

vtellier:
I also forgot to mention that the meaning of the failure event changes a bit. Since the server doesn't sends back an explicit answer to the client, the client only knows that a disconnection happened during the handshake.
This can also happen with network disturbance.

bluca:
That's fine, I marked this as WIP then to give you more time, thanks

vtellier:
It look like the tests are much better, but I didn't had the same results between e6b057f05cf5a861ffc652805b025482554a5d34 and 339b826f3021a8b0ccdd154797c9a23d145bb04c and the discrepency is just the `tests/README.md` file.

@bluca if you think that everything looks good, feel free to merge.

bluca:
OSX tests with lots of resources (test_many_sockets for example) will fail from time to time since the Travis VMs are heavily over-subscribed, so that's fine.

Could you please rebase your branch and push -f to take away the recursive merge node?

vtellier:
Absolutely, sorry this is my first contribution on Github, i'm still not fluent with the best practices.
I just did it.

bluca:
No problem, thank you!

PR #2287, opened 2017-01-01
Problem: build fails on GNU/Hurd as it does not support IPV6_TCLASS

bluca:
Solution: check if IPV6_TCLASS is defined so that when Hurd adds
support it will just work. Also it will avoid tripping over this on
other similar legacy systems.

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -173,8 +173,8 @@ void zmq::set_ip_type_of_service (fd_t s_, int iptos)
     errno_assert (rc == 0);
 #endif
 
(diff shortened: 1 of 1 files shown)


c-rack:
I assume the failing test is unrelated to this PR.

bluca:
Thanks! Yes I think it's a side effect of https://github.com/zeromq/libzmq/pull/2280

bluca:
But it happens only with DRAFT APIs so I'm not too worried for now as it's still WIP

PR #2294, opened 2017-01-06
Windows cmake build fixed

### LINKED ISSUE #2253: Precompiled header doesn't work using CMake to generate MSVC project
```
cd /zmq/src/dir
mkdir build
cd build
cmake -G "Visual Studio 14 2015 Win64" .. 
```

I can compile the code in vs2015, but when I check the project, it doesn't use precompiled headers.

Using QtCreator with MSVC2015 kit, it tries to build the project with precompiled headers on, but failed to file the file.

> libzmq\src\address.cpp(30): fatal error C1083: Cannot open precompiled header file: 'precompiled.pch': No such file or directory

In the root CMake file at line 692, it supposes to enable the precompiled header option for precompiled.hpp file, if I understand it correctly.

Am I doing something wrong?

JohanMabille:
Fixes #2253.

@bluca This solves only the build, not the installation on windows. I will make another PR for this problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,6 +469,7 @@ endif ()
 # source files
 
 set (cxx-sources
+        precompiled.cpp
         address.cpp
         client.cpp
         clock.cpp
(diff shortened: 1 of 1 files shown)


bluca:
Thank you!

Issue #2228, opened 2016-11-28
Problem: infinite hwm of version 4.2 lead to catch no msg 


laplaceyang:
In release version of 4.2, we got a change in `zmq::pipe_t::set_hwms`, that 0 means infinite hwm.
```C++
 void zmq::pipe_t::set_hwms (int inhwm_, int outhwm_)
 {
     int in = inhwm_ + inhwmboost;
     int out = outhwm_ + outhwmboost;
 
     // if either send or recv side has hwm <= 0 it means infinite so we should set hwms infinite
     if (inhwm_ <= 0 || inhwmboost <= 0)
         in = 0;
 
     if (outhwm_ <= 0 || outhwmboost <= 0)
         out = 0;
 
     lwm = compute_lwm(in);
     hwm = out;
 }
```
but in some case ,I use the dealer mode, and **set sndhwm after connect**, lead to the hwm and lwm eq 0. which is not my origin idea.

Then if one side (dealer mode)(side A) is infinite hwm, and another (router mode)(side B) set hwm to a limit number. Then A side will never send_activate_write  to the peer, accoding to the follow code.
```C++
    if (lwm > 0 && msgs_read % lwm == 0)
        send_activate_write (peer, msgs_read);
```

Finally, we got a full hwm by check_hwm, and reset_pollin. Since then, no message can be caught.

bluca:
This was changed to fix a bug, can't remember which, I'll dig it up later today

bluca:
So I was thinking of this: https://github.com/zeromq/libzmq/pull/1649 but that's unrelated, I think.

Looking at the documentation, 0 meant infinite already in 2.1, so this hasn't really changed: http://api.zeromq.org/2-1:zmq-setsockopt

What changed is that now it's possible to modify the HWMs even after a bind/connect, since this PR: https://github.com/zeromq/libzmq/pull/1426 Is this what's causing you a problem?

Would you be able to upload to gist/pastebin a minimal test case that reproduces the issue you are facing? Thanks!

IA-32:
@bluca I just write a test case follow
[https://gist.github.com/IA-32/40374e2b40a5bfec56c7ce69749418eb](url)

IA-32:
@bluca  I just write a test case: 
[https://gist.github.com/IA-32/40374e2b40a5bfec56c7ce69749418eb](https://gist.github.com/IA-32/40374e2b40a5bfec56c7ce69749418eb)

[...]

PR #2299, opened 2017-01-09
Problem: set hwm after connect lead to infinite hwm

laplaceyang:
This pr is wrote to fix bug descibe here [Problem: infinite hwm of version 4.2 lead to catch no msg](https://github.com/zeromq/libzmq/issues/2228)

The pr can be divide into two part:
**1. fix: set hwm after connect lead the hwm of pipe wrong**
**2. fix: set hwm after lead to infinite hwm**
For example:
we set sndhwm to 20, rcvhwm to 20 after connect
```
[before]
socket_base ( hwm 1000, lwm 500 )
session_base ( hwm 1000, lwm 500 )

[after]
socket_base ( hwm 0, lwm 0 )
session_base ( hwm 1000, lwm 500 )
```

The behavior is not expected, so we need also to set peer hwm

after part 1, the result is descibe as follow
```
[before]
socket_base ( hwm 1000, lwm 500 )
session_base ( hwm 1000, lwm 500 )

[after]
socket_base ( hwm 0, lwm 0 )
session_base ( hwm 0, lwm 0 )
```

now our hwm is calc right, but infinite hwm is also not expected.
so, we include this change
```
-    inhwmboost(0),
-    outhwmboost(0),
+    inhwmboost(1),
+    outhwmboost(1),
```

Then 
```
[before]
socket_base ( hwm 1000, lwm 500 )
session_base ( hwm 1000, lwm 500 )

[after]
socket_base ( hwm 20, lwm 10 )
session_base ( hwm 20, lwm 10 )
```

Now the result is what we expected

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -1414,6 +1414,7 @@ void zmq::socket_base_t::update_pipe_options(int option_)
         for (pipes_t::size_type i = 0; i != pipes.size(); ++i)
         {
             pipes[i]->set_hwms(options.rcvhwm, options.sndhwm);
+            pipes[i]->set_peer_hwms(options.sndhwm, options.rcvhwm);
         }
     }
 
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -83,8 +83,8 @@ zmq::pipe_t::pipe_t (object_t *parent_, upipe_t *inpipe_, upipe_t *outpipe_,
     out_active (true),
     hwm (outhwm_),
     lwm (compute_lwm (inhwm_)),
-    inhwmboost(0),
-    outhwmboost(0),
+    inhwmboost(1),
+    outhwmboost(1),
     msgs_read (0),
     msgs_written (0),
     peers_msgs_read (0),
@@ -520,6 +520,11 @@ void zmq::pipe_t::set_hwms (int inhwm_, int outhwm_)
 
     lwm = compute_lwm(in);
     hwm = out;
+}
+
+void zmq::pipe_t::set_peer_hwms (int inhwm_, int outhwm_)
+{
+    peer->set_hwms(inhwm_, outhwm_);
 }
 
 void zmq::pipe_t::set_hwms_boost(int inhwmboost_, int outhwmboost_)
--- a/src/pipe.hpp
+++ b/src/pipe.hpp
@@ -133,6 +133,9 @@ namespace zmq
         //  Set the high water marks.
         void set_hwms (int inhwm_, int outhwm_);
 
+        //  Set the high water marks for peer.
+        void set_peer_hwms (int inhwm_, int outhwm_);
+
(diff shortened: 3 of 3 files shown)


bluca:
Thank you! For your next contribution, please add the details to the commit message text rather than the PR description, so that they are available in the git log

src/socket_base.cpp
```diff
@@ -1414,6 +1414,7 @@ void zmq::socket_base_t::update_pipe_options(int option_)
         for (pipes_t::size_type i = 0; i != pipes.size(); ++i)
         {
             pipes[i]->set_hwms(options.rcvhwm, options.sndhwm);
+            pipes[i]->set_peer_hwms(options.sndhwm, options.rcvhwm);
```
bluca: I don't think this is safe. In the inproc case, this might be owned and used from another thread. @ricnewton any opinion?

laplaceyang: @bluca  how about add mutex lock to the function set_hwms, since set_peer_hwms is actually called the set_hwms, and the set_hwms function is not called frequently

bluca: I'm afraid that won't work either, as the values are read in the hot path and there must be no locks there

laplaceyang: @bluca You are right, I had not considered inproc case before.
I think there are two solution:
[1] disable update hwm after connect, since this is not a common usecase
[2] add a command type in command.hpp to support modify pipe by it's owner thread. If we detect the protocol is inproc, we can set the command, otherwises, set_peer_hwm directly.

bluca: I think the command is the best option. I'd need to double check as can't remember exactly, but I think even with other transports this might be unsafe as the peer pipe is used by the I/O thread, unless I'm mistaken

laplaceyang: well, I find something. since the send_activate_write is called where `msgs_read % lwm == 0`
If we change hwm after connect, the behavior may be strange.

For example:
```
[Side A]
hwm = 150;
lwm = 75
msg_written = 100;
msg_read = 0;

[Side B]
hwm = 150;
lwm = 75
msg_read = 50;
```
If we lower the sndhwm / rcvhwm, so that the hwm = 90, and lwm = 45
Then, we will get a full hwm, but actually we only has 50 pkg;

laplaceyang: @bluca  well ,you are right, I review the code, and found the peer pipe is also used by other thread.

And I find something. since the send_activate_write is called where `msgs_read % lwm == 0`
If we change hwm after connect, the behavior may be strange.

For example:
```
[Side A]
hwm = 150;
lwm = 75
msg_written = 100;

[Side B]
hwm = 150;
lwm = 75
msg_read = 50;
```
If we lower the sndhwm / rcvhwm, so that the hwm = 90, and lwm = 45
Then, we will get a full hwm (now the msg_written > 90), but actually we only has 50 pkg (50 unconfirmed);

laplaceyang: @bluca  good idea

PR #2308, opened 2017-01-16
Updated buildbase and buildall to work with visual studio 2017

chuggafan:
This PR would add support for Visual Studio RC 2017 when building with buildall.bat and buildbase.bat


evoskuil:
I'd recommend dropping this line when VS2017 is released.

PR #2311, opened 2017-01-19
cmake: Set output name to libzmq for the libzmq-static target

### LINKED ISSUE #2310: Different name for the static library between cmake and [configure + make]
The name of the static library generated by cmake is `libzmq-static/[a|lib]` while the regular makefile creates `libzmq.[a|lib]`.

Ping @JohanMabille @minrk.

SylvainCorlay:
This fixes #2310 .

The problem was that the cmake installation produces different artefacts from the regular installation (the static library generated was called libzmq-static.[a|lib]) instead of libzmq.[a|lib].

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,6 +731,8 @@ if (MSVC)
     RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/bin"
     COMPILE_DEFINITIONS "DLL_EXPORT")
   add_library (libzmq-static STATIC ${sources})
+  # NOTE: on windows platform the static library name cannot be the same as the shared library name.
+  # since the .dll also requires a .lib companion file for linking.
   set_target_properties (libzmq-static PROPERTIES
     PUBLIC_HEADER "${public_headers}"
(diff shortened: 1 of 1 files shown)


JohanMabille:
That will break the build on Windows, static and dynamic library can't have the same name because dynamic library also produces a .lib.

SylvainCorlay:
@JohanMabille what do you think of this new version?

JohanMabille:
Looks good to me, CMake build and Linux Makefile builds are consistent.

bluca:
Could you please follow the template set up by the PR  and change the commit message to follow the guidelines?

Also can cmake create a symlink for backward compatibility?

SylvainCorlay:
I have updated the commit message to match the recommended template.

SylvainCorlay:
Regarding the symlink, are you sure this is necessary?

 - This makes the cmake behave like regular the makefile build (and the resulting artefact are now the same as the various package manager.) So there is no regression there.
 - Cmake users make use of `find_package` which will provide the right path.

bluca:
I know it was broken and it's correct now, but still if it breaks existing builds it's not good enough.
But if you say nothing will be broken then it's all good, thanks.

Issue #2307, opened 2017-01-14
Visual Studio: no way to check if libzmq was compiled with draft api


Asmod4n:
On Unix like systems pkg-config can tell you if libzmq was compiled with ZMQ_BUILD_DRAFT_API, cannot find something like that for Visual Studio on Windows.

Was there a reason zmq.h isn't adapting for it? (Leave all draft api declarations out if they aren't enabled)

bluca:
Can't VS support different "configuration" types? I know there's "release" and "debug". Wouldn't that be the better way to implement something like this?

Asmod4n:
If there is a runtime option to check if one is using libzmq from a "draft" build that would be ok.

bluca:
Can't those mechanism add compiler flags? I would imagine things like optimization or debugging symbols need to be added at compile time

Asmod4n:
That would be good too :)

Asmod4n:
Or no, those compiler flags would have to be added to all projects, or it cannot be checked outside of the libzmq build(?)

At the moment you have to guess if the draft api is enabled and see if the linker returns a error.

evoskuil:
@bluca VS configurations exist, but the chosen configuration is not registered in a central location in the operating system after a build (unlike pkg-config).

The proper way to expose a build option in a vs project is to define a property, which then becomes selectable in the VS IDE. A VS configuration (like the Release and Debug configurations I have defined) is just a named collection of such properties.

But there is no mechanism to project the result of one build as a dependency of a dependent build. This is something that requires a registry such as provided by package config. But that of course produces problems where registered builds produce conflicts. For example, someone builds a lib with a certain configuration (e.g. dynamic) and a dependent package needs a different configuration (e.g. static). So pkg-config is really not a solution either. I now only use pkg-config in a prefix directory in which I build and register all dependencies, coordinating their configuration via a generated installer. Otherwise build breaks are a constant problem.

So with VS it amounts to the same thing. Something external must coordinate the build configurations. This is easily accomplished on the build command line, as long as configuration options are properly exposed as properties. This is where xplat projects tend to break VS conventions (since they aren't generally understood) and get ugly. This is the analogy to the proper use of autotools configure.

Asmod4n:
What about just simply adding "draft" to zmq_has?

Asmod4n:
Oh, there is another issue.
The draft api cannot be used with ffi

bluca:
[...]

PR #2313, opened 2017-01-19
Problem: cannot check if libzmq was build with draft api

Asmod4n:
As discussed in #2307, adds "draft" capability to zmq_has

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -1517,6 +1517,10 @@ int zmq_has (const char *capability)
     if (strcmp (capability, "vmci") == 0)
         return true;
 #endif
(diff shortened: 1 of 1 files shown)


bluca:
Thanks! Could you please also fix the docs and the test?

https://github.com/zeromq/libzmq/blob/master/doc/zmq_has.txt
https://github.com/zeromq/libzmq/blob/master/tests/test_capabilities.cpp

Asmod4n:
@bluca btw, how could i get a binding under the zeromq org here?

bluca:
If you have a binding that you want in the zeromq github org just let me know the address and I will fork it in, and then add you as a collaborator and as the admin of the binding

PR #2317, opened 2017-01-21
Problem: cmake artefacts different from make generated artefacts

SylvainCorlay:
Solution: fix target output names.

The comment I added in #2311 about the windows output names was not correct since there is already a `-s` suffix differentiating the static target from the share target.

This PR makes the generated static library the same as with the regular `msbuild` build.


SylvainCorlay:
ping @johanmabille

PR #2322, opened 2017-01-24
Problem: zmq_z85_decode does not validate its input 

### LINKED ISSUE #2321: zmq_z85_decode documentation is ambiguous about error behaviour
The documentation of zmq_z85_decode is ambiguous about its behaviour in case of invalid input.
It states "The encoding shall follow the ZMQ RFC 32 specification." and "The zmq_z85_decode() function shall return dest if successful, else it shall return NULL."

As implemented, it returns NULL if the length of the input is invalid, but not if it contains invalid characters. This should at least be clarified in the documentation.

However, I think that the method should be changed such that it either 
* assumes that the input is completely valid (and have undefined behaviour otherwise), or 
* validates its input more thoroughly. 

I am not sure which of these alternatives better fits in with the overall API design.

sigiesec:
Solution: added checks for invalid characters and overflows

Fixes #2321


bluca:
There are merge conflicts so I can't merge. Could you please rebase your branch?

tests/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ set(tests
         test_xpub_manual
         test_xpub_welcome_msg
         test_timers
+		test_utils_z85
```
bluca: whitespace

tests/test_utils_z85.cpp
```diff
@@ -0,0 +1,111 @@
+/*
+    Copyright (c) 2007-2015 Contributors as noted in the AUTHORS file
```
bluca: Should be just 2017 since it's a new file

bluca:
I've added a couple minor comments since it needs to be rebased anyway, thanks for fixing this up

sigiesec:
I have been struggling for several hours with git now. I somehow seem to have rebased my branch now, but I fear I did something wrong in the very beginning, which I did not manage to repair yet.

bluca:
The build on Windows + CMake appears to be broken:

https://ci.appveyor.com/project/zeromq/libzmq/build/build-1017/job/mo712pt999mh6c66

src/zmq_utils.cpp
```diff
@@ -110,6 +111,18 @@ static uint8_t decoder [96] = {
     0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18,
     0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F, 0x20,
     0x21, 0x22, 0x23, 0x4F, 0x00, 0x50, 0x00, 0x00
```
bluca: missing trailing comma here is breaking the build

src/zmq_utils.cpp
```diff
@@ -110,6 +111,18 @@ static uint8_t decoder [96] = {
...
+    0xFF, 0x44, 0xFF, 0x54, 0x53, 0x52, 0x48, 0xFF,
```
bluca: Are you sure the intention was not to change the previous declaration rather than appending? It's a 96 bytes array with a static initialization of 192 bytes now

evoskuil:
I don't understand why there is a new Visual Studio (2015 only) project added here.

Also, there is an existing set of [base-85 tests](https://github.com/zeromq/libzmq/blob/master/tests/test_base85.cpp).

sigiesec:
@evoskuil I accidentally committed the VS2015 project, and removed it again. I needed to add it locally because there are no VS2015 test projects at all. I merged my tests into the existing test_base85.cpp file. In general, I am not a big fan of goto, but I also noticed that libzmq makes use of it, and I think it is an acceptable tradeoff when doing error handling not based on exceptions.

evoskuil:
@sigiesec Yes, my apologies - I confused the test structure in czmq with that of libzmq. I see why you added it. I did something similar locally when I wrote the original base85 tests.

I devised the VS projects for each, but unlike czmq I did not create test projects for libzmq. libzmq has a main in each test source file, which basically means one project per source file. That's too much to maintain, especially across each version of VS, without automation - which I didn't have time to add.

I don't like this test organization. A few test macros could pull this together into a proper test suite with one main. This would speed up both the build and the test pass for all platforms. IMO the time would be better spent doing that than adding the test project files.

bluca:
Thanks for the fixes! There's still a missing ```#include <stdint.h>``` as it's failing to build with:

```
../src/zmq_utils.cpp: In function 'uint8_t* zmq_z85_decode(uint8_t*, const char*)':
../src/zmq_utils.cpp:163:13: error: 'UINT32_MAX' was not declared in this scope
../src/zmq_utils.cpp:174:41: error: 'UINT32_MAX' was not declared in this scope
```

Could you please fix it?

https://travis-ci.org/zeromq/libzmq/jobs/195123046

bluca:
Uhm it's still complaining:

```
../src/zmq_utils.cpp: In function 'uint8_t* zmq_z85_decode(uint8_t*, const char*)':
../src/zmq_utils.cpp:164:13: error: 'UINT32_MAX' was not declared in this scope
../src/zmq_utils.cpp:175:41: error: 'UINT32_MAX' was not declared in this scope
```

Isn't stdint.h where those are defined?

https://travis-ci.org/zeromq/libzmq/jobs/195139360

bluca:
Great, thanks for persevering and fixing all issues!

PR #2327, opened 2017-01-28
fix divide by zero in zmq::lb_t::sendpipe

chutz:
This is based on https://github.com/zeromq/zeromq3-x/pull/116/files

### DIFF
--- a/src/lb.cpp
+++ b/src/lb.cpp
@@ -137,7 +137,9 @@ int zmq::lb_t::sendpipe (msg_t *msg_, pipe_t **pipe_)
     more = msg_->flags () & msg_t::more? true: false;
     if (!more) {
         pipes [current]->flush ();
-        current = (current + 1) % active;
+
+        if (++current >= active)
+            current = 0;
     }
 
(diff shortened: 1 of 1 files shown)


bluca:
In https://github.com/zeromq/libzmq/pull/982 it was mentioned that it's not possible for active to be 0 in that place - what issue are you seeing? Do you have a test case that can reproduce it?

chutz:
@bluca No we do not have a test case. We were getting occasional random crashes in a python app that uses zeromq. As far as we can tell, there must be some sort of race condition that is hard to reach, since it does not happen very often.

The kernel log was showing this message
```
traps: XXXXXXXXXXX[XXXXXXX] trap divide error ip:66230246c4fa sp:6622bb73d640 error:0 in libzmq.so.5.0.1[662302447000+72000]
```
That maps in the source code to the line we are changing with this patch. Given the similarity to the previous issue, we figured it must be related.

bluca:
Ok, thanks for clarifying. Make sure not to use the same socket from multiple threads, as that's the number one cause of race conditions.

PR #2332, opened 2017-02-03
Problem: v4-in-v6 mapped addresses cause parsing failure

bluca:
Solution: in the address validator accept an endpoint that starts
with :, eg: ::ffff:192.168.0.1:12345 as it's a valid mapped
address.

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -842,7 +842,7 @@ int zmq::socket_base_t::connect (const char *addr_)
         //  Following code is quick and dirty check to catch obvious errors,
         //  without trying to be fully accurate.
         const char *check = address.c_str ();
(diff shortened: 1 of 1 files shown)


bluca:
anyone for a merge?

jimklimov:
Hi, just noticed this. Just in case (and for posterity), does "accept" here mean not crashing on such addresses - not that an IPv6 enabled instance would be able to communicate with an IPv4 instance?

bluca:
Yes, given the connect is asynchronous there is this initial basic sanity check to make sure a user is not passing something that can never work
Then if the other side cannot accept a connection or does not exist or is not reachable that's a separate event that can be checked with the socket monitoring

Issue #2341, opened 2017-02-14
Build VS2010 - LINK : fatal error LNK1104: cannot open file 'libzmq.lib'


dcube9:
Hi,
during build versione 4.2.1 with VS2010 (Both Batch and VStudio) error LNK1104 occours in all "perf" tools.
To fix the problem : modify paths in AdditionalLibraryDirectories and SourceFiles of file **.....builds\msvc\vs2015\libzmq.import.props**

**AdditionalLibraryDirectories**
Original : `$(ProjectDir)..\..\..\..\..\libzmq\bin\$(PlatformName)\Debug\$(PlatformToolset)\$(Linkage-libzmq)\`
Modify in : `$(ProjectDir)..\..\..\..\bin\$(PlatformName)\Debug\$(PlatformToolset)\$(Linkage-libzmq)\`

**SourceFiles in Target Name="Linkage-libzmq-dynamic"**
All : `$(ProjectDir)..\..\..\..\..\libzmq\bin\$(PlatformName)\Debug\$(PlatformToolset)\dynamic\`
Modify in : `$(ProjectDir)..\..\..\..\bin\$(PlatformName)\Release\$(PlatformToolset)\dynamic\`

Regards
P

PS : Why all VS201x use the file libzmq.import.props in VS2015 path ?

bluca:
@bbdb68 @evoskuil any suggestion?

dcube9:
1° Fix file libzmq.import.props in Master
2° Move file *.props from VS2015 in a common path (like generic props in `....\builds\msvc\properties`)

Dont you ?

evoskuil:
Originally the import props were pulled from each build independently. There was a PR a while back that changes that behavior in order to reduce duplication. I have in the past managed the duplication using local scripts that emit the various VS platform versions by making small modifications to one of them that is used as a master. In one of my more recent updates I accommodated the change that centralized the properties files. It's not my preferred approach as I like the VS versions to be independent (once generated).

One option would be to revert that change, otherwise there needs to be a correction to link properly to the master (VS2015) repo from the others. I'm a little confused though because I thought I did that (the latter) already.

bbdb68:
For a newcomer, having both a CMakeLists and solutions is confusing IMHO... why are both maintained ?

evoskuil:
I personally don't know anything about cmake. The VS projects have been around for a long time. I started maintaining them along with a bunch of others and developed the techniques in place in order to make them manageable. and reliable.

[...]

PR #2347, opened 2017-02-21
#2341 fix

zstang:
the root path name specified in this file cause the #2341 problem, the root path name is not need.


bluca:
Thanks! For your next contributions please follow the template set by the PR page for the commit message

dcube9:
Thanks.

PR #2349, opened 2017-02-22
cmake WITH_LIBSODIUM option is broken

boringuy:
- Fixed the variable name in platform.hpp.in

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -133,7 +133,7 @@ zmq::ctx_t::~ctx_t ()
 
     //  If we've done any Curve encryption, we may have a file handle
     //  to /dev/urandom open that needs to be cleaned up.
-#if defined (ZMQ_USE_TWEETNACL)
+#ifdef ZMQ_HAVE_CURVE
     randombytes_close ();
 #endif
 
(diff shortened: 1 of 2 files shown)


src/ctx.cpp
```diff
@@ -133,7 +133,7 @@ zmq::ctx_t::~ctx_t ()
 
     //  If we've done any Curve encryption, we may have a file handle
     //  to /dev/urandom open that needs to be cleaned up.
-#ifdef ZMQ_HAVE_CURVE
+#if defined (ZMQ_USE_TWEETNACL)
```
bluca: are you sure about this? sodium has this function too
https://download.libsodium.org/doc/generating_random_data/

boringuy: Ah... I fixed this before I fix the CMAKE variable then. So, this error was cause by the same problem. I will change this back. Thanks.

boringuy:
Ok. Fixed the pull request, it ends up to be just a one line change in platform.hpp.in

bluca:
Thanks! For your next contributions, if you have fixup commits, please rebase and squash them into the original one.

Also please follow the guidelines set in the PR template for the commit message:

```
Problem: x is broken

Solution: do y and z to fix it
```

Issue #2348, opened 2017-02-22
Problem: In zmq::socket_base_t::send method with or without ZMQ_ROUTER_MANDATORY option set


reza-ebrahimi:
This is the send method in `socket_base_t `class, and there is a bug here, in the case of `ROUTER `socket, we have two processes, one is `ROUTER `process and another is `DEALER `process, code below is the `ROUTER `process:
 
    //  Send a message to an unknown peer with mandatory routing
    //  This will fail
    int mandatory = 1;
    int rc = zmq_setsockopt (router, ZMQ_ROUTER_MANDATORY, &mandatory, sizeof (mandatory));
    assert (rc == 0);

	while (true) {
		//  Get message from dealer to know when connection is ready
		char buffer[255];
		rc = zmq_recv(router, buffer, 255, 0); // for identity
		rc = zmq_recv(router, buffer, 255, 0); // for another part of message

		Sleep(6000); // in this time gap we close DEALER process
  
		//  Send a message to disconnected dealer now
            //  and it will block here in zmq_send method
		rc = zmq_send(router, "X", 1, ZMQ_SNDMORE);
		assert(rc == 1);
		rc = zmq_send(router, "Hello", 5, 0);
		assert(rc == 5);
	}

and there problem is in below send method, due to `ZMQ_ROUTER_MANDATORY` option, it might return `EHOSTUNREACH` or `-1`, but it goes to last while loop to process remained commands and blocks in `select` system call untill another `DEALER `process coming up! 

    int zmq::socket_base_t::send (msg_t *msg_, int flags_)
    {

    scoped_optional_lock_t sync_lock(thread_safe ? &sync : NULL);

    //  Check whether the library haven't been shut down yet.
    if (unlikely (ctx_terminated)) {
        errno = ETERM;
        return -1;
    }

    //  Check whether message passed to the function is valid.
    if (unlikely (!msg_ || !msg_->check ())) {
        errno = EFAULT;
        return -1;
    }

    //  Process pending commands, if any.
    int rc = process_commands (0, true);
    if (unlikely (rc != 0)) {
        return -1;
    }

    //  Clear any user-visible flags that are set on the message.
    msg_->reset_flags (msg_t::more);

    //  At this point we impose the flags on the message.
    if (flags_ & ZMQ_SNDMORE)
        msg_->set_flags (msg_t::more);

    msg_->reset_metadata ();

    //  Try to send the message using method in each socket class
    rc = xsend (msg_);
[...]

PR #2352, opened 2017-02-24
Commit: Issue #2348

reza-ebrahimi:
Commit for Issue #2348
Problem: In zmq::socket_base_t::send method with or without ZMQ_ROUTER_MANDATORY option set

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -212,12 +212,27 @@ int zmq::router_t::xsend (msg_t *msg_)
 
             if (it != outpipes.end ()) {
                 current_out = it->second.pipe;
-                if (!current_out->check_write ()) {
+
+                // Check whether pipe is full or not
+                if (!current_out->check_hwm()) {
                     it->second.active = false;
                     current_out = NULL;
+
                     if (mandatory) {
                         more_out = false;
                         errno = EAGAIN;
+                        return -1;
+                    }
+                }
+                // Check whether pipe is closed or not
+                else 
+                if (!current_out->check_write()) {
+                    it->second.active = false;
+                    current_out = NULL;
+
+                    if (mandatory) {
+                        more_out = false;
+                        errno = EHOSTUNREACH;
                         return -1;
                     }
                 }
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -1120,7 +1120,13 @@ int zmq::socket_base_t::send (msg_t *msg_, int flags_)
     if (rc == 0) {
         return 0;
     }
-    if (unlikely (errno != EAGAIN)) {
+
+    // In case of ZMQ_ROUTER_MANDATORY option set and peer disconnected
+    if (likely(errno == EHOSTUNREACH)) {
+        return -1;
+    }
+
+    if (unlikely(errno != EAGAIN)) {
         return -1;
     }
 
--- a/tests/test_router_mandatory_hwm.cpp
+++ b/tests/test_router_mandatory_hwm.cpp
@@ -84,11 +84,15 @@ int main (void)
     const int BUF_SIZE = 65536;
     char buf[BUF_SIZE];
     memset(buf, 0, BUF_SIZE);
+
     // Send first batch of messages
     for(i = 0; i < 100000; ++i) {
-        if (TRACE_ENABLED) fprintf(stderr, "Sending message %d ...\n", i);
+        if (TRACE_ENABLED) 
+            fprintf(stderr, "Sending message %d ...\n", i);
+
         rc = zmq_send (router, "X", 1, ZMQ_DONTWAIT | ZMQ_SNDMORE);
-        if (rc == -1 && zmq_errno() == EAGAIN) break;
+        if (rc == -1 && zmq_errno() == EHOSTUNREACH)
+            break;
         assert (rc == 1);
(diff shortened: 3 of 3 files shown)


bluca:
This breaks the test_router_mandatory_hwm pretty much everywhere, could you please have a look?

https://travis-ci.org/zeromq/libzmq/builds/204913151

Also please rebase and squash the fixup commits in the first one

reza-ebrahimi:
@bluca I'll check it.

hoditohod:
This breaks MANDATORY router behavior when HWM is reached: socket send should block or signal EAGAIN when ZMQ_DONTWAIT is used.

I think the tests updated in this commit were correct, and they were in sync with documentation.

bluca:
Yes I agree, changing this would be an API breakage, and not sure we want it - EAGAIN is the right error when the pipe is full.
I think a change in check_write would be needed to differentiate between a failure because the pipe is gone or full, as it was mentioned on the mailing list a few days ago.
@somdoron any suggestions?

somdoron:
I think the fix I made for NetMQ didnt change the api and fix the issue. Also I didnt change socket_base. I think we can apply the exact same change here

somdoron:
https://github.com/zeromq/netmq/commit/30095a72ae3a9cfe8695f8010a94ff23be9fab7e

With this change, only if pipe is not active (during termination) host unreachable will be returned instead of eagain. Which doesnt break the api

reza-ebrahimi:
I suggest we should check both pipe full or pipe closing state in the method and return proper error code, this will not break API, only documentation should be updated.


	if (it != outpipes.end()) {
		current_out = it->second.pipe;

		// check wheter pipe is full or not
		if (!current_out->check_hwm()) {
			it->second.active = false;
			current_out = NULL;
			if (mandatory)
				errno = EAGAIN;
		}
		// check wheter pipe is closed or not
		else if (!current_out->check_write()) {
			it->second.active = false;
			current_out = NULL;
			if (mandatory)
				errno = EHOSTUNREACH;
		}

		if (mandatory) {
			more_out = false;
			return -1;
		}
	}
	else if (mandatory) {
		more_out = false;
		errno = EHOSTUNREACH;
		return -1;
	}

somdoron:
Looks good, one comment, not really critical.
Merging this one, can you also port the test from netmq in a new PR?

reza-ebrahimi:
@somdoron Thanks, I'll be create netmq test, but I have a question,

This part of code is check when sending first frame of message as peer identity, so for main part of message this part is not reached and always returns with size of the written message and completely seems to be a legal operation whenever the peer is dead and pipe is not active and maybe destroyed?

bluca:
@reza-ebrahimi @somdoron shouldn't the changes to the test be reverted? They are now failing

reza-ebrahimi:
@bluca which test do you mean?

bluca:
tests/test_router_mandatory_hwm.cpp

bluca:
it's fine, I can send a PR to revert the changes to the test

reza-ebrahimi:
@bluca Yes, I've checked test, it should reverted back to EAGAIN (retrying) state.

reza-ebrahimi:
I'm just ask my question again,

This part of code is check when sending first frame of message as peer identity, so for second part of message, this part is not reached and always returns with size of written message and completely seems to be a legal operation whenever the peer is dead and pipe is not active and maybe destroyed?

First Frame --> send() // returns -1 , Wow! seems to be peer is dead (with ZMQ_ROUTER_MANDATORY)!
Second Frame --> send() // returns exactly message size! where message is sent??!

I think this check might be applied for second part of message too and both calls should return -1.

bluca:
https://github.com/zeromq/libzmq/pull/2353

somdoron:
@reza-ebrahimi I think there is issue with the fix. Because now we might not call check_write (actually most of the time), important piece of code which mark the pipe as in active is not going to run. Actually with this fix, the pipe will never overcome the full state and will block forever. I think it even be easy to produce this issue.

I suggest to first call check_write, and inside to if block to call check_hwm.

somdoron:
the process_active_write which suppose to make the pipe writeable again will not do so:

https://github.com/zeromq/libzmq/blob/master/src/pipe.cpp#L266

reza-ebrahimi:
@somdoron I'm faced with that too, so I think this a wrong code inside check_write, so by its name check_write is a method that checks pipe is writable or not, so this is a bad code to set any state inside that,

```
bool zmq::pipe_t::check_write ()
{
    if (unlikely (!out_active || state != active))
        return false;

    bool full = !check_hwm();

    if (unlikely (full)) {
        out_active = false;
        return false;
    }

    return true;
}
```

I'm understand that piece of code inside ` if (unlikely (full)) {` was not calling moste of the time, I'm thinking about the solution.

reza-ebrahimi:
@somdoron I have fixed the problem with a new method in pipe_t class, can I commit from this PR or create another one?

somdoron:
New one...

somdoron:
New one...

reza-ebrahimi:
@somdoron PR issued. #2355

reza-ebrahimi:
I think line 219 in write method is redundant, pipe_t class exposed checking methods public, so user should be applied checks before calling write method, in many places it called twice.

PR #2353, opened 2017-02-24
Problem: unnecessary change to router_mandatory_hwm test

bluca:
Solution:

Revert "Problem with ZMQ_ROUTER_MANDATORY option when peer lost"

This partially reverts commit 06740b5d8c1fbf13205c9b6c63ba7e7ef4350a04.

The change in the mentioned commit was fixed to be backward
compatible so test changes are no longer necessary.
Partially revert the commit, leaving only the bug fix and taking out
the test change and the superfluous change in socket_base.


bluca:
@somdoron applied the change to socketbase you suggested as well

PR #2354, opened 2017-02-25
Problem: router pipe can block forever

bluca:
Solution: invert check_write and check_hwm calls in router xsend
function so that check_write can mark the pipe as inactive if
necessary.

@somdoron is this what you had in mind? Tried adding a test case for this but couldn't quite make it fail without the fix


somdoron:
Yes...

Issue #2319, opened 2017-01-23
Problem: There is no API or mechanism to manage ROUTER socket connected clients


reza-ebrahimi:
There is 10 Clients (DEALER socket) that connected to a ROUTER socket, if all of clients send message to ROUTER socket and in ROUTER side for sending reply response to exactly such DEALER it might know about its handle, to do that there is no API to manage DEALER handles in ROUTER side, so manager should do some tasks like adding, removing or query a handle.

bluca:
Isn't the identity used for that purpose?

reza-ebrahimi:
@bluca  Yes, of course! but identity doesn't enough in dealing with some situations, for example, one problem happened when a DEALER socket lost (disconnected) its connection, so that socket identity still be in ROUTER socket side and exists their identity state in application code, if you send any data to disconnected DEALER socket it will face with access violation error. so in cause of the error, application code doesn't know about disconnected socket state or doesn't notified about that to consider another tasks such as removing socket identity handle.

I think this is a ZeroMQ kernel task to provide a identity manager in ROUTER socket side and do proper tasks (adding, removing, querying), by now there is no notifying mechanism from ROUTER socket about disconnection of ROUTER clients.

reza-ebrahimi:
@bluca have you any idea?

bluca:
Isn't ROUTER_MANDATORY https://github.com/zeromq/libzmq/blob/master/doc/zmq_setsockopt.txt#L705 a solution for that already? It will return an error in such cases

mrvn:
The socket monitor will send a disconnect event and the application can remove the DEALER. Isn't that all you need?

reza-ebrahimi:
Yes, That is! I need to know which peer disconnected, to do this zeromq might provide me at least peer's identity and another useful information such as socket lost reason:

Disconnection event structure:
  - socket identity
  - disconnection reason

bluca:
PRs are most welcome :)

reza-ebrahimi:
If we want to modify zmq monitoring system, we should use ZMQ_PAIR socket, 
documentation says:
> A socket of type ZMQ_PAIR can only be connected to a single peer at any one time. No message routing or filtering is performed on messages sent over a ZMQ_PAIR socket.

in this case we should create and connect multiple instance of ZMQ_PAIR socket based on how many peers connected to ROUTER socket, I think this isn't efficient, any idea are welcome!

nishsharma001:
Guys, 
[...]

PR #2355, opened 2017-02-25
Change out pipe state processing method

reza-ebrahimi:


somdoron:
I think @bluca already solved that...

somdoron:
Also, changing check write is problematic. The method is being used else where without calling check_hwm

Luca fix is simpler and does change pipe, which is the core of libzmq, I'm afraid to change it

bluca:
I agree with Doron, thanks for the PR but I think it's better to avoid major refactors of these classes unless it's absolutely necessary, there's a lot of risk and not much reward

reza-ebrahimi:
this is `check_hwm` body:

`bool full = hwm > 0 && msgs_written - peers_msgs_read >= uint64_t (hwm);`

and I placed following expression where above expression variables are changing, so calling  `check_hwm` multiple times will have same previous operation, is not?

and for reward, I think many places in zeromq have bad code and if you want to ignore them it will face zeromq with further development problems.

reza-ebrahimi:
most of the zeromq class doesn't exposed to the outside world as an API and consider changes in code as API brokerage is not good idea, by the life time of zeromq ROUTER socket considered as one connected client and there is no API for resolving this problem #2319.

somdoron:
@reza-ebrahimi changing behavior which can crash/effect users applications is API breakage.

The problem of changing the pipe class is that is used all over the code, from all socket types and also from the background IO thread. We cannot know what the full effect of changing the pipe class and should avoid if possible.

If there is a simpler solution which doesn't change the pipe class, we need to prefer it.

somdoron:
@bluca solution is simpler and solve the problem, take a look here:
https://github.com/zeromq/libzmq/commit/a73d5cd6c683733eada457cfe44d4e2959da65d0

bluca:
One of the issues of these classes being internal-only is that there's no direct unit (or module?) testing for them - so refactoring is very dangerous as we don't have any direct and immediate verification for correctness. This is why we need to be careful, and when there's a bug like in this case, trying to apply the smallest change possible is the safest course of action.

I believe it would be possible, however, to add module testing. Perhaps a selftest method inside the class itself, and a new small test binary built by automake without linking to the full libzmq shared library.
@reza-ebrahimi we are very grateful for your help on these issues, and if you would like to refactor these internal classes as you said, I would suggest starting by adding this module selftest functionality. As you I'm sure you know it's widely accepted as an engineering good practive to have comprehensive unit testing in place before engaging in refactoring.

PR #2360, opened 2017-03-03
Problem: Assertion failed in zmq::signaler_t::send

nexcvon:
Solution: Change to the way it does below on non-Windows platforms, retry when send returns -1.

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -187,9 +187,13 @@ void zmq::signaler_t::send ()
     errno_assert (sz == sizeof (inc));
 #elif defined ZMQ_HAVE_WINDOWS
     unsigned char dummy = 0;
-    int nbytes = ::send (w, (char *) &dummy, sizeof (dummy), 0);
-    wsa_assert (nbytes != SOCKET_ERROR);
-    zmq_assert (nbytes == sizeof (dummy));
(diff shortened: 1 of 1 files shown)


bluca:
Thank you!

hartcw:
I see a similar abort from zmq_connect() on windows, but the callstack shows it happens in zmq::signaler_t::recv().

I suspect a similar fix is needed there, to handle the case where ::recv() returns an unexpected value.

bluca:
Could you please test a fix and send a PR, if it solves your problem?

bjovke:
@nexcvcon Are you sure that it's good to potentially loop forever if `send ()` returns `SOCKET_ERROR`?
It is now the same behavior as before #2362 so it kind of reverts that PR.
Why should there be any asserts for failed `send ()` if there's a possibility for it to fail?
And what are the guarantees that it will succeed if retried?
On Linux in the same function there's this code:
```
errno = EINTR;
break;
```
So it is possible to just set `errno `and return in case of `send ()` error?
Furthermore, `zmq_assert (nbytes == sizeof (dummy)); `might also be the cause of troubles on Windows.
I'm not 100% familiar with libzmq internals, but if code looks strange on first sight there's a high potential that it actually is not good.

Issue #2358, opened 2017-03-01
zmq connect aborts intermittently


rabrahamkns:
Using libzmq 4.1.5, built with VS2010, win32, debug.
To reproduce:
My code creates a zmq socket and calls connect(). No Server is running.
The host address passed to connect is something like: "tcp://localhost:5111"
It aborts intermittently.
zmq::signaler_t::send() returns -1.

Call stack:
```

                _RaiseException@16() + 0x52 bytes        
               zmq::zmq_abort(const char * errmsg_=0x050c030c) Line 80 + 0x15 bytes              C++
>             zmq::signaler_t::send() Line 182 + 0x49 bytes     C++
               zmq::mailbox_t::send(const zmq::command_t & cmd_={...}) Line 66       C++
               zmq::ctx_t::send_command(unsigned int tid_=3, const zmq::command_t & command_={...}) Line 372  C++
               zmq::object_t::send_command(zmq::command_t & cmd_={...}) Line 436            C++
               zmq::object_t::send_own(zmq::own_t * destination_=0x1ceedf88, zmq::own_t * object_=0x3f8a7cc0) Line 222         C++
               zmq::own_t::launch_child(zmq::own_t * object_=0x3f8a7cc0) Line 90   C++
               zmq::socket_base_t::add_endpoint(const char * addr_=0x3f534430, zmq::own_t * endpoint_=0x3f8a7cc0, zmq::pipe_t * pipe=0x408e6558) Line 722            C++
               zmq::socket_base_t::connect(const char * addr_=0x3f534430) Line 715 C++
               zmq_connect(void * s_=0x1ceedf88, const char * addr_=0x3f534430) Line 316 + 0xc bytes           C++
               zmq::socket_t::connect(const char * addr_=0x3f534430) Line 581 + 0x12 bytes  C++
               ZmqClient::connectSession(IDvzTransportSessionCallback * cb=0x00000000) Line 76       C++
               CGSUCDatavizHelper::_StartDatavizTransport(IDvzTransport * & p_rTransportType=0x41702f00, wchar_t * p_sSessionId=0x0186e5b0) Line 320 + 0x17 bytes             C++
               CGSUCDatavizHelper::StartAllDatavizServices() Line 240 + 0x11 bytes      C++
               CMFEXMainApplication::_InitInstance(void * p_pLargeMemoryBlock=0x41d70000) Line 283 + 0x8 bytes                C++
[...]

Issue #2339, opened 2017-02-14
assertion failure in signaler_t.cpp


vkhalidov:
Hello,
I'm using 0MQ (4.2.0) with wine (wine-1.8) on Ubuntu 14.04.5 to perform socket communication. The communication is performed between a DLL module (that is shut down and relaunched regularly) that is loaded by an executable that runs under wine and another process that runs directly on Ubuntu. Thus the sockets used in the communication are regularly closed and reopened. Everything works as expected 99% of time. However, sometimes the DLL produces assertion failure when shutting down the socket and blocks, requiring manual restart:
Assertion failed: Bad file number (........\src\signaler.cpp:185)

After having investigated a bit, I found out that there is an io_thread inside 0MQ that tries to send a command after the socket has been destroyed, which causes the assertion failure (the winsock send method returns an error code when trying to send anything to a closed socket, which is checked against in the assert statement).

I observe the following sequence of calls before the problem occurs:

The main thread (DLL module) creates a signaler that calls unblock_socket on the system write socket W
Soon an io_thread is launched in a system thread A (that performs zmq::select_t::loop), it calls signaler_t::send regularly, using this system socket handler W.
After some time a reaper io_thread is created in a system thread B (also performs zmq::select_t::loop) that calls zmq::socket_base_t::start_reaping and destroys the socket (zmq::socket_base_t::check_destroy, signaler_t::~signaler_t), which closes the system write socket W
All of the sudden, after the socket is already destroyed, the io_thread in a system thread A calls signaler_t::send with the system write socket W that has already been closed, which produces the assertion.
It looks like the io_thread that used system thread A is somehow not aware of the fact that the socket has already been destroyed. This is reproduced only with a particular sequence of calls, so sometimes it can run for 2 days without any problem (having hundreds or socket open/close operations succeeded) and all of the sudden encounter the assertion. I did not go deep into the code to analyze how io_threads implement safety measures in order to prevent sending anything to closed sockets. But any insight on where to look and which invariant to check would be of great help!

Best regards,
Vasil

vkhalidov:
A particular situation when this assertion happens:

[...]

PR #2362, opened 2017-03-06
Problem: zmq::signaler_t::send may loop forever

nexcvon:
Solution: restore the wsa_assert statement which was previously removed.

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -189,6 +189,7 @@ void zmq::signaler_t::send ()
     unsigned char dummy = 0;
     while (true) {
         int nbytes = ::send (w, (char*) &dummy, sizeof (dummy), 0);
+        wsa_assert (nbytes != SOCKET_ERROR);
         if (unlikely (nbytes == SOCKET_ERROR))
             continue;
         zmq_assert (nbytes == sizeof (dummy));


src/signaler.cpp
```diff
@@ -189,6 +189,7 @@ void zmq::signaler_t::send ()
     unsigned char dummy = 0;
     while (true) {
         int nbytes = ::send (w, (char*) &dummy, sizeof (dummy), 0);
+        wsa_assert (nbytes != SOCKET_ERROR);
         if (unlikely (nbytes == SOCKET_ERROR))
```
bluca: given the assert above, can nbytes be ever == SOCKET_ERROR?

nexcvon: I am not so sure it won't be. I am watching this in my project, things go pretty well so far. Without this patch, `zmq_assert` below fails very often in my program.

bluca: seems strange, but ok

nexcvon:
#2339 is on Ubuntu but with wine.

The assertion failed **is** zmq_assert. But in this problem (at least mine for sure), nbytes is -1, not another value less than sizeof(dummy).

The statement `wsa_assert (nbytes != SOCKET_ERROR);` doesn't work exactly as it looks. It will check the return value from WSAGetLastError(), and won't fail if it gets 0 or EWOULDBLOCK.

a4z:
Thanks for the info about wine @nexcvon , this maks at least this clear.
So all 3 Problems are on Windows, one via log as stated in the zmq_assert, the other as reported maybe in the wsa_assert.

All 3 reported indicate that send had a problem. Is ignoring the problem the right solution?

and finally, what does the current code solve?
```
    while (true) {
        int nbytes = ::send (w, (char*) &dummy, sizeof (dummy), 0);
        wsa_assert (nbytes != SOCKET_ERROR);
        if (unlikely (nbytes == SOCKET_ERROR))
            continue;
        zmq_assert (nbytes == sizeof (dummy));
        break;
    }
```
it does not solve #2339
and I am unsure if it solves any problem at all,  except that the code still looks like it can endless loop.

`wsa_assert` does  not work as you say, it checks if there is some error string, what is very likely to be the case.
https://github.com/zeromq/libzmq/blob/master/src/err.hpp#L74

I really suggest to revert the code to the state as it has been before #2361 , 
* rephrase the problem that lead to the assertion failure
*  try to reproduce it
* if possible, create a test case for it
*  come than with a solution that has no complex loop where code reader need to reason about what this does for uncertain amount of time, just to find an potential endless loop. 

Please. Thanks!

nexcvon:
I can reproduce this problem using pyzmq 16.0.2 (with zmq 4.1.6). The error code is EWOULDBLOCK (I was wrong about it may be 0), so wsa_assert doesn't fail (see https://github.com/zeromq/libzmq/blob/master/src/err.cpp#L95 and https://github.com/zeromq/libzmq/blob/master/src/err.cpp#L119).

I don't known why EWOULDBLOCK is returned, and how it should be handled correctly. Retrying seems to be better than crashing for now, before someone reports a infinite loop.

I am not familiar with sockets. I hope @a4z you can help improve this.

Following is how I reproduce this problem using Python. I will try to get a c++ version working.

Modified zmq::signaler_t::send:

```
#elif defined ZMQ_HAVE_WINDOWS
    unsigned char dummy = 0;
    while (true) {
        int nbytes = ::send (w, (char*) &dummy, sizeof (dummy), 0);
        if (nbytes == SOCKET_ERROR)
            printf("WSAGetLastError() returns %d\n", WSAGetLastError());
        wsa_assert (nbytes != SOCKET_ERROR);
        if (unlikely (nbytes == SOCKET_ERROR))
            continue;
        zmq_assert (nbytes == sizeof (dummy));
        break;
    }
#else
```

Python code:

```
import time
import zmq

url = "tcp://127.0.0.1:22222"
context = zmq.Context.instance()

for i in range(1000):
    print(i)

    router = context.socket(zmq.ROUTER)
    router.bind(url)

    dealer = context.socket(zmq.DEALER)
    dealer.connect(url)

    dealer.send(b"hello")
    msgs = router.recv_multipart()
    assert msgs[1] == b"hello"

    dealer.close()
    router.close()
    time.sleep(0.1)
```

Output:

```
0
1
2
3
4
<...>
201
202
203
204
WSAGetLastError() returns 10035
205
206
207
208
<...>
446
447
448
449
WSAGetLastError() returns 10035
450
WSAGetLastError() returns 10035
451
452
453
454
<...>
576
577
578
579
WSAGetLastError() returns 10035
580
WSAGetLastError() returns 10035
581
582
583
584
<...>
560
661
662
663
WSAGetLastError() returns 10035
664
665
666
667
<...>
944
945
946
947
WSAGetLastError() returns 10035
948
WSAGetLastError() returns 10035
949
950
951
WSAGetLastError() returns 10035
952
WSAGetLastError() returns 10035
953
WSAGetLastError() returns 10035
954
955
956
957
<...>
995
996
997
998
999
```

a4z:
I am at work with less time unfortunately today, and no windows around me
but I have no idea why your socket might return once every x times it feels bad, but that's leagal.

@nexcvon , could you please test if this solve it?
 It's untested / not compiled code, so maybe adoption required
The key is, it is OK to retry, but please not for ever :-)    

````
    unsigned char dummy = 0;  
    
    const int max_retry = 3 ;
    int nbytes  = 0 ;
    for (int i = 0; i < max_retry ; ++i) {
    
        nbytes = ::send (w, (char *) &dummy, sizeof (dummy), 0);
    
        if (nbytes == SOCKET_ERROR);
            continue ;      
      
        if(nbytes == sizeof (dummy))
            break      
    }
    
    wsa_assert (nbytes != SOCKET_ERROR);
    zmq_assert (nbytes == sizeof (dummy));
````


ps: looking at the code, the Linux version could also need some adoption,
since it it theoretical possible to not send all 8 bytes at once, (MS sends just 1 byte)
in this case we should retry again with the remaining bytes.
but it seems to be super unlikely that this ever happens.
would you like a issue for this @bluca ?
I will not make one as long it does not crash on my machine or I'll fix it someone in the future on the fly

bluca:
This is the signaler, so remember that on Linux it's not a TCP socket like on Windows. I honestly couldn't think of any reason why it should succeed to send, but less than 8 bytes.

But if you want to send a PR to check for that special case please feel free to do so, as long as the overall functionality when there's an actual error is not changed.

a4z:
yes, as unlikely that this ever happen as that 'll touch the linux code now ;-)

rabrahamkns:
update on #2358 - I changed our code ( we use 4.1.5) only for signaler_t::send() to the code in this commit, and we do not see the abort anymore. Thanks.

PR #2365, opened 2017-03-11
problem:zmq_poll is slow and the API is complicated

somdoron:
solution: deprecate zmq_poll in favor of zmq_poller

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -436,7 +436,7 @@ ZMQ_EXPORT int zmq_socket_monitor (void *s, const char *addr, int events);
 
 
 /******************************************************************************/
-/*  I/O multiplexing.                                                         */
+/*  Deprecated I/O multiplexing. Prefer using zmq_poller API                  */
 /******************************************************************************/
 
 #define ZMQ_POLLIN 1


bluca:
I still see the zmq_poller unit test fail 20%-30% of the times on OBS unfortunately, did you see it fails when you run it?

bluca:
With regards to deprecation, we could print warnings from the build system (maybe with a new flag?) when deprecated APIs are used, and after a couple of years (maybe?) retire them.
When I have time I'll start experimenting with zproject

bluca:
Case in point, after this last push it fails on Suse server x86_64:

https://build.opensuse.org/package/live_build_log/network:messaging:zeromq:git-draft/libzmq/SLE_12_SP2/x86_64

```
[  216s] ./config/test-driver: line 95: 20184 Aborted                 "$@" > $log_file 2>&1
[  216s] FAIL: tests/test_poller
```

somdoron:
I will check it. I guess it is the previous PR that vause the issue, as
this is only a comment.

somdoron:
Can you tell when it fails? error or line?

bluca:
Don't think it's either of them, I've seen the same thing happen for the past few months, so it must be an older issue

Unfortunately there's no way to access the ephemeral build workers, so no way to get the logs and tell which line it is, but given it's failing on x86 it should be reproducible

somdoron:
So it is not related to any of the PRs?

I will try to reproduce in my machine (64bit arch linux).

bluca:
Yes I'm 100% sure it's not related to these last PRs. Thanks!

bluca:
This run on Travis failed as well: https://travis-ci.org/zeromq/libzmq/jobs/210223727
It's more interesting as it says which assertion failed, so it might be helpful:

```
FAIL: tests/test_poller
=======================
lt-test_poller: ../tests/test_poller.cpp:43: int main(): Assertion `rc == 0' failed.
```

c-rack:
Link to source: https://github.com/zeromq/libzmq/blob/master/tests/test_poller.cpp#L42-L43

So this port might be already in use. We could try to either use another port or set `SO_REUSEADDR` before binding. Thoughts?

somdoron:
I think we should just change it. It happens  a lot with testing and tcp.

bluca:
Ah yeah didn't even check what it was. Changing sounds good.

The Ubuntu folks have a patch to make all ports dynamic:
http://archive.ubuntu.com/ubuntu/pool/main/z/zeromq3/zeromq3_4.2.1-2ubuntu1.debian.tar.xz
(debian/patches/no-hardcoded-port-in-tests.patch)
It's on my list to try and adapt it (it's a bit sledge-hammery) while leaving at least one test with an hardcoded port, to make sure that codepath is also exercised.

PR #2367, opened 2017-03-12
Problem: ZMQ background threads are unnamed

bluca:
Solution: use pthread API to set the name. For now call every thread
"ZMQ b/g thread". Would be nice to number the I/O threads and name
explicitly the reaper thread, but in reality a bit of internal API
churn would be necessary, so perhaps it's not worth it.
This is useful when debugging a process with many threads.


c-rack:
Would propose to rename it from "ZMQ b/g thread" to "ZMQ background" to avoid the "b/g" abbreviation. If you display thread names, the "thread" part is implied.

bluca:
Sounds reasonable: https://github.com/zeromq/libzmq/pull/2368

PR #2368, opened 2017-03-12
Problem: refactor gone wrong, redundant thread name

bluca:
Solution: fix silly copypasta, rename threads to "ZMQ background"


c-rack:
👍

PR #2375, opened 2017-03-19
Fixed lib and dll paths in visual studio build

mlodew:
As indicated in issue  #2341, the paths to the lib files were broken in libzmq.import.props. Fix #2347 removed the project folder (libzmq) part of the path, to fix this, however the fix leads to problems when building czmq on windows using the build.bat files in the visual studio project folders. The bat files copy the file libzmq.import.props from libzmq to the respective folder in the czmq build directories. Changing the path
`$(ProjectDir)..\..\..\..\..\libzmq\bin\`
to
`$(ProjectDir)..\..\..\..\bin\`
will prevent the czmq build from finding libzmq.lib.
This issue (#2341) should not arise, when the project folder is named 'libzmq'. 

To fix this, the paths were changed such that the libzmq folder is included and a comment was added to the readme.txt.

### DIFF
--- a/builds/msvc/readme.txt
+++ b/builds/msvc/readme.txt
@@ -3,7 +3,7 @@ For building on Windows, use:
      cd build
      ./build.bat
 
-This requires that the CMD.EXE be created using the DevStudio Tools link to create a CMD.EXE window.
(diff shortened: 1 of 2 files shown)


bluca:
Thank you!

For your next contributions, please follow the guidelines and use this format for the commit message:

```
Problem: X is broken

Solution: do Y and Z to fix X
```

PR #2377, opened 2017-03-19
Relicense permission by Paul Colomiets

tailhook:
As asked in the email, I'm granting permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL.

I think it's going to be hard (hopefully not so), but it's a good move. Thanks!


bluca:
Thank you!

PR #2387, opened 2017-03-19
Relisenceing & modifying build so msys is allowed

chuggafan:
MSYS is now allowed if you use configure.ac, also I changed buildall/buildbase to use correct version names for VC++, also I did the re-licensing stuff, only issue with MSYS building is on my machine the headers are off and so they don't immediately compile, but I am unsure if this is an issue with only MSYS2 and will test this more and see if there is a proper solution


bluca:
Thanks!

PR #2423, opened 2017-03-21
RELICENSE: Adam Seering

aseering:


aseering:
Here's a re-license file, as requested.  Please let me know if you would like anything else from me.

I have only contributed one commit.  And the commit was a very small and obvious change.  I'm not sure that I have any legal rights over that sort of change in my jurisdiction, and if I do, I don't want to exercise those rights in a way that would constrain the zeromq project.

c-rack:
Thanks!

PR #2436, opened 2017-03-21
problem: ZMQ_POLLIN & OUT marked depracted but still in use

somdoron:
solution: move them to the zmq_poller block

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -439,11 +439,6 @@ ZMQ_EXPORT int zmq_socket_monitor (void *s, const char *addr, int events);
 /*  Deprecated I/O multiplexing. Prefer using zmq_poller API                  */
 /******************************************************************************/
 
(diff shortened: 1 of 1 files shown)


bluca:
Moving them in the DRAFT section breaks a few things:

```
../src/proxy.cpp: In function ‘int zmq::proxy(zmq::socket_base_t*, zmq::socket_base_t*, zmq::socket_base_t*, zmq::socket_base_t*)’:
../src/proxy.cpp:125:25: error: ‘ZMQ_POLLIN’ was not declared in this scope
         { frontend_, 0, ZMQ_POLLIN, 0 },
```

somdoron:
sorry, that not what I meant...

Just don't want them in  the zmq_poll area, which is marked as obsolete. But still in use by zmq_poller.
Anyway, closing this, we can solve this when we will remove zmq_poll

Issue #2452, opened 2017-03-23
Problem: __STDC_LIMIT_MACROS define precedes PCH include.


evoskuil:
Visual Studio 2013 warning:
```
.\..\..\..\src\zmq_utils.cpp(30): warning C4603: '__STDC_LIMIT_MACROS' : macro is not defined or definition is different after precompiled header use
          Add macro to precompiled header instead of defining here
          ..\..\..\..\src\zmq_utils.cpp(32) : use of precompiled header
```

It's not clear to me whether this inclusion should be strictly limited to `zmq_utils.cpp` and as such I'm hesitant to move it to the precompiled header (PCH). But the PCH should be the first source in the file. Moving this define after the PCH may have unintended consequences as well, since it affects the behavior of includes of the PCH.

bluca:
https://github.com/zeromq/libzmq/commit/c6c21cf1971ba6e0be2fba9ea7637b0b80403d70
> * Define __STDC_LIMIT_MACROS before including stdint.h

Given the commit message, and if I remember correctly, that fixed an issue with stdint.h, so it might be safe to move after the PCH. Want to give it a try?

evoskuil:
The PCH includes `<stdint.h>`, as well as other STD C files, for Windows (I'd really like to remove the Windows condition in the PCH. So the Windows builds would be affected. But it compiles on VS2013 at least.

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/2454

PR #2454, opened 2017-03-23
Problem: __STDC_LIMIT_MACROS before PCH causes VC++ warning.

evoskuil:
Solution: move it after the PCH include. See [discussion](https://github.com/zeromq/libzmq/issues/2452).

### DIFF
--- a/src/precompiled.hpp
+++ b/src/precompiled.hpp
@@ -32,6 +32,8 @@
 
 #include "platform.hpp"
 
+#define __STDC_LIMIT_MACROS 
+
 // This must be included before any windows headers are compiled.
 #if defined ZMQ_HAVE_WINDOWS
 #include "windows.hpp"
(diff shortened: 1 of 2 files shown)


bluca:
```
  CXX    src/src_libzmq_la-zmq_utils.lo
../src/zmq_utils.cpp: In function 'uint8_t* zmq_z85_decode(uint8_t*, const char*)':
../src/zmq_utils.cpp:166:13: error: 'UINT32_MAX' was not declared in this scope
../src/zmq_utils.cpp:177:43: error: 'UINT32_MAX' was not declared in this scope
```

evoskuil:
This means that the PCH is in fact causing the inclusion of `<stdint.h>` - which appears to arise (at least) from the PCH inclusion of `zmq.h`. So the proper resolution would be to define `__STDC_LIMIT_MACROS` in PCH. I would think it would be acceptable (if not desirable) to define the symbol universally given that it impacts the behavior of common includes, which should be consistent.

bluca:
Look good, thanks!

evoskuil:
NP. Is [this](https://travis-ci.org/zeromq/libzmq/jobs/214416605) unrelated?

bluca:
Yes it's unrelated, it's a flacky one

PR #2464, opened 2017-03-25
Added relicense permit

ulikoehler:
I have modified only tw small portion of the template: First, *All* commits by my account are of my own (personal ) copyright. I just want to make it clear that it's not just *a portion* of all commits.

Second, I've included CZMQ in the permission as well as libzmq, as most of my contributions are for CZMQ.

I hope that's OK. If not, I can modify it to your specification.


bluca:
Thanks!

PR #2466, opened 2017-03-26
Problem: insufficient use of const complicates calling.

evoskuil:
Solution: make methods that don't mutate members const.

There are probably more, these were just the obvious ones.


bluca:
I'm a bit concerned by this ABI change, it might not be backward compatible. What's the concrete advantage of this?

evoskuil:
It should always be backward compatible to pass a non-const pointer to a const pointer parameter. The value is that when passing a const parameter as non-const one must either:

1. cast away the const (unsafe)
2. make the parameter mutable (unsafe)
3. make the method which makes the call non-const

The third option is the only safe option. To do the others requires making assumptions that directly contradict the explicit statement by the code that it is unsafe (i.e. the lack of a const qualifier). This choice complicates calling. since the non-const method then infects the entire code base of the caller, forcing const calls to become non-const. This then requires const members to become non-const. This then requires the non-const members to require critical sections in in thread safe classes and causes the API of a higher level library to create the same exact problem for its users, eventually infecting all interfaces of all dependent projects. As a result people generally settle at some point for unsafe code, va option 1 or 2. I decided it would be better to just fix the root cause.

PR #2468, opened 2017-03-26
RELICENSE: Nick Guiffrida grant

goodfella:
This is the relicensing grant that was requested.  Thanks again for the great library.


bluca:
thanks!

goodfella:
You're welcome

PR #2471, opened 2017-03-27
Improve oom handling

t-b:
I had some issues with my application built on top of libzmq in cases of on an out of memory condition. And therefore I also took a look into libzmq itself.

I've fixed a couple of places which used the throwing new version and also added a couple of forgotten asserts. Compiles and all tests pass on debian jessie x64 and x86/x64 Windows 7 with Visual Studio 2015.


t-b:
@bluca Thanks!

Issue #2002, opened 2016-05-19
Problem: curve_server steps on stdout.


evoskuil:
There are about 20 statements in `curve_server_t` that write directly to `stdout`, [such as](https://github.com/zeromq/libzmq/blob/master/src/curve_server.cpp#L105):

``` c
            //  Temporary support for security debugging
            puts ("CURVE I: invalid handshake command");
```

It's marked as temporary. Is there a plan afoot for an implementation without this side effect?

bluca:
Yes we should take these away, especially now that there are socket monitor events.

@vtellier do you think any of those "puts" calls in curve_server.cpp could instead use event monitoring? Perhaps define error numbers and return them as the value of the event

vtellier:
Yes, this really makes sense.

Actually I though about that while working on #2227 but I didn't implemented it because `curve_server` is used by `stream_engine` through the interface `mechanism_t` and the events are signaled by `stream_engine`, regardless the kind of mechanism currently in use.
This makes sense: the first level of information requested in #2227 (success / failure) is defined by `stream_engine`.

A solution for this issue is to extend a bit `mechanism_t` so its implementations have the possibility to give more details.

vtellier:
@bluca Here is one approach to solve this, see the comment of 7c3093dbe7dbf5babfcabad2a725bd9b6b658d95 for details.
The implementation is not completed, this is just a draft.

bluca:
Sounds like a good starting point!
Maybe "undefined" rather than "no_detail"

vtellier:
FYI I don't know if I'll have time to continue on this issue, if anyone wants to go ahead just feel free.

evoskuil:
I got [burned](https://github.com/libbitcoin/libbitcoin-protocol/issues/123) by this issue again. I shipped the 4.2.20 nuget package with these console writes enabled. That's actually a big deal, so I'll need to redeploy. I'm not familiar with the monitoring code, but I would really like to get these console writes out of the lib.

sigiesec:
I started work on this again in my branch https://github.com/sigiesec/libzmq/tree/curve-events

bluca:
Great, thanks!

There was already part of the work done here: https://github.com/zeromq/libzmq/pull/2473

sigiesec:
@bluca, I am aware of that :) I started where @vtellier left.

sigiesec:
[...]

PR #2473, opened 2017-03-28
[WIP DO NOT MERGE] Added handshake error details in monitoring event value (#2002)

vtellier:
This PR aims to fix the issue #2002 by removing the prints in the standart output _(not committed yet)_.
The error details are provided as the value of the event `ZMQ_EVENT_HANDSHAKE_FAILED`.

**Work in progress**
* Need feedback and advices about the event's details. currently the value is and a raw integer with no macro defined in the public API.
* Need to check the CI is not broken
* Need to adapt the test cases

### DIFF
--- a/tests/test_security_curve.cpp
+++ b/tests/test_security_curve.cpp
@@ -208,10 +208,8 @@ int main (void)
     assert (rc == 0);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    int event_value = -1;
-    int event = get_monitor_event (server_mon, &event_value, NULL);
+    int event = get_monitor_event (server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_SUCCEED);
-    assert (event_value == 0);
 #endif
 
     //  Check CURVE security with a garbage server key
@@ -231,10 +229,8 @@ int main (void)
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    event_value = -1;
-    event = get_monitor_event (server_mon, &event_value, NULL);
+    event = get_monitor_event (server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
-    assert (event_value == 2);
 #endif
 
     //  Check CURVE security with a garbage client public key
@@ -253,10 +249,8 @@ int main (void)
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    event_value = -1;
-    event = get_monitor_event (server_mon, &event_value, NULL);
+    event = get_monitor_event (server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
-    assert (event_value == 2);
 #endif
 
     //  Check CURVE security with a garbage client secret key
@@ -275,10 +269,8 @@ int main (void)
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    event_value = -1;
-    event = get_monitor_event (server_mon, &event_value, NULL);
+    event = get_monitor_event (server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
-    assert (event_value == 2);
 #endif
 
     //  Check CURVE security with bogus client credentials
@@ -301,10 +293,8 @@ int main (void)
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    event_value = -1;
-    event = get_monitor_event(server_mon, &event_value, NULL);
+    event = get_monitor_event(server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
-    assert (event_value == 2);
 #endif
 
     //  Check CURVE security with NULL client credentials
@@ -317,10 +307,8 @@ int main (void)
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    event_value = -1;
-    event = get_monitor_event(server_mon, &event_value, NULL);
+    event = get_monitor_event(server_mon, NULL, NULL);
     assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
-    assert (event_value == 2);
 #endif
 
     //  Check CURVE security with PLAIN client credentials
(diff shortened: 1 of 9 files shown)


bluca:
Looks good, thanks!

What about, instead of having an additional set of values and an additional variable, simply having multiple ZMQ_EVENT_HANDSHAKE_FAILED_foo events? ZMQ_EVENT_HANDSHAKE_FAILED_PROTOCOL, ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION etc.

Or is there a particular reason for having the separate set of values?

vtellier:
Indeed that's an idea!

I did it so because I didn't wanted to change the API. But the events to change are still in draft so it should not be a problem.

bluca:
Yes that's the point of the DRAFT system, feel free to change them.

vtellier:
@blucas I have a problem with the linux test that fails, I don't have a linux machine to reproduce it in my environment.
Do you have any idea why would the handshake behave differently on linux and windows?

__Linux:__
```
lt-test_security_curve: tests/test_security_curve.cpp:311: int main(): Assertion `event == 0x2000' failed.
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
```

__Windows:__
```
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
CURVE I: cannot open client INITIATE vouch
```

Also, the _test_many_socket_ goes in timeout, on my environment it takes less than 7 seconds. Is this test a bit sensible or do you think it is an actually regression?

evoskuil:
@vtellier see [recent comments](https://github.com/zeromq/libzmq/pull/2482#issuecomment-290360191) on the test timeout.

bluca:
This does not look like a timing issue though:

```
lt-test_security_curve: tests/test_security_curve.cpp:273: int main(): Assertion `event == 0x2000' failed.
```

bluca:
So that's Valgrind finding out about an uninitialised variable error:

```
==12238== Thread 3:
==12238== Conditional jump or move depends on uninitialised value(s)
==12238==    at 0x4E9FEFC: zmq::stream_engine_t::next_handshake_command(zmq::msg_t*) (stream_engine.cpp:787)
==12238==    by 0x4E9DD73: zmq::stream_engine_t::out_event() (stream_engine.cpp:382)
==12238==    by 0x4E9E01C: zmq::stream_engine_t::restart_output() (stream_engine.cpp:440)
==12238==    by 0x4EA0187: zmq::stream_engine_t::zap_msg_available() (stream_engine.cpp:827)
==12238==    by 0x4E8CF9F: zmq::session_base_t::read_activated(zmq::pipe_t*) (session_base.cpp:285)
==12238==    by 0x4E7B385: zmq::pipe_t::process_activate_read() (pipe.cpp:262)
==12238==    by 0x4E73C51: zmq::object_t::process_command(zmq::command_t&) (object.cpp:78)
==12238==    by 0x4E6453C: zmq::io_thread_t::in_event() (io_thread.cpp:85)
==12238==    by 0x4E61FDD: zmq::epoll_t::loop() (epoll.cpp:188)
==12238==    by 0x4E620E3: zmq::epoll_t::worker_routine(void*) (epoll.cpp:203)
==12238==    by 0x4EA755B: thread_routine (thread.cpp:106)
==12238==    by 0x5304423: start_thread (pthread_create.c:333)
```

I think the fix is the following:

```
--- a/src/curve_server.cpp
+++ b/src/curve_server.cpp
@@ -45,6 +45,7 @@ zmq::curve_server_t::curve_server_t (session_base_t *session_,
     session (session_),
     peer_address (peer_address_),
     state (expect_hello),
+    current_error_detail (no_detail),
     cn_nonce (1),
     cn_peer_nonce(1)
```

Then there is another assert in the test, this time further on, without other memory complaints from Valgrind though:

```
test_security_curve: tests/test_security_curve.cpp:311: int main(): Assertion `event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION' failed.
```

bluca:
It fails at line 311 because the event is _no_detail:

```
#4  0x000000000010b564 in main () at tests/test_security_curve.cpp:311
311	    assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION);
(gdb) p /x event
$2 = 0x800
```

Instead of _encryption which would be 0x2000

vtellier:
Thanks for the details,

I found out the problem, this test case doesn't set any encryption for the client, so the greeting fails as the server expecte CURVE.
So in this case the error is a _protocol_ error due to configuration mismatch between the client and the server.

Currently the test fails because the events are sent several times, I just spottet it, will fix it ASAP.

tests/test_security_curve.cpp
```diff
@@ -229,12 +237,28 @@ int main (void)
...
     //  This will be caught by the curve_server class, not passed to ZAP
+    
+    // This enables to flush the incoming monitoring events that disturbs the test.
+    // Even though the client socket is closed, the server still handles HELLO
+    // messages.
+    Sleep(250);
+    do {
+        event = get_monitor_event(server_mon, NULL, NULL, ZMQ_DONTWAIT);
+    } while(event != -1);
+
```
vtellier: @bluca I had to add this "dirty patch" because the server handles a HELLO message just before the test below.
I believe only the client would reconnect to the server, since the client is supposed to be closed I can't explain this behavior.

tests/test_security_curve.cpp
```diff
@@ -208,8 +211,13 @@ int main (void)
     assert (rc == 0);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    int event = get_monitor_event (server_mon, NULL, NULL);
+    int event = get_monitor_event (server_mon, NULL, NULL, 0);
     assert (event == ZMQ_EVENT_HANDSHAKE_SUCCEED);
+
+    // This event has to be the last one
+    Sleep(250);
```
vtellier: I committed this `Sleep` by mistake, it will be removed in the next one, but I tested locally the behavior I describe in my comment bellow is the same with and without this `Sleep`

vtellier:
I have to close this PR until I find more time to work on it, hopefully quite soon.

Issue #2476, opened 2017-03-29
Problem: assertion causes abort after ZAP stop.


evoskuil:
The following code hits the `zmq_assert (ok);` line [with ok false] sometimes during shutdown. So `zap_pipe` can be non-null and yet the `zap_pipe->write` can fail. The presence of the assertion implies that the write should never fail - yet there are several conditions which can produce the failure, all of which appear to be normal behavior. Please let me know if I'm missing something here. I'll make a PR to fix it if someone can confirm my assumption about the intended use of assert.
```cpp
int zmq::session_base_t::write_zap_msg (msg_t *msg_)
{
    if (zap_pipe == NULL) {
        errno = ENOTCONN;
        return -1;
    }

    const bool ok = zap_pipe->write (msg_);
    zmq_assert (ok);

    if ((msg_->flags () & msg_t::more) == 0)
        zap_pipe->flush ();

    const int rc = msg_->init ();
    errno_assert (rc == 0);
    return 0;
}
```

evoskuil:
The problem propagates to all of the `send_zap_request(...)` methods:
```cpp
void zmq::curve_server_t::send_zap_request (const uint8_t *key)
void zmq::gssapi_server_t::send_zap_request ()
void zmq::null_mechanism_t::send_zap_request ()
void zmq::plain_server_t::send_zap_request (const std::string &username, const std::string &password)
```
None of these perform error handling:
```cpp
void zmq::curve_server_t::send_zap_request (const uint8_t *key)
{
    int rc;
    msg_t msg;

    //  Address delimiter frame
    rc = msg.init ();
    errno_assert (rc == 0);
    msg.set_flags (msg_t::more);
    rc = session->write_zap_msg (&msg);
    errno_assert (rc == 0);

    //  Version frame
    rc = msg.init_size (3);
    errno_assert (rc == 0);
    memcpy (msg.data (), "1.0", 3);
    msg.set_flags (msg_t::more);
    rc = session->write_zap_msg (&msg);
    errno_assert (rc == 0);

    //  Request ID frame
    rc = msg.init_size (1);
    errno_assert (rc == 0);
    memcpy (msg.data (), "1", 1);
    msg.set_flags (msg_t::more);
    rc = session->write_zap_msg (&msg);
    errno_assert (rc == 0);

    //  Domain frame
    rc = msg.init_size (options.zap_domain.length ());
    errno_assert (rc == 0);
[...]

PR #2478, opened 2017-03-29
Problem: assert used for ZAP error handling aborts process.

evoskuil:
See: https://github.com/zeromq/libzmq/issues/2476

I've followed the general existing pattern whereby when a failure results when sending a message the `zmq_msg_t` object is not closed. This looks like a leak to me but it's a more general question than I wanted to resolve in this pull request. Worst case we are temporarily trading a crash for a leak in the case of an error that apparently isn't very common.


bluca:
Why is there a massive diff in src/curve_server.cpp ?

evoskuil:
Yes, it results from the interaction with git automatic CRLF/LF conversion. When a file has a mixture of LF and CRLF characters (which this one does) git does not convert it to CRLF. But then the IDE does convert it, git doesn't then reset it to all LF (because it wasn't initially converted due to being mixed). :/

bluca:
Any chance you could fix it so that the commit doesn't change every line please? Git blame is a really important tool

src/curve_server.cpp
```diff
@@ -310,12 +310,12 @@ int zmq::curve_server_t::process_hello (msg_t *msg_)
                               sizeof hello_box,
                               hello_nonce, cn_client, secret_key);
     if (rc != 0) {
-        //  Temporary support for security debugging
-        puts("CURVE I: cannot open client HELLO -- wrong server key?");
-        errno = EPROTO;
+        //  Temporary support for security debugging
+        puts ("CURVE I: cannot open client HELLO -- wrong server key?");
+        errno = EPROTO;
```
evoskuil: This section had \r\n line endings. Converted to \n (and fixed style).

evoskuil:
All set, squashed commits.

bluca:
Great, thanks!

bluca:
Regarding closing a message on failure to send, usually an error is returned to the caller who then has to take care of it, right? (On the phone so not easy to check around)

evoskuil:
Regarding message closure on failure, I created https://github.com/zeromq/libzmq/issues/2481

PR #2491, opened 2017-04-03
[WIP] windows UWP support

bbdb68:
problem : Universal Windows Platform (UWP) not supported

solution: add UWP platform

this PR is an attempt to add UWP support for native ZeroMQ. I would like some feedback at least about

* the way the ZMQ_HAVE_WINDOWS_UWP flag is setup. I set it up in the cmakelists in order to have it defined in the secondary targets (remote_lat etc) but it is not necessary for libzmq itself

* I did not succed to make if_indextoname being resolved (tcp_address.cpp) : is someone is able to do better ?

* Where should I put some doc about UWP compilation ? (I will remove the cmake.bat file)

thanks for you feedback

Note : UWP is a strange beast : zmq works and is able to bind sockets, but the platform does not allow two processes to communicate throught TCP/IP on the same computer (even if you use the real IP instead of 127.0.0.1). So if in order to perform some tests about this, you will need at least to boxes....


src/windows.hpp
```diff
@@ -39,10 +39,14 @@
 
 //  Set target version to Windows Server 2008, Windows Vista or higher.
 //  Windows XP (0x0501) is supported but without client & server socket types.
-#ifndef _WIN32_WINNT
+#if defined _WIN32_WINNT && !defined ZMQ_HAVE_WINDOWS_UWP
```
bluca: The original is an ```ifndef _WIN32_WINNT``` so shouldn't it be ```!defined (_WIN32_WINNT)``` ?

bluca:
The INSTALL file is a good place to place instructions to build:

https://github.com/zeromq/libzmq/blob/master/INSTALL

src/windows.hpp
bbdb68: right !

bluca:
@bbdb68 is it good to go now?

bbdb68:
@bluca for me it is good, it would be nice to have some feedback from windows build maintainers but if you are ok to merge as is...

bluca:
If there's any comment if can be fixed later.
If you haven't already, could you please open a PR to add your re licensing permission?

bbdb68:
ok - next to come are zproject and czmq ;-)

JSalazar88:
@bbdb68 @bluca:  If I link and use the lib in a C++/WinRT UWP project as supported here, I get an MS error dialog:
> Unable to activate WIndows Store app [app name] The [app name].exe process started, but the activation request failed with error 'The target process aborted before activation completed. More information may be available in the Debug pane of the Output window (Debug->Windows->Output)'.

Even if I only call `zmq_version` and nothing else, the app still crashes, but commenting it out allows the app to run.

I am building via `cmake -G "Visual Studio 16 2019" -A x64 -DWITH_PERF_TOOL=OFF -DZMQ_BUILD_TESTS=OFF -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 -DENABLE_CURVE=OFF -S .. -B .`

And Event Viewer shows:

> Activation for Microsoft.SDKSamples.HttpClient.CPPWINRT_8wekyb3d8bbwe!App failed. Error code: The app didn't start.. Activation phase: COM ActivateExtension

Any thoughts on how I can get libzmq working in an example Windows Store app?
FWIW, I am simply editing the Microsoft "[HttpClient](https://github.com/microsoft/Windows-universal-samples/tree/master/Samples/HttpClient/cppwinrt)" sample application (cppwinrt solution), which otherwise works perfectly fine.

PR #2493, opened 2017-04-04
CMake: Add cross compiling support and export targets to package config

herbrechtsmeier:
Add cross compiling support to CMake by disable runtime checks during cross compile.

Export the targets to the CMake package config and thereby to user projects. This allows other projects to include this project as a sub directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 cmake_minimum_required (VERSION 2.8.12)
 project (ZeroMQ)
 
-list (INSERT CMAKE_MODULE_PATH 0 "${CMAKE_SOURCE_DIR}")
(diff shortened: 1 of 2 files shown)


bluca:
@herbrechtsmeier thanks but this seems to cause a regression in Windows, could you please have a look?

https://ci.appveyor.com/project/zeromq/libzmq/build/build-1326

herbrechtsmeier:
@bluca The current windows test have a problem with the order of the include paths because of multiple platform.hpp files (`CMAKE_ SOURCE_DIR`, `build/msvc`). Additionally the AppVeyor test always test the same configuration because of a mismatch between command line option (`WITH_SODIUM`) and CMake option (`WITH_LIBSODIUM`).

bluca:
Would you be able to help fix it, please?

bluca:
@herbrechtsmeier looks like it's working now. Is it good to go?

PR #2499, opened 2017-04-04
[WIP] Avoid calling zmq_poll() from zmq::proxy() and re-creation of zmq::so…

bjovke:
zmq::poll() experiences huge impact on CPU and I/O on Windows, can be demonstrated with "asyncsrv.cpp" example.
This is because there are two calls to `zmq_poll()` in infinite while loop and in each call `zmq::socket_poller_t` is being created and deleted.

This change is actually code from `zmq_poll()` merged into `zmq::poll()`.
After this change it works like a charm, no high CPU usage nor high I/O.

### DIFF
--- a/src/proxy.cpp
+++ b/src/proxy.cpp
@@ -283,29 +283,29 @@ int zmq::proxy (
             if (unlikely (rc < 0)) {
                 return close_and_return (&msg, -1);
             }
-        }
 
-        //  Transform poller events into zmq_pollitem events.
-        j_start = 0;
-        found_events = rc;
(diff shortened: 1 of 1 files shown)


bjovke:
Hold on with this, there's something wrong.
Some more things need to be reinitialized before reuse. I will take a look at this.

bluca:
No problem - added [WIP] to the title, feel free to remove it when you are ready

bjovke:
I will test another version of this a little bit before I make a pull request.
This one sucks.

PR #2502, opened 2017-04-05
appveyor: Use correct CURVE security options

herbrechtsmeier:
The AppVeyor build test options mismatch between command line options
and CMake CURVE security options. Rename the command line options to
the correct names.


bluca:
Thanks!

PR #2504, opened 2017-04-05
cmake: Use LOCATION property only for imported targets

herbrechtsmeier:
Newer CMake versions allow use of the LOCATION target property only for
imported targets (CMP0026). Because the old package variables are only
needed for backward compatibility after configuration they could be
deactivated for super projects which build ZeroMQ together with other
projects.


bluca:
@herbrechtsmeier could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

Issue #1239, opened 2014-10-31
zmq_poll for ZMQ_POLLOUT utilized 100% for certain sockets.


Kentzo:
I'm on ZeroMQ 4.0.4

Polling sockets of certain types for ZMQ_POLLOUT can utilize 100% of the CPU. Interesting that for some sockets problem only exists when they are used to connect.

Here is the code I used for testing:

```
#include <assert.h>
#include <zmq.h>

int main()
{
    void *c = zmq_ctx_new();

    void *s = zmq_socket(c, ZMQ_REQ);
    int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is 100%
    // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage less than 1%

    // void *s = zmq_socket(c, ZMQ_REP);
    // int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is less than 1%
    // // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage less than 1%

    // void *s = zmq_socket(c, ZMQ_DEALER);
    // int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is 100%
    // // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage less than 1%

    // void *s = zmq_socket(c, ZMQ_ROUTER);
    // int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is 100%
    // // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage is 100%

    // void *s = zmq_socket(c, ZMQ_PUB);
    // int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is 100%
    // // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage is 100%

    // void *s = zmq_socket(c, ZMQ_PUSH);
    // int rc = zmq_connect(s, "tcp://127.0.0.1:*"); // CPU usage is 100%
    // // int rc = zmq_bind(s, "tcp://127.0.0.1:*"); // CPU usage less than 1%

    assert(rc == 0);

    zmq_pollitem_t items[1];
    items[0].socket = s;
    items[0].events = ZMQ_POLLOUT;

    while(1)
    {
        rc = zmq_poll(items, 1, 100);
        assert (rc >= 0);
    }

    return 0;
}
```

PR #2506, opened 2017-04-05
[WIP] Reworked zmq::proxy() for improved performance.

bjovke:
There's a huge CPU impact on running `zmq_poll()` inside `zmq::proxy()`.
There are many malloc/free's and FD sets are created each poll.
`zmq::proxy()` is reworked by "inlining" the code from `zmq_poll()`, setting everything to work and optimizing a little bit.
The goal is that in main `while()` loop there's reuse of all variables/objects.

Since I'm not familiar with broader ZeroMQ inner workings, it might be that I didn't do it completely right.

There's also a strange case, at least on Windows, when `frontend_ `has `ZMQ_POLLIN `and `ZMQ_POLLOUT `events and backend_ none and the other way around. It might be that my test is to blame but it shouldn't have happened anyway. In this case there's a loop causing 100% CPU.

Please comment on this pull request.

### DIFF
--- a/src/proxy.cpp
+++ b/src/proxy.cpp
@@ -52,7 +52,7 @@
 #include "socket_base.hpp"
 #include "err.hpp"
 
-#ifdef ZMQ_BUILD_DRAFT_API
+#ifdef ZMQ_HAVE_POLLER
 
 #include "socket_poller.hpp"
 
@@ -74,7 +74,7 @@
         return close_and_return (&msg, -1);\
     }
 
-#endif // ZMQ_BUILD_DRAFT_API
+#endif // ZMQ_HAVE_POLLER
 
 
 int capture (
@@ -133,7 +133,7 @@ int forward (
 }
 
 
-#ifdef ZMQ_BUILD_DRAFT_API
+#ifdef ZMQ_HAVE_POLLER
 
 int zmq::proxy (
     class socket_base_t *frontend_,
@@ -392,7 +392,7 @@ int zmq::proxy (
     return close_and_return (&msg, 0);
 }
 
-#else // ZMQ_BUILD_DRAFT_API
+#else // ZMQ_HAVE_POLLER
 
 int zmq::proxy (
     class socket_base_t *frontend_,
@@ -497,5 +497,5 @@ int zmq::proxy (
     return close_and_return (&msg, 0);
 }
 
-#endif // ZMQ_BUILD_DRAFT_API
+#endif // ZMQ_HAVE_POLLER
 


somdoron:
I think you will gain same performance improvement if you will use zmq_poller directly

bjovke:
Hmm, the reason for this change is exactly poor performance when using zmq_poll() directly.

somdoron:
because zmq_poll create new zmq_poller (socket_poller) on every call.

Try to create one zmq_poller and reuse it. It is actually exactly what you are trying to develop.

bjovke:
Another change added.
Huge mistake corrected, `socket_poller_t::wait()` returns -1 and sets `errno` to `ETIMEDOUT` in case of timeout, it does not return 0 like zmq_poll().
Corrected handling when ZMQ_POLLIN is available but corresponding ZMQ_POLLOUT not.

bjovke:
@somdoron Yes, that's exactly what I want to achieve. I've created three socket_poller_t and they are being reused in `while()` loop:
- `zmq::socket_poller_t poller_in, poller_out, poller_control;`

When `zmq_poll()` is called, this is the call flow:
`zmq_poll()` -> `zmq_poller_poll()` -> create `zmq::socket_poller_t` -> `zmq_poller_wait_all()` -> `zmq::socket_poller_t::wait()` -> `destroy zmq::socket_poller_t`.
There were two calls to zmq_poll() in original code so this all is done twice, causing high malloc/free rate.

With new code it goes like this:
create `zmq::socket_poller_t` -> `zmq::socket_poller_t::wait()` in loop, `socket_poller_t` reused.
I've checked inside `zmq::socket_poller_t::wait()`, there's no point yet in unwrapping it further.

Another thing I've noticed is that PAUSE/RESUME was not working properly.
With old code, when you pause proxy it causes a 100% CPU loop because all `zmq_poll()`s are signaled and message processing is being ignored so 'zmq_poll()'s stay signalled.
With new code when proxy is paused, only blocking wait on `control_` socket is being done, waiting for RESUME or TERMINATE.

somdoron:
@bjovke just noticed that what you are trying to do.
I actually start working on it few weeks ago. Take a look, it is not finished (I want to organize the code). Anyway take a look, it might help you.

https://github.com/somdoron/libzmq/commit/af202d499cf546caf000d4858ed6e5c9a85faf2d

Let me know what you think

bjovke:
@somdoron Yes, the idea is the same. Your approach seems more like `C` than `C++`. You're using `void* poller = zmq_poller_new ();` but it could be static variable. It doesn't hurt anyway because it's done only once on `zmq::proxy()` startup. You see you caused yourself a lot of work not to leak `poller` with many `zmq_poller_destroy (&poller);`
The main culprit is `zmq_poller_poll()` which you avoided. Your call flow for wait is:
`zmq_poller_wait()` -> `zmq_poller_wait_all()` -> `zmq::socket_poller_t::wait()`
This is good, there's only a little overhead of calling `zmq::socket_poller_t::check_tag()` and checking if event is `NULL`.
If you've tested it and it works good this for sure solves the speed issue.
But it seems now that your and mine changes are conflicting. :)

bjovke:
Change 2 - minor code reformatting....

somdoron:
All test are passing.
And you are right, it is better to use socket_poller directly.
I also tried to close all the edge cases, like RESUME and PAUSE and cases where backend or frontend is not ready to send.

Anyway, if you want I can make the socket_poller change or drop it all together and you can continue and work on your patch.

bjovke:
@somdoron I'll try to run my modified test case against your changes to see how it behaves.
I was not aware you were working on same thing... Which patch do you prefer?

somdoron:
@bjovke you also need to support mode without socket_poller, when zeromq is compiled without the draft mode.

As you already created a pull request yours is preferred. Just make sure you cover the edge cases.

bjovke:
@somdoron Yes, you're right, it needs to work without DRAFT also. I need to rework some stuff, especially the case when there's ZMQ_POLLIN but ZMQ_POLLOUT is not available.

evoskuil:
For reference: https://github.com/zeromq/libzmq/issues/1239

bjovke:
@evoskuil Yes, that's one of the issues currently present which is addressed with PR.

bjovke:
@evoskuil Actually #1239 does not have anything to do with this. @Kentzo is right, code in his test is bad. He's polling for ZMQ_POLLOUT which is always available except in cases when receiving queue is full.

evoskuil:
I'm wondering why I don't see the issue.

bjovke:
@evoskuil Which issue do you mean?

evoskuil:
High or 100% CPU utilization on Windows using `zmq::proxy`.

bjovke:
There are three CPU issues in original proxy:

- Unusually high CPU utilization on higher message rate. There's no 100% CPU here but the code is highly CPU consuming. What it comes down is `zmq_poll()` having high malloc/free rate and I've noticed that one `zmq::socket_poller_t` object has over 190 kB, at least on Windows, and it's being created/deleted on every `zmq_poll()` call.
- 100% CPU utilization when frontend and/or backend sockets are not available for `ZMQ_POLLOUT` due to full queue.
- 100% CPU loop when proxy is paused.

If you have some test program doing high message rate through proxy and/or pausing proxy you will notice it. If you want I can send you one test program.

bjovke:
@somdoron In order not to complicate things further, I've added untouched previous code to be used when DRAFT is not enabled - change 8. I see in `zmq.h` that defining `ZMQ_BUILD_DRAFT_API `automatically defines `ZMQ_HAVE_POLLER `so new code is valid.

bjovke:
One `zmq::socket_poller_t` object has **over 190 kB** - found out why.
There are three `fd_set`s inside. **FD_SETSIZE** is defined as 16384 in libzmq.props for Visual studio 2015.
That means that total of 16384 * 4 * 3 = 196608 bytes are allocated just for one  `zmq::socket_poller_t`, plus some small amount for other member variables. For a lot of new/delete operations for this objects in short time CPU really gets loaded.

bjovke:
I've noticed that `zmq::socket_poller_t::wait()` allocates from stack another three `fd_set`s
-  `fd_set inset, outset, errset;`

They are used as copies of `pollset_in`, `pollset_out `and `pollset_err` so these originals don't get changed by `select()` call. That's another **190 kB** of stack taken during execution of `wait()`.
This is for some future pull request...

bjovke:
Now I've ended up with many new/deletes because `zmq::socket_poller_t`s are too large on Windows to put on stack. :)
This PR seems fine now:

- CPU usage on higher message rate is lowered significantly because `zmq::socket_poller_t`s are created once on proxy startup and reused.
- There's proper wait when any of two (frontend/backend) outgoing queues is full (no 'ZMQ_POLLOUT' available) instead of going into 100% CPU loop.
- PAUSE/RESUME works properly.

bjovke:
#2479 has the same root cause as this PR is trying to improve.

bluca:
Code looks good, but may I ask for a small cleanup? It would be better for the git history and git blame and so on if you had 2 commits in the PR, one adding the new code and a separate one that fixes the formatting of the existing old code. Thanks!

bjovke:
@bluca Yes, of course. Should I close the PR and create these two PRs?

bluca:
No need to, just use the same

bjovke:
Closed to rework and create 2 PRs instead of all of these.

PR #2510, opened 2017-04-08
choose IP(v6) address based on ipv6_ flag in a correct way

lytboris:
Current code treats ipv6_ flag wrong when looking for IP(v6) address from an interface: it will choose IPv4 if it goes first even when ipv6_ flag is set resulting connection breakage.


lytboris:
@bluca , this pull requests being merged breaks ZMQ when you have IPv6 enabled on socket but try to bind IPv4 address.

PR #2517, opened 2017-04-09
[WIP, do not merge]IPv6 source address may be selected from wrong scope

lytboris:
Do source address resolution after destination address is known and scope id is set. Use only addresses from the same scope when selecting an IPv6 address from an interface. Disallow using explicitly specified different scopes.

**Please note that changes related to MS Windows were not tested - I have no platform to build and test these changes**

### DIFF
--- a/src/tcp_address.cpp
+++ b/src/tcp_address.cpp
@@ -647,39 +647,6 @@ int zmq::tcp_address_t::resolve (const char *name_, bool local_, bool ipv6_, boo
           addr_str [addr_str.size () - 1] == ']')
         addr_str = addr_str.substr (1, addr_str.size () - 2);
 
-    // Test the '%' to know if we have an interface name / zone_id in the address
-    // Reference: https://tools.ietf.org/html/rfc4007
-    std::size_t pos = addr_str.rfind('%');
-    uint32_t zone_id = 0;
-    if (pos != std::string::npos) {
-        std::string if_str = addr_str.substr(pos + 1);
-        addr_str = addr_str.substr(0, pos);
-        if (isalpha (if_str.at (0)))
-#if !defined ZMQ_HAVE_WINDOWS_TARGET_XP && !defined ZMQ_HAVE_WINDOWS_UWP 
-            zone_id = if_nametoindex(if_str.c_str());
-#else
-            // The function 'if_nametoindex' is not supported on Windows XP.
-            // If we are targeting XP using a vxxx_xp toolset then fail.
-            // This is brutal as this code could be run on later windows clients
-            // meaning the IPv6 zone_id cannot have an interface name.
-            // This could be fixed with a runtime check.
-            zone_id = 0;
-#endif
-        else
-            zone_id = (uint32_t) atoi (if_str.c_str ());
-
-        if (zone_id == 0) {
-            errno = EINVAL;
-            return -1;
-        }
-
-        // Check both src and dst addresses have the same scope
-        if (ipv6_ && is_src_ && zone_id != address.ipv6.sin6_scope_id) {
-            errno = EINVAL;
-            return -1;
-        }
-    }
-
     //  Allow 0 specifically, to detect invalid port error in atoi if not
     uint16_t port;
     if (port_str == "*" || port_str == "0")
@@ -705,18 +672,16 @@ int zmq::tcp_address_t::resolve (const char *name_, bool local_, bool ipv6_, boo
 
     //  Set the port into the address structure.
     if (is_src_) {
-        if (source_address.generic.sa_family == AF_INET6) {
+        // sin6_addr & sin6_scope_id are set by resolve_*
+        if (source_address.generic.sa_family == AF_INET6)
             source_address.ipv6.sin6_port = htons (port);
-            source_address.ipv6.sin6_scope_id = zone_id;
-        }
         else
             source_address.ipv4.sin_port = htons (port);
     }
     else {
-        if (address.generic.sa_family == AF_INET6) {
+        // sin6_addr & sin6_scope_id are set by resolve_*
+        if (address.generic.sa_family == AF_INET6)
             address.ipv6.sin6_port = htons (port);
-            address.ipv6.sin6_scope_id = zone_id;
-        }
         else
             address.ipv4.sin_port = htons (port);
     }
--- a/src/tcp_address.cpp
+++ b/src/tcp_address.cpp
@@ -178,6 +178,8 @@ int zmq::tcp_address_t::resolve_nic_name (const char *nic_, bool ipv6_, bool is_
 
 //  On these platforms, network interface name can be queried
 //  using getifaddrs function.
+//  This function relies on pre-assigned destination address sockaddr_in6
+//     when ipv6_ and is_src_ flags are set
 int zmq::tcp_address_t::resolve_nic_name (const char *nic_, bool ipv6_, bool is_src_)
 {
     //  Get the addresses.
@@ -210,6 +212,15 @@ int zmq::tcp_address_t::resolve_nic_name (const char *nic_, bool ipv6_, bool is_
             continue;
 
         const int family = ifp->ifa_addr->sa_family;
+
+        // If destination address is link-local, check if Scope Id is equal
+        // before picking up an address
+        if (ipv6_ && is_src_ && family == AF_INET6
+                && IN6_IS_ADDR_LINKLOCAL (&(address.ipv6.sin6_addr))
+                && ((struct sockaddr_in6 *)ifp->ifa_addr)->sin6_scope_id != address.ipv6.sin6_scope_id) {
+            continue;
+        }
+
         if (family == (ipv6_ ? AF_INET6 : AF_INET)
         && !strcmp (nic_, ifp->ifa_name)) {
             if (is_src_)
@@ -286,6 +297,8 @@ int zmq::tcp_address_t::wchar_to_utf8(const WCHAR * src, char ** dest) const {
     return 0;
 }
 
+//  This function relies on pre-assigned destination address sockaddr_in6
+//     when ipv6_ and is_src_ flags are set
 int zmq::tcp_address_t::resolve_nic_name(const char *nic_, bool ipv6_, bool is_src_)
 {
     int rc;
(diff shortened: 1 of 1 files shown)


bluca:
When the address is not link local, is scope id guaranteed to be zero? Or can is it undefined?

lytboris:
It seems I'm treating sin6_scope_id wrong indeed. Give me some more time to dig it.

bluca:
There's a macro to check if an in6_addr pointer is linklocal: IN6_IS_ADDR_LINKLOCAL

lytboris:
Yes, and current code for parsing '%<iface>' should be altered in a way it allows only link-local addresses with scope id.

lytboris:
bluca@, are you aware why fe80:...%... is parsed manually in ZMQ? It seems it's a task for getaddrinfo(). Any MS Windows issues with it?

lytboris:
@syenna, can you please share your comments too? Why %-part of address is parsed manually? getaddrinfo() does fill in sin6_scope_id upon resolution. At least on FreeBSD :)

bluca:
Don't know, sorry, but a multiplatform problem would be a safe guess

lytboris:
I've updated patch addressing LL issue and removed % manual parsing. Windows is the only black horse here and I can not find a single document over the Internet describing getaddrinfo() behavior on Windows in case of IPv6 LL addresses.
https://tools.ietf.org/html/draft-smith-ipv6-link-locals-apps-00#section-5.2 does not mention any restrictions though so I hope this should be OK.

src/tcp_address.cpp
```diff
@@ -622,33 +647,6 @@ int zmq::tcp_address_t::resolve (const char *name_, bool local_, bool ipv6_, boo
           addr_str [addr_str.size () - 1] == ']')
         addr_str = addr_str.substr (1, addr_str.size () - 2);
 
-    // Test the '%' to know if we have an interface name / zone_id in the address
```
bluca: To avoid risks of regressions on weird platforms, I think it would be better to leave this block in place, but use "zone_id" below only if the [source_]address.ipv6.sin6_scope_id variables are 0, so that the value set by the system API has the precedence

lytboris: Well, IPv6 is disabled completely on all weird platforms. I don't think it is good to leave this code in. We can wait for @syenna's reply.

lytboris:
More I look into this, more I see that I'm trying to pull whole rfc6724 in. Binding works more or less OK with IPv4 but is really complicated with IPv6. Once IPv6 security extensions are enabled, not all GUA may be used as a source addresses - they may be in deprecated state and there is no good way to know that from ZMQ perspective.

As a result I would vote for rather disable IPv6 source address binding than re-inventing rfc6724 support in userland.

syenna:
Sorry for the late reply. It's been a while since I submitted zeromq changes. I'll look into it closely today before I comment on this.

Thanks,
Sathish

src/tcp_address.cpp
syenna: zone_id is expected in the passed _name string for non-global/link-local IPv6 addresses. Without the zone_id, a link-local IPv6 address can not be resolved with no ambiguity. Linux OS expects the sin6_scope_id to be set appropriately for a link-local address for it to bind a socket successfully. sin6_scope_id is not always '0' and it depends on the interface and its index in the system. This code was necessary to get zeromq sockets work properly on IPv6 link local addresses.

syenna:
Systems with multiple interfaces need the interface name/index to be passed along with link-local address. Without the interface name/index, OS can not resolve a link-local address with no ambiguity. When a user of this library provides an IPv6 address, the address is expected to have interface name/index in case of link-local addresses. Setting sin6_scope_id must be based on provided address. It can not be set to 0 by default. This library is in use in a system with multiple IPv6 interfaces, hence this change was added.

src/tcp_address.cpp
syenna: It is better to keep this block in place.

lytboris:
@syenna , I know how link-local addresses work and being specified :) My point is that you don't need to set sin6_scope_id manually here - passing whole IPv6 address (with %... suffix) straight to getaddrinfo() works perfect, no need to parse it in ZMQ: getaddrinfo() will set sin6_scope_id for you. Your change contains two parts - one part allows % as a part of the address and it should be there and the other one seems to be doing the same thing getaddrinfo() is for.

syenna:
I do not remember getaddrinfo() handling it properly when I added those changes. If it works as you say, then this block can be removed.

bluca:
I know on Solaris getnameinfo doesn't handle it properly, so I'd rather avoid risking a regression

lytboris:
At least modern Linux works:
`        if( ( rv = getaddrinfo("fe80::beef%enp0s9", "http", &hints, &res) ) ) {`
...
`                printf("%s %d\n", inet_ntop(AF_INET6, &((struct sockaddr_in6 *)p->ai_addr)->sin6_addr, buf, sizeof(buf)),`
`                                ((struct sockaddr_in6 *)(p->ai_addr))->sin6_scope_id);`
resulting
`% ./a.out
fe80::beef 2
%
`

bluca:
Yep - but we do have users on Solaris 10/11, AIX 5/6/7, zOS and more

lytboris:
Are you sure all these OSes require interface index in sin6_scope_id? How do you know that? It is possible they are weird so much so they use different id in sin6_scope_id rather than ifIndex.
Next, you said it's OK first to let getaddrinfo() to set sin6_scope_id and then set it to zone_id if it does not match with sin6_scope_id. This may break even existing systems if algorithm for setting sin6_scope_id is changed for a reason.
Do you have a way to re-check getaddrinfo() behavior for those OSes?
Current code neither allow run in dual-stack mode nor allow multiple A/AAAA- records in remote URL. Leaving things like manual zone_id parsing makes it more complicated to implement a fix for this.

bluca:
>Are you sure all these OSes require interface index in sin6_scope_id? How do you know that? It is possible they are weird so much so they use different id in sin6_scope_id rather than ifIndex.

Because it's working right now

> Next, you said it's OK first to let getaddrinfo() to set sin6_scope_id and then set it to zone_id if it does not match with sin6_scope_id. This may break even existing systems if algorithm for setting sin6_scope_id is changed for a reason.

Not if they don't match, which is unlikely anyway, but if one is not set at all, so as a fallback to avoid regressions on systems where these system calls are not fully implemented, like old Solaris versions or old Linux/glibc versions

> Do you have a way to re-check getaddrinfo() behavior for those OSes?

Unfortunately no,  which is why extra care is needed to avoid regressions

>Current code neither allow run in dual-stack mode nor allow multiple A/AAAA- records in remote URL. Leaving things like manual zone_id parsing makes it more complicated to implement a fix for this.

Sorry, I'm not sure I understand how that is related to having a fallback for getaddrinfo?

lytboris:
So in summary your idea is
1) try passing %-suffixed address into getaddrinfo and check if sin6_scope_id is set.
2) if it is zero - parse %-suffix manually and fill in
3) if if is non-zero - leave it as is.
Correct? 

My point is that IPv6 code is a bit messy and I was trying to clean it up a bit before trying to make a next step - make a ZMQ_IPV6 option one that enables IPv6 and leaves IPv4 working too. Now if you enable IPv6 - all IPv4 addresses are ignored at all, so ZMQ is either IPv4-only or IPv6-only now. If IPv6 is broken, ZMQ will never try to fallback to IPv4 because only first reply from getaddrinfo() is stored in socket structure.

bluca:
> So in summary your idea is
> 
> try passing %-suffixed address into getaddrinfo and check if sin6_scope_id is set.
if it is zero - parse %-suffix manually and fill in
if if is non-zero - leave it as is.
Correct?

Yes - this minimizes the risk of problems where those APIs are not fully implemented

> My point is that IPv6 code is a bit messy and I was trying to clean it up a bit before trying to make a next step - make a ZMQ_IPV6 option one that enables IPv6 and leaves IPv4 working too. Now if you enable IPv6 - all IPv4 addresses are ignored at all, so ZMQ is either IPv4-only or IPv6-only now. If IPv6 is broken, ZMQ will never try to fallback to IPv4 because only first reply from getaddrinfo() is stored in socket structure.

Thanks for the cleanups, improvements are always welcome
But I always enable IPv6 in all my applications in all sockets and can use v4 addresses just fine, so not sure what you mean? Could you provide a test case for this?

lytboris:
> But I always enable IPv6 in all my applications in all sockets and can use v4 addresses just fine, so not sure what you mean? Could you provide a test case for this?
How do you specify remote endpoint? A FQDN having both A- and AAAA- records? Current code works OK  upto the dual-stack remote FQDN.

bluca:
Right - I only use IP addresses, no DNS involved

lytboris:
Now ZMQ holds only one sockaddr structure for source/destination addresses - the only item chosen from a set of items returned by getaddrinfo(). An obvious thing to do is to store all items and ignore IPv6 records if IPv6 is **not** enabled on the socket. But this turns into nightmare if I use, say `tcp://lo1:0;my.remote.hostname.org:9999` when that FQDN resolves into both IPv4 and IPv6 addresses. Socket will never try IPv4 (IPv6) even when IPv6 (IPv4) is broken somehow - brackets reflects ZMQ_IPV6 status. More: lo1 is a problem when IPv6 is on the staage: see my comments on security extensions above.
To sum this all up: you're trying to avoid regressions even for sin6_scope_id but it seems that the only good way to move IPv6 support forward is to disable source interface resolution for IPv6 completely. I.e. let broken systems (ones having incomplete support for rfc6724 for instance) be broken. And in this case sin6_scope_id regression fades away completely.

lytboris:
We can disable (forbid) interface resolution for IPv6 unless IFF_LOOPBACK is set on the interface we're trying to bind to or destination address is not a GUA. This should not bring any regression because current code is broken anyway: it will choose link-local address in non-loopback interface since this is how getifaddrs() sorts addresses. Any other ideas how to make resolve_nic be sane with IPv6?

lytboris:
More binding issues uncovered with IPv6 enabled on a socket. If you enable IPv6 on a socket and try to bind to a IPv4 address - this will fail.

PR #2519, opened 2017-04-10
Case found not covered in latest zmq::proxy() code.

bjovke:
Latest `zmq::proxy()` PR #2518 covered cases when one or both of proxy's `ZMQ_POLLOUT` queues are full.
There's another case, when one `ZMQ_POLLOUT` is full and second one has no pending messages.
In that case `ZMQ_POLLOUT `and `ZMQ_POLLIN `are both signalled either for frontend or backend socket while the other socket is not signalled at all - no `ZMQ_POLLOUT` because the queue is full and no `ZMQ_POLLIN` because there are no pending messages.
This can cause 100% CPU loop.
This case was not covered and this PR adds it.


src/proxy.cpp
```diff
@@ -66,6 +66,8 @@
     delete poller_receive_blocked;\
     delete poller_send_blocked;\
     delete poller_both_blocked;\
+    delete poller_frontend_only;\
+    delete poller_backend_only\
```
bluca: missing trailing ";" (although it's a macro"

bjovke: @bluca Hm, yes, I deliberately removed it because macro is called with a ";" at the end, so it doesn't look odd in the code.
What is the rule for this, I can return the ";"?

bjovke: @bluca `CHECK_RC_EXIT_ON_FAILURE()` has an extra ";" after curly braces when called.
Maybe I should fix that also, removing ";" when called or changing macro so ";" at the end is syntactically correct?

bluca: it's fine if it's intentional

bjovke:
@bluca I've corrected CHECK_RC_EXIT_ON_FAILURE() macro because it can cause issues with some usage scenarios. Take look at: [Swallowing-the-Semicolon](https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html).

bluca:
Yeah I'm familiar with that, actually I think the best solution for macros are braced groups: ({ foo; bar; }) but I'm not sure if MSVC++ supports them?

bjovke:
What about `do { ..code... } while (0)` ?
That also appears on numerous places as recommendation.

bluca:
that's common yes, but most compilers (gcc/icc/clang) support braced groups which have additional benefits, see http://stackoverflow.com/questions/1238016/are-compound-statements-blocks-surrounded-by-parens-expressions-in-ansi-c for some description
the question is only if msvc++ supports it

PR #2521, opened 2017-04-10
Consistency of Windows TCP tuning assertions

jakecobb:
Treat ECONNRESET, ENETRESET and EINTR as non-fatal
failures in TCP Tuning on Windows as they are in the
Unixes.  Also adds ENETRESET to the Unix list that was missing.

Related to PR #2492.

### DIFF
--- a/src/tcp.cpp
+++ b/src/tcp.cpp
@@ -321,11 +321,14 @@ void zmq::tcp_assert_tuning_error(zmq::fd_t s_, int rc_)
     zmq_assert (rc == 0);
     if (err != 0) {
         wsa_assert (err == WSAECONNREFUSED
-                 || err == WSAETIMEDOUT
+                 || err == WSAECONNRESET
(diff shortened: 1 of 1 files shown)


jakecobb:
Test failures appear to be from the poller changes:

```
The following tests FAILED:
	 22 - test_stream_disconnect (Failed)
	 23 - test_disconnect_inproc (Timeout)
	 64 - test_poller (Failed)
	 68 - test_radio_dish (Failed)
```

There is also a timeout of `test_stream_disconnect` after 10 seconds.  It needs about 15 seconds to complete in Debug locally, but it never hits the code path of this commit so I don't think any performance issue there is related to this PR.

bluca:
Yes that's a known issue, been worked on.

For your next contributions, please follow the guidelines for commit messages as instructed by the Pull Request template (Problem: xxx Solution: yyy)

PR #2526, opened 2017-04-11
Problem: GCC 6 build fails due to misleading indentation and visually ambiguous if/else block ordering.

bjovke:
Solution: fixed indentation and if/else block.


bluca:
thanks

Issue #2505, opened 2017-04-05
cmake and windows


diorcety:
There is a bug with cmake and library on Windows if you do a OUTPUT_NAME with the same name that the target the output will have liblib. in this case liblibzmq.dll.a and liblibzmq.dll .
Changing the libzmq target by libzmq-shared (as libzmq-static) will solve the issue.
(tested on msys2)

bluca:
could you please send a PR to fix it?

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/2527

Issue #2376, opened 2017-03-19
Re-licensing libzmq


rlenferink:
Hello libzmq contributors,

I'm starting this issue because of the current libzmq license. At the moment, libzmq is licensed as LGPL with static linking exception and this makes it a complex "one of a kind" license.

Almost a year ago the ZeroMQ project team started to collect license grants to re-license libzmq to a clearer license, namely the MPLv2. Collecting these license grants stopped for a while, because Pieter Hintjens, the former BDFL (Benevolent Dictator for Life) behind ZeroMQ, got sick and passed away.

We want to continue Pieter's work with collecting these contributor license grants. We want to ask you to send us a license grant to permit the libzmq to relicense to a different license.  At the moment, libzmq is licensed as LGPL 3 with a static linking exception (http://zeromq.org/area:licensing).  This makes it hard to make use of libzmq in projects with some other Open Source licenses (e.g. projects using the Apache License).  The existing LGPL 3 with exceptions is also a non-standard license that some corporate lawyers are unwilling to accept for use in their companies.

For relicensing the libzmq project, we need to collect a license grant from each individual contributor who wrote a major piece of code in the development process of libzmq.

We are requesting contributors to send in a license grant to relicense libzmq to one of the options listed below:
1. relicense to the MPLv2 license specifically;
1. relicense to the MPLv2 or any "share alike" Open Source Initiative (OSI) approved license chosen by the current ZeroMQ BDFL;
1. relicense to the MPLv2 or any Open Source Initiative (OSI) approved license chosen by the current ZeroMQ BDFL **(preferred selection)**

If a contribution was made during working hours for an employer then we will also need a license grant (or "quit claim") from the employer.

We would like to follow a certain template as license grant to ensure all the necessary information is included.  There are three templates, one for each option.  Please choose the template that best represents your wishes.

The templates can be found at: https://github.com/zeromq/libzmq/tree/master/RELICENSE/templates

To include the license grant, please:
- fork the original libzmq repository
- create a new file in the RELICENSE directory, named after your name and/or your employers name, with the license grant as content, using the template chosen
- commit this file to the repository, using the same email address as your earlier contributions
- create a pull request on libzmq to have your license grant merged into the repository.

See the existing files in:
https://github.com/zeromq/libzmq/tree/master/RELICENSE

or this specific example: https://github.com/zeromq/libzmq/blob/master/RELICENSE/rlenferink.md

for a guide for how to fill out the template.

---

[...]

PR #2527, opened 2017-04-12
Fix compilation on windows using msys2

diorcety:
The first commit fix an issue on Windows regarding python.
Why using python for running asciidoc?

The second commit fix the issue #2505 
Using and empty prefix "" doesn't work. The lib prefix is duplicated


bluca:
1) Please send a relicense agreement as per https://github.com/zeromq/libzmq/issues/2376
2) For your next PRs, please follow the guidelines as explained in the PR template and use the Problem: xx Solution: yy format for the git commit messages

PR #2533, opened 2017-04-14
Problem: FD set copying for Windows still not optimal in some places.

bjovke:
Solution: Improved `memcpy()` of FD sets for Windows builds on three locations in code where FD sets are being copied.

### DIFF
--- a/src/select.cpp
+++ b/src/select.cpp
@@ -401,16 +401,36 @@ zmq::select_t::fds_set_t::fds_set_t ()
 
 zmq::select_t::fds_set_t::fds_set_t (const fds_set_t& other_)
 {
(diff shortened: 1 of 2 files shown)


pijyoi:
In zmq.cpp, 
"fd_set pollset_in  = { 0 };" should likewise be converted to use FD_ZERO.

pijyoi:
Hmm, it was commit b3d5fa63a0643afdf7de44dc1739725766ae459e that changed usage of FD_ZERO to manually zeroing out the fdset in order to silence Coverity.

On Windows, FD_ZERO merely sets the number of entries to 0.

bluca:
In some cases static analyzers don't get it right, so it's fine to override them if you think it's the best way

PR #2534, opened 2017-04-14
Problem: Minor inconsistency in macro syntax. Solution: Code fixed.

bjovke:


src/proxy.cpp
```diff
@@ -69,7 +69,7 @@
         delete poller_both_blocked;\
         delete poller_frontend_only;\
         delete poller_backend_only;\
-    } while (0)
+    } while (true)
```
bluca: surely you meant while (false) ?

bjovke: @bluca Actually yes.....

bjovke: I can't believe I did this...

bjovke:
@bluca Sorry. Now it should be OK. Two brains are better than one.

bluca:
no worries

Issue #2531, opened 2017-04-13
ZMQ_GSSAPI_PRINCIPAL sockopt has no effect on client side


garlick:
I am trying to get GSSAPI security working with my application, and indeed seem have it going with libzmq 4.1.4 and czmq 3.0.2.  (My compliments to the author for getting this integrated with the libzmq security framework!)  However, this one quirk puzzles me:

Setting ZMQ_GSSAPI_PRINCIPAL (e.g. with czmq 3.0.2 `zsock_set_gssapi_principal()`  seems not to matter on the client end.  I can set it to a random string, or not set it all, and authentication still works using the default principal, presumably figured out by the underlying libkrb5.   It does not seem possible to influence the choice of principal, even when several are available in the credential cache.

This would seem contrary to how  [@cbusbey's gist](https://gist.github.com/cbusbey/11265987) indicates it should work.

If anyone else is working with GSSAPI perhaps they can confirm these results, or explain why this is OK.  Meanwhile I will try to put together a simple reproducer.

TL;DR downstream issue is flux-framework/flux-core#758

bluca:
I can't say I'm familiar with GSSAPI, but the only thing that ZMQ_GSSAPI_PRINCIPAL variable seem to be used for is this call to acquire_credential:

https://github.com/zeromq/libzmq/blob/master/src/gssapi_client.cpp#L63

Does it ring a bell?

garlick:
Maybe, though I'm also not too familiar.  Still I'll have a go :-)

According to [zmq_gssapi(7)](http://api.zeromq.org/4-2:zmq-gssapi) ZMQ_GSSAPI_PRINCIPAL sockopt is supposed to be set on both client and server side. `acquire_credential()` definitely works correctly on the server side to establish the service principal that must match the client's service principal (set with a different sockopt).

On the client side, the "service_name" arg to `acquire_credential()` is the client principal yet the GSS_C_NT_HOSTBASED_SERVICE flag is unconditionally set in the name provided to the second arg of `gss_acquire_cred()`.

According to https://web.mit.edu/kerberos/krb5-1.12/doc/appdev/gssapi.html

> If the desired name for a krb5 initiator credential is a host-based name, it is converted to a principal name of the form service/hostname in the local realm, where hostname is the local hostname if not specified. 

Hmm, maybe that's causing `gss_acquire_cred()` to fail?

[...]

PR #2541, opened 2017-04-20
fix misc. bugs in GSSAPI support

garlick:
As discussed in #2531, the ZMQ_GSSAPI_PRINCIPAL socket option cannot be used on the client side due to an incorrect flag in the call to `gss_acquire_cred()` which makes a successfully acquired cred unusable in `gss_init_sec_context()`.  This probably went unnoticed because client creds are acquired with the GSS_C_NT_HOSTBASED_SERVICE flag which would fail for a user principal.  Failure to acquire a credential is not treated as a fatal error - the code proceeds to call `gss_init_sec_context()` with a NULL cred, which indicates that the default credentials should be used.   This PR fixes the incorrect flag, and changes the logic to abort authentication if credentials cannot be acquired for the desired principal.

The code is written so that not setting ZMQ_GSSAPI_PRINCIPAL on either client or server causes default credentials to be used, and in fact this may be desirable.  On the client side, not specifying it is the only way to use a user principal, and on the server side, according to https://web.mit.edu/kerberos/krb5-1.13/doc/appdev/gssapi.html
> The simplest choice is to pass GSS_C_NO_CREDENTIAL as the acceptor credential. In this case, clients may authenticate to any service principal in the default keytab (typically DEFKTNAME, or the value of the KRB5_KTNAME environment variable). This is the recommended approach if the server application has no specific requirements to the contrary.

The zmq_gssapi(7) manual page is updated to indicate that ZMQ_GSSAPI_PRINCIPAL is optional.

Finally, configure.ac is updated to fix an missing AC_DEFINE that was preventing the GSSAPI support from being compiled when `--with-libgssapi_krb5` is configured.

I've tested GSSAPI with these fixes in two scenarios:

1) My user principal authenticating to a service principal (keytab), with ZMQ_GSSAPI_PRINCIPAL not set on client, and both with and without ZMQ_GSSAPI_PRINCIPAL option set on server.

2) Service principal (keytab) to service principal (keytab), with and without ZMQ_GSSAPI_PRINCIPAL option set on client and server.  (client and server can use same or different principals).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -427,7 +427,8 @@ AC_ARG_WITH([libgssapi_krb5], [AS_HELP_STRING([--with-libgssapi_krb5],
 # conditionally require libgssapi_krb5
 if test "x$require_libgssapi_krb5_ext" != "xno"; then
     AC_CHECK_HEADERS(gssapi/gssapi_generic.h)
-    AC_SEARCH_LIBS([gss_init_sec_context], [gssapi_krb5 gssapi],,
+    AC_SEARCH_LIBS([gss_init_sec_context], [gssapi_krb5 gssapi],
+        AC_DEFINE(HAVE_LIBGSSAPI_KRB5, [1], [Enabled GSSAPI security]),
         AC_MSG_ERROR(libgssapi_krb5 is needed for GSSAPI security))
 fi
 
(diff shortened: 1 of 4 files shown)


bluca:
Thanks.

Is reverting to the default credential bad from a security point of view? If so, could you please backport to zeromq4-1 which is also affected?

garlick:
(sorry for the multiple edits of my description while you were possibly reading it!)

I don't _think_ it's necessarily a security problem, more of a usability one.  The client's principal still has to be valid (even if it's not the intended one), and the server still has a to allow that principal to connect (via ZAP).  The connection is still secure.

bluca:
Ok, thanks

PR #2544, opened 2017-04-21
RELICENSE: Hendrik Beskow grant 

Asmod4n:
Problem: no relicense grant for Hendrik Beskow

Solution: add relicense grant


c-rack:
Thanks!

Issue #2542, opened 2017-04-20
GSSAPI only allows host-based principals to be explicitly selected


garlick:
Principals specified with ZMQ_GSSAPI_PRINCIPAL (client and server), as well as ZMQ_GSSAPI_SERVICE_PRINCIPAL (client) are looked up with the GSS_C_NT_HOSTBASED_SERVICE name type.

As described in  https://web.mit.edu/kerberos/krb5-1.13/doc/appdev/gssapi.html, this means principals must be specified as "service" or "service@host", and are internally canonicalized (at least in MIT krb5) to principals named "service/host" in the local realm, with host set to the local hostname if unspecified.

This means that a principal that is not tied to a host (for example an "application principal") cannot be used with zeromq, unless it is the default credential on the client side (used when the client does not set ZMQ_GSSAPI_PRINCIPAL).  Maybe there is a good reason not to allow that, and in fact @fichtner may be asserting this in
> commit 8c09ae6e49b29a731dc4f0b123101baecdace113
> Author: Franco Fichtner <franco@lastsummer.de>
> Date:   Mon May 5 21:59:11 2014 +0200
> 
>    gssapi: RFC 2744 mandates GSS_C_NT_HOSTBASED_SERVICE
>   
>   Tested on FreeBSD 10.

(I can't seem to find the passage he is referring to in [RFC 2744](https://tools.ietf.org/html/rfc2744))

It would seem like a useful thing to allow principals to be optionally looked up with the GSS_C_NT_USER_NAME name type.  I'm not sure of the right way to do this though.

Option 1:  add two new socket options (true by default for backwards compat), e.g. 
* ZMQ_GSSAPI_PRINCIPAL_NT_HOST_BASED
* ZMQ_GSSAPI_SERVICE_PRINCIPAL_NT_HOST_BASED

Option 2:  special prefix character for princpals, e.g. if `@` is the first character, strip it off and look up the name with GSS_C_NT_USER_NAME

bluca:
New options are easier to document imho

PR #2545, opened 2017-04-21
add GSSAPI NAMETYPE socket options

garlick:
Add two new socket options as discussed in #2542 that allow the GSSAPI name type to be changed from "host type" to "user name" or "krb5 principal".

There is a description of the various name types from the MIT kerberos perspective at https://web.mit.edu/kerberos/krb5-1.12/doc/appdev/gssapi.html

The upshot is it allows zeromq to be a bit more flexible with respect to the type of principals used for GSSAPI authentication.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -21,7 +21,8 @@ MAN3 = zmq_bind.3 zmq_unbind.3 zmq_connect.3 zmq_disconnect.3 zmq_close.3 \
     zmq_atomic_counter_value.3 zmq_atomic_counter_destroy.3
 
 MAN7 = zmq.7 zmq_tcp.7 zmq_pgm.7 zmq_inproc.7 zmq_ipc.7 \
-    zmq_null.7 zmq_plain.7 zmq_curve.7 zmq_tipc.7 zmq_vmci.7 zmq_udp.7
+    zmq_null.7 zmq_plain.7 zmq_curve.7 zmq_tipc.7 zmq_vmci.7 zmq_udp.7 \
+    zmq_gssapi.7
 
 MAN_DOC =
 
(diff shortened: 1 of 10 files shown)


garlick:
This appveyor failure seems unrelated to the changes in this pr, in fact most of it is probably compiled out.  I'm not keeping track of what tests might spuriously fail in the CI here.  Could this be one?
```
test 7
      Start  7: test_hwm
7: Test command: C:\projects\build_libzmq\bin\Release\test_hwm.exe
7: Test timeout computed to be: 10
7: Assertion failed: send_count == recv_count, file C:\projects\libzmq\tests\test_hwm.cpp, line 134
 7/71 Test  #7: test_hwm .........................***Exception: Other  0.99 sec
```

bluca:
Yeah don't worry about that one.

Could you please:

1) Move the new options in the DRAFT section (and add them to src/zmq_drafts.h as well)? We add new options as draft initially so that they can be changed until stabilized
2) Add the new options to zmq_getsockopt as well
3) Document them in the doc/zmq_[g|s]etsockopt manpage

bluca:
@garlick ping

garlick:
Sure, I'm away from my test environment right now - monday OK?

bluca:
Of course, no rush, thanks

bluca:
@garlick these new APIs have been in DRAFT state for almost a year and a public release, so it's time to declare them as STABLE:

```

/*  GSSAPI principal name types                                               */
#define ZMQ_GSSAPI_NT_HOSTBASED 0
#define ZMQ_GSSAPI_NT_USER_NAME 1
#define ZMQ_GSSAPI_NT_KRB5_PRINCIPAL 2
```

Are you happy with that? It means that they will not be allowed to change anymore

garlick:
I think it's probably OK, thanks for checking.

PR #2547, opened 2017-04-24
Targets broken by some parts of f7d1c159d4f96686ce994319280c1bd3a2eab994

diorcety:
Superfluous modifications which occur issues with target names


bluca:
@diorcety @RPGillespie6 @stefanherbrechtsmeier this has gone back and forth a few times since 4.2.2, all 3 of you made changes related, please double check this latest version and make sure it works for everyone so that we can have an agreement

RPGillespie6:
Still working for me. On a side note though, shouldn't we use a CMake variable at this point to hold the target names? I mean, when you are going though and changing 20+ instances of `zmq-shared` to `libzmq` ...

diorcety:
The issue is in https://github.com/zeromq/libzmq/blob/master/ZeroMQConfig.cmake.in 
the target names was not changed
The zmq-shared changes was not neccesary to fix the "liblib" issue on msys2 env

herbrechtsmeier:
The target names must be fixed because they are exported to CMake configuration file and used by other projects. Therefore the project must agree on fixed targets / names. At the moment the project uses different names:
ZeroMQ: project name; CMake package name and variables
libzmq: pkg-config name; CMake targets
zmq: install directory under share; CMake option prefix

Regarding the shared library I think that it makes only sense to append a `–shared` if we want to add a dynamic alias target to the shared or static library. Additionally the pkg-config file needs to be adapted if we make the shared library optional. Instead of two build options the project should respect the `BUILD_SHARED_LIBS` variable and only use an additional `ZMQ_ENABLE_STATIC` option.

PR #2548, opened 2017-04-24
move GSSAPI NAMETYPE options to DRAFT, etc.

garlick:
Per feedback in PR #2545, this PR
* moves new GSSAPI options to DRAFT section of header
* adds the options to `zmq_getsockopt()`
* documents the options in `zmq_getsockopt(3)`, `zmq_setsockopt(3)` man pages

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -368,13 +368,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
 #define ZMQ_VMCI_BUFFER_MAX_SIZE 87
 #define ZMQ_VMCI_CONNECT_TIMEOUT 88
 #define ZMQ_USE_FD 89
-#define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
-#define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
-
-/*  GSSAPI principal name types                                               */
-#define ZMQ_GSSAPI_NT_HOSTBASED 0
-#define ZMQ_GSSAPI_NT_USER_NAME 1
-#define ZMQ_GSSAPI_NT_KRB5_PRINCIPAL 2
 
 /*  Message options                                                           */
 #define ZMQ_MORE 1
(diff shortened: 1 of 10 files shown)


bluca:
Thanks! A couple more comments, sorry

include/zmq.h
```diff
@@ -638,6 +631,18 @@ ZMQ_EXPORT int   zmq_timers_reset (void *timers, int timer_id);
 ZMQ_EXPORT long  zmq_timers_timeout (void *timers);
 ZMQ_EXPORT int   zmq_timers_execute (void *timers);
 
+/******************************************************************************/
+/*  GSSAPI socket options to set name type                                    */
+/******************************************************************************/
+
+#define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 1090
```
bluca: I think it's better to have the definitive value immediately (90, 91) so that when it's declared stable folks who are already using it with DRAFT flag will not have any breakage, it will just keep working - a bit less disruption

src/gssapi_client.cpp
```diff
@@ -51,8 +51,12 @@ zmq::gssapi_client_t::gssapi_client_t (const options_t &options_) :
     service_name = static_cast <char *>(malloc(service_size+1));
     assert(service_name);
     memcpy(service_name, options_.gss_service_principal.c_str(), service_size+1 );
-    service_name_type = convert_nametype (options_.gss_service_principal_nt);
 
+#ifdef ZMQ_BUILD_DRAFT_API
```
bluca: The definitions are available internally, so you can remove all te \#ifdefs in the internal code - this is the reason we added src/zmq_drafts.h, so that internal code would be easier to write and maintain without lots and lots of ifdefs.
If users manually inspect the code and find out the magic number and use it it's their problem anyway - we guarantee a stable API, not safety from silly hacks

garlick:
Makes sense!  Another PR coming shortly.  By the way, you don't need to merge my PR's to keep me interested :-)  I can address feedback and squash before merging to reduce history noise if that's preferable.

bluca:
As you prefer, it's fine to merge things if they are not completely broken and do refactors afterwards, it's our policy, but the sender's choice matters more so next time I'll comment inline

PR #2550, opened 2017-04-26
add simple GSSAPI test for make check

garlick:
Since there is no test coverage for GSSAPI, I propose here a simple test patterned after the CURVE test.  It's not comprehensive, doesn't test the new socket options, and requires a kerberos environment containing a specific test principal.  If environment variables pointing to a test keytab file are not set, the test is skipped (also if gssapi was not configured).

In the future maybe we can find a way to build a standalone kerberos environment just for CI testing, but for now, better to have _something_?

### DIFF
--- a/tests/test_security_gssapi.cpp
+++ b/tests/test_security_gssapi.cpp
@@ -1,0 +1,325 @@
+/*
+    Copyright (c) 2007-2017 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


tests/test_security_gssapi.cpp
```diff
@@ -0,0 +1,303 @@
...
+        free (identity);
+        free (mechanism);
+        free (principal);
+    }
+    zmq_close (handler);
+}
+
+int main (void)
+{
+#ifndef HAVE_LIBGSSAPI_KRB5
```
bluca: Instead of having the IFDEF here it can be in the makefile.am directly, like other tests are doing, eg tipc. This way it's skipped and doesn't risk a false positive.

tests/test_security_gssapi.cpp
```diff
@@ -0,0 +1,303 @@
...
+//  variables KRB5_KTNAME and KRB5_CLIENT_KTNAME to FILE:/path/to/your/keytab.
+//  The test will use it both for client and server roles.
+//
+//  The test is derived in large part from test_security_curve.cpp
+
+const char *name = "zmqtest2";
+
+static volatile int zap_deny_all = 0;
+
+#ifdef ZMQ_BUILD_DRAFT_API
```
bluca: I don't this this ifdef is needed?

bluca:
Having more tests is definitely good, thanks.

The CI runs in Ubuntu 12.04 or 14.04, we can install any package we want on it. If you have a set of commands to run to configure Kerberos on Ubuntu we can easily add it.

garlick:
Thanks for the feedback @bluca.  I tjust pushed some changes that hopefully address your comments.  The last commit is an unsolicited cleanup of the long test `main()`, breaking each connection attempt out to separate functions.

Let me know when you're happy and I'll squash down the incremental development in this PR.

bluca:
Looks good to me, thanks!

bluca:
Regarding the CI, do you think it's possible to have a working kerberos instance on a single Ubuntu 14.04 machine?

If so it should be pretty easy to set up another Travis run for it. I can help if you are not familiar with it.

garlick:
Just forced a push with changes squashed.   I also snuck in a change to set the principal on the client side (rather than letting krb5 infer it from the credential cache or keytab) and set the principal name type to the default value on both client and server, just to exercise the new socket option.

Will look into the KDC in travis but I will need to get back to that later and will definitely ask for help if I get stuck.  Thanks!

bluca:
Great, thanks

PR #2553, opened 2017-04-28
Fix WITH_DOC

### LINKED ISSUE #2552: ASCIIDOC_EXECUTABLE used even if not found
if not ASCIIDOC_FOUND then WITH_DOC is forced to false
but the command using ASCIIDOC_EXECUTABLE is still defined. 
it should be conditionnal to WITH_DOC

cf https://github.com/diorcety/libzmq/commit/fcf992a7e73850d9cc9aa4ca782385cb8cf7ef43

diorcety:
Problem: Even if WITH_DOC is false the ASCIIDOC_EXECUTABLE is called, leading to and error.
Solution: Make the ASCIIDOC_EXECUTABLE command conditionnal to the WITH_DOC variable

Fix #2552

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -674,19 +674,19 @@ foreach (txt ${docs})
   string (REGEX REPLACE ".*/(.*)\\.txt" "\\1.html" html ${txt})
   set (src ${txt})
   set (dst doc/${html})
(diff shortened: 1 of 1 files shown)


bluca:
thanks

PR #2560, opened 2017-05-01
Problems: cannot use wildcard port with source address, tests bind to hardcoded TCP ports, ASAN CI run could use some improvements

bluca:
Solutions: see commits

The diff is quite large, but it does the same thing to most tests: instead of binding to an hardcoded TCP port or IPC file, uses wildcards and ZMQ_LAST_ENDPOINT.
This allows running tests in parallel, and removes a great source of false negatives on overbooked machines.
Ubuntu has been carrying a patch implementing a form of this change, I've taken it and made several modifications. Thanks to Steve Langasek for the original patch in Ubuntu.
A handful of tests still bind to an hardcoded port to test a few more codepaths, but they are now listed in testutil.hpp so that we can make sure they are at least all different.
Finally some improvements for the address sanitizer run.


bluca:
geh, MSVC problems, will fix

bluca:
Let's see if it's happy now...

bluca:
All good. It's absurd it took Microsoft so long to implement something as basic as snprintf...

bluca:
Ping!

PR #2572, opened 2017-05-10
Problem: REP leaves label msgs for dead REQ in pipe

### LINKED ISSUE #2567: Memory leak in REP socket handling
Problem happens in a REP server, when a REQ client disconnects before the response is sent. According to valgrind, about 500B leaks per response on my test setup. I've discovered the problem in ZMQ 4.2.1, then confirmed also in master and 4.1.6.
The app I'm working on is a simple REQ/REP server, working on Linux, using tcp/ipc protocols. It occasionally requires significant amount of time to process the request, due to external hardware problems, during which our client timeouts and leaves the server with small, but growing leak.

I've traced the problem to a partial message being left in a pipe:
1. When a request is received, the rep_t::xrecv() calls router_t::xsend() to copy all the labels to response message.
2. The rest of message is passed to user's app, which works hard to prepare the response
3. In the meantime, the client goes away
4. The response is sent using rep_t::xsend(), which calls router_t::xsend(), which correctly notices the pipe had gone away and frees the response (thanks to fix from #1313), however the labels already put into the pipe in step "1" are left there and never freed.

I suppose the session_base_t::clean_pipes() should remove them in step "3", however it does not happen for some reason... Unfortunately, I don't have enough time to get familiar with libzmq code, so I cannot came with a patch at hand... But I would gratefully test one.

Please see attached valgrind output, and a minimal server + client test apps for reproducing the issue (my real server/client uses czmq 4 and pyzmq 14.x, however that should not matter much). The sample server is hardcoded with "processing" time (just a sleep) set to 1s, and the sample client has RCVTIMEO set to 0.5s. With such values I get the memleak every request. When receive timeout is raised enough that client reads the response, the problem stops occurring (obviously).

[libzmq_valgrind.txt](https://github.com/zeromq/libzmq/files/986552/libzmq_valgrind.txt)
[testapps.zip](https://github.com/zeromq/libzmq/files/986558/testapps.zip)

bluca:
Solution: roll back the pipe if writing messages other than the
first fails in router::xsend.
Also add test case that reproduces the memory leak when ran with
valgrind.
Fixes #2567

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -273,6 +273,9 @@ int zmq::router_t::xsend (msg_t *msg_)
             // Message failed to send - we must close it ourselves.
             int rc = msg_->close ();
             errno_assert (rc == 0);
+            // HWM was checked before, so the pipe must be gone. Roll back
+            // messages that were piped, for example REP labels.
+            current_out->rollback ();
             current_out = NULL;
         } else {
           if (!more_out) {
--- a/tests/test_reqrep_ipc.cpp
+++ b/tests/test_reqrep_ipc.cpp
@@ -29,9 +29,48 @@
 
 #include "testutil.hpp"
 
-int main (void)
+void test_leak (void)
 {
-    setup_test_environment();
+    char my_endpoint[256];
+    void *ctx = zmq_ctx_new ();
+    assert (ctx);
+
+    void *sb = zmq_socket (ctx, ZMQ_REP);
+    assert (sb);
+    int rc = zmq_bind (sb, "ipc://*");
+    assert (rc == 0);
+    size_t len = sizeof(my_endpoint);
+    rc = zmq_getsockopt (sb, ZMQ_LAST_ENDPOINT, my_endpoint, &len);
+    assert (rc == 0);
+
+    void *sc = zmq_socket (ctx, ZMQ_REQ);
+    assert (sc);
+    rc = zmq_connect (sc, my_endpoint);
+    assert (rc == 0);
+
+    rc = s_send (sc, "leakymsg");
+    assert (rc == strlen ("leakymsg"));
+
+    char *buf = s_recv (sb);
+    free (buf);
+
+    rc = zmq_close (sc);
+    assert (rc == 0);
+
+    msleep (SETTLE_TIME);
+
+    rc = s_send (sb, "leakymsg");
+    assert (rc == strlen ("leakymsg"));
+
+    rc = zmq_close (sb);
+    assert (rc == 0);
+
+    rc = zmq_ctx_term (ctx);
+    assert (rc == 0);
+}
+
+void test_simple (void)
+{
     char my_endpoint[256];
     void *ctx = zmq_ctx_new ();
     assert (ctx);
@@ -59,6 +98,15 @@ int main (void)
 
     rc = zmq_ctx_term (ctx);
     assert (rc == 0);
+}
+
+int main (void)
+{
+    setup_test_environment();
+
+    test_simple ();
+
+    test_leak ();
 
     return 0 ;
 }


bluca:
@somdoron I think this is the right thing to do, but I could be missing something, could you please have a quick look when you have time?

somdoron:
I think the fix is incorrect,  take a look at the pipe terminated method,  current out will be null and all msg will be dropped (future and past)

bluca:
Thanks, will look again

bjovke:
@somdoron `current_out = NULL;` happens after `current_out->rollback ();` with this fix anyway.
The issue could be only that `rollback()` removes all messages from pipe, I really don't know if this is correct behavior?
"_all msg will be dropped (future and past)_" `current_out = NULL;` existed before this patch. What exactly do you mean by this?

somdoron:
Ok Sorry, I'm surprised we only find out about this now. 
We also need to call the rollback in the xpipe_terminated (or only there and remove the current_out=null from the xsend, letting the xpipe_termianted handle it)

Also calling rollback in the router version is a mistake because it set the more_out to false, which is not what we want here (because it is a silent drop, setting more_out to false will cause an error on the next part, which the router will think is a routing id).

So I would just call outpipe.rollback() in both places.

bjovke:
What I've discovered in my test:

- When message is received by `ZMQ_REP` server, `zmq::stream_engine_t::in_event ()` is called, which eventually calls `zmq::metadata_t::add_ref ()` for metadata of message.
- `ref_cnt` of metadata grows to 4.
- Client is killed before server tries to respond.
- `ref_cnt` of metadata is decreased to 1 after whole cleanup has ended and `metadata `is never freed.
- With patch from @bluca ref_cnt correctly decreases to 0 and `LIBZMQ_DELETE(metadata);` is called in `zmq::stream_engine_t` destructor.
- This correctly corresponds to something left referencing `metadata`, probably one message which is not destroyed.

bjovke:
I'm also not sure why there's `current_out = NULL;` `zmq::router_t::xsend`, `zmq::router_t::rollback` and `zmq::router_t::xpipe_terminated` both set this to `NULL `in a more appropriate way.

bluca:
Thanks for the comments&insights

As far as I can see, setting current_out to NULL in xsend has been around since forever, I can trace it back to before router.cpp existed, in xrep: https://github.com/zeromq/libzmq/commit/b79d07b8bc844135c44c1ff8b84b50dc08f56925

> Also calling rollback in the router version is a mistake because it set the more_out to false, which is not what we want here (because it is a silent drop, setting more_out to false will cause an error on the next part, which the router will think is a routing id).

Note that I intentionally called outpipe's rollback rather than router for this reason, already in this diff.

So if I followed correctly, the proposal is to resend this same PR, with the addition of calling pipe_->rollback() in xpipe_terminated ?

somdoron:
Yes

On May 17, 2017 12:23 AM, "Luca Boccassi" <notifications@github.com> wrote:

Thanks for the comments&insights

As far as I can see, setting current_out to NULL in xsend has been around
since forever, I can trace it back to before router.cpp existed, in xrep:
b79d07b
<https://github.com/zeromq/libzmq/commit/b79d07b8bc844135c44c1ff8b84b50dc08f56925>

Also calling rollback in the router version is a mistake because it set the
more_out to false, which is not what we want here (because it is a silent
drop, setting more_out to false will cause an error on the next part, which
the router will think is a routing id).

Note that I intentionally called outpipe's rollback rather than router for
this reason, already in this diff.

So if I followed correctly, the proposal is to resend this same PR, with
the addition of calling pipe_->rollback() in xpipe_terminated ?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/zeromq/libzmq/pull/2572#issuecomment-301920021>, or mute
the thread

bluca:
Here it is: https://github.com/zeromq/libzmq/pull/2577

PR #2580, opened 2017-05-18
Use OBJECT_DEPENDS and OBJECT_OUTPUTS for precompiled.hpp

diorcety:
This fix, allows to have a correct dependency tree for precompiled header.
The issue is triggered when using ninja build system, which doesn't take the source files in the given order (fixed for make in #2294).
This allow to use ninja for building libzmq on windows (on linux it works as it is)


bluca:
Thanks!

PR #2596, opened 2017-06-13
Problems: ZAP can be set up incorrectly, CURVE and GSSAPI can be used only with ZAP

### LINKED ISSUE #2594: Problem: CurveZMQ connection fails without ZAP
In commit a0ccdc866efbac5889d8872c7f27c53e9c0a7c5b from #2482 the logic of `curve_server.cpp` was changed so that `zmq::curve_server_t::process_initiate` returns -1 if `zap_connect` fails.  Previously, this was allowed and the ZAP logic was only processed if that returned 0.  Was this deliberate?  Is ZAP mandatory now?

bluca:
Solutions: see commits

Fixes #2594

### DIFF
--- a/src/curve_server.cpp
+++ b/src/curve_server.cpp
@@ -494,21 +494,23 @@ int zmq::curve_server_t::process_initiate (msg_t *msg_)
     //  encryption without authentication), but if it was requested and it does
     //  not work properly the program will abort.
     rc = session->zap_connect ();
-    if (rc != 0)
-        return -1;
-    rc = send_zap_request (client_key);
-    if (rc != 0)
-        return -1;
-    rc = receive_and_process_zap_reply ();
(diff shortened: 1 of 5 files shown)


bluca:
@evoskuil please let me know if this looks good to you - thanks!

bluca:
Ping :-)

bjovke:
@bluca I can't help you here, I'm not familiar how this code works...

evoskuil:
Thanks for the ping. I'm traveling for the next couple weeks and I can't give a very good review of this with just my iphone. The principle is sound but just looking it over in the browser it is hard to say for sure.

bluca:
Thanks. Then I would recommend to go ahead with the changes to unblock folks using curve, and then we can revisit and revise if necessary.

Issue #2602, opened 2017-06-21
intermittent memory leak for req/rep send/recv.


Matthew-Jemielity:
Hi,

While unit testing my project using zmq I found randomly occuring memory leak for a REQ/REP socket pair doing zmq_send{,_const} and zmq_recv. I'm attaching an almost-minimal self-contained program which reproduces this issue:
[zmq_recv_req_rep.txt](https://github.com/zeromq/libzmq/files/1092725/zmq_recv_req_rep.txt)
Hopefully I'm using libzmq correctly.
There seems to be some kind of race condition involved, since checking with valgrind can have four different outcomes:
- no error (often)
- memleak only (very rare)
- invalid read + write only (sometimes)
- invalid read + write and memleak (rare)

The errors seem easier to reproduce on 32-bit system.

ZMQ I'm using was compiled from source, commit 2d83acceffdef71c62c936487db9478b5478a00c , through `./autogen.sh && ./configure --prefix=/usr && make && make check && sudo make install` .
Tested on systems:
1. Ubuntu 14.04.05 64-bit: `4.2.0-42-lowlatency #49~14.04.1-Ubuntu SMP PREEMPT Wed Jun 29 21:21:58 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux, gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4, valgrind-3.10.1`
2. qemu vm with Ubuntu 14.04.05 32-bit: `4.4.0-31-generic #50~14.04.1-Ubuntu SMP Wed Jul 13 01:06:37 UTC 2016 i686 i686 i686 GNU/Linux, gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4, valgrind-3.10.1`

Source was compiled with:
`gcc -Wall -Wextra -std=c99 -pedantic zmq_recv_req_rep.c -lzmq -pthread -o test`
and checked with:
`valgrind --leak-check=full ./test`

valigrind message:
```
==18183== Memcheck, a memory error detector
==18183== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==18183== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==18183== Command: ./test
==18183== 
==18183== Thread 2:
==18183== Invalid write of size 1
==18183==    at 0x4E6D239: zmq::pipe_t::terminate(bool) (pipe.cpp:393)
[...]

PR #2609, opened 2017-06-27
Problem: intermittent memory leak for req/rep send/recv. #2602 Solution: memory leak fixed.

bjovke:


bjovke:
Solution for issue #2602.

Issue #2612, opened 2017-06-28
Don't undef _WIN32_WINNT


diorcety:
If the user set an _WIN32_WINNT definition we should avoid to change it.
https://github.com/zeromq/libzmq/blob/master/src/windows.hpp#L54
https://github.com/zeromq/libzmq/blob/master/include/zmq.h#L71
An error message is maybe better.

bjovke:
@diorcety I'm not sure that I can see the issue.
`_WIN32_WINNT` will be defined/redefined only if it's already not defined or if features require some minimum Windows version:

- `ZMQ_HAVE_WINDOWS_UWP` defined -> minimum Windows 10.
- `__MINGW32__` defined -> if `_WIN32_WINNT` is less than `0x0600` redefine it to `0x0600`.

This code maybe needs some rearrangement to be more readable but it's nothing critical.

```
//  Set target version to Windows Server 2008, Windows Vista or higher.
//  Windows XP (0x0501) is supported but without client & server socket types.
#if !defined _WIN32_WINNT && !defined ZMQ_HAVE_WINDOWS_UWP
#define _WIN32_WINNT 0x0600
#endif

#if defined ZMQ_HAVE_WINDOWS_UWP
#define _WIN32_WINNT _WIN32_WINNT_WIN10 
#endif

#ifdef __MINGW32__
//  Require Windows XP or higher with MinGW for getaddrinfo().
#if(_WIN32_WINNT >= 0x0600)
#else
#undef _WIN32_WINNT
#define _WIN32_WINNT 0x0600
#endif
#endif
```

diorcety:
If the user define 0x0501 i don't see the the point to undef it. The issue behind is a preprocessor nightmare that I had. I have 0x0501 as value, I include one header which include windows.h, the vista functions are not defined. I include an another header which include libzmq, the value is changed to 0x0600. I include an other header which use _WIN32_WINNT do modify some code using vista functions, which trigger errors.

bjovke:
@diorcety To be honest I'm not sure why for MinGW Vista is minimum Windows version (0x0600). Even the comment is confusing:
```
//  Require Windows XP or higher with MinGW for getaddrinfo().
#if(_WIN32_WINNT >= 0x0600)
```
This is related to #1381 and #1383 and it seems that this part with MinGW is skipped.
I will make a PR with some changes for this.

bjovke:
[...]

PR #2614, opened 2017-06-30
[WIP] Problem: _WIN32_WINNT macro issues when compiling on Windows XP and MinGW, issue #2612. Solution: Do not undefine _WIN32_WINNT if user has defined it.

bjovke:


bjovke:
Solution for #2612.

bjovke:
Tests won't compile with this change, it has to be further investigated.

PR #2619, opened 2017-07-07
Partial Windows XP support

diorcety:
PR for bug #2612

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -66,10 +66,9 @@ extern "C" {
 
 #ifdef __MINGW32__
 //  Require Windows XP or higher with MinGW for getaddrinfo().
-#if(_WIN32_WINNT >= 0x0600)
+#if(_WIN32_WINNT >= 0x0501)
 #else
-#undef _WIN32_WINNT
-#define _WIN32_WINNT 0x0600
+#error You need at least Windows XP target
 #endif
 #endif
 #include <winsock2.h>
--- a/src/condition_variable.hpp
+++ b/src/condition_variable.hpp
@@ -39,9 +39,22 @@
 #ifdef ZMQ_HAVE_WINDOWS
 
 #include "windows.hpp"
+#if defined(_MSC_VER)
+#if _MSC_VER >= 1800
+#define _SUPPORT_CONDITION_VARIABLE 1
+#else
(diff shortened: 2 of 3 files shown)


somdoron:
Wont this make new socket types unsusable on windows?

bluca:
Yes, all new sockets test cases are failing

somdoron:
I think it is better to leave the default as windows vista. If someone one to compile for windows xp they can set the WINNT flag.

diorcety:
I will try to resolve that. I'am a bit busy now

diorcety:
Every thing seems good now. MSVC doesn't set correctly _cplusplus. Using _MSC_VER we can enable or not the conditional_variable usage or not.

bjovke:
@diorcety I'm not sure this is OK. It changes default Windows version from Vista to XP for the whole library, breaking the compatibility.
The offending part of the code is this one:
```
#undef _WIN32_WINNT
#define _WIN32_WINNT 0x0600
```
In my opinion, only this should be changed. If user has defined `_WIN32_WINNT` and it is XP or higher (Windows 10 in case of UWP), then it should not be undefined and set to different value. This means that user must explicitly enable XP support while default behavior should be Vista and Windows 10 in case of UWP.

bjovke:
@diorcety Generally this PR is ok, [this](https://github.com/zeromq/libzmq/pull/2619/commits/fa0f3889fa18b8e8a65772bafb899e423b6c7e41#diff-e81114d231ae71d13f16d1b6970abb37R64) and [this](https://github.com/zeromq/libzmq/pull/2619/commits/fa0f3889fa18b8e8a65772bafb899e423b6c7e41#diff-7af18dafe332fc0e0aee2c444d29231eR43) are the only offending lines. It should be changed to `0x0600`.

diorcety:
Done

bluca:
@diorcety could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

diorcety:
@bluca Done with #3666

PR #2622, opened 2017-07-14
Fix ROUTER's xhas_out() in MANDATORY mode

msune:
Before this commit, xhas_out() was returning true regardless. This
was correct before the ZMQ_ROUTER_MANDATORY flag as introduced.
However, ZMQ_POLLOUT.

With this commit, _if_ ZMQ_ROUTER_MANDATORY is set, xhas_out() will
return false if ALL peer's outgoing pipes are full.

There is an outstanding high-level design question:

If ZMQ_ROUTER_MANDATORY is set, and zmq_poll() waits for ZMQ_POLLOUT
events, zmq_poll() will immediately wake up if only 1 pipe has
room to send, regardless of the peer, creating a busy loop of
zmq_poll() wake-up, zmq_send() (EAGAIN). There is no way for
the application to selectively wait for ZMQ_POLLOUT for specific
peer(s), which seems somehow necessary in ZMQ_ROUTER_MANDATORY.

This discussion will be addressed in a separate issue.

Signed-off-by: Marc Sune <marc@voltanet.io>
Signed-off-by: Fredi Raspall <fredi@voltanet.io>

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -423,10 +423,19 @@ bool zmq::router_t::xhas_in ()
 
 bool zmq::router_t::xhas_out ()
 {
-    //  In theory, ROUTER socket is always ready for writing. Whether actual
-    //  attempt to write succeeds depends on which pipe the message is going
-    //  to be routed to.
-    return true;
+    //  In theory, ROUTER socket is always ready for writing (except when
+    //  MANDATORY is set). Whether actual attempt to write succeeds depends
+    //  on whitch pipe the message is going to be routed to.
+
+    if(!mandatory)
+        return true;
+
+    bool has_out = false;
+    outpipes_t::iterator it;
+    for (it = outpipes.begin (); it != outpipes.end (); ++it)
+        has_out |= it->second.pipe->check_hwm();
+
+    return has_out;
 }
 
 zmq::blob_t zmq::router_t::get_credential () const


bluca:
Thanks - it's a small change in behaviour but given it's only when mandatory is set, and it makes a bit more sense, I think we can let it pass

Also for your next PRs, please use the recommended format for commit messages:

```Problem: x is broken

Solution: do y and z to fix x
```

somdoron:
I also think we can accept this.  As nobody really used router for polling out as it is always true

bluca:
Yep! @msune could you please modify the relative doc files? For the MANDATORY option as a starter I'd say

msune:
@bluca Sure I can do that. Could you briefly point out which exact ones should I be modifying (sorry rather new in the community)?

@ all: I've opened this issue to follow-up on the `ZMQ_POLLOUT` issue: https://github.com/zeromq/libzmq/issues/2623. Let me know your thoughts

msune:
@bluca Also are there any chances this can be backported to v4.X?

Thanks

bluca:
The setsockopt paragraph would be the first place: https://github.com/zeromq/zeromq4-x/blob/master/doc/zmq_setsockopt.txt#L394

Also perhaps it's a good time to add a clarification on POLLOUT  bahaviour to the ROUTER zmq_socket paragraph: https://github.com/zeromq/zeromq4-x/blob/master/doc/zmq_socket.txt#L133

Regarding the older branches, feel free to cherry-pick and backport the commit, but also add another commit with a note in the NEWS file please

PR #2639, opened 2017-07-31
Problems: using Wine requires system tuning, ZMQ_BINDTODEVICE not draft

### LINKED ISSUE #2638: make_fdpair hangs trying to send large chunk of data
Faced this while using ZMQ under Wine on OS X. it just hangs on this line https://github.com/zeromq/libzmq/blob/1d58a00992fa3f2930bb8c6602db600e24a13756/src/signaler.cpp#L543 trying to send 1MB into a socket.
Quick workaround:
```
$ sudo sysctl -w net.inet.tcp.sendspace=1300000
net.inet.tcp.sendspace: 131072 -> 1300000
```
But probably it would be better to make ZMQ working regardless of system buffer size. Let me know if any further details/debugging are required.

bluca:
Solution: document it in the INSTALL file
Fixes #2638

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -368,7 +368,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
 #define ZMQ_VMCI_BUFFER_MAX_SIZE 87
 #define ZMQ_VMCI_CONNECT_TIMEOUT 88
 #define ZMQ_USE_FD 89
-#define ZMQ_BINDTODEVICE 90
 
 /*  Message options                                                           */
 #define ZMQ_MORE 1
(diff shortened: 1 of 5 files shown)


include/zmq.h
```diff
@@ -368,7 +368,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
 #define ZMQ_VMCI_BUFFER_MAX_SIZE 87
 #define ZMQ_VMCI_CONNECT_TIMEOUT 88
 #define ZMQ_USE_FD 89
-#define ZMQ_BINDTODEVICE 90
```
jimklimov: Perhaps add a comment that numbering continues elsewhere? Wouldn't an `enum` be better to avoid conflicts (e.g. two values later with same numbers)?

But even if so, this is a subject for subsequent PR :)

bluca: We intentionally avoid enums as it's way, way too easy to break ABI

Issue #2643, opened 2017-08-03
Misleading error code in case ZAP handler sends an invalid status code


sigiesec:
According to the ZAP protocol specification, the allowed status codes are 200, 300, 400 and 500. If the ZAP handler sends something like "blah" instead, this is a violation of the ZAP protocol rather than a rejected authentication request.

The error code should therefore be EPROTO rather than EACCES.

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/2646

PR #2646, opened 2017-08-03
Fix error code when ZAP handler sends an invalid status code

sigiesec:
Problem: Misleading error code in case ZAP handler sends an invalid status code (#2643)

Solution: Use EPROTO instead of EACCES error code in that case


bluca:
thanks - fixed the merge conflict via github (fancy new feature!)

PR #2647, opened 2017-08-03
Problems: too many OSX travis builds, curve test uses hard-coded TCP port

bluca:
Solutions: see commits


bluca:
That's fine don't worry, it's probably better to have an additional drafty build

PR #2648, opened 2017-08-03
Problem: no tests for ZAP handler that has an invalid protocol

sigiesec:
Solution: add tests

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -169,6 +169,9 @@ foreach(test ${tests})
   set_tests_properties(${test} PROPERTIES TIMEOUT 10)
 endforeach()
 
+#override timeout for these tests
+set_tests_properties(test_security_curve PROPERTIES TIMEOUT 20)
+
 #Check whether all tests in the current folder are present
 file(READ "${CMAKE_CURRENT_LIST_FILE}" CURRENT_LIST_FILE_CONTENT)
 file(GLOB ALL_TEST_SOURCES "test_*.cpp")
--- a/tests/test_security_curve.cpp
+++ b/tests/test_security_curve.cpp
@@ -137,7 +137,13 @@ int get_monitor_event_with_timeout (void *monitor,
 enum zap_protocol_t
 {
   zap_ok,
-  zap_wrong_version
+  // ZAP-compliant non-standard cases
+  zap_status_internal_error,
+  // ZAP protocol errors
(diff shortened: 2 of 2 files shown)


sigiesec:
@bluca the CI jobs are now frequently failing due to test timeouts. The timeout seems to be fixed at 10 seconds. I don't know where to change that. test_security_curve now contains several test cases that each need to wait some time for remaining events. Since the CI nodes are terribly slow sometimes, it would not help to reduce these times. Do you see any other feasible option than increasing the test timeout?

bluca:
I'm afraid it's testbed-specific and I really have no idea where the cloudapp stuff is.

Is it waiting for timeouts in connections?

bluca:
@sigiesec actually it's probably a settings in cmake, that you might try and change: https://cmake.org/cmake/help/v3.0/command/set_tests_properties.html

sigiesec:
Thanks for the cmake hint, will try that :)

bluca:
@sigiesec I've seen another timeout on Windows:

https://ci.appveyor.com/project/zeromq/libzmq/build/job/2qo955y3n176fl75

bump to 30?

bluca:
https://github.com/zeromq/libzmq/pull/2654

sigiesec:
Hm... hope that helps... unless there is some case where no event is emitted at all, and the test waits forever... then the test or library must be fixed

bluca:
Yeah if it still timeouts with 60 seconds it's definitely a hang

sigiesec:
:( ok, will try to fix this. unfortunately, the tests behave quite differently on my local machine, so it is hard to reproduce for me. will add some more debug output

bluca:
Yup it timed out again:

```
      Start 30: test_security_curve
30: Test command: C:\projects\build_libzmq\bin\Debug\test_security_curve.exe
30: Test timeout computed to be: 60
30/71 Test #30: test_security_curve ..............***Timeout  60.01 sec
```

Win32, debug, libsodium

https://ci.appveyor.com/project/zeromq/libzmq/build/build-1567/job/b0nei74paefqsomj

sigiesec:
I added some more debug output, hope thats sheds some light onto the cause. 

https://ci.appveyor.com/project/sigiesec/libzmq/build/build-56 is running

But I fear I cannot continue before Monday.

bluca:
no worries, it's not urgent

PR #2649, opened 2017-08-04
Problem: Message property names are duplicated

sigiesec:
Solution: Introduce defines in the (DRAFT) API, this also makes them accessible to libzmq users


bluca:
Thanks. This should be documented in zmq_msg_gets, could you please do that?

sigiesec:
Sure, see #2651

bluca:
great, thanks

PR #2650, opened 2017-08-04
Problem: sporadic failure in test_curve_security_with_null_client_credentials

sigiesec:
Solution: add workaround for EPIPE error also here

### DIFF
--- a/tests/test_security_curve.cpp
+++ b/tests/test_security_curve.cpp
@@ -417,9 +417,11 @@ void test_curve_security_with_null_client_credentials (void *ctx,
     close_zero_linger (client);
 
 #ifdef ZMQ_BUILD_DRAFT_API
-    int event = get_monitor_event (server_mon, NULL, NULL, 0);
+    int err;
+    int event = get_monitor_event (server_mon, &err, NULL, 0);
 
-    assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ZMTP);
+    assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_ZMTP
+            || (event == ZMQ_EVENT_HANDSHAKE_FAILED_NO_DETAIL && err == EPIPE));
 #endif
 }
 
@@ -436,6 +438,8 @@ void test_curve_security_with_plain_client_credentials (void *ctx, void *server)
     assert (rc == 0);
     expect_bounce_fail (server, client);
     close_zero_linger (client);
+
+    // TODO add assertion here as in test_curve_security_with_null_client_credentials
 }
 
 void test_curve_security_unauthenticated_message (char *my_endpoint,


bluca:
@sigiesec thanks for all the fixes. There are 4 TODO's in test_security_curve.cpp, what's the plan for them?

sigiesec:
> tests\test_security_curve.cpp(451):    // TODO add another event type ZMQ_EVENT_HANDSHAKE_FAILED_AUTH for this case?
> tests\test_security_curve.cpp(732):    //  TODO is this usable? EAGAIN does not appear to be an appropriate error 

I need some feedback on these. I think it would be a good idea to add another ZMQ_EVENT_HANDSHAKE_FAILED_AUTH event type, but there are no more bits left. The event type is only 16 bit.

For an internal error reported by the ZAP handler, ZMQ_EVENT_HANDSHAKE_FAILED_NO_DETAIL or ZMQ_EVENT_HANDSHAKE_FAILED_ZAP could be appropriate with a different error code, but I am not sure which one.

> tests\test_security_curve.cpp(496):    // TODO add assertion here as in test_curve_security_with_null_client_credentials

this overlaps with your changes in #2647, I wanted to wait until they are merged

> tests\test_security_curve.cpp(544):    // TODO remove code duplication

I can approach this soon.

bluca:
For the _FAILED_AUTH, wouldn't _FAILED_ZAP already cover it?

If not, the even type is actually an int so it shoud be fine to add a new one. As far as I can see it's used everywhere as a bitmask, so it should work out of the box.

ZMQ_EVENT_ALL would need to be changed though, which would be an ABI change in theory - but on the other hand, older builds would still use a correct mask corresponding to the events that were available at the time, so we could consider it compatible in that sense and avoid bumping the revision. We could keep it as-is for the time being and ponder over it.

Regarding the errno, EAGAIN wouldn't be too bad, but perhaps EACCES is more appropriate?

Finally, the other PR is now merged.

sigiesec:
My idea with _FAILED_ZAP was/is, that the ZAP handler violated the ZAP protocol, i.e. an error outside of libzmq, but on the server side.
If the ZAP handler returns status code 400, this is not a violation of the ZAP protocol, but a rejected authentication request, which is something significantly different IMHO. The cause (not really an error at all) is on the client side, which supplied credentials that are not accepted.

According to https://github.com/zeromq/libzmq/blob/master/doc/zmq_socket_monitor.txt, "The first frame contains an event number (16 bits)". And, it is also specified as a bitmask there ("The 'events' argument is a bitmask of the socket events you wish to monitor"). So, there cannot be more than 16 events. Changing this to 32 bits would break existing socket monitors.

EAGAIN is perfect for the ZAP status code 300, which indicates a *temporary* failure, but 500 indicates some unknown problem, which might be persistent.
EACCES would be confusing IMO, since this is explicitly not used to indicate a rejected authentication. EFAULT sounds better, but its description "Bad address" does not.

bluca:
> My idea with _FAILED_ZAP was/is, that the ZAP handler violated the ZAP protocol, i.e. an error outside of libzmq, but on the server side.
If the ZAP handler returns status code 400, this is not a violation of the ZAP protocol, but a rejected authentication request, which is something significantly different IMHO. The cause (not really an error at all) is on the client side, which supplied credentials that are not accepted.
>
> According to https://github.com/zeromq/libzmq/blob/master/doc/zmq_socket_monitor.txt, "The first frame contains an event number (16 bits)". And, it is also specified as a bitmask there ("The 'events' argument is a bitmask of the socket events you wish to monitor"). So, there cannot be more than 16 events. Changing this to 32 bits would break existing socket monitors.

Right I was considering the ABI, rather than the message sent back which is 2 bytes as you said.

So the only way to extend it would be to add a new context option, disabled by default, that enables a larger event type.

So for now I would suggest leaving them conflated, and document it.

> EAGAIN is perfect for the ZAP status code 300, which indicates a temporary failure, but 500 indicates some unknown problem, which might be persistent.
EACCES would be confusing IMO, since this is explicitly not used to indicate a rejected authentication. EFAULT sounds better, but its description "Bad address" does not.

The description is often not fully appropriate, so EFAULT would be ok, again as long as it's documented.

sigiesec:
Ok, the last 2 TODOs mentioned are now resolved by #2652 

Will work on the others.

bluca:
Great stuff, thank you!

bluca:
@sigiesec  new failure on Windows:

```
30: Test command: C:\projects\build_libzmq\bin\Release\test_security_curve.exe
30: Test timeout computed to be: 20
30: test_garbage_server_key
30: count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 4
30: test_garbage_client_public_key
30: count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 3
30: test_garbage_client_secret_key
30: count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 3
30: Ignored event: 800 (err = 32)
30: Ignored event: 800 (err = 32)
30: Unexpected event: 800 (err = 108)
30: Assertion failed: false, file C:\projects\libzmq\tests\test_security_curve.cpp, line 307
30/71 Test #30: test_security_curve ..............***Exception: Other 13.56 sec
test 31
      Start 31: test_iov
```

https://ci.appveyor.com/project/zeromq/libzmq/build/build-1566/job/fypn4vl3sg24ydec

sigiesec:
108 is ECONNRESET

bluca:
@sigiesec I've seen this on Linux too, can somewhat reproduce once in every ~10 runs by artificially slowing everything down to a crawl:

```
sudo LD_LIBRARY_PATH=/tmp/test/lib cpulimit -c 1 -l 1 -m -f -- valgrind tests/.libs/test_security_curve
```

cpulimit at 1 core and 1% plus valgrind :-)

Backtrace points to here:

https://github.com/zeromq/libzmq/blob/master/src/stream_engine.cpp#L506

A TCP read is returning ECONNRESET. If I understand correctly, it means that the other side has not yet read what the engine has sent. Given the extremely slow environment I guess this is to be expected, so we can expect it just like EPIPE.
What do you think?

BTW in case you haven't come across it yet, we have a function that can print the current backtrace using libunwind-dev: ```zmq::print_backtrace ();``` from err.hpp. Quite useful to stick it in places that return errors like in this case to quickly find out what's going on.

bluca:
It also got it stuck on another run, I think it was when shutting down the context in shutdown_context_and_server_side after line 718, after the test_curve_security_zap_protocol_error

sigiesec:
Yes, I think ECONNRESET can be handled analogously to EPIPE as well. I added that in my branch already.

sigiesec:
Thanks for the hint on the place where the tests get stuck. I will investigate that further.

bluca:
A new one on Travis valgrind job:

```
test_garbage_server_key
count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 2
test_garbage_client_public_key
count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 1
test_garbage_client_secret_key
count of ZMQ_EVENT_HANDSHAKE_FAILED_ENCRYPTION events: 1
Ignored event: 800 (err = 32)
lt-test_security_curve: tests/test_security_curve.cpp:572: void test_curve_security_zap_unsuccessful(void*, char*, void*, void*, int, int): Assertion `1 <= zmq_atomic_counter_value (zap_requests_handled)' failed.
```

bluca:
https://github.com/zeromq/libzmq/pull/2656

bluca:
Got it to get stuck again via valgrind and vgdb, at the shutdown after test_curve_security_zap_unsuccessful

```
valgrind --tool=memcheck --track-origins=yes --expensive-definedness-checks=yes --vgdb=yes --vgdb-error=0 tests/.libs/test_security_curve
```

This is really slow and also runs in gdb.

It's stuck waiting for the term_mailbox:

https://github.com/zeromq/libzmq/blob/master/src/ctx.cpp#L179

I need to skedaddle now, will keep looking tomorrow

bluca:
The test that gets stuck seems to vary, it's not always the same.

The socket that gets stuck seems to be the dealer client:

```
(gdb) bt
#0  0x00000000060636ad in poll () at ../sysdeps/unix/syscall-template.S:84
#1  0x0000000004e7c5b2 in zmq::signaler_t::wait (this=this@entry=0x693d7a8, timeout_=timeout_@entry=-1)
    at src/signaler.cpp:233
#2  0x0000000004e61376 in zmq::mailbox_t::recv (this=this@entry=0x693d740, cmd_=cmd_@entry=0xffefffbc0, 
    timeout_=timeout_@entry=-1) at src/mailbox.cpp:81
#3  0x0000000004e5060d in zmq::ctx_t::terminate (this=this@entry=0x693d6a0) at src/ctx.cpp:179
#4  0x0000000004e9a62b in zmq_ctx_term (ctx_=0x693d6a0) at src/zmq.cpp:175
#5  0x000000000010af03 in shutdown_context_and_server_side (ctx=0x693d6a0, zap_thread=0x69443a0, server=0x6944400, 
    server_mon=<optimized out>) at tests/test_security_curve.cpp:423
#6  0x000000000010bb1d in main () at tests/test_security_curve.cpp:697
(gdb) p sockets
$12 = {items = std::vector of length 1, capacity 8 = {0x6944400}}
(gdb) p *(socket_base_t *)0x6944400
$13 = {<zmq::own_t> = {<zmq::object_t> = {_vptr.object_t = 0x50b9018 <vtable for zmq::dealer_t+16>, ctx = 0x693d6a0, 
      tid = 4}, options = {sndhwm = 1000, rcvhwm = 1000, affinity = 0, identity_size = 6 '\006', 
      identity = "IDENT", '\000' <repeats 250 times>, rate = 100, recovery_ivl = 10000, multicast_hops = 1, 
      multicast_maxtpdu = 1500, sndbuf = -1, rcvbuf = -1, tos = 0, type = 5, linger = 0, connect_timeout = 0, 
      tcp_maxrt = 0, reconnect_ivl = 100, reconnect_ivl_max = 0, backlog = 100, maxmsgsize = -1, rcvtimeo = 250, 
      sndtimeo = 250, ipv6 = false, immediate = 0, filter = false, invert_matching = false, recv_identity = false, 
      raw_socket = false, raw_notify = true, socks_proxy_address = "", tcp_keepalive = -1, tcp_keepalive_cnt = -1, 
      tcp_keepalive_idle = -1, tcp_keepalive_intvl = -1, tcp_accept_filters = std::vector of length 0, capacity 0, 
      zap_ipc_creds = false, ipc_uid_accept_filters = std::set with 0 elements, 
      ipc_gid_accept_filters = std::set with 0 elements, ipc_pid_accept_filters = std::set with 0 elements, 
      mechanism = 2, as_server = 1, zap_domain = "", plain_username = "", plain_password = "", 
      curve_public_key = '\000' <repeats 31 times>, 
      curve_secret_key = "'\224A֫>\203\272\224}\203\315mX\355\252_7\027Â\274\312ٯV\317?*\233\335\063", 
      curve_server_key = '\000' <repeats 31 times>, gss_principal = "", gss_service_principal = "", 
      gss_principal_nt = 0, gss_service_principal_nt = 0, gss_plaintext = false, socket_id = 41, conflate = false, 
      handshake_ivl = 30000, connected = true, heartbeat_ttl = 0, heartbeat_interval = 0, heartbeat_timeout = -1, 
      use_fd = -1, bound_device = ""}, terminating = true, sent_seqnum = {value = 2}, processed_seqnum = 2, 
    owner = 0x0, owned = std::set with 0 elements, term_acks = 2}, <zmq::array_item_t<0>> = {
    _vptr.array_item_t = 0x50b9168 <vtable for zmq::dealer_t+352>, array_index = 0}, <zmq::i_poll_events> = {
    _vptr.i_poll_events = 0x50b9188 <vtable for zmq::dealer_t+384>}, <zmq::i_pipe_events> = {
    _vptr.i_pipe_events = 0x50b91c0 <vtable for zmq::dealer_t+440>}, connect_rid = "", 
  endpoints = std::multimap with 1 elements = {["tcp://127.0.0.1:38143"] = {first = 0x69450c0, second = 0x0}}, 
  inprocs = std::multimap with 0 elements, tag = 3735928559, ctx_terminated = true, destroyed = false, 
  mailbox = 0x6944a70, pipes = {items = std::vector of length 1, capacity 1 = {0x6965580}}, poller = 0x6940490, 
  handle = 0x697ca30, last_tsc = 4715393263635, ticks = 1, rcvmore = false, clock = {last_tsc = 4715393129734, 
    last_time = 1367271}, monitor_socket = 0x0, monitor_events = 0, last_endpoint = "tcp://127.0.0.1:38143", 
  thread_safe = false, reaper_signaler = 0x0, sync = {mutex = {__data = {__lock = 0, __count = 0, __owner = 0, 
        __nusers = 0, __kind = 1, __spins = 0, __elision = 0, __list = {__prev = 0x0, __next = 0x0}}, 
      __size = '\000' <repeats 16 times>, "\001", '\000' <repeats 22 times>, __align = 0}, attr = {
      __size = "\001\000\000", __align = 1}}, monitor_sync = {mutex = {__data = {__lock = 0, __count = 0, 
        __owner = 0, __nusers = 0, __kind = 1, __spins = 0, __elision = 0, __list = {__prev = 0x0, __next = 0x0}}, 
      __size = '\000' <repeats 16 times>, "\001", '\000' <repeats 22 times>, __align = 0}, attr = {
      __size = "\001\000\000", __align = 1}}}
```

bluca:
I _think_ the problem was a race with the ZAP handler thread and the inproc socket. I've ran my proposed solution many times and it hasn't hanged: https://github.com/zeromq/libzmq/pull/2659

PR #2664, opened 2017-08-07
Problem: some test cases may still fail because of EPIPE/ECONNABORT/ECONNRESET

sigiesec:
Solution: uniformly handle these error conditions in all test cases


bluca:
I assume they were used for debugs, so should the fprintfs be dropped?

bluca:
From the diff there seems to be only debug printfs? Or am I missing something?

sigiesec:
The behavioural change is that all calls to get_monitor_event are replaced by get_monitor_event_with_timeout, which handles EPIPE/... 

I don't think the fprintf's should be dropped. They allow to see which test case was executed when an assertion fails. Using a test framework would remove the need for such boilerplate code.

bluca:
ok, thanks, merged

PR #2665, opened 2017-08-08
Problem: ZAP status codes != 200 do not result in an appropriate monitor event

sigiesec:
Solution: move emitting of monitor events to the point where zap reply is processed and use appropriate errnos

Also fix some more EPIPE/ECONNRESET/ECONNAGAIN failures

### DIFF
--- a/src/stream_engine.cpp
+++ b/src/stream_engine.cpp
@@ -785,23 +785,6 @@ int zmq::stream_engine_t::next_handshake_command (msg_t *msg_)
         if (rc == 0)
             msg_->set_flags (msg_t::command);
 
-#if 0
-        // FIXME emitting the event here leads to extra EAGAIN errors
-#ifdef ZMQ_BUILD_DRAFT_API
-        if (mechanism->status () == mechanism_t::error) {
-            int err = errno;
-            if (mechanism->error_detail () == mechanism_t::zmtp)
-                socket->event_handshake_failed_zmtp (endpoint, err);
-            else if (mechanism->error_detail () == mechanism_t::zap)
-                socket->event_handshake_failed_zap (endpoint, err);
-            else if (mechanism->error_detail () == mechanism_t::encryption)
-                socket->event_handshake_failed_encryption (endpoint, err);
-            else
-                socket->event_handshake_failed_no_detail (endpoint, err);
-        }
-#endif
-#endif
-
         return rc;
     }
 }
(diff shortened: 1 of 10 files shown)


bluca:
@sigiesec thanks!
It looks like your fork is a bit messy with recursive merges etc, could you please reset it so that it's back to a coherent state and the git history is mergeable? This way I don't have to squash it.

```
git fetch upstream
git checkout master
git reset --hard upstrema/master
```

src/stream_engine.cpp
```diff
@@ -784,19 +784,6 @@ int zmq::stream_engine_t::next_handshake_command (msg_t *msg_)
 
         if (rc == 0)
             msg_->set_flags (msg_t::command);
-#ifdef ZMQ_BUILD_DRAFT_API
```
bluca: Why was this block dropped?

sigiesec: `mechanism->status () == mechanism_t::error` becomes true asynchronously in the IO thread that processes the ZAP reply. Here, no useful error code is available. So, I moved emitting such events to the point where the error is identified. 
This also changes from "edge-triggering" to "level-triggering", which might emit multiple events for the same error condition.

sigiesec:
@bluca I noticed that something got messy in my branch. I think it occurred when I rebased it after changes to zeromq/master. However, I don't know what I did wrong. I executed the commands now, but this doesn't fix the branch at hand, right?

bluca:
Your local master branch should be on par with the zeromq one, and any new branch you create from that point will be fine.

If you don't already, I would recommend creating a ```git lg``` alias:

```
git config --global alias.lg "log --graph --abbrev-commit --decorate --date=relative --format=format:'%C(bold blue)%h%C(reset) %G? - %C(bold green)(%ar)%C(reset) %C(red)%s%C(reset) %C(dim red)- %an%C(reset)%C(dim blue)%d%C(reset)' --all"
```

which will show all branches in a tree-like format, and it then becomes very easy to visualize.

bluca:
if you have already a branch you need to fix, do an interactive rebase:

```
git rebase -i upstream/master
```

src/stream_engine.cpp
bluca: I see, thanks.
Does the multiple event change means that there is a public API change?

sigiesec: Only to the DRAFT API, all handshake events are only defined in the DRAFT API.

bluca: Perfect, thanks for confirming

sigiesec:
Thanks for the git hints, @bluca :) I hope I will not run into these problems next time.

jimklimov:
...and in the end you `git push -f origin` to force-update the current branch of your workspace to the "remote" repo named `origin` (assuming it is your github fork, which the PR is based on) and the PR will magically get a cleaner commit-history.

If you work in several environments, don't forget to `git pull` on your other workspaces to avoid mess and "conflicts apparent" later on. There is a reason that force-pushes are discouraged in general shared-source projects ;) 

PR #2671, opened 2017-08-10
Problem: log for bad ZAP status code is confusing

AlainODea:
Solution: log that handler sent bad status code to clarify ZAP debugging

### DIFF
--- a/src/null_mechanism.cpp
+++ b/src/null_mechanism.cpp
@@ -335,7 +335,7 @@ int zmq::null_mechanism_t::receive_and_process_zap_reply ()
     //  Status code frame
     if (msg [3].size () != 3) {
         //  Temporary support for security debugging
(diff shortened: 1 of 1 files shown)


bluca:
Thanks - @sigiesec we should probably convert the null mechanism and others to the events as well

AlainODea:
My pleasure. I'm really enjoying the clarity of this codebase. I found it easy to get into and you folks are really welcoming to newcomers. I appreciate that. Thank you :)

sigiesec:
@bluca, yes, we should do that :) I think the ZAP handling code should be refactored in that context, it is now somewhat duplicated across the mechanisms. I will have a look at that

bluca:
Great, thanks!

@AlainODea welcome to the club :-)

PR #2680, opened 2017-08-14
[WIP, do not merge] Problem: insufficient tests for ZMTP-CURVE protocol errors

sigiesec:
Solution: add tests with misbehaving ZMTP-CURVE client

### DIFF
--- a/src/curve_server.cpp
+++ b/src/curve_server.cpp
@@ -102,6 +102,12 @@ int zmq::curve_server_t::process_handshake_command (msg_t *msg_)
             rc = process_initiate (msg_);
             break;
         default:
+            // TODO I think this is not a case reachable with a misbehaving 
+            // client. It is not an "invalid handshake command", but would be 
+            // trying to process a handshake command in an invalid state, 
+            // which is purely under control of this peer. 
+            // Therefore, it should be changed to zmq_assert (false);
+
             // CURVE I: invalid handshake command
             current_error_detail = zmtp;
             errno = EPROTO;
@@ -242,6 +248,9 @@ int zmq::curve_server_t::decode (msg_t *msg_)
 
 int zmq::curve_server_t::zap_msg_available ()
 {
+    //  TODO I don't think that it is possible that this is called in any 
+    //  state other than expect_zap_reply. It should be changed to
+    //  zmq_assert (state == expect_zap_reply);
     if (state != expect_zap_reply) {
         errno = EFSM;
         return -1;
@@ -371,6 +380,13 @@ int zmq::curve_server_t::produce_welcome (msg_t *msg_)
     rc = crypto_box (welcome_ciphertext, welcome_plaintext,
                      sizeof welcome_plaintext,
                      welcome_nonce, cn_client, secret_key);
+
+    //  TODO I think we should change this back to zmq_assert (rc == 0);
+    //  as it was before https://github.com/zeromq/libzmq/pull/1832
+    //  The reason given there was that secret_key might be 0ed.
+    //  But if it were, we would never get this far, since we could
+    //  not have opened the client's hello box with a 0ed key.
+
     if (rc == -1)
         return -1;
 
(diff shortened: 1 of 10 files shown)


sigiesec:
There is a build failure in the non-cmake builds. I don't understand how it is caused. Anyone got a clue?

See, e.g. https://travis-ci.org/zeromq/libzmq/jobs/264284870:
{code}
make[1]: *** No rule to make target `tests/test_security_curve.c', needed by `distdir'.  Stop.
{code}

There is no file tests/test_security_curve.c, but tests/test_security_curve.*cpp*

jimklimov:
Did you generate the file as a zproject 'main'? Should be `.cc` (not `.cpp`) and there may be a mixup of underscore and dash in recipes for binary and source file.

sigiesec:
@jimklimov I did not generate the file at all, it already existed. I tried to change the project files so that it is conditionally compiled only when ZMQ_HAS_CURVE is true. All of the source files in the repo have a .cpp suffix, .cc is not used here.

sigiesec:
The build problem relating to test_security_curve.c has disappeared.
Now, only the libsodium builds are still failing, but this is rather as expected, as I force-use tweetnacl.c in the tests. This should probably use the same library as in libzmq itself, as configured.

tests/test_security_curve.cpp
```diff
@@ -40,6 +40,10 @@
 #  include <unistd.h>
 #endif
 
+#include "../src/tweetnacl.h"
```
bluca: Sorry but I have some doubts about letting internal state&code leak in the tests, see README: https://github.com/zeromq/libzmq/blob/master/tests/README.md

src/curve_server.cpp
```diff
@@ -102,6 +102,12 @@ int zmq::curve_server_t::process_handshake_command (msg_t *msg_)
             rc = process_initiate (msg_);
             break;
         default:
+            // TODO I think this is not a case reachable with a misbehaving
```
sigiesec: Can this be changed to zmq_assert?

src/curve_server.cpp
```diff
@@ -242,6 +248,9 @@ int zmq::curve_server_t::decode (msg_t *msg_)
 
 int zmq::curve_server_t::zap_msg_available ()
 {
+    //  TODO I don't think that it is possible that this is called in any 
+    //  state other than expect_zap_reply. It should be changed to
+    //  zmq_assert (state == expect_zap_reply);
```
sigiesec: Can this be changed to zmq_assert?

src/curve_server.cpp
```diff
@@ -371,6 +380,13 @@ int zmq::curve_server_t::produce_welcome (msg_t *msg_)
     rc = crypto_box (welcome_ciphertext, welcome_plaintext,
                      sizeof welcome_plaintext,
                      welcome_nonce, cn_client, secret_key);
+
+    //  TODO I think we should change this back to zmq_assert (rc == 0);
```
sigiesec: Can this be changed to zmq_assert?

src/curve_server.cpp
```diff
@@ -581,6 +603,11 @@ int zmq::curve_server_t::produce_error (msg_t *msg_) const
 
 int zmq::curve_server_t::send_zap_request (const uint8_t *key)
 {
+    // TODO  I don't think the rc can be -1 anywhere below.
```
sigiesec: Can these be changed to zmq_assert?

sigiesec:
I think I am done implementing the test cases I wanted to add in this PR.

tests/test_security_curve.cpp
bluca: But the code is leaking - personally I would much, much rather prefer copying part of the implementation in the test case client. Having to jump through the hoops to include source code directly in the test binary is a bit messy and I'm afraid it will cause problems in funkier build systems like visual studio :-/ I can take care of that in a while.

sigiesec: Yes, the code is leaking. I did not deny that ;)
What do you think about putting such code into a separate library that both libzmq and specific tests link against?
I am very much against duplicating code, even for things that appear trivial. And the code implementing the CURVE protocol is not trivial at all.

bluca: I'm afraid that would be worse - it would break all applications that do static builds since there's now an extra library, and would make all build systems even more complicated.

It's already broken now - for example, it's no longer possible to build with libunwind:

```
  CXXLD    tests/test_security_curve
/usr/bin/ld: src/err.o: undefined reference to symbol '_Ux86_64_getcontext'
//usr/lib/x86_64-linux-gnu/libunwind.so.8: error adding symbols: DSO missing from command line
```

Likely some flags missing somewhere.

sigiesec: I can imagine only few things worse than duplicating code ;)
I don't think it would break applications... libzmq can still contain the inner library (static libraries can be merged).

Issue #2681, opened 2017-08-14
Possible buffer overflow in CURVE mechanism handshake


sigiesec:
E.g., the function zmq::curve_client_t::produce_initiate contains the following lines:
```
    //  Assume here that metadata is limited to 256 bytes
    uint8_t initiate_plaintext [crypto_box_ZEROBYTES + 128 + 256];
```

One part of the metadata is the socket identity, which alone may be 255 bytes, so this is definitely too small.

The other mechanisms suffer from similar risks, but use larger buffers (512 bytes). Still, they do not check for overruns.

PR #2683, opened 2017-08-15
Problem: Possible buffer overruns related to metadata in various mechanisms

sigiesec:
Solution: compute actual size to represent actual metadata and allocate buffers accordingly
Reduce code duplication related to metadata handling
Added test case that reproduces one instance of the problem with the CURVE mechanism

### DIFF
--- a/src/curve_client_tools.hpp
+++ b/src/curve_client_tools.hpp
@@ -146,9 +146,6 @@ struct curve_client_tools_t
                                  const uint8_t *metadata_plaintext,
                                  const size_t metadata_length)
     {
-        const size_t max_metadata_length = 256;
-        zmq_assert (metadata_length < max_metadata_length);
-
         uint8_t vouch_nonce[crypto_box_NONCEBYTES];
         uint8_t vouch_plaintext[crypto_box_ZEROBYTES + 64];
         uint8_t vouch_box[crypto_box_BOXZEROBYTES + 80];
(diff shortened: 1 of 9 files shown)


sigiesec:
This PR fixes #2681

src/curve_client_tools.hpp
```diff
@@ -146,9 +146,6 @@ struct curve_client_tools_t
                                  const uint8_t *metadata_plaintext,
                                  const size_t metadata_length)
     {
-        const size_t max_metadata_length = 256;
-        zmq_assert (metadata_length < max_metadata_length);
```
bluca: won't this assert ensure that there can never be an overflow?

bluca:
Could you please rebase? It looks like it's branched off your other PR, so it can't be merged

src/curve_client_tools.hpp
sigiesec: At this point, yes. I inserted this check as part of my previous PR #2680. Sorry, I did not reflect that properly in the description of this PR. Still, the assertion could be triggered by correct code using libzmq, so this needs to be changed.

bluca: Ah ok, missed that it wasn't there before, sorry

sigiesec:
@bluca I rebased the branch.

bluca:
Uhm I still see all the commits from the previous branch though?

You'll want something like this:

```
git branch backup
git fetch upstream
git rebase --onto upstream/master HEAD^
```

If it's not only the very last commit you want, add another ```^``` per commit

bluca:
Oh and then git push -f

sigiesec:
Hm... I can try that again... I don't get it why it does not just do a fast-forward though. Is this related to the previous PR being squash-merged into the master?

sigiesec:
@bluca does it look better now?

bluca:
Great, thanks!

In terms of workflow it's great to have many small commits, but individually they should still work.

You can squash commits that fix things broken by earlier ones in the same branch with an interactive rebase:

```
git rebase -i upstream/master
```

The editor will open, and you can reorder the commits as you want, and can squash or fixup them (to merge 2 or more commits into a single one).

It's important to try and have each commit in the history at least buildable, because otherwise git bisect will not work anymore, and it's a very, very powerfull tool to track down regressions.

So when in a PR I see that intermediate commits are not fully working I tend to squash when merging to avoid breaking bisect. Sorry if it causes problems.

sigiesec:
@bluca Great :) I will try to be more careful with the branch history next time.

sigiesec:
Fixes #2681

PR #2688, opened 2017-08-16
Problem: missing flags break build of test_security_curve

bluca:
Solution: add all the required compiler flags since the test includes
source code from the library directly

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -638,9 +638,13 @@ tests_test_security_curve_SOURCES += \
 endif
 
 tests_test_security_curve_LDADD = \
-	src/libzmq.la
+	src/libzmq.la $(LIBUNWIND_LIBS)
+tests_test_security_curve_CPPFLAGS = \
+	${LIBUNWIND_CFLAGS}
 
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -638,9 +638,13 @@ tests_test_security_curve_SOURCES += \
 endif
 
 tests_test_security_curve_LDADD = \
-	src/libzmq.la
+	src/libzmq.la $(LIBUNWIND_LIBS)
```
jimklimov: Sanity check : Wasn't libunwind an optional dependency (or am I mistaking projects)? Should this either become required dependency for this test to get built, or `#if HAVE_LIBUNWIND` trickery is needed?

bluca: The variable will be empty if it's not found or disabled

jimklimov: And will compilation break if the library is missing, then?

bluca: It's ifdef'd in err.hpp - which is included by the test, so it was breaking when the library was there because the test was not being linked against it, despite the ifdef being true

PR #2689, opened 2017-08-17
Problem: ZAP client code duplicated across mechanisms with behaviour deviations

sigiesec:
Solution: Extracted the ZAP handling code into classes used by all mechanisms, and added tests for ZAP errors for NULL and PLAIN mechanisms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -565,7 +565,8 @@ set (cxx-sources
         udp_engine.cpp
         udp_address.cpp
         scatter.cpp
-        gather.cpp)
+        gather.cpp
+		zap_client.cpp)
 
 set (rc-sources version.rc)
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -523,7 +523,10 @@ tests_test_security_null_LDADD = src/libzmq.la
 tests_test_security_plain_SOURCES = tests/test_security_plain.cpp
 tests_test_security_plain_LDADD = src/libzmq.la
 
-tests_test_security_zap_SOURCES = tests/test_security_zap.cpp
+tests_test_security_zap_SOURCES = \
+	tests/test_security_zap.cpp \
+	tests/testutil_security.hpp \
+	tests/testutil.hpp
 tests_test_security_zap_LDADD = src/libzmq.la
 
 tests_test_spec_req_SOURCES = tests/test_spec_req.cpp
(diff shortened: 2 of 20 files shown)


sigiesec:
In particular, this adds ZMQ_EVENT_HANDSHAKE_FAILED_* events for ZAP protocol errors for the mechanisms other than CURVE, partially addressing the first comment in #2671

bluca:
Thanks!

@sigiesec I noticed you haven't sent a relicense grant, could you please do so when you have time? https://github.com/zeromq/libzmq/tree/master/RELICENSE

sigiesec:
@bluca Sorry, I missed that until now. Just submitted the agreement as #2690

bluca:
@sigiesec I've seen test_security_zap fail a few times now - unfortunately in CI or OBS so no useful logs. On Linux too. Have you seen it fail while testing?

sigiesec:
No, I did see have a single failure locally. Have you got some links to the failing run logs? Can we see if it is an assertion failure, or segfault, or timeout?

bluca:
They look like asserts:

https://travis-ci.org/zeromq/libzmq/jobs/265536824
https://travis-ci.org/zeromq/libzmq/jobs/265575582

sigiesec:
Hm... why is there no output at all? The test outputs "NULL mechanism" immediately after calling setup_test_environment.

bluca:
autoconf's make check redirects all output to the log file - which then is conveniently lost given it's on an ephemeral CI.

To make it spit it out on error, it needs to be called with verbose, eg:

```make VERBOSE=1 check```

**Comment by bluca** [comment]
The coverage job didn't have that: https://github.com/zeromq/libzmq/pull/2693

**Comment by bluca** [comment]
@sigiesec 

```
FAIL: tests/test_security_zap
=============================
NULL mechanism
test_zap_protocol_error wrong_version
Unexpected event: 8000 (err = 32)
lt-test_security_zap: ../tests/testutil_security.hpp:518: int expect_monitor_event_multiple(void*, int, int): Assertion `false' failed.
```

https://travis-ci.org/zeromq/libzmq/jobs/265657413

```
FAIL: tests/test_security_zap
=============================
NULL mechanism
test_zap_protocol_error wrong_version
test_zap_protocol_error wrong_request_id
test_zap_protocol_error wrong_status_invalid
Unexpected event: 8000 (err = 32)
lt-test_security_zap: tests/testutil_security.hpp:518: int expect_monitor_event_multiple(void*, int, int): Assertion `false' failed.
```

https://travis-ci.org/zeromq/libzmq/jobs/265657422

sigiesec:
Ah great, that's something to work with ;) I will have a look.

PR #2695, opened 2017-08-18
Problem: Values in ZMQ_EVENT_HANDSHAKE_FAILED_* are not helpful

### LINKED ISSUE #2682: Value in ZMQ_EVENT_HANDSHAKE_FAILED_* monitor events
The value in the ZMQ_EVENT_HANDSHAKE_FAILED_* monitor events is currently an errno error code.
This does not allow to distinguish error situations in a fine grained way.

Another approach would be to define custom error codes for every situation where a handshake error is encountered, at least for all situations currently subsumed into EPROTO. This would allow for better tests, and for better diagnostics in production.

Currently, this is a DRAFT API, so it can still be changed without breaking a finalized API.

Any opinions on this?

### LINKED ISSUE #2694: Can't build on Windows
I have several versions of Visual Studio installed but can't seem to get past the build stage.  Following the instructions on the wiki are tough because they mention running configure.bat which doesn't exist.  In any event, if I run build.bat it fails building with VS2013 with unresolved externals, same behavior if I open the solution file and attempt to build that way.  Here's one of the externals it's complaining about:

Error	LNK2019	unresolved external symbol "public: void __thiscall zmq::zap_client_t::send_zap_request(char const *,unsigned int,unsigned char const *,unsigned int)" (?send_zap_request@zap_client_t@zmq@@QAEXPBDIPBEI@Z) referenced in function "private: void __thiscall zmq::curve_server_t::send_zap_request(unsigned char const *)" (?send_zap_request@curve_server_t@zmq@@AAEXPBE@Z)	libzmq	C:\source\libzmq\builds\msvc\vs2013\libzmq\curve_server.obj	1

sigiesec:
Solution: Redesign ZMQ_EVENT_HANDHAKE_FAILED_* events and use dedicated values for ZMQ_EVENT_HANDHAKE_FAILED_PROTOCOL

Fixes #2682 

It also adds the monitoring events to the other mechanisms as suggested in #2671 

This also fixes #2694, and it should fix the spurious failures mentioned in #2689


builds/msvc/vs2008/libzmq/libzmq.vcproj
```diff
@@ -121,6 +121,7 @@
       <File RelativePath="..\..\..\..\src\kqueue.cpp" />
       <File RelativePath="..\..\..\..\src\lb.cpp" />
       <File RelativePath="..\..\..\..\src\mailbox.cpp" />
```
bluca: @sigiesec what about curve_mechanism_base.cpp in this file?

bluca:
Great work, thanks!

builds/msvc/vs2008/libzmq/libzmq.vcproj
sigiesec: The VS2008 project doesn't even contain curve_client.cpp and curve_server.cpp ...

bluca: agh, ok let me have a look

PR #2696, opened 2017-08-18
Problem: gssapi_mechanism_base_t aborts the application when it receives a message it cannot decode

sigiesec:
Solution: emit an ZMQ_EVENT_HANDSHAKE_FAILED_PROTOCOL event instead of using zmq_assert

### DIFF
--- a/src/gssapi_mechanism_base.cpp
+++ b/src/gssapi_mechanism_base.cpp
@@ -181,12 +181,18 @@ int zmq::gssapi_mechanism_base_t::decode_message (msg_t *msg_)
     maj_stat = gss_unwrap(&min_stat, context, &wrapped, &plaintext,
                           &state, (gss_qop_t *) NULL);
 
-    //  TODO I don't think it is a good idea to use zmq_assert here. If 
-    //  decryption fails, gss_unwrap returns GSS_S_BAD_SIG. This opens up 
-    //  to DoS attacks by clients! Instead, a 
-    //  ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC event should be emitted.
-
-    zmq_assert(maj_stat == GSS_S_COMPLETE);
+    if (maj_stat != GSS_S_COMPLETE)
+    {
+        //  TODO is it correct to release the plaintext buffer if gss_unwrap 
+        //  did not succeed?
+        gss_release_buffer (&min_stat, &plaintext);
+        free (wrapped);
+        session->get_socket ()->event_handshake_failed_protocol (
+          session->get_endpoint (),
+          ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC);
+        errno = EPROTO;
+        return -1;
+    }
     zmq_assert(state);
 
     // Re-initialize msg_ for plaintext


sigiesec:
Can someone please verify that it is correct to call gss_release_buffer on the output buffer if gss_unwrap failed?

bluca:
Could you please rebase? I merged your other PR

sigiesec:
@bluca Rebase done. I think slowly I get accustomed with the git workflow :)

src/gssapi_mechanism_base.cpp
```diff
@@ -181,12 +181,18 @@ int zmq::gssapi_mechanism_base_t::decode_message (msg_t *msg_)
...
-    //  decryption fails, gss_unwrap returns GSS_S_BAD_SIG. This opens up 
-    //  to DoS attacks by clients! Instead, a 
-    //  ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC event should be emitted.
-
-    zmq_assert(maj_stat == GSS_S_COMPLETE);
+    if (maj_stat != GSS_S_COMPLETE)
+    {
+        //  TODO is it correct to release the plaintext buffer if gss_unwrap 
+        //  did not succeed?
+        gss_release_buffer (&min_stat, &plaintext);
```
bluca: @garlick do you know if this is the correct thing to do?

garlick: Probably not correct.  If `gss_unwrap()` fails,  `plaintext` is likely not initialized.

bluca: Thanks. What is the correct thing to do to cleanly handle that error?

bluca:
No worries, and thanks.

src/gssapi_mechanism_base.cpp
sigiesec: Based on the documentation of gss_unwrap, I could not tell.
But I found several code examples that also do call gss_release_buffer in an error case.
For example: https://github.com/happyhappiness/vscode/blob/650a7dbf7303fa37e5096b1ef667d2e7aae04872/second/download/CMake/CMake-old-new/CMake-old-new/Kitware_CMake_old_file_98.cpp#L482

bluca: So going at the source directly:

https://github.com/krb5/krb5/blob/9c18f6c93b602ec69ca2640e27ab2ccbb7f45a3c/src/lib/gssapi/mechglue/g_unseal.c#L55

gss_unwrap as the very first thing checks that plaintext is not a null pointer, which in our case it's true given it's on the stack, and then initialises its members to 0 length and null ptr.

So it should be safe to release it in all cases, and the release API seems to check again if it's not a null pointer and then if the members are 0 length and null pointer it's a no-op.

bluca: And it seems to be the case in the oldest release available in github, from 2007, as well: 
https://github.com/krb5/krb5/blob/krb5-1.5.4-final/src/lib/gssapi/mechglue/g_unseal.c#L55

sigiesec:
@garlick When you have the time, please have a closer look. If I understand it correctly, the original
```
zmq_assert(maj_stat == GSS_S_COMPLETE);
```
meant that the process is killed when a message is processed that cannot be decrypted. 
Since this (typically) comes from outside the process, it is not acceptable to assume that it is correct. It may be undecryptable because of an erroneous implementation of the peer or because of a malicious peer.

src/gssapi_mechanism_base.cpp
bluca: I've removed the TODO for now, @garlick when you have more time to double check let me know if my interpretation is wrong and we can reconsider.
https://github.com/zeromq/libzmq/pull/2700

garlick:
I agree the assertion is inappropriate.

PR #2697, opened 2017-08-18
Problem: missing files from VS2008 solution

bluca:
Solution: generate list of sources and headers with:

 ```ls -1 src/*cpp | sort | sed 's|src/\(.*\)|      <File RelativePath="..\\..\\..\\..\\src\\\1" />|g'```
 ```ls -1 src/*hpp | sort | sed 's|src/\(.*\)|      <File RelativePath="..\\..\\..\\..\\src\\\1" />|g'```

and add them.


bluca:
Would be great if someone with VS2008 could test it afterwards :-)

evoskuil:
FWIW, I've long recommended that the VS2008 solution be dropped. The file formats are incompatible with the technique used in the others, and they are six generations behind. I've never built them.

Issue #2667, opened 2017-08-08
curve_server should not send an error in case of a ZAP response with status 300


sigiesec:
Status code 300 indicates a temporary failure. https://rfc.zeromq.org/spec:23/ZMTP section Error Handling specified that a peer should not send an ERROR in case of a temporarily failure, but simply close the connection.

Issue #2586, opened 2017-05-24
zeromq can hang indefinitely in zmq::ctx_t::terminate() despite LINGER=0


tcwalther:
I have an edge-case in a Python program where ZeroMQ can hang indefinitely. I have an extremely hard time isolating the bug; the most reliable way for me to reproduce it is to repeatedly run our entire test suite until that suite hangs (every 10th time on average). I've already posted a corresponding bug report at https://github.com/zeromq/pyzmq/issues/1003, but I think I've found out a bit more now which hopefully justifies creating an issue here.

The basic problem is that I have created contexts and sockets all with `LINGER=0` and `BLOCKY=False`, and still, sometimes it hangs forever in `zmq::ctx_t_terminate()`. I am using zeromq 4.2.1. Looking at the GDB output, I can see that it hangs in https://github.com/zeromq/libzmq/blob/v4.2.1/src/ctx.cpp#L194. Here is the GDB output:

```
(gdb) frame
#4  0x00007f1d61ffcccd in zmq::ctx_t::terminate (this=0x4984290) at src/ctx.cpp:194
194	        int rc = term_mailbox.recv (&cmd, -1);
(gdb) p *this
$13 = {tag = 2880293630, sockets = {items = std::vector of length 48, capacity 64 = {0x48c6860, 0x5ca93d0,
      0x49d4bc0, 0x5c93cc0, 0x5bf5ff0, 0x4941160, 0x5ce6780, 0x5daee30, 0x5d03450, 0x5d0e030, 0x5dbf220, 0x49ccb20,
      0x49026d0, 0x4f375c0, 0x5d0ae60, 0x5e60cd0, 0x5d1b370, 0x5b00010, 0x4a14470, 0x4b0fb40, 0x4f247b0, 0x5c4fd60,
      0x48dd020, 0x5c4ed80, 0x4988070, 0x5d11ca0, 0x5c7b840, 0x4976780, 0x5aed1b0, 0x5cb16e0, 0x4944f50, 0x5d34bb0,
      0x4d79b60, 0x4a2fc90, 0x4c3b020, 0x5e64e60, 0x4eedc40, 0x4f87570, 0x5e5cd50, 0x4f41b40, 0x5e1ea80, 0x5afacd0,
      0x4acf650, 0x48c4640, 0x4f40a10, 0x5b57c60, 0x5b38b80, 0x5e66cc0}},
[...]

Issue #2305, opened 2017-01-13
Assertion on closing context when queues run full


sigiesec:
The following code produces an assertion when closing the context with ZeroMQ 4.2.1, but was fine with ZeroMQ 4.1.5:
```
Assertion failed: Invalid argument (..\..\..\..\src\select.cpp:97)
```

The code is as follows:
```
#include "zmq.hpp"
#include <iostream>
#include <string>
#include <thread>

int amount = 5000;

// main func
int main(int argc, char **argv) {
    zmq::context_t ctx(1);
    zmq::socket_t backend_sock_(ctx, ZMQ_ROUTER);
    backend_sock_.bind("tcp://127.0.0.1:8888");

    zmq::socket_t frontend_sock_(ctx, ZMQ_DEALER);
    frontend_sock_.setsockopt(ZMQ_IDENTITY, "1", 1);
    frontend_sock_.connect("tcp://127.0.0.1:8888");

    // identity probe
    frontend_sock_.send(nullptr, 0);
    zmq::message_t msg;
    backend_sock_.recv(&msg);
    assert(msg.more());
    backend_sock_.recv(&msg);
    assert(!msg.more());

    const std::string str("msg to be sent");

    int i = 0;
    while (i++ < amount) {
      backend_sock_.send("1", 1, ZMQ_SNDMORE);
      backend_sock_.send("", 0, ZMQ_SNDMORE);
      backend_sock_.send(str.data(), str.size(), 0);
    }
  return 0;
}
```

If I reduce amount to ca. 1050 or lower, the code works fine.

Environment is Windows 7, Visual Studio 2015, 64 Bit, Debug Runtime

bluca:
Can you provide a snippet that reproduces it without the C++ bindings?

sigiesec:
Yes, I hope this is equivalent (minus any error handling):
```
#include "zmq.h"
#include <iostream>
#include <string>
#include <thread>

int amount = 5000;

// main func
int main(int argc, char **argv) {
  void *ctx = zmq_ctx_new();
  void *backend_sock_ = zmq_socket(ctx, ZMQ_ROUTER);
  zmq_bind(backend_sock_, "tcp://127.0.0.1:8888");

  void *frontend_sock_ = zmq_socket(ctx, ZMQ_DEALER);
  zmq_setsockopt(frontend_sock_, ZMQ_IDENTITY, "1", 1);
  zmq_connect(frontend_sock_, "tcp://127.0.0.1:8888");

[...]

PR #2698, opened 2017-08-18
Problem: no tests for client-side events for successful handshake and authentication failure in handshake

sigiesec:
Solution: added tests for CURVE, add emitting of client-side event in curve_client_t; add ZAP code 300/500 tests for all mechanisms; suppress sending an error message for ZAP code 300

### DIFF
--- a/src/curve_client.cpp
+++ b/src/curve_client.cpp
@@ -275,17 +275,21 @@ int zmq::curve_client_t::process_error (
     }
     if (msg_size < 7) {
         session->get_socket ()->event_handshake_failed_protocol (
-          session->get_endpoint (), ZMQ_PROTOCOL_ERROR_ZMTP_MALFORMED_COMMAND_ERROR);
+          session->get_endpoint (),
+          ZMQ_PROTOCOL_ERROR_ZMTP_MALFORMED_COMMAND_ERROR);
         errno = EPROTO;
         return -1;
     }
     const size_t error_reason_len = static_cast <size_t> (msg_data [6]);
     if (error_reason_len > msg_size - 7) {
         session->get_socket ()->event_handshake_failed_protocol (
-          session->get_endpoint (), ZMQ_PROTOCOL_ERROR_ZMTP_MALFORMED_COMMAND_ERROR);
+          session->get_endpoint (),
+          ZMQ_PROTOCOL_ERROR_ZMTP_MALFORMED_COMMAND_ERROR);
         errno = EPROTO;
         return -1;
     }
+    const char *error_reason = reinterpret_cast<const char *> (msg_data) + 7;
+    handle_error_reason (error_reason, error_reason_len);
     state = error_received;
     return 0;
 }
--- a/src/mechanism_base.cpp
+++ b/src/mechanism_base.cpp
@@ -52,3 +52,14 @@ int zmq::mechanism_base_t::check_basic_command_structure (msg_t *msg_)
     return 0;
 }
 
+void zmq::mechanism_base_t::handle_error_reason (const char *error_reason,
+                                                 int error_reason_len)
+{
+    if (error_reason_len == 3 && error_reason[1] == '0'
+        && error_reason[2] == '0' && error_reason[0] >= '3'
+        && error_reason[0] <= '5') {
+        // it is a ZAP status code, so emit an authentication failure event
+        session->get_socket ()->event_handshake_failed_auth (
+          session->get_endpoint (), (error_reason[0] - '0') * 100);
+    }
+}
--- a/src/mechanism_base.hpp
+++ b/src/mechanism_base.hpp
@@ -43,6 +43,8 @@ class mechanism_base_t : public mechanism_t
     session_base_t *const session;
 
     int check_basic_command_structure (msg_t *msg_);
+
+    void handle_error_reason (const char *error_reason, int error_reason_len);
 };
 }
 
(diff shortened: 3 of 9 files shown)


sigiesec:
In the valgrind build, the test case expecting a client-side authentication failed (status 500) event apparently does not emit such an event. Is it possible that the server closes the connection before sending the ERROR command to the client? (and this happens with the valgrind build because everything is much slower there?) Is this behaviour acceptable? If it is, the tests should allow it, otherwise something must be fixed, probably within stream_engine_t

bluca:
Not sure what exactly is happening to be able to tell. But there's another test hanging (killed by sigalarm) without valgrind. Maybe it's the same issue?

sigiesec:
This PR fixes #2667

sigiesec:
@bluca Yes, probably it is the same issue. 

What do we do? Accept the apparent behaviour and relax the test assumption for now?

bluca:
Given it's existing behaviour, I think it's fine to accept it for now, and perhaps file an issue to track it in case we need to revise.

bluca:
I've now seen test_security_curve hang a few times (killed by sigalarm) in some builds on OBS, I'd say one in 20 times, on various distro and architectures (x86_64 too, so not only emulated slow environments), have you seen it hang too?

sigiesec:
@bluca If it happens only in non-DRAFT builds, maybe this is due to some race condition that is avoided/masked when DRAFT is enabled, which involves waiting for the new events, which might provide the required synchronization, or at least involve additional delays. I will try to run the tests without DRAFT on my machine repeatedly.

sigiesec:
I fixed one more issue that probably only affects the DRAFT builds related to the tests added by this PR. 

I will try tomorrow to run the non-DRAFT build test repeatedly to see if I can reproduce that. The discussion is a bit misplaced here, because it refers to the state *before* this PR, right?

bluca:
Yes, indeed it's current state of affairs in the master branch, sorry for the confusion.

First started seen the hangs repeatedly on the 16th if that can help.

bluca:
The plain and null got an update here, should the gssapi be updated as well?

sigiesec:
Yes, gssapi should probably changed accordingly.

sigiesec:
I checked it again: gssapi_server_t does not send any error to the client in case ZAP authentication failed. So there is no place to change gssapi_client_t accordingly. Maybe gssapi_server_t (and the GSSAPI mechanism protocol) could be extended to allow for ERROR messages, but this is a change with a larger scope.

bluca:
Ok, thanks. I guess it's fine as-is for now.

bluca:
@sigiesec test_security_zap also occasionally hangs:

```
[  224s] ./config/test-driver: line 112: 20072 Alarm clock             "$@" > $log_file 2>&1
[  224s] FAIL: tests/test_security_zap
```

Again in stable builds, and again before this PR too.

sigiesec:
@bluca I tried to reproduce the hangs (without DRAF) on my Windows machine, put it under load, but was not able to get any hang so far. However, I got the error mentioned in #2305 again once, with an assertion failure within the Windows-specific section of zmq::select_t::rm_fd

bluca:
agh, I thought that one was fixed :-(

I'll try to reproduce it myself today or tomorrow, perhaps it's linux-specific

sigiesec:
I thought so too :( Haven't seen it for quite a while.

sigiesec:
Just a thought: are you sure the tests are really stuck? or are they just to slow? I remember we increased the test timeout for test_security_curve and test_security_zap for the CMake builds, but I don't think we did so for the default builds. I haven't found a place in Makefile.am to configure the test timeout though.

bluca:
I suppose that's a good point, but it's 60 seconds now which is the same in CMake:

https://github.com/zeromq/libzmq/blob/master/tests/testutil.hpp#L321

bluca:
I can reproduce it once every ~50 runs in valgrind + vgdb. It's stuck terminating the ctx here: https://github.com/zeromq/libzmq/blob/master/tests/test_security_curve.cpp#L662

As it happened before, it's waiting to receive on the mailbox at https://github.com/zeromq/libzmq/blob/master/src/ctx.cpp#L179

bluca:
And as before it's the server socket (it has identity "IDENT")

bluca:
@somdoron do you know off the top of your head what object could be which term_ack is missed?
It's a TCP dealer socket

sigiesec:
@bluca I changed the source links in your comment to the permalinks, so they won't break when the master is changed.

This is the same code line where a hang was reported as #2586 in version 4.2.1:
https://github.com/zeromq/libzmq/blob/7005f22726d4a6ca527f27560a0a132394fdbbcc/src/ctx.cpp#L194

jimklimov:
Perhaps you can add some progress monitor to report how far the test went, and see if it is just a matter of not enough time due to fixed 60sec limit vs. getting stuck on some deadlock or borked on a race?

bluca:
@sigiesec yes that's right, so the fix I did to try and make the zap handler thread in the test less racy didn't help unfortunately.

@jimklimov it's definitely stuck where I posted, I can repro in GDB - that's where the dumps come from.

There is nothing special about that server dealer socket - apart from the fact that it uses zap. Other tests that bind a dealer socket don't hang at shutdown.
This suggests it might be a problem with the internal zap inproc - that's about the only difference.

I'll try and find out what is the second object that the socket_base_t "owns" and is waiting on an ack from - the first one is the tcp_listener_t and that's terminated correctly. I don't think the internal zap handler gets owned by the socket, as it's a global one per context.

Issue #2699, opened 2017-08-18
Problem: threadsafe socket types don't support SNDMORE / multi-frame messages


minrk:
Probably the most valuable feature in libzmq for me is the multi-frame message format, where chunks of a message can be serialized separately, without concatenation, but the message is still delivered as a unit.

The new threadsafe sockets have abandoned support for SNDMORE, presumably in the interest of threadsafety. However, as far as I am concerned, this means all the new sockets may as well not exist since they can't support messages in this format.

Is there any hope of non-contiguous messages in the threadsafe sockets? Ideally via the same SNDMORE mechanism as all the already working sockets?

If there are no plans for multi-frame message support, the docs should probably be updated to remove the note about RADIO-DISH replacing PUB-SUB and SERVER-CLIENT replacing ROUTER-DEALER, since they can't really aspire to do that without multiframe message support.

bluca:
Unfortunately I don't think it can be done with thread safe sockets, simply because there's no way to guarantee any ordering - if 2 threads send 2 frames each, there's no telling in which order they'll get to the I/O thread with the current implementation :-(

bluca:
@somdoron what do you think?

sigiesec:
I can understand that this is not possible to implement. However, I would like to second that then, ROUTER/DEALER cannot be deprecated/removed. For our use, multi-frame messages are essential as well.

somdoron:
I don't think we can implement SNDMORE with thread safe sockets. some scenario
1. While I write multipart message can I receive in another socket?
2. While I write multipart message can another socket send?
3. What if it take me a long time sending the last time, does all the rest block?

I'm not a big fan of multipart and I think we can live without them, however I understand some people like them. I think the thread safe socket alternative is important and we do see people use them.

We can drop the plan to deprecate or remove not thread safe sockets and fix the docs.

I did have a plan for multipart like support for thread safe socket through a binary message, like czmq bsend and zproto protocol. I even started implement that a year ago, I can try and complete it.

minrk:
I'd trade threadsafe for SNDMORE on the new sockets any day, but I totally understand that it solves a problem for some. I'm fine if all that's done is removing the deprecation comment in the docs.

I could imagine, for instance, a probably janky solution where `send(SNDMORE)` acquired a socket-specific message-assembly lock and the final `send(0)` released it  (don't ask me how to keep track of the lock!), preserving the atomicity of a multipart message.

[...]

PR #2700, opened 2017-08-18
Problems: free of stack variable, TODO left to evaluate

### LINKED ISSUE #2701: pgm build via cmake doesn't link libzmq with libpgm
steps to reproduce on linux

```
mkdir -p build && cd build && cmake -DWITH_OPENPGM=ON .. && make
sudo ldd lib/libzmq.so
```

The listing of linked libraries doesn't have the libpgm.

The build via `./autogen.sh && ./configure --with-pgm` works fine

bluca:
Solutions: see commits

Fixes #2701 #2699


sigiesec:
@bluca, thanks for correcting my mistake in the call to free. It is quite hard to modify the source without having a chance to run or even compile it.

bluca:
No problem!

bluca:
@sigiesec if you are on Linux, with the changes in this PR you can install the library and build with it:

```
sudo apt install libkrb5-dev
./configure .... --with-libgssapi_krb5=yes
```

It won't run the test unless you setup the whole kerberos thing, but at least you can check it builds.

I'll see if I can add it by default to the packages builds on OBS, and then on a couple of CI jobs as well that already install packages.

tests/test_security_gssapi.cpp
```diff
@@ -179,19 +185,23 @@ void test_unauth_creds (void *ctx, void *server, void *server_mon, char *endpoin
...
     rc = zmq_connect (client, endpoint);
     assert (rc == 0);
 
     expect_bounce_fail (server, client);
     close_zero_linger (client);
 
+#ifdef ZMQ_BUILD_DRAFT_API
     int event = get_monitor_event (server_mon, NULL, NULL);
-    assert (event == ZMQ_EVENT_HANDSHAKE_FAILED);
+    assert (event == ZMQ_EVENT_HANDSHAKE_FAILED_AUTH);
```
bluca: @sigiesec I took a very wild guess here and in the next events, is this correct?

bluca:
There's a few errors in the GSSAPI code with older GCC on Debian 7, Ubuntu 12.04 and RHEL 7. I'll have a look during the weekend:

https://build.opensuse.org/package/live_build_log/home:bluca:branches:network:messaging:zeromq:git-draft/libzmq/Debian_7.0/x86_64

tests/test_security_gssapi.cpp
sigiesec: This one looks good.

tests/test_security_gssapi.cpp
```diff
@@ -205,8 +215,10 @@ void test_null_creds (void *ctx, void *server, void *server_mon, char *endpoint)
...
```
sigiesec: This should be ZMQ_EVENT_HANDSHAKE_FAILED_PROTOCOL with value ZMQ_PROTOCOL_ERROR_ZMTP_MECHANISM_MISMATCH  as in https://github.com/zeromq/libzmq/blob/7283574ceedbd4569f514d8a8a86117a0efbf65f/tests/test_security_curve.cpp#L178

Maybe expect_zmtp_mechanism_mismatch should be moved to testutil_security.hpp and reused here (and for the PLAIN client test as well)

tests/test_security_gssapi.cpp
```diff
@@ -292,23 +304,30 @@ int main (void)
...
     rc = zmq_bind (server, "tcp://127.0.0.1:*");
     assert (rc == 0);
     rc = zmq_getsockopt (server, ZMQ_LAST_ENDPOINT, my_endpoint, &len);
     assert (rc == 0);
 
+#ifdef ZMQ_BUILD_DRAFT_API
     //  Monitor handshake events on the server
     rc = zmq_socket_monitor (server, "inproc://monitor-server",
-            ZMQ_EVENT_HANDSHAKE_SUCCEED | ZMQ_EVENT_HANDSHAKE_FAILED);
+            ZMQ_EVENT_HANDSHAKE_SUCCEEDED | ZMQ_EVENT_HANDSHAKE_FAILED_AUTH);
```
sigiesec: Here, ZMQ_EVENT_HANDSHAKE_FAILED_PROTOCOL must also be enabled.

tests/test_security_gssapi.cpp
bluca: Thanks, sent a quick fix here: https://github.com/zeromq/libzmq/pull/2706

PR #2702, opened 2017-08-19
Problems: GSSAPI builds broken with picky compilers, Debian/RPM packages do not use GSSAPI, no CI jobs that test GSSAPI

bluca:
Solutions: see commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
       apt:
         packages:
         - valgrind
-  - env: BUILD_TYPE=default CURVE=libsodium GSSAPI=enabled
+  - env: BUILD_TYPE=default CURVE=libsodium GSSAPI=enabled PGM=enabled
     os: linux
     addons:
       apt:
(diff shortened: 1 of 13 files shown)


bluca:
@sigiesec when you have time, please have a look at the GSSAPI mechanism changes, I think they match but you never know.
Also a while ago you asked for CI jobs to test GSSAPI, I've added it to a couple of runs (PGM also) where it does not impact the runtime as it was already installing packages.

c-rack:
@bluca so is this ready for merge or shall I wait for feedback from @sigiesec ?

bluca:
@c-rack it is ready, as it builds and runs as far as I can tell. If I botched anything we can revise :-)

c-rack:
👍

bluca:
Thanks!

sigiesec:
@bluca Your changes look good :)

Great that we now have GSSAPI and PGM builds in the CI, this allows making changes with more confidence :)

bluca:
Are you developing on Linux? If so you can install the packages and have the code build on your machine too

sigiesec:
No, unfortunately I am developing under Windows.

bluca:
Ok. PGM should work, at the very least CMake has support for building with it under Windows, gut GSSAPI is autotools-only at the moment I'm afraid.

PR #2712, opened 2017-08-22
Problem: no tests for socket-related functions passing NULL as socket

sigiesec:
Solution: added tests


sigiesec:
Why are the DRAFT builds failing? https://travis-ci.org/sigiesec/libzmq/jobs/267200744#L1009

bluca:
don't include <zmq.h>, it's already in testutil.hpp and it should be after the platform.hpp

sigiesec:
@bluca Great, thanks :)

Issue #2713, opened 2017-08-22
Use of EAGAIN vs. ETIMEDOUT


sigiesec:
Almost all libzmq public API functions that work with a timeout return EAGAIN when the timeout has been reached. However, zmq_poller_wait, zmq_poller_wait_all and zmq_poller_poll return ETIMEDOUT instead. 

At least under Windows ETIMEDOUT resp. WSAETIMEDOUT is used only to indicate connection timeouts, which is also an argument for EAGAIN.

Is there a specific reason for using ETIMEDOUT? Otherwise I would change these uses to EAGAIN as well.

bluca:
This is one for @somdoron I think :-)

somdoron:
Good point. I think only those method explicitly accept timeout as parameter while the rest don't.

somdoron:
We can consider the change, however the API already in use (although marked as draft) so people might hate us.

sigiesec:
Eh... if we cannot change it then why is it marked as draft? 

Apart from that, I think it is a good thing that zmq_poller_* is still marked as draft, since e.g. #2623 is still open, which might involve more significant changes than aligning the error codes.

bluca:
Yeah I agree, if it's marked as DRAFT it's fair game, it's documented well enough (I think?) why that system exists and what it implies to enable those APIs

bluca:
For the docs I opened this to track it: https://github.com/zeromq/libzmq/issues/2188

somdoron:
PR is of coursed welcomed

sigiesec:
@somdoron I already fixed it, will send a PR tonight.

I wanted to "assign" the issue (and others) to myself, but apparently I cannot do that. Isn't it sufficient that I am a "member"?

paddor:
The change makes sense. Not hating anyone, barely annoyed. But no hate. :) It's marked as draft, after all.

PR #2714, opened 2017-08-22
Problem: insufficient tests for zmq_poller_*

sigiesec:
Solution: added test cases, aligned behaviour for some corner cases, removed unreachable code


sigiesec:
Fixes #2713

PR #2718, opened 2017-08-25
Problem: bug reports are often missing relevant information

sigiesec:
Solution: provide template for bug reports


bluca:
Thanks, wanted to do this for a while but never got around to it

Issue #2679, opened 2017-08-14
zthread dies after clone() with Permission denied


jimklimov:
Got hit by a strange issue while debugging an application server: I compile it as unprivileged user, then to simplify testing I `sudo` it to start as root and use production configs, where it drops privilege to webserver user. Later on it has bits talking ZMQ, and with such way of launching it fails.

The trace looks like this in the end:

````
[pid 32203] clone(child_stack=0x7f7203ffefb0, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tidptr=0x7f7203fff9d0, tls=0x7f7203fff700, child_tidptr=0x7f7203fff9d0) = 32758
[pid 32203] sched_getparam(32758, { 0 }) = 0
[pid 32203] sched_getscheduler(32758)   = 0 (SCHED_OTHER)
[pid 32203] sched_setscheduler(32758, SCHED_OTHER, { 0 }) = 0
[pid 32203] open("/proc/self/task/32758/comm", O_RDWR) = -1 EACCES (Permission denied)
[pid 32203] write(2, "Permission denied (../libzmq/src/thread.cpp:178)\n", 85) = 85
[pid 32203] rt_sigprocmask(SIG_UNBLOCK, [ABRT], NULL, 8) = 0
[pid 32203] tgkill(30551, 32203, SIGABRT) = 0
````

seems like it cloned off a thread to talk to, but could not manage it?

We have no issue with the same codebase running under normal conditions (a systemd service starting as `root` and dropping privilege to webserver), so one unproven guess is that this may be linked to one extra `sudo` that developer-runs start with, and/or that systemd sets up some CAP's and privileges that we don't in command line...

bluca:
This is weird... I also have applications that drop privileges and then start threads and I've never seen this

sigiesec:
This happens when setting the name of the thread: https://github.com/zeromq/libzmq/blob/5c981be9f838958900b911232233644fb744e243/src/thread.cpp#L177

Since this is not essential, it would probably be better to ignore errors when trying to set the thread name?

bluca:
Well spotted, and yes ignoring the error sounds reasonable.

bluca:
@jimklimov want to send a PR to fix it?

jimklimov:
[...]

PR #2719, opened 2017-08-25
Problem: cosmetic setThreadName() can be fatal

jimklimov:
Solution: do not assert() the values returned; if we failed to set the thread name - we just have harder debugging, not flawed production conditions.

Closes github issue #2679

Signed-off-by: Jim Klimov <EvgenyKlimov@eaton.com>


c-rack:
Thanks!

Issue #2623, opened 2017-07-14
Inconsistencies with ZMQ_POLLOUT and ZMQ_ROUTER in ZMQ_MANDATORY mode


msune:
(Related to PR #2622; partial fix, API and ABI compatible)

I am copying the description of the problem here from #2622:

> If `ZMQ_ROUTER_MANDATORY` is set, and `zmq_poll()` waits for `ZMQ_POLLOUT` events, zmq_poll() will immediately wake up if only 1 pipe has room to send, regardless of the peer, creating a busy loop of `zmq_poll()` wake-up, `zmq_send()` (`EAGAIN`). There is no way for the application to selectively wait for `ZMQ_POLLOUT` for specific peer(s), which seems somehow necessary in `ZMQ_ROUTER_MANDATORY`.

Some possible solutions come to my mind now (obviously, all of them breaking API and ABI unless implemented in parallel):

1) Change `zmq_poll_item_t` to accept a list of `ZMQ_IDENTITY` to `ZMQ_POLLOUT` as an optional parameter. If none is specified, then current behaviour should be maintained.

2) Create `zmq_poll`able peer `zmq_sock_t` (fake), and open APIs to get those `zmq_sock_t` given `ZMQ_IDENTITY` to be placed as regular `zmq_poll_item_t`.

I tend to think 1) is a bit easier to implement and simpler to use. I would like to know the opinions. We could try to work a bit on it, either on `zmq_poll()` itself or a parallel version. But I want to make sure there are chances to be accepted before committing any efforts.

Thoughts?

bluca:
Sorry but we can't break zmq_poll compatibility, as its API is finalised.

But we have in development a new polling implementation, you can find it as zmq_poller - I think that would be a good place to implement this functionality.
That API has not yet been declared STABLE, so it can be changed if needed.

https://github.com/zeromq/libzmq/blob/master/include/zmq.h#L585
https://github.com/zeromq/libzmq/blob/master/src/zmq.cpp#L754
https://github.com/zeromq/libzmq/blob/master/src/socket_poller.cpp

msune:
@bluca I will have a look into the poller, and see if it can be easily implemented. We can keep this issue for discussion or you can close.

bluca:
Thanks! It's fine to keep this open for discussion. Let us know if you need any help.

somdoron:
I think I might have a simple solution, right now we have  zmq_poller_add which accept events to listen to.
[...]

PR #2730, opened 2017-09-01
Problem: unable to query state of a router for a particular peer

sigiesec:
Solution: add zmq_socket_get_peer_state function

This addresses #2623 at least partially.

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -443,6 +443,27 @@ zmq::blob_t zmq::router_t::get_credential () const
     return fq.get_credential ();
 }
 
+int zmq::router_t::get_peer_state (const void *identity,
+                                   size_t identity_size) const
+{
+    int res = 0;
+
+    blob_t identity_blob ((unsigned char *) identity, identity_size);
+    outpipes_t::const_iterator it = outpipes.find (identity_blob);
+    if (it == outpipes.end ()) {
+        errno = EHOSTUNREACH;
+        return -1;
+    }
+
+    const outpipe_t &outpipe = it->second;
+    if (outpipe.pipe->check_hwm ())
+        res |= ZMQ_POLLOUT;
+
+    /** \todo does it make any sense to check the inpipe as well? */
+
+    return res;
+}
+
 bool zmq::router_t::identify_peer (pipe_t *pipe_)
 {
     msg_t msg;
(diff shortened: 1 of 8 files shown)


bluca:
The new API def should also be added to src/drafts.hpp

sigiesec:
@bluca You probably refer to src/zmq_draft.h, right?

bluca:
Yeah, sorry

sigiesec:
NP, done :)

src/router.cpp
```diff
@@ -443,6 +443,27 @@ zmq::blob_t zmq::router_t::get_credential () const
...
+    if (it == outpipes.end ()) {
+        errno = EHOSTUNREACH;
+        return -1;
+    }
+
+    const outpipe_t &outpipe = it->second;
+    if (outpipe.pipe->check_hwm ())
+        res |= ZMQ_POLLOUT;
+
+    /** \todo does it make any sense to check the inpipe as well? */
```
somdoron: I think it does make sense

src/zmq.cpp
```diff
@@ -247,6 +247,16 @@ int zmq_ctx_destroy (void *ctx_)
 
 // Sockets
 
+static zmq::socket_base_t *as_socket_base_t (void *s_)
```
somdoron: should this be a macro instead?

sigiesec: What would be the benefit of a macro? Using a function is more readable/debuggable/...

A macro could be used to further reduce the code duplication at the call sites, as it could include the method return. But this might impair readability.

somdoron: Actually I don't really care, it just been the standard in the project so far. I'm good with both.
Anyway two benefits of macro:
1. Performance - those methods are being called a lot (millions of time a second), I'm not sure how the compiler will treat them, we just added a function call. It is probably nothing. however I saw in the past how inlining and macro improve performance.
2. You would'n need the if check on the return value of the function. Just call the MACRO in the beginning of every usage.

bluca:
@sigiesec the zmq_socket_get_peer_state API should get a manpage at some point (currently compiling rel notes for 4.2.3)

sigiesec:
@bluca Certainly... maybe when it's clear how #2742 will be handled.

bluca:
yes makes sense

PR #2734, opened 2017-09-04
Variable is assigned but is not used until the end of the function

pavel-pimenov:
Fix https://www.viva64.com/en/w/V1001/print/

### DIFF
--- a/src/mechanism.cpp
+++ b/src/mechanism.cpp
@@ -166,7 +166,7 @@ void zmq::mechanism_t::make_command_with_basic_properties (
     memcpy (ptr, prefix, prefix_len);
     ptr += prefix_len;
 
-    ptr += add_basic_properties (
+    add_basic_properties (
       ptr, command_size - (ptr - (unsigned char *) msg_->data ()));
 }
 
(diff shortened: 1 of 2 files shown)


jimklimov:
I think these could have their use in a couple of cases:
* debugging with stepping through code and looking at the variables' values (e.g. where `ptr` points to now);
* extending these functions in the future so new code looks at the proper pointed-to location.

Perhaps a comment (esp. to cater for case 2) or a special comment to make compiler/linter happier would be a better way to address such cliffhangers.

bluca:
Sounds like a good idea to me. But I'm not sure how to make that static analyser ignore the warning.

bluca:
@pavel-pimenov any idea?

pavel-pimenov:
for PVS-studio: //-Vxxx
https://www.viva64.com/en/m/0017/#ID0E1G

Example:

#ifdef _DEBUG
    ptr += //-V1001
#endif
        add_basic_properties (
      ptr, command_size - (ptr - (unsigned char *) msg_->data ()));

.....

    *ptr++ = static_cast <unsigned char> (password.length ());
    memcpy (ptr, password.c_str (), password.length ());
#ifdef _DEBUG
    ptr += password.length (); //-V1001
#endif

Issue #2736, opened 2017-09-05
Add "STATISTICS" command to zmq_proxy_steerable()


f18m:
# Issue description
Currently there is no visibility on how many packets / bytes a ZMQ proxy has processed.
A way to retrieve such information would be important, at the very least, for debugging purpose.
Attached is the proposed patch to ZMQ master branch to add such feature.

# Patch description

On libzmq/src/proxy.cpp the following changes are done:

- the forward() function keeps track of how many packets/bytes are forwarded through the proxy
- the proxy control socket now handles an additional "STATISTICS" command; a reply is sent over the control socket (REP/REQ type should be used in such case for the control socket) with both frontend and backend socket stats.

On the tests_proxy.cpp file:

- small bugfix: when verbose = 1 the control-socket commands were not NUL-terminated. The patch adds NUL-termination that makes stdout more clear
- the control sockets for CLIENTS and SERVER WORKERS are split in 2 sockets; this allows to stop all threads except the proxy; this way the test can be more reliable since we can ensure that the proxy has processed all messages before assert()ing on the number of messages that went through it.
- a "STOP" command is added for CLIENTS and SERVER WORKERS: when the STOP command is received everybody stops doing zmq_send() and just listen on its sockets. In this way the chattering stops and we can reliably assert that the number on the number of processed packets.
- each CLIENT and SERVER WORKER keeps track using zmq atomics of how many requests/replies has sent
- the main() will use the STATISTICS command of the proxy to verify that all messages sent by clients and server workers were correctly processed by the proxy.

# What's the expected result?

"make check" should pass all tests.

f18m:
Attached patch file:

[libzmq_proxy_statistics.patch.txt](https://github.com/zeromq/libzmq/files/1277046/libzmq_proxy_statistics.patch.txt)

bluca:
Don't record the stats unconditionally - add an ON/OFF command (eg: STATS_ON STATS_OFF). Measurements have impact on performance, and if somebody is not interested there is no point in wasting resources.

Also the statistics message, instead of packing everything into a single buffer which makes it very akward toe extract, use one frame per statistics in a multipart msg.

Otherwise looks good, thansk!

bluca:
Also maybe rather than packets, call them messages in the stats, since packets might be misleading given it's counting messages

f18m:
Hi Luca,
thanks for taking a look at the patch so quickly!

> Don't record the stats unconditionally - add an ON/OFF command (eg:
[...]

PR #2737, opened 2017-09-05
Add "STATISTICS" command to zmq_proxy_steerable() 

f18m:
# Pull Request Notice

```
Problem:
Currently there is no visibility on how many packets / bytes a ZMQ proxy has processed on its frontend/backend socket. The only possible way is using the "capture" socket but that seems overkill to just keep track of packets/bytes passed through the proxy.
A way to retrieve such information would be important, at the very least, for debugging purpose.

Solution:
Attached is the proposed patch to ZMQ master branch to add such feature.
The solution is just to increment counters in the proxy.cpp forward() function and add a new command on the proxy control socket to retrieve them.
```

Fixes https://github.com/zeromq/libzmq/issues/2736

### DIFF
--- a/src/proxy.cpp
+++ b/src/proxy.cpp
@@ -154,9 +154,9 @@ int forward (
 
     // A multipart message counts as 1 packet:
     from_stats->msg_in++;
-    from_stats->bytes_in+=complete_msg_size;
+    from_stats->bytes_in += complete_msg_size;
     to_stats->msg_out++;
-    to_stats->bytes_out+=complete_msg_size;
+    to_stats->bytes_out += complete_msg_size;
 
     return 0;
 }
@@ -169,11 +169,11 @@ int reply_stats(
     // first part: frontend stats
 
     zmq::msg_t stats_msg1, stats_msg2;
-    int rc = stats_msg1.init_size( sizeof(zmq_socket_stats_t) );
+    int rc = stats_msg1.init_size (sizeof(zmq_socket_stats_t));
     if (unlikely (rc < 0))
         return close_and_return (&stats_msg1, -1);
 
-    memcpy( stats_msg1.data(), (const void*) frontend_stats, sizeof(zmq_socket_stats_t) );
+    memcpy (stats_msg1.data(), (const void*) frontend_stats, sizeof(zmq_socket_stats_t));
 
     rc = control_->send (&stats_msg1, ZMQ_SNDMORE);
     if (unlikely (rc < 0))
@@ -181,10 +181,10 @@ int reply_stats(
 
     // second part: backend stats
 
-    rc = stats_msg2.init_size( sizeof(zmq_socket_stats_t) );
+    rc = stats_msg2.init_size (sizeof(zmq_socket_stats_t));
     if (unlikely (rc < 0))
         return close_and_return (&stats_msg2, -1);
-    memcpy( stats_msg2.data(), (const void*) backend_stats,  sizeof(zmq_socket_stats_t) );
+    memcpy (stats_msg2.data(), (const void*) backend_stats, sizeof(zmq_socket_stats_t));
 
     rc = control_->send (&stats_msg2, 0);
     if (unlikely (rc < 0))
@@ -570,7 +570,7 @@ int zmq::proxy (
                         {
                             rc = reply_stats(control_, &frontend_stats, &backend_stats);
                             if (unlikely (rc < 0))
-                                return -1;
+                                return close_and_return (&msg, -1);
                         }
                         else {
                             //  This is an API error, we assert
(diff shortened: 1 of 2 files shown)


src/proxy.cpp
```diff
@@ -129,12 +147,57 @@ int forward (
...
+    if (unlikely (rc < 0))
+        return close_and_return (&stats_msg1, -1);
+
+    memcpy( stats_msg1.data(), (const void*) frontend_stats, sizeof(zmq_socket_stats_t) );
+
+    rc = control_->send (&stats_msg1, ZMQ_SNDMORE);
+    if (unlikely (rc < 0))
+        return close_and_return (&stats_msg1, -1);
+
+    stats_msg1.close();
```
bluca: don't close messages unless there was an error - otherwise the buffer will be deallocated under the I/O thread's feet

src/proxy.cpp
```diff
@@ -129,12 +147,53 @@ int forward (
...
+
         if (more == 0)
             break;
     }
+
+    // A multipart message counts as 1 packet:
+    from_stats->msg_in++;
+    from_stats->bytes_in+=complete_msg_size;
+    to_stats->msg_out++;
+    to_stats->bytes_out+=complete_msg_size;
```
sigiesec: Spaces are missing around the operator

f18m:
You're right! Fixed it! thanks



2017-09-05 15:42 GMT+02:00 Luca Boccassi <notifications@github.com>:

src/proxy.cpp
```diff
@@ -129,12 +147,53 @@ int forward (
...
+    rc = control_->send (&stats_msg1, ZMQ_SNDMORE);
+    if (unlikely (rc < 0))
+        return close_and_return (&stats_msg1, -1);
+
+    // second part: backend stats
+
+    rc = stats_msg2.init_size( sizeof(zmq_socket_stats_t) );
+    if (unlikely (rc < 0))
+        return close_and_return (&stats_msg2, -1);
+    memcpy( stats_msg2.data(), (const void*) backend_stats,  sizeof(zmq_socket_stats_t) );
```
sigiesec: There should be a space after the function name, but no spaces inside the braces

src/proxy.cpp
```diff
@@ -491,16 +566,24 @@ int zmq::proxy (
...
                         state = terminated;
                     else {
-                        //  This is an API error, so we assert
-                        puts ("E: invalid command sent to proxy");
-                        zmq_assert (false);
+                        if (msg.size () == 10 && memcmp (msg.data (), "STATISTICS", 10) == 0)
+                        {
+                            rc = reply_stats(control_, &frontend_stats, &backend_stats);
+                            if (unlikely (rc < 0))
+                                return -1;
```
sigiesec: Shouldn't this be `return close_and_return (&msg, -1);` ?

tests/test_proxy.cpp
```diff
@@ -48,13 +48,31 @@
 #define ID_SIZE_MAX 32
 #define QT_WORKERS    5
 #define QT_CLIENTS    3
-#define is_verbose 0
+#define is_verbose 1
```
sigiesec: Macros should be all-uppercase

tests/test_proxy.cpp
```diff
@@ -173,13 +208,13 @@ server_task (void *ctx)
     assert (rc == 0);
 
     // Control socket receives terminate command from main over inproc
-    void *control = zmq_socket (ctx, ZMQ_SUB);
+    void *control = zmq_socket (ctx, ZMQ_REP);
     assert (control);
-    rc = zmq_setsockopt (control, ZMQ_SUBSCRIBE, "", 0);
-    assert (rc == 0);
+    /*rc = zmq_setsockopt (control, ZMQ_SUBSCRIBE, "", 0);
+    assert (rc == 0);*/
```
sigiesec: Remove the commented-out code, or add a comment explaining why it makes sense to keep it.

tests/test_proxy.cpp
bluca: it's a pre-existing one, so it's fine to avoid mixing refactor and new code additions

sigiesec: You're right, I did not see that

f18m:
Thanks @sigiesec  for the review, I fixed all the points hopefully!

bluca:
Thanks @f18m 

Now for your git workflow, here's a one-time step:

```
git remote add upstream https://github.com/zeromq/libzmq.git
```

Now you'll need to sync your local repository with the changes:

```
git fetch upstream
git checkout master
git reset --hard upstream/master
```
Careful as the reset --hard throws away all the local changes on the checked out branch.

You need it now because you used your master branch instead of another branch.

Next time you have a piece of work to do, create a new branch instead:

```
git checkout master
git pull upstream master
git checkout -b foo
```

f18m:
Thanks! That was a very fast integration!
About git branches: I will keep that in mind for future works.

PR #2742, opened 2017-09-06
[WIP][discussion] Adding _peer functions for zmq_poller #2623

msune:
As recommended by @somdoron, opening this PR for discussion.

Here uou can find some preliminary, incomplete work here, but I think sufficient to trigger discussion. I will continue tomorrow, but I would appreciate early feedback before going any much further.

## Short summary

Key 4 additional methods are:

```
 +ZMQ_EXPORT int  zmq_poller_add_peer (void *poller, void *socket, void *zid,
 +                                                            int zid_len,
 +                                                            void *user_data,
 +                                                            short events);
 +ZMQ_EXPORT int  zmq_poller_modify_peer (void *poller, void *socket, void *zid,
 +                                                            int zid_len,
 +                                                            short events);
 +ZMQ_EXPORT int  zmq_poller_remove_peer (void *poller, void *socket, void *zid,
 +                                                            int zid_len);

 +ZMQ_EXPORT int zmq_recv_peer (void *s, void *zid, int zid_len, void *buf,
 +                                                               size_t len, 
 +                                                               int flags);
```

I think the usage is clear. One can use regular poller methods to add the entire socket or (or in addition) use _peer to add specific ZMQ identity poller items.

To read from a specific ZMQ identity, `zmq_recv_peer()`

There is probably a better name than `peer` (identity?). Open to suggestions.

## Missing

A lot:

* Validation in all `poller_peer()` methods to check that socket is multi-peer
* Implementation of `zmq_recv_peer()` and support for `ZMQ_PEER_EVENTS` in `getsockopt()` on sockets with multi-peer support
* Tests and validation
* Documentation

## What I don't like

I dislike using `getsockopt()` with an additional struct `zmq_gso_peer_events_t` for `ZMQ_PEER_EVENTS`. I was following the current approach.

I would very much prefer opening an API to the socket base to get that, and expose it through regular C bindings. Comments?

## Some other stuff, unrelated to this feature

* I removed duplicated blobs of code in here: https://github.com/zeromq/libzmq/blob/master/src/socket_poller.cpp#L633-L670 and here: https://github.com/zeromq/libzmq/blob/master/src/socket_poller.cpp#L672-L678 with `zmq::socket_poller_t::zero_trail_events()` and `zmq::socket_poller_t::check_events()`
*  I can't understand why poller and user items are different (but it is late here...). Is there any good reason for doing all this copy, ordering and mangling?

## Thoughts?

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -266,6 +266,13 @@ ZMQ_EXPORT int zmq_msg_init_data (
   zmq_msg_t *msg, void *data, size_t size, zmq_free_fn *ffn, void *hint);
 ZMQ_EXPORT int zmq_msg_send (zmq_msg_t *msg, void *s, int flags);
 ZMQ_EXPORT int zmq_msg_recv (zmq_msg_t *msg, void *s, int flags);
+
+#ifdef ZMQ_BUILD_DRAFT_API
+ZMQ_EXPORT int zmq_mute_peer (void *s_, const void *routing_id,
+				    const int routing_id_len,
+                                    int mute);
+#endif //ZMQ_BUILD_DRAFT_API
+
 ZMQ_EXPORT int zmq_msg_close (zmq_msg_t *msg);
 ZMQ_EXPORT int zmq_msg_move (zmq_msg_t *dest, zmq_msg_t *src);
 ZMQ_EXPORT int zmq_msg_copy (zmq_msg_t *dest, zmq_msg_t *src);
@@ -376,6 +383,23 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
+
+#define ZMQ_PEER_EVENTS 93
+
+#ifdef ZMQ_BUILD_DRAFT_API
+
+/*  getsockopt additional struct to recover peer events                       */
+typedef struct zmq_gso_peer_events_t
+{
+    void *routing_id;
+    int routing_id_len;
+    short events;
+    short reserved;
+}  zmq_gso_peer_events_t;
+
+#define ZMQ_GSO_PEER_EVENTS_SIZE sizeof(zmq_gso_peer_events_t)
+
+#endif //ZMQ_BUILD_DRAFT_API
 
 /*  Message options                                                           */
 #define ZMQ_MORE 1
@@ -670,20 +694,41 @@ typedef struct zmq_poller_event_t
 #else
     int fd;
 #endif
+
+#ifdef ZMQ_BUILD_DRAFT_API
+    void *routing_id;
+    int routing_id_len;
+#endif //ZMQ_BUILD_DRAFT_API
+
     void *user_data;
     short events;
 } zmq_poller_event_t;
 
 ZMQ_EXPORT void *zmq_poller_new (void);
-ZMQ_EXPORT int zmq_poller_destroy (void **poller_p);
-ZMQ_EXPORT int
-zmq_poller_add (void *poller, void *socket, void *user_data, short events);
-ZMQ_EXPORT int zmq_poller_modify (void *poller, void *socket, short events);
-ZMQ_EXPORT int zmq_poller_remove (void *poller, void *socket);
-ZMQ_EXPORT int
-zmq_poller_wait (void *poller, zmq_poller_event_t *event, long timeout);
-ZMQ_EXPORT int zmq_poller_wait_all (void *poller,
-                                    zmq_poller_event_t *events,
+ZMQ_EXPORT int  zmq_poller_destroy (void **poller_p);
+ZMQ_EXPORT int  zmq_poller_add (void *poller, void *socket, void *user_data,
+                                    short events);
+ZMQ_EXPORT int  zmq_poller_modify (void *poller, void *socket, short events);
+ZMQ_EXPORT int  zmq_poller_remove (void *poller, void *socket);
+
+#ifdef ZMQ_BUILD_DRAFT_API
+ZMQ_EXPORT int  zmq_poller_add_peer (void *poller, void *socket,
+                                    void *routing_id,
+                                    int routing_id_len,
+                                    void *user_data,
+                                    short events);
+ZMQ_EXPORT int  zmq_poller_modify_peer (void *poller, void *socket,
+                                    void *routing_id,
+                                    int routing_id_len,
+                                    short events);
+ZMQ_EXPORT int  zmq_poller_remove_peer (void *poller, void *socket,
+                                    void *routing_id,
+                                    int routing_id_len);
+#endif //ZMQ_BUILD_DRAFT_API
+
+ZMQ_EXPORT int  zmq_poller_wait (void *poller, zmq_poller_event_t *event,
+                                    long timeout);
+ZMQ_EXPORT int  zmq_poller_wait_all (void *poller, zmq_poller_event_t *events,
                                     int n_events,
                                     long timeout);
 
@@ -699,9 +744,11 @@ ZMQ_EXPORT int zmq_poller_modify_fd (void *poller, int fd, short events);
 ZMQ_EXPORT int zmq_poller_remove_fd (void *poller, int fd);
 #endif
 
+#ifdef ZMQ_BUILD_DRAFT_API
 ZMQ_EXPORT int zmq_socket_get_peer_state (void *socket,
                                           const void *routing_id,
-                                          size_t routing_id_size);
+                                          const size_t routing_id_size);
+#endif //ZMQ_BUILD_DRAFT_API
 
 /******************************************************************************/
 /*  Scheduling timers                                                         */
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -228,16 +228,31 @@ zmq::socket_base_t::socket_base_t (ctx_t *parent_,
     }
 }
 
-int zmq::socket_base_t::get_peer_state (const void *routing_id_,
-                                        size_t routing_id_size_) const
-{
-    LIBZMQ_UNUSED (routing_id_);
-    LIBZMQ_UNUSED (routing_id_size_);
+#ifdef ZMQ_BUILD_DRAFT_API
+
+int zmq::socket_base_t::mute_peer (const void *routing_id,
+                                const int routing_id_size,
+                                const bool mute){
+    LIBZMQ_UNUSED (routing_id);
+    LIBZMQ_UNUSED (routing_id_size);
 
     //  Only ROUTER sockets support this
     errno = ENOTSUP;
     return -1;
 }
+
+int zmq::socket_base_t::get_peer_state (const void *routing_id,
+                                        size_t routing_id_size) const
+{
+    LIBZMQ_UNUSED (routing_id);
+    LIBZMQ_UNUSED (routing_id_size);
+
+    //  Only ROUTER sockets support this
+    errno = ENOTSUP;
+    return -1;
+}
+
+#endif // ZMQ_BUILD_DRAFT_API
 
 zmq::socket_base_t::~socket_base_t ()
 {
@@ -424,6 +439,11 @@ int zmq::socket_base_t::getsockopt (int option_,
         return do_getsockopt<int> (optval_, optvallen_,
                                    (has_out () ? ZMQ_POLLOUT : 0)
                                      | (has_in () ? ZMQ_POLLIN : 0));
+    }
+
+    if (option_ == ZMQ_PEER_EVENTS) {
+            errno = EINVAL;
+            return -1;
     }
 
     if (option_ == ZMQ_LAST_ENDPOINT) {
--- a/src/socket_poller.cpp
+++ b/src/socket_poller.cpp
@@ -99,13 +99,25 @@ bool zmq::socket_poller_t::check_tag ()
 }
 
 int zmq::socket_poller_t::add (socket_base_t *socket_,
-                               void *user_data_,
+#ifdef ZMQ_BUILD_DRAFT_API
+                               void *routing_id,
+                               int routing_id_len,
+#endif // ZMQ_BUILD_DRAFT_API
+
+                               void* user_data_,
                                short events_)
 {
     for (items_t::iterator it = items.begin (); it != items.end (); ++it) {
+#if ! defined(ZMQ_BUILD_DRAFT_API)
         if (it->socket == socket_) {
-            errno = EINVAL;
-            return -1;
+#else
+        if (it->socket == socket_ && it->routing_id_len == routing_id_len) {
+            if(memcmp(it->routing_id, routing_id, routing_id_len) == 0)
+#endif // ZMQ_BUILD_DRAFT_API
+            {
+                errno = EINVAL;
+                return -1;
+            }
         }
     }
 
@@ -132,6 +144,10 @@ int zmq::socket_poller_t::add (socket_base_t *socket_,
         0,
         user_data_,
         events_
+#ifdef ZMQ_BUILD_DRAFT_API
+        , routing_id,
+        routing_id_len
+#endif // ZMQ_BUILD_DRAFT_API
 #if defined ZMQ_POLL_BASED_ON_POLL
         ,
         -1
@@ -157,6 +173,10 @@ int zmq::socket_poller_t::add_fd (fd_t fd_, void *user_data_, short events_)
         fd_,
         user_data_,
         events_
+#ifdef ZMQ_BUILD_DRAFT_API
+        , NULL,
+        0
+#endif // ZMQ_BUILD_DRAFT_API
 #if defined ZMQ_POLL_BASED_ON_POLL
         ,
         -1
@@ -168,13 +188,23 @@ int zmq::socket_poller_t::add_fd (fd_t fd_, void *user_data_, short events_)
     return 0;
 }
 
-int zmq::socket_poller_t::modify (socket_base_t *socket_, short events_)
+int zmq::socket_poller_t::modify (socket_base_t  *socket_,
+#ifdef ZMQ_BUILD_DRAFT_API
+                               void *routing_id,
+                               int routing_id_len,
+#endif // ZMQ_BUILD_DRAFT_API
+                               short events_)
 {
     items_t::iterator it;
 
     for (it = items.begin (); it != items.end (); ++it) {
-        if (it->socket == socket_)
-            break;
+        if (it->socket != socket_)
+            continue;
+#ifdef ZMQ_BUILD_DRAFT_API
+        if(it->routing_id_len == routing_id_len &&
+            memcmp(it->routing_id, routing_id, routing_id_len) == 0)
+#endif // ZMQ_BUILD_DRAFT_API
+        break;
     }
 
     if (it == items.end ()) {
@@ -210,13 +240,24 @@ int zmq::socket_poller_t::modify_fd (fd_t fd_, short events_)
 }
 
 
-int zmq::socket_poller_t::remove (socket_base_t *socket_)
+int zmq::socket_poller_t::remove (socket_base_t *socket_
+#ifdef ZMQ_BUILD_DRAFT_API
+                               , void *routing_id,
+                               int routing_id_len
+#endif // ZMQ_BUILD_DRAFT_API
+                               )
 {
     items_t::iterator it;
 
     for (it = items.begin (); it != items.end (); ++it) {
-        if (it->socket == socket_)
-            break;
+        if (it->socket != socket_)
+            continue;
+
+#ifdef ZMQ_BUILD_DRAFT_API
+        if(it->routing_id_len == routing_id_len
+            && memcmp(it->routing_id, routing_id, routing_id_len) == 0)
+#endif // ZMQ_BUILD_DRAFT_API
+        break;
     }
 
     if (it == items.end ()) {
@@ -392,6 +433,11 @@ void zmq::socket_poller_t::zero_trail_events (
         events_[i].fd = 0;
         events_[i].user_data = NULL;
         events_[i].events = 0;
+
+#ifdef ZMQ_BUILD_DRAFT_API
+        events_[i].routing_id = NULL;
+        events_[i].routing_id_len = 0;
+#endif // ZMQ_BUILD_DRAFT_API
     }
 }
 
@@ -412,17 +458,42 @@ int zmq::socket_poller_t::check_events (zmq::socket_poller_t::event_t *events_,
         //  The poll item is a 0MQ socket. Retrieve pending events
         //  using the ZMQ_EVENTS socket option.
         if (it->socket) {
-            size_t events_size = sizeof (uint32_t);
+            size_t events_size;
             uint32_t events;
-            if (it->socket->getsockopt (ZMQ_EVENTS, &events, &events_size)
-                == -1) {
-                return -1;
+
+#ifdef ZMQ_BUILD_DRAFT_API
+            if (it->routing_id_len == 0){
+#endif // ZMQ_BUILD_DRAFT_API
+                //  Regular poller item
+                events_size = sizeof (uint32_t);
+                if (it->socket->getsockopt (ZMQ_EVENTS, &events, &events_size)
+                                           == -1) {
+                    return -1;
+                }
+#ifdef ZMQ_BUILD_DRAFT_API
+            } else {
+                //  Peer poller item
+                zmq_gso_peer_events_t peer_events;
+                peer_events.routing_id = it->routing_id;
+                peer_events.routing_id_len = it->routing_id_len;
+                events_size = ZMQ_GSO_PEER_EVENTS_SIZE;
+
+                if (it->socket->getsockopt (ZMQ_PEER_EVENTS,
+                                           (void*)&peer_events,
+                                           &events_size) == -1)
+                    return -1;
+                events = peer_events.events;
             }
+#endif // ZMQ_BUILD_DRAFT_API
 
             if (it->events & events) {
                 events_[found].socket = it->socket;
                 events_[found].user_data = it->user_data;
                 events_[found].events = it->events & events;
+#ifdef ZMQ_BUILD_DRAFT_API
+                events_[found].routing_id = it->routing_id;
+                events_[found].routing_id_len = it->routing_id_len;
+#endif // ZMQ_BUILD_DRAFT_API
                 ++found;
             }
         }
@@ -460,6 +531,10 @@ int zmq::socket_poller_t::check_events (zmq::socket_poller_t::event_t *events_,
                 events_[found].user_data = it->user_data;
                 events_[found].fd = it->fd;
                 events_[found].events = events;
+#ifdef ZMQ_BUILD_DRAFT_API
+                events_[found].routing_id = NULL;
+                events_[found].routing_id_len = 0;
+#endif // ZMQ_BUILD_DRAFT_API
                 ++found;
             }
         }
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -496,6 +496,12 @@ int zmq_mute_peer (void *s_, const void *routing_id, const int routing_id_len,
     zmq::socket_base_t *s = as_socket_base_t (s_);
     if (!s)
         return -1;
+
+    if (!routing_id || routing_id_len == 0) {
+        errno = EINVAL;
+        return -1;
+    }
+
     return s->mute_peer (routing_id, routing_id_len, mute);
 }
 
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -488,6 +488,19 @@ int zmq_recv (void *s_, void *buf_, size_t len_, int flags_)
     return nbytes;
 }
 
+#ifdef ZMQ_BUILD_DRAFT_API
+
+int zmq_mute_peer (void *s_, const void *routing_id, const int routing_id_len,
+                                     int mute)
+{
+    zmq::socket_base_t *s = as_socket_base_t (s_);
+    if (!s)
+        return -1;
+    return s->mute_peer (routing_id, routing_id_len, mute);
+}
+
+#endif //ZMQ_BUILD_DRAFT_API
+
 // Receive a multi-part message
 //
 // Receives up to *count_ parts of a multi-part message.
@@ -1202,6 +1215,8 @@ int zmq_poller_add (void *poller_, void *s_, void *user_data_, short events_)
       ->add (socket, user_data_, events_);
 }
 
+
+
 #if defined _WIN32
 int zmq_poller_add_fd (void *poller_,
                        SOCKET fd_,
@@ -1241,7 +1256,6 @@ int zmq_poller_modify (void *poller_, void *s_, short events_)
     return ((zmq::socket_poller_t *) poller_)->modify (socket, events_);
 }
 
-
 #if defined _WIN32
 int zmq_poller_modify_fd (void *poller_, SOCKET fd_, short events_)
 #else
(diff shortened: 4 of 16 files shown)


include/zmq.h
```diff
@@ -265,6 +265,9 @@ ZMQ_EXPORT int zmq_msg_init_data (zmq_msg_t *msg, void *data,
     size_t size, zmq_free_fn *ffn, void *hint);
 ZMQ_EXPORT int zmq_msg_send (zmq_msg_t *msg, void *s, int flags);
 ZMQ_EXPORT int zmq_msg_recv (zmq_msg_t *msg, void *s, int flags);
+ZMQ_EXPORT int zmq_recv_peer (void *s, void *zid, int zid_len, void *buf,
```
bluca: Please move new APIs in the DRAFT section of the header (and duplicate them in src/zmq_drafts.h)

include/zmq.h
```diff
@@ -639,17 +654,32 @@ typedef struct zmq_poller_event_t
 #else
     int fd;
 #endif
+    void *zid;
```
bluca: if you move these 2 at the bottom we avoid an ABI breakage for users of DRAFT APIs

src/zmq.cpp
```diff
@@ -520,6 +533,47 @@ int zmq_recv (void *s_, void *buf_, size_t len_, int flags_)
     return nbytes;
 }
 
+int zmq_recv (void *s_, void *zid, int zid_len, void *buf_, size_t len_,
```
bluca: zmq_recv_peer ?

src/zmq.cpp
```diff
@@ -481,6 +481,19 @@ s_recvmsg (zmq::socket_base_t *s_, zmq_msg_t *msg_, int flags_)
...
 
+static int
+s_recvmsg_peer (zmq::socket_base_t *s_, void *zid, int zid_len, zmq_msg_t *msg_,
+                                                                int flags_)
+{
+    int rc = s_->recv_peer (zid, zid_len, (zmq::msg_t *) msg_, flags_);
+    if (unlikely (rc < 0))
+        return -1;
+
+    //  Truncate returned size to INT_MAX to avoid overflow to negative values
```
bluca: this is a new API, so there's no need to constrain the return type. If it needs to be bigger than an int, make it so

include/zmq.h
bluca: NVM I'm talking rubbish, if it's on the stack it's going to break anyway

src/socket_poller.cpp
```diff
@@ -401,6 +413,97 @@ void zmq::socket_poller_t::rebuild ()
...
+        events_[i].user_data = NULL;
+        events_[i].events = 0;
+        events_[i].zid = NULL;
+        events_[i].zid_len = 0;
+    }
+}
+
+int zmq::socket_poller_t::check_events(zmq::socket_poller_t::event_t *events_,
+                                                                 int n_events_,
+                                                                 int& found) {
```
somdoron: you can just return found (or -1)

src/socket_poller.cpp
```diff
@@ -401,6 +413,97 @@ void zmq::socket_poller_t::rebuild ()
...
+                    events_[found].events = it->events & events;
+                    events_[found].zid = it->zid;
+                    events_[found].zid_len = it->zid_len;
+                    ++found;
+                }
+            }
+            //  Else, the poll item is a raw file descriptor, simply convert
+            //  the events to zmq_pollitem_t-style format.
+            else {
+                short revents = pollfds [it->pollfd_index].revents;
```
somdoron: what about windows case?

somdoron:
my only concern here is that you did some refactoring together with the change, I think it better to separate them to two PR. Much easier to review the change that way and to avoid bugs (I know it it WIP but like not implemented the windows version of poller)

include/zmq.h
msune: I started putting it into `zmq_drafts.h`, but then for sockopts I had to define them in in `zmq.h` (otherwise I had to add includes to `zmq_drafts.h` all over). Please advise how to proceed better for getsockopts.

src/zmq.cpp
msune: Yup. Code is not tests, no worries.

src/zmq.cpp
msune: OK thanks. I was confused about return values.

src/socket_poller.cpp
msune: :+1:

src/socket_poller.cpp
msune: Well that code is a straight porting of the two repeated blobs of code:

https://github.com/zeromq/libzmq/blob/master/src/socket_poller.cpp#L468-L514
https://github.com/zeromq/libzmq/blob/master/src/socket_poller.cpp#L633-L670

If anything is missing for Windows, then it's also missing in master. But I don't think so, right?

msune:
@somdoron I agree with the comment on splitting the PRs. I will send a PR only with the refactor of these two utility functions this afternoon and then rebase this into that.

include/zmq.h
bluca: All definitions are duplicated - once in the include/zmq.h inside the \#ifdef DRAFT below, and once in src/zmq_drafts.h  where they are all inside a \#if not defined (DRAFT). These way you can have the definitions available internally even if the library is not built by the user with DRAFT APIs.

src/socket_poller.cpp
somdoron: the code is not exactly the same... take a closer look

somdoron: the handling of FD case

src/socket_poller.cpp
```diff
@@ -465,60 +568,12 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_ev
...
-            }
-            //  Else, the poll item is a raw file descriptor, simply convert
-            //  the events to zmq_pollitem_t-style format.
-            else {
-                short revents = pollfds [it->pollfd_index].revents;
-                short events = 0;
-
-                if (revents & POLLIN)
-                    events |= ZMQ_POLLIN;
-                if (revents & POLLOUT)
```
somdoron: This is posix handling...

src/socket_poller.cpp
```diff
@@ -630,52 +685,12 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_ev
...
-                }
-            }
-            //  Else, the poll item is a raw file descriptor, simply convert
-            //  the events to zmq_pollitem_t-style format.
-            else {
-                short events = 0;
 
-                if (FD_ISSET (it->fd, &inset))
-                    events |= ZMQ_POLLIN;
-                if (FD_ISSET (it->fd, &outset))
```
somdoron: this is windows...

include/zmq.h
```diff
@@ -371,6 +374,18 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
 #define ZMQ_VMCI_BUFFER_MAX_SIZE 87
 #define ZMQ_VMCI_CONNECT_TIMEOUT 88
 #define ZMQ_USE_FD 89
+#define ZMQ_PEER_EVENTS 90
+
+/*  getsockopt additional struct to recover peer events                       */
+typedef struct zmq_gso_peer_events_t
```
sigiesec: I think the C++ style `struct zmq_gso_peer_events_t { ... };` should be preferred.

include/zmq.h
```diff
@@ -371,6 +374,18 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
...
+{
+    void *zid;
```
sigiesec: What does "zid" mean? I think the ZMQ API uses the term "identity" as in ZMQ_IDENTITY. This might be prefixed here with "peer", i.e. peer_identity and peer_identity_len
This applies to all other uses of "zid" as well.

(I am not so sure with _len vs. _size; "len" might suggest the identity is a string, which it is not)

include/zmq.h
sigiesec: What does gso mean?

include/zmq.h
```diff
@@ -371,6 +374,18 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
...
+    int zid_len;
+    short events;
+    short padding;
+}  zmq_gso_peer_events_t;
+
+#define ZMQ_GSO_PEER_EVS_SIZE (size_t)sizeof(zmq_gso_peer_events_t)
```
sigiesec: What does EVS mean?

The type of sizeof(T) is defined to be size_t (http://en.cppreference.com/w/cpp/language/sizeof), so the cast is redundant.

include/zmq.h
```diff
@@ -371,6 +374,18 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg, const char *property)
...
 #define ZMQ_USE_FD 89
+#define ZMQ_PEER_EVENTS 90
+
+/*  getsockopt additional struct to recover peer events                       */
+typedef struct zmq_gso_peer_events_t
+{
+    void *zid;
+    int zid_len;
+    short events;
+    short padding;
```
sigiesec: What's the purpose of the padding?

src/socket_poller.cpp
```diff
@@ -94,12 +94,16 @@ bool zmq::socket_poller_t::check_tag ()
...
+int zmq::socket_poller_t::add (socket_base_t *socket_, void *zid, int zid_len,
+                                                       void* user_data_,
+                                                       short events_)
 {
     for (items_t::iterator it = items.begin (); it != items.end (); ++it) {
-        if (it->socket == socket_) {
-            errno = EINVAL;
-            return -1;
+        if (it->socket == socket_ && it->zid_len == zid_len) {
+            if(memcmp(it->zid, zid, zid_len) == 0){
```
sigiesec: missing whitespace in this line

src/socket_poller.cpp
```diff
@@ -149,12 +153,16 @@ int zmq::socket_poller_t::add_fd (fd_t fd_, void *user_data_, short events_)
...
+                                                           int zid_len,
+                                                           short events_)
 {
     items_t::iterator it;
 
     for (it = items.begin (); it != items.end (); ++it) {
-        if (it->socket == socket_)
+        if (it->socket != socket_)
+            continue;
+        if(it->zid_len == zid_len && memcmp(it->zid, zid, zid_len) == 0)
```
sigiesec: missing whitespace in this line

src/socket_poller.cpp
msune: @somdoron I will have a closer look when doing the refactor PR first. Thanks

include/zmq.h
msune: GetSockOpt. Trying to shortening struct name.

include/zmq.h
msune: zid = ZMQ Identity

id is to much overloaded, pid (peer id) was too confusing with posix PID. If you have a better name for it, I am open to change it.

_len vs _size; I can change to _size, if that's making it clear.

include/zmq.h
msune: EVS = events. Again trying to shorten names.

Agree on the redundancy comment. Saw it afterwards

src/socket_poller.cpp
msune: :+1:

src/socket_poller.cpp
msune: :+1: 

I did not check in detail style. Please note this is an RFC PR.

include/zmq.h
sigiesec: Ah ok :) In particular in the public API, abbreviations should be used reluctantly to avoid confusion. 

I am not sure if it can just be named zmq_peer_events_t, this is even shorter, and maybe sufficiently precise.

include/zmq.h
sigiesec: Again, using an abbreviation here is not a good idea at all IMO.

zid/ZMQ identity is confusing, since this prefix is not used for any other field.

The use of the term "identity" may be confusing (I think this was discussed somewhere else), but it is the term the ZMQ API uses. Using another term at some places would be even more confusing.

include/zmq.h
sigiesec: Apart from the general abbreviations discussion, it would be a really good idea to use the same spelling as the name of the type this refers to.

include/zmq.h
somdoron: I think rounting_id might be a solution. It is something that started being used instead of identity few years ago around zmq projects. We can continue and use it (zproto, czmq, SERVER socket type)

sigiesec: Ok... but then ZMQ_ROUTING_ID should be added as a synonym for ZMQ_IDENTITY, and ZMQ_IDENTITY should be declared deprecated.

somdoron: sounds good to me

sigiesec: Hm, and what about ZMQ_MSG_PROPERTY_IDENTITY and its value, "Identity"?

sigiesec: Eh... There already is a pipe_t::routing_id with another meaning. This would need to be renamed to avoid a conflict.

somdoron:
Actually it is same meaning...

include/zmq.h
msune: I don't get it; are you arguing about the use of `typedef`?

Aside from not being a C++ style thing, the entire `zmq.h` has `typedef`s for each and every single of the structs.

Please clarify.

include/zmq.h
msune: Please clarify, whn you reach a consesus

include/zmq.h
sigiesec: Forget about this. This is the C zmq.h, not an internal C++ header. It must be kept as it is.

include/zmq.h
sigiesec: @somdoron Your comment was somehow placed outside this thread... You said it is the same meaning. That cannot be true. pipe_t::routing_id and router_t::next_rid are both uint32_t, rather than a blob. pipe_t has members routing_id and identity, so I cannot rename identity to routing_id, since this already names another member.

somdoron:
Well, not exactly the same, routing id in pipe is like identity but for
SERVER socket type. You call it routerRoutingId and serverRoutingId.

sigiesec:
@somdoron Your comment was placed outside the review again.

ok, I will rename them according to that.

include/zmq.h
msune: Align to 4 byte multiple and save space for any eventual addition (binary compatibility). But it can be removed.

msune:
Waiting for #2744 on one side, and to clarify the identity issues of #2743 to rebase to that code and continue the work.

include/zmq.h
sigiesec: @msune I think we agreed on "routing_id" :) Sorry for the confusion.

include/zmq.h
sigiesec: I think you are making several assumptions on the size of the members and on the alignment that may not be true on all platforms.

Reserving something for addition without breaking binary compatibility may still be a good idea, but then I would rename the member to "reserved".

somdoron:
@sigiesec probably because I answered directly from the email

src/socket_base.cpp
```diff
@@ -436,6 +436,11 @@ int zmq::socket_base_t::getsockopt (int option_, void *optval_,
         return 0;
     }
 
+    if (option_ == ZMQ_PEER_EVENTS) {
```
sigiesec: This is redundant. Any option not specifically handled will lead to EINVAL.

bluca:
@msune Sorry for the delay - severe slowdowns on Travis are holding things back. The identity pr is now merged.

msune:
@bluca Thanks.

I plan to rebase and start to implement it by mid-end of this week. I will keep you posted.

sigiesec:
Even if this PR is updated/completed/merged, I wouldn't declare the newly added functions STABLE at the same time as the existing zmq_poller_* functions. The label can/should be removed.

bluca:
I'm fine either way - just had the impression that this was a desired feature for the new zmq_poller before it could be considered complete, but if it's an optional addition then feel free to remove the label

sigiesec:
It is desired for certain use cases, involving ROUTER sockets, but I think it is not at all clear that this is the desired API. The remainder of the zmq_poller_* API is usable for many use cases without this, and it is quite urgent to have a replacement for zmq_poll.

sigiesec:
I removed the label/milestone.

msune:
I sort of agree this is more of an addition, but it's also true it has substantial additions to the API.  If that wouldn't be consider "stable", and assuming code is solid and tested enough, how much time would it involve to go to stable? What is the API/ABI policy for this?

I already asked this to @bluca, but @bluca @sigiesec is there a timeframe for these new apis to go upstream? I need to allocate time for this, and so far I undestood this was not urgent. Let me know otherwise.

sigiesec:
When the PR is ready to be merged, it will be merged, and they will be automatically included in the next release as DRAFT.

In the RFC draft ( ;) ) it is stated that they may become STABLE when they remained unchanged in one public release AND for at least 6 months. See https://github.com/zeromq/rfc/blob/master/44/README.md, section 2.6 no. 9

bluca:
> I already asked this to @bluca, but @bluca @sigiesec is there a timeframe for these new apis to go upstream? I need to allocate time for this, and so far I undestood this was not urgent. Let me know otherwise.

The process was not defined before, so we have recently amended C4:

https://rfc.zeromq.org/spec:44/C4

TL;DR: must be documented, must have been in at least one public release as DRAFT and must have been published as DRAFT for at least 6 months without changes to the API signatures, and the submitter must not object to the change of state.

I have recently started documenting via issues and milestones the missing steps for the current DRAFT APIs so that it would be clearer for users and devs, hence why the change in labelling

msune:
@sigiesec @bluca Thank you for the pointers. I haven't seen this before.

I find that for new APIs, which don't _break_ existing APIs,  this is a bit to rigid. It can easily go to a 12 months time span between PR is merged and is marked as stable.

Anyway, if that's the case; please let me know what is the plan for new poller APIs to be marked as stable and try to best fit in that window (i.e. I would like to get in the current window and not next one).

bluca:
This is the current tracker:

https://github.com/zeromq/libzmq/milestone/3

It might not be complete - hoping to get help from others, as I really haven't used the new API much, so feel free to create new issues and let me know and I'll tag them

sigiesec:
@msune Indeed, it can easily take 12 months for new APIs to become STABLE, but IMO that is a good thing. There should be a chance to see if the DRAFT API is usable in the field, by many users, in many contexts. Personally, I think that 12 months might even be too short for features that are not widely used within that period. Once the API is STABLE, the API cannot be changed anymore.  The case you cite is essentially the only case. Breaking existing API cannot be done, only new API added and the old API deprecated. Stability of the API is not tied to the production-readiness of the implementation. There is no fixed cycle for changing the state of the API. 

However, I am just curious why it is important for you to see the API declared STABLE?

msune:
> However, I am just curious why it is important for you to see the API declared STABLE?

Sort of a pragmatic thing. Non-stable APIs (i.e. ifdefed) tend to be, in some projects, 2nd class citizens, and prone to be broken. I didn't track close enough libzmq project to know if that's really the case (I know there are tests with and without DRAFT apis, but sometimes this doesn't cut it), 

If I invest (my personal) time on this, is to convince part of my team to adopt these new APIs, as they solve issues we find with the current poller, rather than work-arounding them as we do now. I wouldn't want them to be broken from time to time.

sigiesec:
@msune I understand your concerns. However, tests for the DRAFT APIs are as important as tests for other features, and future PRs will not be merged if they break tests for DRAFT APIs. Insofar, they are 1st class citizens. Furthermore, the tests being stable is one precondition to declare them STABLE.

What can happen is that the API is modified (syntax or semantics) while in DRAFT state, which will require users to adapt. However, the only way to get new STABLE APIs is to go through DRAFT APIs. It would be much more troublesome, if they were declared STABLE prematurely, and then new replacement DRAFT APIs would need to be introduced, the old STABLE APIs deprecated, and both be maintained for a significant time. It can only be speculated how likely this is in case of these APIs.

I hope this is acceptable for you.

Apart from that, I don't think that much work is necessary to get the PR to a mergable state. Some comments from the reviews are open, but I think they are only minor. For rebasing, probably most necessary changes are due to the reformatting of the source code. This will be much less likely in the future, since the formatting is now automatically checked.

msune:
I've rebased this PR and I am working on the missing pieces. Essentially:

* #ifdef everything under DRAFT, as per discussed
* Implement `recv_peer` for `ZMQ_ROUTER`. I don't have much time, so I won't implement it for any other socket (not sure if it's applicable either)
* Add tests and make sure DRAFT and non-DRAFT passes
* Final rebase an turn this into a real PR.

About `identity` vs `routing-id`; sorry I am confused. I saw a PR integrated and then reverted, so I am not sure if the final decision was to stick to identity. @bluca @sigiesec ?

bluca:
it's routing-id

msune:
@bluca 

> it's routing-id

I hate to bang you again with this, specially after refactoring all the code to use `routing-id`.

I've seen in some APIs:

http://api.zeromq.org/4-2:zmq-msg-set-routing-id

`routing-id` is actually an `uint32_t` (`ZMQ_SERVER`).

In the `*peer` APIs, as `routing-id` (at least for `ZMQ_ROUTER`) will actually be what ZMQ manual calls "identity":

>ZMQ_ROUTER
A socket of type ZMQ_ROUTER is an advanced pattern used for extending request/reply sockets. When receiving messages a ZMQ_ROUTER socket shall prepend a message part containing the **identity** of the originating peer to the message before passing it to the application. Messages received are fair-queued from among all connected peers. When sending messages a ZMQ_ROUTER socket shall remove the first part of the message and use it to determine the identity of the peer the message shall be routed to. If the peer does not exist anymore the message shall be silently discarded.
> ...
> When a ZMQ_REQ socket is connected to a ZMQ_ROUTER socket, in addition to the **identity** of the originating peer each message received shall contain an empty delimiter message part. Hence, the entire structure of each received message as seen by the application becomes: one or more identity parts, delimiter part, one or more body parts. When sending replies to a ZMQ_REQ socket the application must include the delimiter part.

That's not an `uint32_t` in the general case, rather a fixed number of bytes, without bounds.

I think using `routing-id` here can be confusing. Can you reconfirm this?

bluca:
Note that page has been updated too, it does not mention identity anymore: https://github.com/zeromq/libzmq/blob/master/doc/zmq_socket.txt#L491

Can it not be 32 bit too?

bluca:
Note that the set_routing_id API is still DRAFT so it can be changed to take a different type if needed

msune:
> Note that page has been updated too, it does not mention identity anymore: >https://github.com/zeromq/libzmq/blob/master/doc/zmq_socket.txt#L491
>
>Can it not be 32 bit too?

I am not familiar with `ZMQ_SERVER`.

We have applications that use `zmq_identity` in `ZMQ_ROUTER` sockets, which is something we _anyway_ have to send from DEALER -> ROUTER to encode the _real_ identity of our application, which in our case is a `uint64_t`. But I can think other may use it as a string or something similar, without a fixed amount of bytes.

So given that, historically, `zmq_identity` was a set of bytes, constraining it to 32 bits, sounds not the best. (Not saying, necessarily, that having `zmq_identity` the way it is was the best of the decisions).

Not sure if the concept is different, but, at least based on the APIs as of now, `zmq_identity` != `routing-id`

bluca:
That's fine, it's a DRAFT API as I mentioned so it can be changed.
@somdoron will know more about the difference in how it is used and why it was picked as an uint32 for ZMQ_SERVER.

sigiesec:
I was also somewhat confused by that. I think that either the integer routing ids should be renamed, or the concepts should be merged, but I think that's outside the scope of this PR. Or do you need to touch the integer routing IDs?

msune:
@sigiesec I don't need to rename it, I think. It's just a matter of putting the right name to the `_peer` APIs

msune:
Guys. I've pushed an updated version. I am still missing the poller tests with `routing-id` and docs. But before I continue;

There is a change on the original proposal that I would like to discuss with you @somdoron @sigiesec @bluca .

I did not like much the effects on having `zmq_recv_peer()`. A user could have called `zmq_recv()` and `zmq_recv_peer` in any particular order. In multi-part messages or on ROUTER sockets (or any other using routing-ids), things were getting complicated; what to do? Fail if we were in the middle of a recv? handle recvs in "parallell"? Library implementation and user code was not so clean.

As an alternative, I've implemented something that can achieve the same result, and perhaps, can be useful on its own `zmq_mute_peer()`. The idea is simple; take out from the fq scheduler pipes referring to a zmq routing-id so that `recv` will not read more from those routing-ids.

Thoughts on this?

I need to add more logic to test multi-part and routing-ids, given the prefetcher on router sockers etc...

sigiesec:
Without digging deeper, It think it would be acceptable to disallow mixing zmq_recv and zmq_recv_peer in the middle of a multipart message, and get an assertion or even undefined behaviour in that case.

somdoron:
> That's fine, it's a DRAFT API as I mentioned so it can be changed.
@somdoron will know more about the difference in how it is used and why it was picked as an uint32 for ZMQ_SERVER.

Actually router routing id are also uint32 (and always have been), however, because router have to also support custom identities we encode the uint32 as 5 bytes, first one indicate if it custom or auto-counter one.

The reason for calling it routingId, is to go away from the identity concept which is buggy (where client decide on his identity). The correct way to have "identity" for a peer socket is by handshake and not by remote peer setting it.

Calling it routing id actually started in the zproto project and we then try to rename the concept to routingId and deprecate the identity thing.

somdoron:
I'm not sure I'm following the mute concept.

I think `zmq_recv_peer` can work, but few semantics are needed:
- `zmq_recv_peer` can only be used as the first calls in a multipart message.
- if not called as first it will return invalid error
- all continues calls must be zmq_recv.

This API will work both for ROUTER and SERVER.

sigiesec:
> The reason for calling it routingId, is to go away from the identity concept which is buggy (where client decide on his identity). The correct way to have "identity" for a peer socket is by handshake and not by remote peer setting it.

What do you mean by "the identity concept [...] is buggy"? The overall system must ensure that the identities are unique. We have cases where there is a fixed set of clients with specific names, and other cases where we generate UUIDs for each client. I am not sure if deprecating this option is the right way.

But maybe I miss something about the bugginess of the concept?

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+    bool d1 = strncmp(buffer, "X", strlen("X")) == 0;
+    rc = zmq_recv (router, buffer, 255, 0);
+    assert (rc == 6);
+	if(d1)
+	    assert (strncmp(buffer, "HelloX", strlen("HelloX")) == 0);
+	else
+	    assert (strncmp(buffer, "HelloY", strlen("HelloY")) == 0);
+    }
+
+    // Mute and unseen routing-id
```
sigiesec: "and" -> "an" ?

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+    rc = zmq_mute_peer (router, "Z", 1, 0x1);
+    assert (rc == -1);
+    assert (errno == EHOSTUNREACH);
+
+    //  Mute X
+    rc = zmq_mute_peer (router, "X", 1, 0x1);
```
sigiesec: This can be simplified (and error reporting improved) by using TEST_ASSERT_SUCCESS_ERRNO as above

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    //  Nothing left on the buffer
+    rc = zmq_recv (router, buffer, 255, ZMQ_DONTWAIT);
+    assert (rc == -1);
+    assert (errno == EAGAIN);
+
+    //  Poller should NOT have IN events
+    rc = zmq_poll (poll_items, 1, 10);
+    assert (rc == 0);
+}
```
sigiesec: The test sockets should be closed using test_context_socket_close, otherwise a warning will be output by the test

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+    //  Poller should have IN events
+    rc = zmq_poll (poll_items, 1, 10);
+    assert (rc == 1);
+
+    //  Now we should only receive from X
+    rc = zmq_recv (router, buffer, 255, 0);
+    assert (rc == 1);
+
+    assert (strncmp(buffer, "X", 1) == 0);
+    rc = zmq_recv (router, buffer, 255, 0);
```
sigiesec: This can be simplified using recv_string_expect_success

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+    setup_test_context ();
+}
+
+void tearDown ()
+{
+    teardown_test_context ();
+}
+
+
+void test_mute_peer ()
```
sigiesec: I am not completely sure, but does this contain multiple independent test cases? It would be desirable to split these up, to allow the tests to run and possibly fail independently.

msune:
> I'm not sure I'm following the mute concept.

`zmq_mute_peer()` will mute in_pipe for that routing-id, so that subsequent calls to `zmq_recv()` or alike, will not read from that pipe. `zmq_poll()` and alike will not give `ZMQ_POLLIN` events either. If a multi-part or routing-id+message is half read, `zmq_recv()` will allow the partially read message to be read, and then mute that routing-id.

One use-case where this is useful; when 2 ROUTER sockets are working with MANDATORY flag, and the application reads from socket1 routingidY and processes and outputs a message on socket2 routingidX for a certain routing-id, if output HWM is hit for that routing-idX in socket2, one could want to keep reading from socket1 except from routingidY.

`zmq_mute_peer()` is complementary to `zmq_recv_peer()` in the sense that one could achieve the same result as mute_peer by manually polling over _the rest_  of identities. However when this group is big, `zmq_mute_peer()` makes things simpler for the user.

>I think zmq_recv_peer can work, but few semantics are needed:
>
>zmq_recv_peer can only be used as the first calls in a multipart message.
> - if not called as first it will return invalid error
> - all continues calls must be zmq_recv.
> - This API will work both for ROUTER and SERVER.

That's what I wanted to discuss. I think this is the only _sane_ way to implement `zmq_recv_peer` and alike. As long as these set of constraints are acceptable, then I will have a look. It will require, much more code.

Another aside topic; maybe someone can shed some light on this. Any idea on why fair-scheduler `fq.hpp` doesn't have any map for routing-id, so that a pipe can be located by it's routing-id instead of looping? I will have to add this for `zmq_recv_peer()`

tests/test_peer_funcs.cpp
msune: Got it

tests/test_peer_funcs.cpp
msune: Will have a look

tests/test_peer_funcs.cpp
msune: Right

tests/test_peer_funcs.cpp
```diff
@@ -0,0 +1,157 @@
...
+    rc = zmq_recv (router, buffer, 255, 0);
+    assert (rc == 6);
+	if(d1)
+	    assert (strncmp(buffer, "HelloX", strlen("HelloX")) == 0);
+	else
+	    assert (strncmp(buffer, "HelloY", strlen("HelloY")) == 0);
+    }
+
+    // Mute and unseen routing-id
+    rc = zmq_mute_peer (router, "Z", 1, 0x1);
```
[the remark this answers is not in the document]
msune: Will have a look

tests/test_peer_funcs.cpp
msune: No. This is a single test case, as it tests the entire mute functionality (and things are missing; e.g. mute in between a multi-part message or routing-id already read).

Splitting this in multiple subtestcases only bloats the code, as it needs init/finit sequence on each test case, or you end up cascading test cases (without fini/init), in which case the main reason to split - "to allow the tests to run and possibly fail independently" - is not anymore valid.

sigiesec: Ok

somdoron:
> But maybe I miss something about the bugginess of the concept?

What if for example you started another client with the same identity?
And this can actually happen in production if the identity is part of a configuration file and such.
Just take a look at the issues, we had a lot about that.

Router actually have two modes to overcome this issue, hijack and reject. 
Hijack came later, to fix the buginess of router, but only for some users, which actually want this behavior.

Identity (in ROUTER paradigm) is a mix of both identity and routing id, which should be separate into two different fields. Routing id assigned from the server, identity comes from the client.

A lot of zeromq users had identity hell, I just had paid client few month ago suffering from such a problem and another one approached me few weeks ago.

With SERVER socket type, we changed that, routing ids are only assigned by the server. Identity is higher level concept and is not handled by SERVER socket type.

include/zmq.h
```diff
@@ -670,20 +694,41 @@ typedef struct zmq_poller_event_t
...
-zmq_poller_wait (void *poller, zmq_poller_event_t *event, long timeout);
-ZMQ_EXPORT int zmq_poller_wait_all (void *poller,
-                                    zmq_poller_event_t *events,
+ZMQ_EXPORT int  zmq_poller_destroy (void **poller_p);
+ZMQ_EXPORT int  zmq_poller_add (void *poller, void *socket, void *user_data,
+                                    short events);
+ZMQ_EXPORT int  zmq_poller_modify (void *poller, void *socket, short events);
+ZMQ_EXPORT int  zmq_poller_remove (void *poller, void *socket);
+
+#ifdef ZMQ_BUILD_DRAFT_API
```
bluca: this section of the header is already in an ifdef, so you can remove this

bluca:
@msune most of the ifdefs can be removed - they are necessary in the header, as that's where the symbols for export are defined, but not elsewhere. It should simplify the diff a bit.

include/zmq.h
```diff
@@ -377,6 +384,23 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg,
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
 
+#define ZMQ_PEER_EVENTS 93
```
bluca: this should be in the DRAFT section as well - please move it and the block below wit the struct after line 619, where "DRAFT Socket options." is - and you can remove the ifdefs

PR #2743, opened 2017-09-06
Problem: term "identity" is confusing

sigiesec:
Solution: replace by "routing id"


sigiesec:
There is some inconsistency, test_stream is failing. I am working on that, along with changing terminology in all of the tests. Will send another PR soon

sigiesec:
The problem is that ZMTP uses "Identity" as a string as well. Of course we cannot change that before a potential ZMTP 4.0

sigiesec:
See #2745

bluca:
Sorry I'll be without laptop until Monday, I can look next week

bluca:
But I can see from my mobile on OBS that all stable builds are failing in the test_stream and all draft builds are not compiling

somdoron:
@sigiesec I will send a revert PR, so please submit a PR again when the issue is solved.

PR #2744, opened 2017-09-06
Problem: duplicated socket_poller::wait() code

msune:
Incorporated feedback from discussing in #2742. Refactor even more code to save some lines here and there.

# Description
zmq::socket_poller_t::wait() had an important set of common lines
between POLL and SELECT variant.

Solution: refactor zmq::socket_poller_t::wait() and add the
following methods:

zmq::socket_poller_t::zero_trail_events()
zmq::socket_poller_t::check_events()
zmq::socket_poller_t::adjust_timeout()

### DIFF
--- a/src/socket_poller.cpp
+++ b/src/socket_poller.cpp
@@ -401,7 +401,135 @@ void zmq::socket_poller_t::rebuild ()
     need_rebuild = false;
 }
 
-int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_events_, long timeout_)
+void zmq::socket_poller_t::zero_trail_events (
+                                         zmq::socket_poller_t::event_t *events_,
+                                         int n_events_,
+                                         int found)
+{
+    for (int i = found; i < n_events_; ++i) {
+        events_[i].socket = NULL;
+        events_[i].fd = 0;
+        events_[i].user_data = NULL;
+        events_[i].events = 0;
+    }
+}
+
+#if defined ZMQ_POLL_BASED_ON_POLL
+int zmq::socket_poller_t::check_events (zmq::socket_poller_t::event_t *events_,
+                                                              int n_events_)
+#elif defined ZMQ_POLL_BASED_ON_SELECT
+int zmq::socket_poller_t::check_events (zmq::socket_poller_t::event_t *events_,
+                                                              int n_events_,
+                                                              fd_set& inset,
+                                                              fd_set& outset,
+                                                              fd_set& errset)
+#endif
+{
+    int found = 0;
+    for (items_t::iterator it = items.begin (); it != items.end () &&
+                                                  found < n_events_; ++it) {
+
+        //  The poll item is a 0MQ socket. Retrieve pending events
+        //  using the ZMQ_EVENTS socket option.
+        if (it->socket) {
+            size_t events_size = sizeof (uint32_t);
+            uint32_t events;
+            if (it->socket->getsockopt (ZMQ_EVENTS, &events, &events_size)
+                                                                        == -1) {
+                return -1;
+            }
+
+            if (it->events & events) {
+                events_[found].socket = it->socket;
+                events_[found].user_data = it->user_data;
+                events_[found].events = it->events & events;
+                ++found;
+            }
+        }
+        //  Else, the poll item is a raw file descriptor, simply convert
+        //  the events to zmq_pollitem_t-style format.
+        else {
+
+#if defined ZMQ_POLL_BASED_ON_POLL
+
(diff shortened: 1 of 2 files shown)


msune:
Awaiting CI. Local tests passing.

src/socket_poller.cpp
```diff
@@ -464,117 +592,28 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_ev
...
-        }
+        int found = check_events(events_, n_events_);
         if (found) {
-            for (int i = found; i < n_events_; ++i) {
-                events_[i].socket = NULL;
-                events_[i].fd = 0;
-                events_[i].user_data = NULL;
-                events_[i].events = 0;
-            }
+            if (found > 0)
```
sigiesec: found is always non-negative, so this is redundant with the condition above

src/socket_poller.cpp
```diff
@@ -464,117 +592,28 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_ev
...
+                zero_trail_events(events_, n_events_, found);
```
sigiesec: missing space before (

src/socket_poller.cpp
msune: No, it isn't.

`zmq::socket_poller_t::check_events()` can return -1, as a way to bail out (as the old code was doing in the body of the function).

With this branch, you avoid calling the cleanup function which doesn't support values < 0.

src/socket_poller.cpp
msune: Is there any `checkpatch.pl`-like which does style checkings on the patch before PRing? It is really annoying (sorry I contribute to many different projects with various styles)

msune:
Travis CI errors doesn't seem to be related to the patch.

src/socket_poller.cpp
sigiesec: Ok, sorry I did not see that. I thought it was an artifact of the refactoring :)

src/socket_poller.cpp
sigiesec: I think so too. I made a clang-format config file for myself, but clang-format does not support everything in the ZeroMQ style. In particular, it does not support adding spaces before indexing square brackets :( I would really appreciate if the style could be moved to something that can be expressed using clang-format, which is a very good tool IMO, and one of the few (the only one?) that works with a real C++ parser.

sigiesec:
Yes, I think so too. Those jobs that fail are failing for some other reason.

msune:
I would like to collect some complete reviews before doing the final style changes. So please, if you did a complete review, express your conditional ACK.

src/socket_poller.hpp
```diff
@@ -81,6 +81,21 @@ namespace zmq
         bool check_tag ();
 
     private:
+        void zero_trail_events(zmq::socket_poller_t::event_t *events_,
+                                                          int n_events_,
+                                                          int found);
+        int check_events(zmq::socket_poller_t::event_t *events_, int n_events_
```
sigiesec: Maybe put the whole check_events definition within each #if branch. It is somewhat hard to read this way.

sigiesec:
I added a minor remark on the function definition in the header. Apart from that, looks good to me :) Ack ;)

src/socket_poller.hpp
msune: @sigiesec Not against if there is consensus. So if someone else can do +1 or -1 on this, I will add it in the style refactor.

msune:
Thanks @sigiesec. Someone else so we can finalise this?

msune:
Pull request amended with the style issues spoted. Please, have a look. CI passing as before. 

Can someone confirm if failures in Jenkins are related to this PR or not (e.g. `master` is also affected)? I have the feeling they are not related, but since it's on the Windows side, I believe, I am really not sure.

bluca:
Thanks!

src/socket_poller.cpp
```diff
@@ -401,7 +401,135 @@ void zmq::socket_poller_t::rebuild ()
     need_rebuild = false;
 }
 
-int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_, int n_events_, long timeout_)
+void zmq::socket_poller_t::zero_trail_events (
```
bluca: What about adding ```static inline``` descriptors? As @somdoron mentioned these functions get called a lot, so a hint to the compilers can't hurt.

bluca:
Not sure about the Jenkins one, sorry, but a previous one without cosmetic changes was fine

PR #2752, opened 2017-09-18
Problem: missing tests and inconsistent behaviour for ZAP corner cases

### LINKED ISSUE #2711: ZAP handler communication errors should be handled consistently
The current implementations of the mechanisms are not consistent with regard to ZAP handler communication errors, most importantly
1. there is no ZAP handler at all
1. the ZAP handler has not yet started / connected to the ZAP socket
1. the ZAP handler has disconnected from the ZAP socket **before** the session tries to bind to the ZAP socket
1. the ZAP handler disconnects from the ZAP socket **after** the session has bound to the ZAP socket
1. the ZAP handler does not read the ZAP request
1. the ZAP handler does not send a ZAP response

The first two error conditions cannot be distinguished. The third could in principle be distinguished from the first and second, but this is not currently implemented, and seems rather arbitrary.

The NULL and CURVE mechanisms simply ignore errors 1-3, and accept the connection attempt. The CURVE mechanism claims to support the Stonehouse pattern by this, however this attributes to error condition 1 only, which appears rather insecure.

The PLAIN mechanism rejects any connection attempt in case of error conditions 1-3. 

Condition 4 will likely result in some server-side error. Error conditions 5 and 6 will probably cause a client-side timeout, which might be acceptable.

I suggest that it can be explicitly configured whether a ZAP handler is required by a dedicated socket option. Inconsistent configurations (e.g. a ZAP domain is defined, but ZAP handler is not configured as required) should result in an error.

Error condition 2 can be avoided by the client code (like it is done in testutil_security.hpp), but it is rather easy to get this wrong (it was also wrong in the tests until some days ago), and hard to notice the effect. Error condition 3 is also easily triggered by errors in the ZAP handler code, and may remain unnoticed, and allow unauthenticated clients.

Should these cases use server assertions (and abort the process), or should they simply emit monitor events? They are not libzmq-internal errors, but still severe process-internal errors, which cannot be signalled synchronously to the libzmq-client code.

sigiesec:
Solution: added tests & aligned behaviour

Fixes #2711

### DIFF
--- a/tests/testutil_security.hpp
+++ b/tests/testutil_security.hpp
@@ -154,7 +154,9 @@ enum zap_protocol_t
     zap_wrong_request_id,
     zap_status_invalid,
     zap_too_many_parts,
-    zap_disconnect
+    zap_disconnect,
+    zap_do_not_recv,
+    zap_do_not_send
 };
 
 void *zap_requests_handled;
@@ -182,8 +184,11 @@ void zap_handler_generic (void *ctx,
       {handler, 0, ZMQ_POLLIN, 0},
     };
 
+    // if ordered not to receive the request, ignore the second poll item
+    const int numitems = (zap_protocol == zap_do_not_recv) ? 1 : 2;
+
     //  Process ZAP requests forever
-    while (zmq_poll (items, 2, -1) >= 0) {
+    while (zmq_poll (items, numitems, -1) >= 0) {
         if (items[0].revents & ZMQ_POLLIN) {
             char *buf = s_recv (control);
             assert (buf);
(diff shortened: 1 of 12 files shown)


tests/testutil_security.hpp
```diff
@@ -183,6 +202,8 @@ void zap_handler_generic (void *ctx,
         char *version = s_recv (handler);
         if (!version)
             break; //  Terminating - peer's socket closed
+        if (zap_protocol == zap_disconnect)
+            break;
```
bluca: this leaks the version string

sigiesec: Fixed this in 77f76a4

Issue #1590, opened 2015-09-24
Behaviour of REQ with disconnect


ghost:
The bevavior of REQ socket in the face of disconnect() looks strange. Sometimes it hangs, sometimes it works depending on the URL the disconnect() is done on (and probably many other parameters - process scheduling...).

The test case comes from an adaptation of the ZMQ Guide Hello World. Basically, the client connects to multiple server (REQ), and perform send()/recv() in sequence. After some times, it disconnects from one server URL but continue to send()/recv(). Sometimes the client hangs, sometimes it does not depending on the URL that was disconnected. The expected behavior is that it must not hang.

Compile:

``` bash
gcc -o server server.c -lzmq
gcc -o client client.c -lzmq
```

Launching:

``` bash
client ipc:///tmp/server1 ipc:///tmp/server2 &
server ipc:///tmp/server1 &
server ipc:///tmp/server2 &
```

Wait for the 10th send()/recv(). It hangs. It should not. If you edit the client.c and change argv[2] by argv[1] it works as expected.

Client code:

``` C
// Hello World client
#include <zmq.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>

int
main (int argc, char ** argv)
{
  printf ("Connecting to hello world server…\n");
  void *context = zmq_ctx_new ();
  void *requester = zmq_socket (context, ZMQ_REQ);
  int i;
  for (i = 1; i < argc; i++) {
        zmq_connect (requester, argv[i]);
  }

  int request_nbr;
  for (request_nbr = 0; request_nbr != 1000; request_nbr++)
    {
      char buffer[100];
      memset(buffer, 0, 100);
      printf ("Sending Hello %d…\n", request_nbr);
      zmq_send (requester, "Hello", 5, 0);
      zmq_recv (requester, buffer, 100, 0);
      printf ("Received World %d: %s\n", request_nbr, buffer);
      if (request_nbr == 10) {
        printf("Disconnecting from %s\n", argv[2]);
        zmq_disconnect(requester, argv[2]);
      }
    }
  zmq_close (requester);
  zmq_ctx_destroy (context);
  return 0;
}
```

Server Code:

``` C
// Hello World server

#include <zmq.h>
#include <stdio.h>
#include <unistd.h>
#include <string.h>
#include <assert.h>

int
main (int argc, char** argv)
{
// Socket to talk to clients
  void *context = zmq_ctx_new ();
  void *responder = zmq_socket (context, ZMQ_REP);
[...]

Issue #883, opened 2014-02-11
Reconnect due to ZAP authentication failure ignores ZMQ_RECONNECT_IVL_MAX.


hashstat:
When a server performing ZAP authentication rejects a client with a 400 (or other non-200) error, the client will use ZMQ_RECONNECT_IVL to calculate the reconnect timeout but never performs the exponential backoff strategy according to the ZMQ_RECONNECT_IVL_MAX option. Contrast this with a client that cannot connect to the server at all. Therefore, failed authentication can cause an indefinite number of connection attempts, especially when considered with issue #882.

But don't worry. I have a fix.

sigiesec:
I had a look at your changes, but I do not quite understand how they relate to ZAP errors. Aren't they related to any reconnects during handshaking?

Would you care to submit them as a PR?

PR #2756, opened 2017-09-19
Problem: zmq_connect fails after disconnect due to RECONNECT_IVL == -1

bluca:
Solution: when a connection breaks and ZMQ_RECONNECT_IVL is set to -1,
which means a reconnection will not be attempted, send a message from
the I/O thread to the application thread to make the socket call
term_endpoint, which is the equivalent of manually calling
zmq_disconnect.
This way subsequent zmq_connect call to the same endpoint will attempt
again to do a connection.
Otherwise, for some socket types like SUBs, those new connects will
fail as the endpoint is recorded, despite the connection having been
permanently closed.

Add test cases to exercise this corner case with TCP and IPC.

### DIFF
--- a/src/command.hpp
+++ b/src/command.hpp
@@ -30,6 +30,7 @@
 #ifndef __ZMQ_COMMAND_HPP_INCLUDED__
 #define __ZMQ_COMMAND_HPP_INCLUDED__
 
+#include <string>
 #include "stdint.hpp"
 
 namespace zmq
@@ -69,6 +70,7 @@ namespace zmq
             term_req,
             term,
             term_ack,
+            term_endpoint,
             reap,
             reaped,
             inproc_connected,
@@ -152,6 +154,12 @@ namespace zmq
             //  shut down.
             struct {
             } term_ack;
+
+            //  Sent by session_base (I/O thread) to socket (application thread)
+            //  to ask to disconnect the endpoint.
+            struct {
+                std::string *endpoint;
+            } term_endpoint;
 
             //  Transfers the ownership of the closed socket
             //  to the reaper thread.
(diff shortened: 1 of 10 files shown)


sigiesec:
@bluca Do you think this fix is related to #1590 and #883?

bluca:
Mmh I don't think so, this is specifically related to automated disconnects (or failed connections) and that socket option

sigiesec:
ok, just saw the issues again, and they seemed superficially related

src/session_base.cpp
```diff
@@ -536,6 +536,11 @@ void zmq::session_base_t::reconnect ()
     //  Reconnect.
     if (options.reconnect_ivl != -1)
         start_connecting (true);
+    else {
+        std::string *ep = new (std::string);
```
a4z: when/how will this string be deleted?

src/command.hpp
```diff
@@ -153,6 +155,12 @@ namespace zmq
             struct {
             } term_ack;
 
+            //  Sent by session_base (I/O thread) to socket (application thread)
+            //  to ask to disconnect the endpoint.
+            struct {
+                std::string *endpoint;
```
a4z: I think this should simply be a std::string, and maybe not a pointer to a string.
It seem to me it  is very unclear who is responsible for this pointer

src/session_base.cpp
bluca: By the application thread on receipt of the message

src/command.hpp
bluca: It must be either a pointer or a basic type, as it is part of a union

src/session_base.cpp
bluca: The pointer is stored in the union args_t which is part of struct command_t, and the receiver deletes it here:

https://github.com/zeromq/libzmq/pull/2756/files#diff-4243b1744be5585239d76c7e7d18017fR1428

Issue #2764, opened 2017-10-03
IPC sockets do not reconnect to active bind


Kentzo:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description

# Environment

* libzmq version (commit hash if unreleased): 4.1.6
* OS: macOS 10.12.6
* pyzmq: 16.0.2

# Minimal test code / Steps to reproduce the issue

Here are 2 python files:

```python
# pub.py

import zmq
import time

c = zmq.Context.instance()
s = c.socket(zmq.PUB)
s.bind('ipc:///tmp/socket')
time.sleep(10)  # helps to run 2 processes simultaneously
s.send(b'42')
```

```python
# sub.py

import zmq

c = zmq.Context.instance()
s = c.socket(zmq.SUB)
s.subscribe('')
s.connect('ipc:///tmp/socket')

while True:
    print(s.recv())
```

To reproduce the issue:
1. Run `sub.py`
2. Run 2 `pub.py` simultaneously

# What's the actual result? (include assertion message & call stack if applicable)

`sub.py` received only one message (expected), but subsequent calls to `pub.py` will not cause `sub.py` to receive anything.

# What's the expected result?

`sub.py` should notice that previously connected socket is closed, reconnect, and be able to receive messages from subsequent `pub.py`.

bluca:
Sounds similar to https://github.com/zeromq/libzmq/issues/2763 can't remember right now, but I think AF_UNIX sockets have the same behaviour as AF_INET w.r.t. dead peer detection.

Could you please try with the ZMQ_HEARTBEAT* options?

Kentzo:
@bluca I'll try that.

While debugging I also came a across that ZeroMQ allows the same process to bind to the same IPC address twice. Not sure that should be allowed.

bluca:
Yes 0 length read closes the socket: https://github.com/zeromq/libzmq/blob/master/src/stream_engine.cpp#L316

The problem is the kernel won't give you that in a lot of cases. For AF_UNIX as I mentioned I can't remember what's the exact behaviour right now.

Regarding double binding, it's because as it is customs the library will unlink the socket file before binding. That's because those files are never automatically cleaned up - that's just how AF_UNIX works I'm afraid.

Kentzo:
I wonder if that problem also happens due to unlinking. I just tried the same test, but instead of running 2 parallel processes I bound 2 sockets to the same address around a single connect.

Kentzo:
[...]

Issue #2733, opened 2017-09-01
Problem: test_security_curve/zap hang on slow machines with no DRAFT APIs


bluca:
# Issue description

test_security_curve and test_security_zap hang 1 in 5 times on slow environments, like running through Valgrind or on overcrowded build VMs on OBS.

# Environment

* libzmq version (commit hash if unreleased): master
* OS: Linux

# Minimal test code / Steps to reproduce the issue

```
while valgrind --tool=memcheck --track-origins=yes --expensive-definedness-checks=yes tests/.libs/test_security_curve; do echo; echo; done
```

When built with autotools. Adding --vgdb=yes --vgdb-error=0 allows to attach gdb from another terminal and inspect the running process. Compiling with ```-O0 -ggdb3``` is recommended.

# What's the actual result? (include assertion message & call stack if applicable)

The test that actually hangs might vary, but the problem is always the same: the server socket (DEALER with identity "ident") does not close properly and the context hangs on zmq_ctx_term.

 It's stuck terminating the ctx here: https://github.com/zeromq/libzmq/blob/9c8844fd0840b699194ac89dca9967b6e8d32dec/tests/test_security_curve.cpp#L662

As it happened before, it's waiting to receive on the mailbox at https://github.com/zeromq/libzmq/blob/9c8844fd0840b699194ac89dca9967b6e8d32dec/src/ctx.cpp#L179

```
(gdb) bt
#0  0x00000000060636ad in poll () at ../sysdeps/unix/syscall-template.S:84
#1  0x0000000004e7c5b2 in zmq::signaler_t::wait (this=this@entry=0x693d7a8, timeout_=timeout_@entry=-1)
    at src/signaler.cpp:233
#2  0x0000000004e61376 in zmq::mailbox_t::recv (this=this@entry=0x693d740, cmd_=cmd_@entry=0xffefffbc0, 
    timeout_=timeout_@entry=-1) at src/mailbox.cpp:81
#3  0x0000000004e5060d in zmq::ctx_t::terminate (this=this@entry=0x693d6a0) at src/ctx.cpp:179
#4  0x0000000004e9a62b in zmq_ctx_term (ctx_=0x693d6a0) at src/zmq.cpp:175
#5  0x000000000010af03 in shutdown_context_and_server_side (ctx=0x693d6a0, zap_thread=0x69443a0, server=0x6944400, 
    server_mon=<optimized out>) at tests/test_security_curve.cpp:423
[...]

PR #2765, opened 2017-10-04
Problem: Race condition in IPC sockets

Kentzo:
Solution: Don't unlink file on close

File may not belong to the socket at that point.

Refs #2764

### DIFF
--- a/src/ipc_listener.cpp
+++ b/src/ipc_listener.cpp
@@ -294,21 +294,12 @@ int zmq::ipc_listener_t::close ()
 
     s = retired_fd;
 
-    //  If there's an underlying UNIX domain socket, get rid of the file it
-    //  is associated with.
-    //  MUST NOT unlink if the FD is managed by the user, or it will stop
-    //  working after the first client connects. The user will take care of
-    //  cleaning up the file after the service is stopped.
     if (has_file && options.use_fd == -1) {
         rc = 0;
 
-        if ( !filename.empty () ) {
-          rc = ::unlink(filename.c_str ());
-        }
-
         if ( rc == 0 && !tmp_socket_dirname.empty() ) {
-          rc = ::rmdir(tmp_socket_dirname.c_str ());
-          tmp_socket_dirname.clear();
+            rc = ::rmdir(tmp_socket_dirname.c_str ());
+            tmp_socket_dirname.clear();
         }
 
         if (rc != 0) {
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ test_last_endpoint
 test_pair_inproc
 test_pair_ipc
 test_pair_tcp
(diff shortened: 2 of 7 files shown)


Kentzo:
A number of tests failed under certain configurations. Should look for errors or these tests are known to fail?

bluca:
Thanks, but could you please split the refactor and the behaviour changes in 2 different commits? This will make it easier to track eventual regressions.

bluca:
Also the test failures are new and are those that use IPC, so yes please have a look, thanks

Kentzo:
There are 3 changes:

1. ZMQ_HAVE_IPC
2. cleanup in ipc_listener
3. bufgix

I can add 2 commits with 1 and 2,3.

Kentzo:
Strange that it only fails with autotools.

src/ipc_listener.cpp
```diff
@@ -210,115 +211,130 @@ int zmq::ipc_listener_t::get_address (std::string &addr_)
 
 int zmq::ipc_listener_t::set_address (const char *addr_)
 {
-    //  Create addr on stack for auto-cleanup
+    printf("a\n");
+    if (options.use_fd != -1) {
```
bluca: this was done later otherwise the address is not saved in the endpoint class variable

bluca: ditto for filename

Kentzo: If understood the code correctly, when use_fd is set address is irrelevant.

bluca: it's used for the events, for example, like a few lines below

Kentzo:
I undid all of the refactoring as unfortunately don't have time to mess with it.

However, there seems to be a bug when [reporting the close event](https://github.com/zeromq/libzmq/blob/01a3f3955c04803e0131ea5e7e08788a138924d8/src/ipc_listener.cpp#L320): by that time `s` will be [reassigned to `retired_fd`](https://github.com/zeromq/libzmq/blob/01a3f3955c04803e0131ea5e7e08788a138924d8/src/ipc_listener.cpp#L295).

Kentzo:
@bluca Can I backport this fix to 4.1 and 4.2 (what repo)?

Kentzo:
@bluca Do you think `test_security_zap` failed due to my changes?

bluca:
> However, there seems to be a bug when reporting the close event: by that time s will be reassigned to retired_fd.

Good point, I can take care of that, thanks.

> @bluca Can I backport this fix to 4.1 and 4.2 (what repo)?

Sure, for 4.2 this is the right repo already

> @bluca Do you think test_security_zap failed due to my changes?

That's https://github.com/zeromq/libzmq/issues/2733 it's a known (tricky) problem

PR #2778, opened 2017-10-13
Background threads enhancements

f18m:
# Pull Request Notice

Problem: 
 - right now there's no way to set background thread CPU affinity
 - right now it's impossible to set the NICE value of background threads when using SCHED_OTHER Linux default scheduling

Solution: 
 - add option ZMQ_THREAD_AFFINITY to allow setting CPU affinity
 - add special casing for SCHED_OTHER that triggers a call to the nice() syscall, to allow setting the NICE of background threads.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -557,10 +557,6 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 #ifdef ZMQ_BUILD_DRAFT_API
 
-/* DRAFT Context options */
-#define ZMQ_THREAD_AFFINITY 6
-#define ZMQ_THREAD_AFFINITY_DFLT -1
-
 /*  DRAFT Socket types.                                                       */
 #define ZMQ_SERVER 12
 #define ZMQ_CLIENT 13
@@ -615,6 +611,8 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_MSG_T_SIZE 6
+#define ZMQ_THREAD_AFFINITY 7
+#define ZMQ_THREAD_AFFINITY_DFLT -1
 
 /*  DRAFT Socket methods.                                                     */
 ZMQ_EXPORT int zmq_join (void *s, const char *group);
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -557,6 +557,10 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 #ifdef ZMQ_BUILD_DRAFT_API
 
+/* DRAFT Context options */
+#define ZMQ_THREAD_AFFINITY 6
+#define ZMQ_THREAD_AFFINITY_DFLT -1
+
 /*  DRAFT Socket types.                                                       */
 #define ZMQ_SERVER 12
 #define ZMQ_CLIENT 13
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -87,6 +87,8 @@ void zmq::thread_t::setThreadName(const char *name_)
 
 #include <signal.h>
 #include <unistd.h>
+#include <sys/time.h>
+#include <sys/resource.h>
 
 extern "C"
 {
@@ -101,8 +103,8 @@ extern "C"
         rc = pthread_sigmask (SIG_BLOCK, &signal_set, NULL);
         posix_assert (rc);
 #endif
-
         zmq::thread_t *self = (zmq::thread_t*) arg_;
+        self->applySchedulingParameters();
         self->tfn (self->arg);
         return NULL;
     }
@@ -122,7 +124,14 @@ void zmq::thread_t::stop ()
     posix_assert (rc);
 }
 
-void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_)
+void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_, int affinity_)
+{
+    thread_priority=priority_;
+    thread_sched_policy=schedulingPolicy_;
+    thread_affinity=affinity_;
+}
+
+void zmq::thread_t::applySchedulingParameters()         // to be called in secondary thread context
 {
 #if defined _POSIX_THREAD_PRIORITY_SCHEDULING && _POSIX_THREAD_PRIORITY_SCHEDULING >= 0
     int policy = 0;
@@ -136,14 +145,17 @@ void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_
     int rc = pthread_getschedparam(descriptor, &policy, &param);
     posix_assert (rc);
 
-    if(priority_ != -1)
+    if(thread_sched_policy != -1)
     {
-        param.sched_priority = priority_;
+        policy = thread_sched_policy;
     }
 
-    if(schedulingPolicy_ != -1)
+    if(thread_priority != -1)
     {
-        policy = schedulingPolicy_;
+        param.sched_priority = thread_priority;
+
+        if (policy == SCHED_OTHER)
+            param.sched_priority = 0;   // this is the only supported priority for SCHED_OTHER!
     }
 
 #ifdef __NetBSD__
@@ -158,6 +170,33 @@ void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_
 #endif
 
     posix_assert (rc);
+
+    if (thread_sched_policy == SCHED_OTHER)
+    {
+        // in this case the thread priority setting can be only zero for the
+
+#define SETPRIORITY_MAX_PRIO                (-20)       // highest prio
+#define SETPRIORITY_MEDIUM_PRIO             (+0)
+
+        int prio = (thread_priority != ZMQ_THREAD_PRIORITY_DFLT) ? SETPRIORITY_MAX_PRIO : SETPRIORITY_MEDIUM_PRIO;
+        //rc = setpriority(PRIO_PROCESS, 0 /* calling process */, prio);
+        rc = nice(prio);
+        errno_assert (rc != -1);
+    }
+
+    if (thread_affinity != -1)        // FIXME in this place and places above I think it would be better to use the various ZMQ_*_DFLT preproc symbols instead of -1 directly
+    {
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+        for (unsigned int cpuidx=0; cpuidx<sizeof(int)*8; cpuidx++)
+        {
+            int cpubit = (1 << cpuidx);
+            if ( (thread_affinity & cpubit) != 0 )
+                CPU_SET( cpuidx , &cpuset );
+        }
+        rc = sched_setaffinity(0 /* use calling thread PID */, sizeof(cpu_set_t), &cpuset);
+        posix_assert (rc);
+    }
 #else
 
     LIBZMQ_UNUSED (priority_);
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -146,17 +146,24 @@ void zmq::thread_t::applySchedulingParameters()         // to be called in secon
     int rc = pthread_getschedparam(descriptor, &policy, &param);
     posix_assert (rc);
 
-    if(thread_sched_policy != -1)
+    if(thread_sched_policy != ZMQ_THREAD_SCHED_POLICY_DFLT)
     {
         policy = thread_sched_policy;
     }
 
-    if(thread_priority != -1)
-    {
-        param.sched_priority = thread_priority;
+    /* Quoting docs:
+       "Linux allows the static priority range 1 to 99 for the SCHED_FIFO and
+       SCHED_RR policies, and the priority 0 for the remaining policies."
+       Other policies may use the "nice value" in place of the priority:
+    */
+    bool use_nice_instead_priority = (policy != SCHED_FIFO) && (policy != SCHED_RR);
 
-        if (policy == SCHED_OTHER)
-            param.sched_priority = 0;   // this is the only supported priority for SCHED_OTHER!
+    if(thread_priority != ZMQ_THREAD_PRIORITY_DFLT)
+    {
+        if (use_nice_instead_priority)
+            param.sched_priority = 0;                   // this is the only supported priority for most scheduling policies
+        else
+            param.sched_priority = thread_priority;     // user should provide a value between 1 and 99
     }
 
 #ifdef __NetBSD__
@@ -172,19 +179,21 @@ void zmq::thread_t::applySchedulingParameters()         // to be called in secon
 
     posix_assert (rc);
 
-    if (thread_sched_policy == SCHED_OTHER &&
+    if (use_nice_instead_priority &&
             thread_priority != ZMQ_THREAD_PRIORITY_DFLT)
     {
         // assume the user wants to decrease the thread's nice value
         // i.e., increase the chance of this thread being scheduled: try setting that to
         // maximum priority.
+        rc = nice(-20);
+
         // IMPORTANT: ignore return code: EPERM is typically returned for unprivileged processes
         //            (CAP_SYS_NICE capability is required or RLIMIT_NICE resource limit should be set)
         //            but in that case there is no much we can do
-        rc = nice(-20);
+
     }
 
-    if (thread_affinity != -1)        // FIXME in this place and places above I think it would be better to use the various ZMQ_*_DFLT preproc symbols instead of -1 directly
+    if (thread_affinity != ZMQ_THREAD_AFFINITY_DFLT)
     {
         cpu_set_t cpuset;
         CPU_ZERO(&cpuset);
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -187,12 +187,13 @@ void zmq::thread_t::applySchedulingParameters()         // to be called in secon
         // maximum priority.
         rc = nice(-20);
 
-        // IMPORTANT: ignore return code: EPERM is typically returned for unprivileged processes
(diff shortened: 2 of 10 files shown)


f18m:
PS: I just tested this on Linux. I didn't try Windows as all the code is POSIX-specific.
Indeed I maybe missing some #ifdef to correctly fence the POSIX-only code... I can add that. But before I would like to hear your opinions on that... thanks.

include/zmq.h
```diff
@@ -557,6 +557,10 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 #ifdef ZMQ_BUILD_DRAFT_API
 
+/* DRAFT Context options */
```
bluca: There's already some below, at line 617 (ZMQ_MSG_T_SIZE), could you please fix it? Also it conflicts

bluca:
Could you please have a look at the Windows CI? The build appears broken:

https://ci.appveyor.com/project/zeromq/libzmq/build/build-1782/job/o287lk85j4g6y6ea

```
C:\projects\libzmq\src\thread.cpp(74): error C2511: 'void zmq::thread_t::setSchedulingParameters(int,int)': overloaded member function not found in 'zmq::thread_t' [C:\projects\build_libzmq\libzmq-static.vcxproj]
  stream.cpp
  c:\projects\libzmq\src\thread.hpp(50): note: see declaration of 'zmq::thread_t'

C:\projects\libzmq\src\thread.cpp(74): error C2511: 'void zmq::thread_t::setSchedulingParameters(int,int)': overloaded member function not found in 'zmq::thread_t' [C:\projects\build_libzmq\libzmq.vcxproj]
  c:\projects\libzmq\src\thread.hpp(50): note: see declaration of 'zmq::thread_t'
```

bluca:
On Linux it fails as well:

```
Operation not permitted (../src/thread.cpp:184)
```

I can't remember off the top of my head, does it require special capabilities?

f18m:
Yes, actually it requires CAP_SYS_NICE capability or a suitable RLIMIT_NICE
resource limit... http://man7.org/linux/man-pages/man2/nice.2.html .

I did not get the error because I have the bad habit of developing and
running my software as root user.
Maybe I should put a check using getrlimit() API before attemping nice()
call... or perhaps we could leave the nice() call to simply fail in case of
insufficient permissions... what do you think?

PS: I will have a look at build failure under Windows but I expected that.
Before getting to a "final PR" I would like to have your feedback... thanks.

2017-10-13 19:46 GMT+02:00 Luca Boccassi <notifications@github.com>:

f18m:
Btw I would like to propose also another (small) background thread enhancement: all background threads right now have the same name, also across different processes.
The problem in my case is that my company is using ZMQ in several different processes (after all ZMQ is all about modularity right? :) ) and each one is using 2-3 ZMQ background threads... when you look at the threads system-wide (top -H) the result is that there are lots of "ZMQ background" as top scorers...

A simple way to help with this would be to add a ZMQ_THREAD_NAME_PREFIX context option and then prepend the user-provided string to the "ZMQ background" string. In this way it would be easy for ZMQ user to provide a (short) identifier of the process.

The problem with this is again that zmq_ctx_set() takes only integers... but it should be possible to add an API like:
   int zmq_ctx_set (void *context, int option_name, const void *option_value, size_t option_len);
similar to int zmq_setsockopt ()... that would allow also for a better implementation of ZMQ_THREAD_AFFINITY option, which would be able to take a cpu_set_t as parameter...

Just my 2 cents...

src/thread.cpp
```diff
@@ -158,10 +178,39 @@ void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_
...
+
+    if (use_nice_instead_priority &&
+            thread_priority != ZMQ_THREAD_PRIORITY_DFLT)
+    {
+        // assume the user wants to decrease the thread's nice value
+        // i.e., increase the chance of this thread being scheduled: try setting that to
+        // maximum priority.
+        rc = nice(-20);
+
+        // IMPORTANT: ignore return code: EPERM is typically returned for unprivileged processes
```
bluca: This should be an assert - if a user sets the priority but it cannot be done due to lack of privileges, it should not be ignored silently as it's a programming error.

src/thread.cpp
```diff
@@ -158,10 +178,39 @@ void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_
...
+    {
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+        for (unsigned int cpuidx=0; cpuidx<sizeof(int)*8; cpuidx++)
+        {
+            int cpubit = (1 << cpuidx);
+            if ( (thread_affinity & cpubit) != 0 )
+                CPU_SET( cpuidx , &cpuset );
+        }
+        rc = sched_setaffinity(0 /* use calling thread PID */, sizeof(cpu_set_t), &cpuset);
```
bluca:
```
(If you are using the POSIX  threads
       API, then use pthread_setaffinity_np(3) instead of sched_setaffinity().)
```

Also given it's non-posix, you'll need to add checks in aclocal.m4 and CMakeList.txt and IFDEF it out if not available, otherwise the build on non-linux system will be broken

bluca:
Sorry but there's still quite a few build errors, so I can't merge it yet - could you please have a look at the Travis logs and fix it?

include/zmq.h
```diff
@@ -611,6 +611,8 @@ ZMQ_EXPORT void zmq_threadclose (void* thread);
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_MSG_T_SIZE 6
+#define ZMQ_THREAD_AFFINITY 7
```
bluca: these need to be copied also in src/zmq_draft.h otherwise the no-draft builds will fail

bluca:
> Maybe I should put a check using getrlimit() API before attemping nice()
call... or perhaps we could leave the nice() call to simply fail in case of
insufficient permissions... what do you think?

Given you'd need to check for the getrlimit availability it's perhaps simpler to document it in the manpage. But the call must assert if it cannot be completed, as it's a programming error.

> Btw I would like to propose also another (small) background thread enhancement: all background threads right now have the same name, also across different processes.
The problem in my case is that my company is using ZMQ in several different processes (after all ZMQ is all about modularity right? :) ) and each one is using 2-3 ZMQ background threads... when you look at the threads system-wide (top -H) the result is that there are lots of "ZMQ background" as top scorers...
>
> A simple way to help with this would be to add a ZMQ_THREAD_NAME_PREFIX context option and then prepend the user-provided string to the "ZMQ background" string. In this way it would be easy for ZMQ user to provide a (short) identifier of the process.
>
> The problem with this is again that zmq_ctx_set() takes only integers... but it should be possible to add an API like:
int zmq_ctx_set (void *context, int option_name, const void *option_value, size_t option_len);
similar to int zmq_setsockopt ()... that would allow also for a better implementation of ZMQ_THREAD_AFFINITY option, which would be able to take a cpu_set_t as parameter...
>
> Just my 2 cents...

Yes the thread names have always been annoying when debugging with GDB. They didn't use to be named at all, I added that a short while back, but they are all named the same as it requires quite a bit of refactoring to pass a useful identifier down (reaper thread, socket thread, I/O thread number X).
If you want to try that refactoring it would be very welcome of course, but as a separate PR so that we can get this completed first.

I would advise against adding new options - we already have way too many, and the more are added the more usability is impacted for new developers.
In this case I think having a meaningful name as I wrote earlier - reaper thread, I/O thread X etc would be enough. Perhaps you can also throw in the PID as well so you can distinguish between processes.
This way it works out of the box for everybody, and it's not yet another option that users have to manually add to make it useful.

If having the proper name is too much work, perhaps just the PID would be a good start for your use case.

What do you think?

src/thread.cpp
```diff
@@ -158,10 +178,39 @@ void zmq::thread_t::setSchedulingParameters(int priority_, int schedulingPolicy_
...
+        posix_assert (rc);
+    }
 #else
 
     LIBZMQ_UNUSED (priority_);
```
bluca: This function does not take parameters anymore, so this \#else branch and the 3 macros have to be removed

```
src/thread.cpp: In member function 'void zmq::thread_t::applySchedulingParameters()':
src/thread.cpp:211:20: error: 'priority_' was not declared in this scope
     LIBZMQ_UNUSED (priority_);
                    ^
src/macros.hpp:6:37: note: in definition of macro 'LIBZMQ_UNUSED'
 #define LIBZMQ_UNUSED(object) (void)object
                                     ^
src/thread.cpp:212:20: error: 'schedulingPolicy_' was not declared in this scope
     LIBZMQ_UNUSED (schedulingPolicy_);
                    ^
src/macros.hpp:6:37: note: in definition of macro 'LIBZMQ_UNUSED'
 #define LIBZMQ_UNUSED(object) (void)object
                                     ^
src/thread.cpp:213:20: error: 'affinity_' was not declared in this scope
     LIBZMQ_UNUSED (affinity_);
                    ^
src/macros.hpp:6:37: note: in definition of macro 'LIBZMQ_UNUSED'
 #define LIBZMQ_UNUSED(object) (void)object
```

f18m:
Hi Luca,
I have implemented all the things you pointed out in the comments. Let's see how all builds go. I'm able to test only a small subset of the possible configurations.

>I would advise against adding new options - we already have way too many, and the more are added >the more usability is impacted for new developers.
>In this case I think having a meaningful name as I wrote earlier - reaper thread, I/O thread X etc would >be enough. Perhaps you can also throw in the PID as well so you can distinguish between processes.
>This way it works out of the box for everybody, and it's not yet another option that users have to >manually add to make it useful.
>If having the proper name is too much work, perhaps just the PID would be a good start for your use >case.
>What do you think?

Honestly as ZMQ user, I don't care too much of the specific ZMQ background thread type (e.g., I don't care if that's a reaper thread or something else) since I don't have a clue about what ZMQ is creating and doing in those background threads. As ZMQ user however I would like to see if the top-CPU-usage-scorer threads are ZMQ background threads (and thanks to your change for setting thread name that's now possible) and if so to which of my process they belong. Maybe other users are interested in knowing also the type of ZMQ background thread, don't know.

The idea of adding the PID is better than nothing... but requires you to find out the PID of each of your zmq-enabled processes to understand what's going on. For this reason I would really prefer to have a customizable prefix: that way by looking at 'top -H' you are immediately able to understand the status of your box and if a ZMQ background is top-scorer to identify which of your process is receiving/sending too much network traffic.

Note that the thread name prefix can also be implemented using current zmq_ctx_set() API: prefix has to be really short (3/4 chars I would say is enough; given thread name limit is 15 chars). So we could interpret each byte of an INT option as character and prefix it to thread names. I know it's not that elegant but would workaround the zmq_ctx_set() API limitations.

Btw for now this PR does not contain any change yet to thread names.

Thanks!
Francesco

tests/test_ctx_options.cpp
```diff
@@ -30,15 +30,108 @@
...
+    // However changing the nice value of a process requires appropriate permissions...
+    // check that the current effective user is able to do that:
+    if (is_allowed_to_raise_priority())
+    {
+        rc = zmq_ctx_set(ctx, ZMQ_THREAD_PRIORITY, 1 /* any positive value different than the default will be ok */);
+        assert (rc == 0);
+    }
+
+
+    // test affinity:
```
bluca: Given this option is new and in DRAFT state, you should ifdef it:

```
../tests/test_ctx_options.cpp: In function ‘void test_ctx_thread_opts(void*)’:
../tests/test_ctx_options.cpp:84:27: error: ‘ZMQ_THREAD_AFFINITY’ was not declared in this scope
     rc = zmq_ctx_set(ctx, ZMQ_THREAD_AFFINITY, ZMQ_THREAD_AFFINITY_DFLT);
                           ^
../tests/test_ctx_options.cpp:84:48: error: ‘ZMQ_THREAD_AFFINITY_DFLT’ was not declared in this scope
     rc = zmq_ctx_set(ctx, ZMQ_THREAD_AFFINITY, ZMQ_THREAD_AFFINITY_DFLT);
```

\#ifdef ZMQ_THREAD_AFFINITY
...
\#endif

tests/test_ctx_options.cpp
```diff
@@ -30,15 +30,108 @@
...
+void test_ctx_thread_opts(void* ctx)
+{
+    int rc;
+
+    // verify that setting negative values (e.g., default values) fail:
+    rc = zmq_ctx_set(ctx, ZMQ_THREAD_SCHED_POLICY, ZMQ_THREAD_SCHED_POLICY_DFLT);
+    assert (rc == -1 && errno == EINVAL);
+    rc = zmq_ctx_set(ctx, ZMQ_THREAD_PRIORITY, ZMQ_THREAD_PRIORITY_DFLT);
+    assert (rc == -1 && errno == EINVAL);
+    rc = zmq_ctx_set(ctx, ZMQ_THREAD_AFFINITY, ZMQ_THREAD_AFFINITY_DFLT);
```
bluca: this check as well

bluca:
@f18m merged, thanks for your patience!

I've opened a PR to start the documentation: https://github.com/zeromq/libzmq/pull/2779
Could you please build on that to add detailed documentation about the new option, when you have time?

> Honestly as ZMQ user, I don't care too much of the specific ZMQ background thread type (e.g., I don't care if that's a reaper thread or something else) since I don't have a clue about what ZMQ is creating and doing in those background threads. As ZMQ user however I would like to see if the top-CPU-usage-scorer threads are ZMQ background threads (and thanks to your change for setting thread name that's now possible) and if so to which of my process they belong. Maybe other users are interested in knowing also the type of ZMQ background thread, don't know.
>
> The idea of adding the PID is better than nothing... but requires you to find out the PID of each of your zmq-enabled processes to understand what's going on. For this reason I would really prefer to have a customizable prefix: that way by looking at 'top -H' you are immediately able to understand the status of your box and if a ZMQ background is top-scorer to identify which of your process is receiving/sending too much network traffic.

Ok that's fine, feel free to send a PR to add this

> Note that the thread name prefix can also be implemented using current zmq_ctx_set() API: prefix has to be really short (3/4 chars I would say is enough; given thread name limit is 15 chars). So we could interpret each byte of an INT option as character and prefix it to thread names. I know it's not that elegant but would workaround the zmq_ctx_set() API limitations.

Note that there are many high-level and strongly typed language bindings. These kind of overloads might work in C/C++ but that's not enough, the public API needs to be friendly enough for binding developers to support it as well.

I would suggest to start simple again, and use a simple numeric ID. If you know what you are looking for when debugging it shouldn't matter too much whether it's a string or a number, as long as they are unique.

f18m:
Hi Luca, Hi all,

sorry to revive this thread (should I open a defect or just discuss about
this on the mailing list instead?) but I found a couple of issues that are
really annoying regarding my own patches:

1) if you don't use the ZMQ_THREAD_NAME_PREFIX option, you currently have
threads named "0/ZMQbg/0" which is not really nice. I can modify the code
to avoid the prefix in case it is left to zero.
2) I'm deploying my zmq-based app on servers with more than 32 cores. This
turns out to be a problem since currently zmq_ctx_set() takes an int
(32bits) as the only possible option value. This pushes me again on the
idea that we should have a zmq_ctx_set() function more similar to
zmq_setsockopt(), taking a "void*" and a size_t.

What do you think?

Thanks,
Francesco





2017-10-16 14:15 GMT+02:00 Luca Boccassi <notifications@github.com>:

bluca:
> 1) if you don't use the ZMQ_THREAD_NAME_PREFIX option, you currently have
threads named "0/ZMQbg/0" which is not really nice. I can modify the code
to avoid the prefix in case it is left to zero.

Sure that's fine

> 2) I'm deploying my zmq-based app on servers with more than 32 cores. This
turns out to be a problem since currently zmq_ctx_set() takes an int
(32bits) as the only possible option value. This pushes me again on the
idea that we should have a zmq_ctx_set() function more similar to
zmq_setsockopt(), taking a "void*" and a size_t.

The problem is that void* option parameters make it very hard for bindings developers, so I'd like to avoid it if possible.

What if we rethink this API? Instead of a single option that takes the mask, have ZMQ_THREAD_AFFINITY_CPU_ADD and _REMOVE that take a single CPU ID and add it to the set, which gets built in options.cpp as things are added and removed.

This way there's no size limitation. What do you think?

f18m:
Hi Luca,


2017-10-24 18:40 GMT+02:00 Luca Boccassi <notifications@github.com>:
>
>
>    1. I'm deploying my zmq-based app on servers with more than 32 cores.
>    This
>    turns out to be a problem since currently zmq_ctx_set() takes an int
>    (32bits) as the only possible option value. This pushes me again on the
>    idea that we should have a zmq_ctx_set() function more similar to
>    zmq_setsockopt(), taking a "void*" and a size_t.
>
> The problem is that void* option parameters make it very hard for bindings
> developers, so I'd like to avoid it if possible.
>
> What if we rethink this API? Instead of a single option that takes the
> mask, have ZMQ_THREAD_AFFINITY_CPU_ADD and _REMOVE that take a single CPU
> ID and add it to the set, which gets built in options.cpp as things are
> added and removed.
>
> This way there's no size limitation. What do you think?
>
Sure, that's a possibility and is quite close to POSIX API finally since
such API would resemble the macros CPU_SET() and CPU_CLR().
However this would fix the problem in this specific case of thread affinity
but IMHO having a more generic way to pass >4 bytes to the context would be
wise to allow future developments that do not break ABI (the other
possibility that comes to my mind is again thread name prefix as char*
instead of int, but in future this may open other possibilities).

In any case if that's the way to go, I can provide a patch tomorrow for
_ADD() and _REMOVE()

Thanks for the fast support!
Francesco

bluca:
At some point we might have to cave, but I'd like to try and hold up as long as possible to make bindinng devs life easier.
If a different API is added it won't likely be "free form" like void* but more strictly defined for the same reason, eg: strings only or uint64_t only or something like that.

PR #2779, opened 2017-10-16
Problems: no documentation for new thread affinity and priority options, test_ctx_options only checks global DRAFT flag

bluca:
Solutions: document them and fix the test ifdefs

### DIFF
--- a/doc/zmq_ctx_set.txt
+++ b/doc/zmq_ctx_set.txt
@@ -74,6 +74,16 @@ This option only applies before creating any sockets on the context.
 Default value:: -1
 
 
+ZMQ_THREAD_AFFINITY: Set CPU affinity for I/O threads
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_THREAD_AFFINITY' argument sets CPU affinity for the internal
+context's thread pool. This option is only supported on Linux.
+This option only applies before creating any sockets on the context.
+
+[horizontal]
+Default value:: -1
+
+
 ZMQ_MAX_MSGSZ: Set maximum message size
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


f18m:
I'd suggest to better document affinity thread control; also about THREAD_PRIORITY we can be a little more accurate... I propose something like that:


ZMQ_THREAD_PRIORITY: Set scheduling priority for I/O threads
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The 'ZMQ_THREAD_PRIORITY' argument sets scheduling priority for
internal context's thread pool. This option is not available on windows.
Supported values for this option depend on chosen scheduling policy.
On Linux, when the scheduler policy is SCHED_OTHER, SCHED_IDLE or SCHED_BATCH, the OS scheduler
will not use the thread priority but rather the thread "nice value"; in such cases
the system call "nice" will be used to set the nice value to -20 (max priority) instead of
adjusting the thread priority (which must be zero for those scheduling policies).
Details can be found in sched.h file, or at http://man7.org/linux/man-pages/man2/sched_setscheduler.2.html.
This option only applies before creating any sockets on the context.

[horizontal]
Default value:: -1


ZMQ_THREAD_AFFINITY: Set CPU affinity for I/O threads
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The 'ZMQ_THREAD_AFFINITY' argument sets CPU affinity for the internal
context's thread pool. This option is only supported on Linux.
On Linux, each bit of the 'option_value' argument will represent an enabled
CPU in the corresponding cpu_set_t; for example the value 0x1 will set affinity
of all context's thread pool to CPU 0; the value 0x110 will set affinity
of all context's thread pool to CPU 1 and 2.
This option only applies before creating any sockets on the context.

[horizontal]
Default value:: -1

f18m:
btw I can create a PR if that's easier for you

bluca:
Yes please do, thanks - I just added a skeleton to get it rolling

PR #2781, opened 2017-10-16
Problems: autoconf does not use pkgconfig for NORM, deb packages do not build with libnorm

bluca:
Solutions: use pkg-config and enable debian builds


jimklimov:
Travis got stuck on OSX buids, and fell out in some ZAP tests. Same failed on my recent build of another project, got cleared by rebuild job...

bluca:
Yeah that's the usual hang due to https://github.com/zeromq/libzmq/issues/2733

PR #2784, opened 2017-10-17
Background thread names

f18m:
# Pull Request Notice

Problem: zeromq background thread names cannot be customized. It is difficult to understand which background thread is associated to which process.

Solution: allow the user to set a custom thread name prefix. Currently because of zmq_ctx_set() limitation it is just a number


bluca:
Thanks! The new option should be documented as well in doc/zmq_ctx_set, could you please send a PR for it?

PR #2792, opened 2017-10-18
Clarify usage of zmq_close

sigiesec:


doc/zmq_close.txt
```diff
@@ -21,22 +21,18 @@ behaviour for discarding messages sent by the application with _zmq_send()_ but
...
-The _zmq_close()_ function shall return zero if successful. Otherwise it shall
-return `-1` and set 'errno' to one of the values defined below.
-
-
-ERRORS
-------
-*ENOTSOCK*::
-The provided 'socket' was invalid.
-
+The _zmq_close()_ function always returns zero.
```
bluca: It does check the socket is valid via the tag though?

https://github.com/zeromq/libzmq/blob/master/src/zmq.cpp#L275

doc/zmq_close.txt
```diff
@@ -21,22 +21,18 @@ behaviour for discarding messages sent by the application with _zmq_send()_ but
 not yet physically transferred to the network depends on the value of the
 _ZMQ_LINGER_ socket option for the specified 'socket'.
 
+_zmq_close()_ must be called exactly once for each socket. If it is never called,
+_zmq_ctx_term()_ will block forever. If it is called multiple times for the same 
+socket or if _socket_ does not point to a socket, the behaviour is undefined.
```
bluca: Sockets are either thread safe, so the tag is set to deadbeef inside a critical section, or they are not and so only one zmq_close can be called at any one time on a socket, which means setting the tag will work and subsequent calls will return ENOTSOCK.

Is this not working correctly in some cases?

doc/zmq_close.txt
sigiesec: Ah, right. Good that you noticed that. I went a step too far.

Yes, it checks that, so it might return -1/ENOTSOCK. But this is not really useful for a caller. If the pointer points to something else, which is not a socket at all, it might be smaller than the tag and checking the tag might cause an invalid memory access. If the pointer is garbage, checking the tag might cause an invalid memory access. 

If the pointer pointed to a socket, for which zmq_close has already been called, it might have already been reaped, and then checking the tag might cause an invalid memory access.

But for nullptr you are right, it can return ENOTSOCK, and this is useful.

doc/zmq_close.txt
sigiesec: See my comment above, third paragraph.

doc/zmq_close.txt
sigiesec: I updated/reverted the documentation regarding nullptr.

For the other case, it now still says that behaviour is undefined. That it returns -1/ENOTSOCK in that case is just a special case of undefined behaviour ;)

doc/zmq_close.txt
```diff
@@ -33,9 +37,9 @@ return `-1` and set 'errno' to one of the values defined below.
 
 
 ERRORS
-------
+-------
```
bluca: Shouldn't this line match exactly the one above? Otherwise the web doc formatting will break

sigiesec: Yes, fixed that

PR #2813, opened 2017-11-02
Problem: Sometimes on OSX we get ETIMEDOUT instead of EAGAIN

AntonBarwald:
Cannot reproduce/test this in libzmq. But i ran czmq/zloop self-test: https://github.com/zeromq/czmq/blob/ed902413453c0e50c65f887c40a5e9052619a224/src/zloop.c#L1  

If i print the "Value of zmq_error(): " here: 
https://github.com/zeromq/libzmq/blob/812e756264cc92d735dc56d91eaf8db2843f77ce/src/zmq.cpp#L783 i get:

```
Running czmq test 'zloop'...
 * zloop: D: 17-11-01 15:24:18 zloop: register timer id=1 delay=1000 times=1
D: 17-11-01 15:24:18 zloop: register timer id=2 delay=5 times=1
D: 17-11-01 15:24:18 zloop: register timer id=3 delay=20 times=1
D: 17-11-01 15:24:18 zloop: register PAIR reader
D: 17-11-01 15:24:18 zloop polling for 5 msec
Value of zmq_errno(): 35
D: 17-11-01 15:24:18 zloop: call timer handler id=2
D: 17-11-01 15:24:18 zloop: cancel timer id=1
D: 17-11-01 15:24:18 zloop polling for 14 msec
Value of zmq_errno(): 35
D: 17-11-01 15:24:18 zloop: call timer handler id=3
D: 17-11-01 15:24:18 zloop polling for 9976 msec
D: 17-11-01 15:24:18 zloop: call PAIR socket handler
Value of zmq_errno(): 60
Value of zmq_errno(): 60
OK
```

### DIFF
--- a/src/socket_poller.cpp
+++ b/src/socket_poller.cpp
@@ -553,6 +553,10 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_,
 #elif defined ZMQ_HAVE_ANDROID
         usleep (timeout_ * 1000);
         return -1;
+#elif defined ZMQ_HAVE_OSX
+        usleep (timeout_ * 1000);
+        errno = EAGAIN;
+        return -1;
 #else
         usleep (timeout_ * 1000);
         return -1;


ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

sigiesec:
@AntonBarwald Which poller implementation are you using?

AntonBarwald:
@sigiesec Okey, I will look in to the specific poller implementation. Should I close this PR and create a issue instead? 

I think I use kqueue ? Can that be correct?

bluca:
Yes, on OSX the default is kqueue. Thanks! Feel free to leave this open and just force push a new commit, as you prefer.

sigiesec:
The socket_poller_t does not use kqueue, but only select or poll, if I see this correctly. At first sight, I don't see where an ETIMEDOUT might result from socket_poller_t::wait. Maybe the problem is a bit deeper than just the wrong error code being returned.

AntonBarwald:
What I found is that after calling to usleep, errno == 60:
https://github.com/zeromq/libzmq/blob/812e756264cc92d735dc56d91eaf8db2843f77ce/src/socket_poller.cpp#L557  

I don't know if this is a good enough "workaround"? But I have no idea how to continue troubleshooting, any thoughts?
```
#else
        usleep (timeout_ * 1000);
        errno = EAGAIN;
        return -1;
#endif
```

sigiesec:
@AntonBarwald Yes, I think setting errno to EAGAIN explicitly is the right solution. Great that you found that out! It is by no means obvious that usleep changes errno, and this is documented nowhere for Darwin/Mac OS X. However, looking at its implementation, this is obviously the case: https://opensource.apple.com/source/Libc/Libc-583/gen/nanosleep.c (nanosleep is called by usleep)

ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

AntonBarwald:
@sigiesec Yes you are right, when looking at the implementation it make sense. 

Should I update the title of the PR or change anything else ?

ricnewton:
Merged build finished.

sigiesec:
@AntonBarwald Thanks a lot :) I think the rest can stay as it is :) I suggest just to wait until the Mac OS X travis builds have completed successfully, and then the PR can be merged.

AntonBarwald:
@sigiesec No worries, thank you for guiding me in the right direction :)

Okay, sounds good.

PR #2816, opened 2017-11-02
Problem: Build fails on systems without accept4()

barsnick:
Solution: Add a fallback to accept() for systems without accept4()

The accept4() system call was originally introduced as a proprietary Linux
extension starting with Linux 2.6.28; support in glibc was available
starting with version 2.10. Others also introduced it later (e.g. FreeBSD
from 10.0).

This uses the same fallback to POSIX's accept() as already used for
systems without SOCK_CLOEXEC.

Tested with Linux 2.6.27, glibc 2.9.

# Pull Request Notice

Before sending a pull request make sure each commit solves one clear, minimal,
plausible problem. Further each commit should have the following format:

```
Problem: X is broken

Solution: do Y and Z to fix X
```

git fetch upstream
git rebase upstream/master
git push -f

git rebase -i HEAD~10

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -617,7 +617,7 @@ AC_LANG_POP([C++])
 
 # Checks for library functions.
 AC_TYPE_SIGNAL
-AC_CHECK_FUNCS(perror gettimeofday clock_gettime memset socket getifaddrs freeifaddrs fork posix_memalign mkdtemp)
+AC_CHECK_FUNCS(perror gettimeofday clock_gettime memset socket getifaddrs freeifaddrs fork posix_memalign mkdtemp accept4)
 AC_CHECK_HEADERS([alloca.h])
 
 # pthread_setname is non-posix, and there are at least 4 different implementations
--- a/src/ipc_listener.cpp
+++ b/src/ipc_listener.cpp
@@ -382,7 +382,7 @@ zmq::fd_t zmq::ipc_listener_t::accept ()
     //  The situation where connection cannot be accepted due to insufficient
     //  resources is considered valid and treated by ignoring the connection.
     zmq_assert (s != retired_fd);
-#if defined ZMQ_HAVE_SOCK_CLOEXEC
+#if defined ZMQ_HAVE_SOCK_CLOEXEC && defined HAVE_ACCEPT4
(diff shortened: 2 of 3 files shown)


ricnewton:
Merged build started.

ricnewton:
Merged build triggered.

ricnewton:
Merged build triggered.

ricnewton:
Merged build finished.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

bluca:
Hi, thanks for the PR, but unfortunately there's a couple of problems that would break using CLOEXEC in 2 cases:

1) When using CMake - could you please add a check_function_exists for accept4 in CMakeList.txt ?
2) When there is SO_CLOEXEC but no accept4, the following if below the change in src/tcp_listener.cpp will be false: ```!defined ZMQ_HAVE_SOCK_CLOEXEC && defined FD_CLOEXEC``` so the fallback to FD_CLOEXEC won't be used. I guess it should be changed to something like ```(!defined ZMQ_HAVE_SOCK_CLOEXEC || !defined HAVE_ACCEPT4) && defined FD_CLOEXEC```

barsnick:
Thanks for the review.

1. Okay, two build systems, maintenance hell, wow. ;-) Sorry, I missed that, Fedora (where I got the build process from) makes use of autoconf. I'll try to polish my CMake knowledge and fix that, as adding a test for a function should be next to trivial.
2. I'll have another look at the logic.

bluca:
The CMake addition should  be pretty easy, there's already usage of check_function_exists which is the equivalent of AC_CHECK_FUNC, it should be just a matter of adding that and the stub define in builds/cmake/platform.hpp.in

bluca:
@barsnick any news? If you need help with CMake I can take care of it, but please do review the ifdef logic

bluca:
@barsnick I'd like to have this ready for 4.2.3 which is imminent, so I have fixed the logic and added cmake support here: https://github.com/zeromq/libzmq/pull/2842 so closing this one. Thanks for brinding the problem to our attention!

PR #2825, opened 2017-11-08
Problem: REQ socket with ZMQ_REQ_RELAXED does not report ZMQ_POLLOUT when queried for events after first message.

### LINKED ISSUE #2824: ZMQ_REQ socket does not report ZMQ_POLLOUT when ZMQ_REQ_RELAXED is set
# Issue description

When creating a ZMQ_REQ socket with ZMQ_REQ_RELAXED enabled, the ZMQ_EVENTS socket option does not indicate the socket is writable after sending the first message. In spite of that, a call to `zmq_send()` does in fact succeed without errors (as expected).

# Environment

* libzmq version: 4.4.2
* OS: macOS High Sierra

# Minimal test code

```
#include <zmq.h>
#include <unistd.h>

int main() {
    void* ctx = zmq_ctx_new();
    void* req = zmq_socket(ctx, ZMQ_REQ);
    void* rep = zmq_socket(ctx, ZMQ_REP);

    int res;
    int events;
    size_t events_len;

    int relaxed = 1;
    res = zmq_setsockopt(req, ZMQ_REQ_RELAXED, &relaxed, sizeof(relaxed));
    assert(res == 0);

    res = zmq_connect(req, "tcp://127.0.0.1:12345");
    assert(res == 0);

    res = zmq_bind(rep, "tcp://127.0.0.1:12345");
    assert(res == 0);

    sleep(1);

    res = zmq_getsockopt(req, ZMQ_EVENTS, &events, &events_len);
    assert(res == 0);
    printf("%i\n", (events & ZMQ_POLLOUT) == ZMQ_POLLOUT);

    res = zmq_send(req, "foo", 3, 0);
    assert(res == 3);

    res = zmq_getsockopt(req, ZMQ_EVENTS, &events, &events_len);
    assert(res == 0);
    printf("%i\n", (events & ZMQ_POLLOUT) == ZMQ_POLLOUT);

    res = zmq_send(req, "bar", 3, 0);
    assert(res == 3);
    return 0
}
```

# What's the actual result?

The output of this program is:

```
1
0
```

That is; `(events & ZMQ_POLLOUT) == ZMQ_POLLOUT` evaluates as true for the first message, but false after sending the second message. 

# What's the expected result?

My expectation is that the output would be:

```
1
1
```

In other words; `(events & ZMQ_POLLOUT) == ZMQ_POLLOUT` for the REQ socket if ZMQ_REQ_RELAXED is set.

This seems like a problem in https://github.com/zeromq/libzmq/blob/master/src/req.cpp#L216 – I believe the code should check for strictness just like sending does; therefore I think it should be:

```
if (receiving_reply && strict)
    return false;
```

Does this make sense? Am I right in expecting that the REQ socket with ZMQ_REQ_RELAXED enabled should be writable according to the ZMQ_EVENTS socket option?

rolftimmermans:
Solution: Check for strictness before returning false if no reply has been received.

Fixes #2824.

### DIFF
--- a/src/req.cpp
+++ b/src/req.cpp
@@ -213,7 +213,7 @@ bool zmq::req_t::xhas_in ()
 
 bool zmq::req_t::xhas_out ()
 {
-    if (receiving_reply)
+    if (receiving_reply && strict)
         return false;
 
     return dealer_t::xhas_out ();
--- a/tests/test_req_relaxed.cpp
+++ b/tests/test_req_relaxed.cpp
@@ -54,6 +54,16 @@ static void bounce (void *socket)
     } while (more);
 }
 
+static int get_events (void *socket)
+{
+    int rc;
+    int events;
+    size_t events_size = sizeof(events);
+    rc = zmq_getsockopt (socket, ZMQ_EVENTS, &events, &events_size);
+    assert (rc == 0);
+    return events;
+}
+
 int main (void)
 {
     setup_test_environment ();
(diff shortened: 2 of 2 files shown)


ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

rolftimmermans:
Looking at the build failures, they seem unrelated. Should the builds just be restarted?

bluca:
Yeah it's a known tracked issue, unrelated. Thanks for the PR!

PR #2840, opened 2017-11-16
Problem: Build fails in Visual Studio 2008 without `stdint.h`

ccpaging:
```
Problem: Build fails on Visual Studio 2008 without `stdint.h`

Solution: Cleaning include files which include `stdint.h` except `stdint.hpp`.
Define UINT32_MAX in `stdint.hpp`.
```

git fetch upstream
git rebase upstream/master
git push -f

git rebase -i HEAD~10

### DIFF
--- a/src/precompiled.hpp
+++ b/src/precompiled.hpp
@@ -63,7 +63,6 @@
 #include <signal.h>
 #include <stdarg.h>
 #include <stddef.h>
-#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

jimklimov:
I think the proper way would be to detect if the system has the header - like detected for many other things - and use it ifdef it's available. The library's definitions are all ifndef on purpose - they are fallbacks (system definitions preferred).

ccpaging:
Detecting and handling `stdint.h` is in `stdint.hpp` also.

It is not necessary to include `stdint.h` again in other include files. I just clean them and test it OK in Visual Studio 2008.

bluca:
yeah I think this is fine - stdint is already handled "differently" - probably easier than adding checks in every build system

PR #2842, opened 2017-11-18
Problem: accept4 not available on all platforms

bluca:
Solution: check for availability in CMake and autoconf before using it


ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

Issue #2817, opened 2017-11-02
ISO C++ forbids braced-groups within expressions


jdemeyer:
# Issue description

On my Gentoo Linux system, zeromq does not build due to
```
src/msg.cpp:400:5: error: ISO C++ forbids braced-groups within expressions [-Werror=pedantic]
     assert (metadata_ != NULL);
```
The reason is that glibc 2.25 uses GNU extensions to define `assert`. Since zeromq does not use `-std=c++XY` or `-ansi`, such GNU extensions should be allowed.

But the **real** reason for the broken build is `-Werror`, see #2818 for that.

# Environment

* libzmq version (commit hash if unreleased): 4.2.2
* OS: `Linux tamiyo 3.17.7-gentoo #2 SMP PREEMPT Fri Dec 23 18:13:49 CET 2016 x86_64 Intel(R) Core(TM) i7-2640M CPU @ 2.80GHz GenuineIntel GNU/Linux`
* Compiler: g++ (Gentoo 4.9.4 p1.0, pie-0.6.4) 4.9.4
* glibc: glibc-2.25 (+ Gentoo patches)

# Minimal test code / Steps to reproduce the issue

1.  Build zeromq

# What's the actual result? (include assertion message & call stack if applicable)

```
src/msg.cpp:400:5: error: ISO C++ forbids braced-groups within expressions [-Werror=pedantic]
     assert (metadata_ != NULL);
```

# What's the expected result?

A working build of zeromq

bluca:
std=gnu11 is used by default, with fallback to gnu99, not sure why is it falling for that?

jdemeyer:
> std=gnu11 is used by default, with fallback to gnu99, not sure why is it falling for that?

But zeromq is a C++ library, not C. So the C standard flag seems irrelevant.

bluca:
The c++ equivalents, of course

bluca:
@jdemeyer I cannot reproduce on Debian sid x86_64 with glibc 2.25 and gcc 7:

```
root@BRA-6G85P12:/zeromq3-4.2.2# ldd debian/tmp/usr/lib/x86_64-linux-gnu/libzmq.so.5.1.2 
	linux-vdso.so.1 (0x00007ffe37909000)
	libeatmydata.so => /usr/lib/x86_64-linux-gnu/libeatmydata.so (0x00007fbf3ada6000)
	/usr/lib/cowdancer/libcowdancer.so (0x00007fbf3aba0000)
[...]

PR #2844, opened 2017-11-19
Problems: CMake rebuilds everything twice, warnings with Tweetnacl, no autoconf option to disable Werror

### LINKED ISSUE #2818: Stop using -Werror
# Issue description

zeromq is built with the `-Werror` compiler flag. This means that any kind of warning in the build toolchain (compiler, assembler, linker) breaks the build. The option `-Werror` may be great when you have complete control over the operating system and the toolchain (for example, on a CI system). But who knows on what kind of systems and with which compilers people will compile zeromq. The `-Werror` option is breaking zeromq on systems where it would otherwise build and work perfectly fine. The fact that zeromq already disables `-Werror` on some systems should also be a sign that the option is unreliable.

Concretely, my build is broken due to #2817. Regardless of whether/how you fix that issue, I'm sure that `-Werror` will cause other breakage for no good reason.

bluca:
Solutions: fix warnings, add autoconf --disable-Werror option and do intermediate CMake build to avoid rebuilding all objects.

The CMake fix is on *nix only - the cloudapp CI reported a failure on Windows, but it's now down and I cannot access the log anymore. Help would be greatly appreciated.

Fixes #2818


ricnewton:
Merged build triggered.

ricnewton:
Merged build started.

ricnewton:
Merged build finished.

bluca:
ping - anyone for a merge?

bluca:
@sigiesec any chance you could please give this a test on Windows? I managed to log one last time in cloudapp and solve the problem there, but just in case

sigiesec:
@bluca I made a brief check building via cmake, and it is working. I am sorry but I am very busy with other stuff right now, so I cannot give it a deeper look :(

bluca:
That's fine, thanks!

jdemeyer:
IMHO, `-Werror` should never be the default. Activating such a dangerous option should require explicit action. However, I don't care enough to make a pull request.

bluca:
All compiler warnings must either be solved, or in a very very small number of cases explicitly and individually disabled with pragmas when they are false positives, so Werror is very helpful as a default.

jdemeyer:
> All compiler warnings must be solved

by **developers**

> so Werror is very helpful as a default

for **developers**.

As a **user**, it's not at all helpful if my build of `zeromq` fails because I happen to have an OS or toolchain which emits unexpected warnings.

bluca:
As a user you can report what problems you find so that they get actually fixed, rather than ignored, be it in the toolchain or elsewhere.

For example gcc 7 got stricter in format parsing, and thanks to the user that reported them several bugs which could be critical were fixed immediately rather than being ignored for years

jdemeyer:
> As a user you can report what problems you find so that they get actually fixed

Fixed like in #2817?

In all seriousness: as a user, I don't care about compiler warnings. All I want is compiling zeromq on my system. The build system should always be as user-friendly as possible. And intentionally breaking a build which would otherwise work just fine isn't user friendly.

> For example gcc 7 got stricter in format parsing, and thanks to the user that reported them several bugs which could be critical were fixed immediately rather than being ignored for years

A better solution would be to test with GCC 7 on a continuous integration system and enable `-Werror` there. Then you wouldn't need to bother users with your bugs.

bluca:
> Fixed like in #2817?

> in the toolchain or elsewhere

Have you reported it to the toolchain maintainer? Either that's a bug in the toolchain, or you are using a broken combination of libc + gcc, both of which are in your interest to get fixed.

> In all seriousness: as a user, I don't care about compiler warnings. All I want is compiling zeromq on my system. The build system should always be as user-friendly as possible. And intentionally breaking a build which would otherwise work just fine isn't user friendly.

Ignoring compiler errors does not mean it's working fine - it just means sweeping problems under the carpet and hoping they don't come back and bite you further down the line. Especially in a library that implements encryption algorithms and does network I/O that's a really bad idea. But now you have a flag that allows you to do just that, if you wish.

PR #2852, opened 2017-12-01
Problem: Debian packages are missing zmq.hpp

bluca:
Solution: install it if available. Download it from its repo as part
of the OBS source run.
Debian and Ubuntu always ship zmq.hpp in libzmq-dev, so do the same.
In the RPM world it is in its own separate package, so don't do the
same for RPM.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -52,6 +52,12 @@ else
 endif
 endif
 
+override_dh_auto_install:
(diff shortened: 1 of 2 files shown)


stephanlachnit:
Note: this has been changed in Debian 12 / Ubuntu 24.04, `zmq.hpp` is now in `cppzmq-dev`.

PR #2861, opened 2017-12-12
export ZMQ_STATIC compile flag to depending projects (MSVC)

glaure:
Problem: If libzmq is used as static library on windows all using projects need to define "ZMQ_STATIC".
In case libzmq is part of a larger project, this patch automatically exports the needed flag using cmake functionality.

Solution: export flag using target_compile_definitions
```


glaure:
Luca I just wanted to thank you (and all others too)! It is nice to be able to contribute to this project!

bluca:
No problem, and thanks for all the fixes!

PR #2868, opened 2017-12-14
Problem: Build is broken on OpenBSD

jasperla:
Solution: re-instate block for OpenBSD erroneously removed in 9622a830

### DIFF
--- a/src/precompiled.hpp
+++ b/src/precompiled.hpp
@@ -39,6 +39,10 @@
 #include "windows.hpp"
 #endif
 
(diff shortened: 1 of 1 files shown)


bluca:
Thanks, sorry for the regression - sadly we have no CI for BSDs

jasperla:
Your quick response makes up a bit for that ;-)

jimklimov:
But now people can fire up Jenkins on their platform of choice and get it tested by just pointing a github organization folder at zeromq org ;) And perhaps register an account so it can update PR statuses.

PR #2870, opened 2017-12-14
Problem: vc++ 11.0 (vs 2012) compile fails on move semantics.

evoskuil:
Solution: disable move semantics at the correct MSVC version.


evoskuil:
Nuget packages for [VS2013](https://www.nuget.org/packages/libzmq_vc120), [VS2015](https://www.nuget.org/packages/libzmq_vc140), [VS2017](https://www.nuget.org/packages/libzmq_vc141) are now published for version 4.2.3. NuGet package versions indicate 4.2.30 which is the result of a numbering mistake a while back in 4.2.x. Once we are past 4.2 it will be correct again.

bluca:
What is the binary file builds/msvc/resource.rc ? Was it added by mistake?

evoskuil:
It's a text file, updated (not added) for the version number change. It's a GUI-generated (or manually) resource definition. Specifically in this case it contains file metadata for DLLs.

bluca:
Ah ok, thanks - for some reasons GitHub says it's a binary file so I got confused for a moment

evoskuil:
NP, maybe the extension and a BOM giving it trouble.

PR #2872, opened 2017-12-20
Fix CMake tests on Windows

cowo78:
CMake tests for some symbols (WSAStartup, UuidCreateSequential, GetAdaptersAddresses) were checked in the wrong way.
Added automatic dependies pull-in for windows/msvc.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -121,20 +121,10 @@ extern "C" {
     #include <poll.h>
 #endif
 
-/* Advise the linker to automatically pull dependencies in */
 #if defined _WIN32
 #   if defined _MSC_VER
-#       if (HAVE_WS2_32)
-#           pragma comment(lib, "ws2_32.lib")
-#       elif (HAVE_WS2)
-#           pragma comment(lib, "ws2.lib")
-#       endif  // HAVE_WS2_32
-#       if (HAVE_IPHLAPI)
-#           pragma comment(lib, "iphlpapi.lib")
-#       endif  // HAVE_IPHLAPI
-#       if (HAVE_RPCRT4)
-#           pragma comment(lib, "rpcrt4.lib")
-#       endif  // HAVE_RPCRT4
+#       pragma comment(lib, "ws2_32.lib")
+#       pragma comment(lib, "iphlpapi.lib")
 #   endif  // _MSC_VER
 #endif  // _WIN32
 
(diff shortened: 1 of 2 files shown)


include/zmq.h
```diff
@@ -121,6 +121,16 @@ extern "C" {
     #include <poll.h>
 #endif
 
+/* Advise the linker to automatically pull dependencies in */
+#if defined _WIN32
+#   if defined _MSC_VER
+#       pragma comment(lib, "ws2_32.lib")
```
bluca: This is the public header, can it be done elsewhere? If not, are you 110% sure this is compatible with all versions of Windows from XP up, on all architectures, with all versions of VS/VC++/etc? Once it's in, it cannot be taken out

sigiesec: I think this is also unnecessary when libzmq is built as a DLL. 

Also, in case of building libzmq as a static library, I don't think this is generally desirable. Depending on the context, one might want to control this explicitly.

Supporting such auto-linking might be an optional feature. Boost, e.g., also supports autolinking optionally. The solution is quite sophisticated as the library naming ensures that the correct variant of a library is selected. See http://www.boost.org/doc/libs/1_66_0/boost/config/auto_link.hpp

cowo78: Of course it can be done elsewhere. I don't think it has anything to do with the OS SDK, while AFAIK it's supported on msvc since ~10 years.

@sigiesec it does not matter whether the target is static or dynamic (agree I could have split into 2 pull requests). AFAIK those libs are always dynamic (import libraries in msvc terms) so you always need to link them. The boost solution is designed to link the correct version of the boost library, not OS/3rd party dependencies. Agreed it can be an optional feat.

sigiesec: ws2_32 is always dynamic, i.e. ws2_32.lib is always only an import library.

But a dynamic libzmq.dll only uses ws2_32.lib internally and does not expose this dependency to its clients, so they do not required to link against ws2_32.

cowo78: @sigiesec I see your point, agreed. Will wrap the pragmas in ZMQ_STATIC. I was not suggesting to add auto-link facility like boost (handy stuff but too much work involved).

cowo78: @bluca Had a look at the zmq.h header but since it does not include private stuff but only declares prototypes what I meant to do must be done either in zmq.h or in something that zmq.h directly includes.
To sum up:
* PRO Don't need to explicitly link zmq
* CON Need to clutter zmq.h or include private "auto-link" file on win32/msvc

maybe it's just easier to drop the auto-link feature and keep the CMake patch if you like it. Your shout, the auto-link was essentially candy, the real stuff is the CMake patch.

bluca: I'd suggest small steps - let's start with the cmake part

Issue #788, opened 2013-12-14
SUB accepts multiple connects to same PUB endpoint and result is nonsensical


trevorbernard:

WallStProg:
I've recently run into a problem related to this one, and would like to get some feedback on a potential solution.

The problem is that for pub/sub the original fix for this issue only covers the case where the socket is a SUB socket -- PUB sockets are allowed to connect multiple times to the same endpoint, and when they do that, they receive duplicate messages.

Now, I know that there are issues related to PUB sockets connecting to SUB sockets, but in some architectures that is unavoidable, and the library/application will need to deal with those issues.

Given that, wouldn't it make sense to include PUB sockets in the `is_single_connect` test?

FWIW, that approach appears to be working in my environment, and I'm happy to submit a PR if the community is interested.

bluca:
I think it makes sense, feel free to send a PR, thanks

ilyakrasnovsky:
Does it also make sense to add `XSUB`  and `XPUB` sockets to the `is_single_connect` test? I've noticed the same message-duplication problem when I `connect` more than once to an existing address from an `XSUB` socket. I can open a new issue if that helps, but wanted to jump off of a relevant discussion. Thanks!

WallStProg:
I’m not familiar with XPUB/XSUB, as I don’t use them (except w/zmq_proxy), but a couple of points:

- It would seem that hard-coding this behavior could interfere with manual handling of subscribe/unsubscribe requests — e.g. a particular application might want to use ZMQ_XPUB_VERBOSE, ZMQ_XPUB_MANUAL, etc. to handle the case of duplicate subscribe requests
- For that reason, it might be better to have this behavior controlled by a socket option?

While we’re on the topic (no pun intended), an argument can be made that silently ignoring duplicate subscription requests is not the best approach, and that it might be better to return a specific errno value (EEXIST?) in the case where a duplicate subscription is being ignored.  Of course, this would break existing code and so should also be a socket option.  Not a pain point for me, but I can imagine that it might be useful in some cases.

ilyakrasnovsky:
[...]

PR #2879, opened 2017-12-29
prevent duplicate connections from PUB sockets also 

WallStProg:
(see https://github.com/zeromq/libzmq/issues/788)

The original fix for the duplicate connect issue only covered SUB connecting to PUB, but the same problem exists in the other direction.  (Irrespective of any other issues related to PUB connecting to SUB).

Note to Luca -- let me know if you need/want a test pgm for this, or if this behavior should be documented somewhere (e.g.in zmq_connect.txt?)

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -823,7 +823,7 @@ int zmq::socket_base_t::connect (const char *addr_)
         return 0;
     }
     bool is_single_connect = (options.type == ZMQ_DEALER ||
-                              options.type == ZMQ_SUB ||
+                              options.type == ZMQ_SUB    || options.type == ZMQ_PUB ||
                               options.type == ZMQ_REQ);
     if (unlikely (is_single_connect)) {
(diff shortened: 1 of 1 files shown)


bluca:
Thanks! A PGM test sounds good - will be useful in any case. Adding more details to the docs sounds good as well.

WallStProg:
I meant test program, not PGM test ;-)  I'll look at whipping something together -- assume that should go in tests directory.  FWIW, I don't see anything there now that tests this feature.

As for docs, I'll put something in zmq_connect.txt, then if you want to move it elsewhere you can.

bluca:
Yes the test directory is the right place. Can it be self-contained (provided libpgm is installed) or does it need some special setup in the system?

WallStProg:
Sorry it's taken so long ... finally had time to put together doc changes and a small test program -- please see https://github.com/zeromq/libzmq/pull/3117

PR #2885, opened 2018-01-10
removing hard-coded path of librt

JohanMabille:
The full path to librt.so is hard-coded in the generated file ZeroMQTargets.cmake. While this is not problematic when installing from source, this may prevent to link with ZeroMQ if the package has been built on another machine with a different folder / system structure.

This behavior is the result of linking with the result of the `find_library` instruction, which returns a full path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -870,7 +870,7 @@ if (BUILD_SHARED)
   endif ()
 
   if (RT_LIBRARY)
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
+1000 on this. The bug that this fixes has caused issues in the context of packaging for conda.

SylvainCorlay:
Great!

PR #2888, opened 2018-01-13
Fix inclusion of cmake files in dists

minrk:
The absence of these files means that zeromq releases cannot be built with cmake.

move a few cmake-specific files into the cmake build directory so that it's self contained.


SylvainCorlay:
Thanks a lot for this. I wanted to fix this but had to be AFK most of the day.

PR #2901, opened 2018-01-27
Enable ccache for faster builds

thedrow:
Because, why not?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ os:
 
 dist: trusty
 
+cache: ccache
+
 env:
   matrix:
     - BUILD_TYPE=default CURVE=tweetnacl DRAFT=enabled


bluca:
I'm afraid it's not so simple, the CI scripts need to be updated to optionally set CC and CXX to ccache like in here: https://github.com/zeromq/czmq/blob/master/ci_build.sh

thedrow:
I'm pretty sure that ccache overrides the default gcc and clang with symlinks.

jimklimov:
I am surprised I never added it here. So much for not dogfeeding with zproject here ;)

Indeed, much of that magic for ccache variable setup may be superfluous nowadays (was different a few years back), at least assuming there is only one true compiler. Or two, both covered by ccache. This is quite likely on Travis CI cloud, may be not so on private deployments battered to work on different OSes and distros and tools - so I'd rather not rip that out.

But here - yay, good catch :)

PR #2903, opened 2018-01-31
Problem: in case of exhausted resources on creation of a context, assertions are triggered

### LINKED ISSUE #2895: Assertion failure: !(pollset [i].revents & POLLNVAL) (poll.cpp:157)
# Issue description

An assertion failure occurs in the following line:
https://github.com/zeromq/libzmq/blob/3226b8ebddd9c6c738ba42986822c26418a49afb/src/poll.cpp#L157

The fd member of the offending entry in pollset in that case has the value -1 / retired_fd.

Apparently, this was already the case when poll was called, which caused poll to set the revents of the entry to POLLNVAL.

I found that it is possible to call add_fd with a fd_ parameter of retired_fd/-1, which is not checked, so I assume this has happened somehow, but I do not know via which call chain this happened.

# Environment

* libzmq version (commit hash if unreleased): 4.2.3 built with ZMQ_USE_POLL
* OS: Windows

Unfortunately, I have not been able to reproduce this issue. It occurs if some networking issues happen, such as firewall configurations between peers change, or the local network settings are reconfigured. I don't know yet what exactly triggers the issue, but it has occurred several times on different machines.

Assertion failure: !(pollset [i].revents & POLLNVAL) within a zmq::poll_t::worker_routine thread.

# What's the expected result?

No assertion failure, operation continues.

sigiesec:
Solution: signal error to caller, and apply appropriate cleanup

Fixes #2895 (I hope)

### DIFF
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -161,7 +161,12 @@ void *zmq_ctx_new (void)
 
     //  Create 0MQ context.
     zmq::ctx_t *ctx = new (std::nothrow) zmq::ctx_t;
-    alloc_assert (ctx);
+    if (ctx) {
+        if (!ctx->valid ()) {
+            delete ctx;
+            return NULL;
+        }
+    }
     return ctx;
 }
 
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -127,6 +127,11 @@ zmq::ctx_t::~ctx_t ()
     tag = ZMQ_CTX_TAG_VALUE_BAD;
 }
 
+bool zmq::ctx_t::valid () const
+{
+    return term_mailbox.valid ();
+}
+
 int zmq::ctx_t::terminate ()
 {
     slot_sync.lock();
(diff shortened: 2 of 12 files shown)


bluca:
could you please move the whitespace changes to a different commit? Especially when touching something so critical that will likely have repercussions we should keep the history clean for easier bisecting and so on

sigiesec:
Yeah sorry... I accidentally reformatted everything, and I thought I reverted that but apparently I missed several places.

sigiesec:
Ok, I think I reverted all irrelevant changes now.

src/zmq.cpp
```diff
@@ -161,7 +161,12 @@ void *zmq_ctx_new (void)
 
     //  Create 0MQ context.
     zmq::ctx_t *ctx = new (std::nothrow) zmq::ctx_t;
-    alloc_assert (ctx);
+    if (ctx) {
+        if (!ctx->valid ()) {
+            delete ctx;
+            return NULL;
```
bluca: I'm not sure about this - it never returned null, so nobody will be checking for it, which means we might get a flurry of segfaults reports...

t-b: @bluca The documentation states that:

´´´
The zmq_ctx_new() function shall return an opaque handle to the newly created context if successful. Otherwise it shall return NULL and set errno to one of the values defined below.
´´´

so I would expect good code to already check for null.

PR #2910, opened 2018-02-02
Problem: race conditions for options.linger

sigiesec:
Solution: make options.linger atomic

Note: this requires #2908 to be merged first.

### DIFF
--- a/src/atomic_ptr.hpp
+++ b/src/atomic_ptr.hpp
@@ -34,8 +34,8 @@
 #define ZMQ_ATOMIC_PTR_MUTEX
 #elif defined ZMQ_HAVE_ATOMIC_INTRINSICS
 #define ZMQ_ATOMIC_PTR_INTRINSIC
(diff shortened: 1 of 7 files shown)


sigiesec:
Fixes part of #2733

src/atomic_ptr.hpp
```diff
@@ -34,7 +34,8 @@
 #define ZMQ_ATOMIC_PTR_MUTEX
 #elif defined ZMQ_HAVE_ATOMIC_INTRINSICS
 #define ZMQ_ATOMIC_PTR_INTRINSIC
-#elif (defined __cplusplus && __cplusplus >= 201103L)
+#elif (defined __cplusplus && __cplusplus >= 201103L                           \
```
bluca: might be better with a couple more brackets for easier reading

sigiesec: Ack. Will add them.

PR #2918, opened 2018-02-03
Problem: test_timers and test_timeo frequently fail on travis-ci

sigiesec:
Solution: relaxed test assumptions

Note: this PR requires #2908 to be merged first.


include/zmq.h
```diff
@@ -542,7 +542,11 @@ ZMQ_EXPORT void zmq_atomic_counter_destroy (void **counter_p);
 ZMQ_EXPORT void *zmq_stopwatch_start (void);
 
 /*  Stops the stopwatch. Returns the number of microseconds elapsed since     */
-/*  the stopwatch was started.                                                */
+/*  the stopwatch was started, but does not stop or deallocate the stopwatch. */
```
bluca: > Stops the stopwatch....  but does not stop or deallocate the stopwatch

Looks a bit wrong :-) Also it's a new API so it should be DRAFT

sigiesec: Yes :D I will fix that.

sigiesec: Fixed in #2926

Issue #2919, opened 2018-02-03
Windows Intel compiler build broken


tetsuh:
# Issue description

Windows Intel compiler build broken.
```
include\zmq.h(250): error : expected a ";"
    unsigned char _ [64] __attribute__ ((aligned (sizeof (void *))));
```

# Environment

* libzmq version (commit hash if unreleased):  v4.2.0 or later
* OS:  Windows 7 or Windows 10
* Compiler: Visual Studio 2017 + Intel Parallel Studio XE 2018

# Minimal test code / Steps to reproduce the issue

1.  build libzmq from source by MS VC++
2.  prepare any C++ a source code which include "zmq.h"
3.  Compile it by Intel compiler

tetsuh:
It looks ifdef order issue of struct zmq_msg_t in zmq.h
#2920 
I confirmed following environments are okay but the other is unsure.
Windows MS VC++, Windows Intel compiler, Linux GCC, Linux Clang, Linux Intel compiler.

PR #2920, opened 2018-02-03
Problem: Windows Intel compiler build broken

tetsuh:
Problem: #2919 Windows Intel compiler build error by include zmq.h 

Solution: 
  Modified ifdef condition order of typedef struct zmq_msg_t alignement.
  Placed if defined(_MSC_VER) before the other compiler.


bluca:
Thanks!

bluca:
Sorry - have to temporarily revert, nothing to do with your changes but just to avoid merge conflicts in a fairly large style reformatting happening in #2908 - I will take care of re-applying your fix afterwards. Should not take long.

tetsuh:
sure, thanks.

PR #2923, opened 2018-02-05
Add support for SIO_LOOPBACK_FAST_PATH on Windows

### LINKED ISSUE #2899: Use TCP loopback fastpath in zmq::signaler_t::make_fdpair if available
From Windows 8 / Server 2012, a new socket option is available that improves performance of TCP loopback connections, which could improve performance of zmq::signaler_t.

For more information, see https://blogs.technet.microsoft.com/wincat/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path/

sigiesec:
Fixes #2899 

This PR requires #2908 to be merged first.

### DIFF
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -384,6 +384,33 @@ void zmq::signaler_t::forked ()
 }
 #endif
 
+#if defined ZMQ_HAVE_WINDOWS
+static void tune_socket (const SOCKET socket)
+{
+    BOOL tcp_nodelay = 1;
+    int rc = setsockopt (socket, IPPROTO_TCP, TCP_NODELAY,
+                         (char *) &tcp_nodelay, sizeof tcp_nodelay);
+    wsa_assert (rc != SOCKET_ERROR);
+
+    int sio_loopback_fastpath = 1;
+    DWORD numberOfBytesReturned = 0;
+
+    rc = WSAIoctl (socket, SIO_LOOPBACK_FAST_PATH, &sio_loopback_fastpath,
+                   sizeof sio_loopback_fastpath, NULL, 0,
+                   &numberOfBytesReturned, 0, 0);
+
+    if (SOCKET_ERROR == rc) {
+        DWORD lastError = ::WSAGetLastError ();
+
+        if (WSAEOPNOTSUPP == lastError) {
+            // This system is not Windows 8 or Server 2012, and the call is not supported.
+        } else {
+            wsa_assert (false);
+        }
+    }
+}
+#endif
+
 //  Returns -1 if we could not make the socket pair successfully
 int zmq::signaler_t::make_fdpair (fd_t *r_, fd_t *w_)
 {
@@ -483,10 +510,8 @@ int zmq::signaler_t::make_fdpair (fd_t *r_, fd_t *w_)
     int rc = setsockopt (listener, SOL_SOCKET, SO_REUSEADDR,
                          (char *) &so_reuseaddr, sizeof so_reuseaddr);
     wsa_assert (rc != SOCKET_ERROR);
-    BOOL tcp_nodelay = 1;
-    rc = setsockopt (listener, IPPROTO_TCP, TCP_NODELAY, (char *) &tcp_nodelay,
-                     sizeof tcp_nodelay);
-    wsa_assert (rc != SOCKET_ERROR);
+
+    tune_socket (listener);
 
     //  Init sockaddr to signaler port.
     struct sockaddr_in addr;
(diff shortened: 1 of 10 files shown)


tests/test_pair_tcp.cpp
```diff
@@ -29,23 +29,39 @@
...
 #include "testutil.hpp"
 
-int main (void)
+typedef void (*extra_func_t) (void *socket);
+
+void set_sockopt_fastpath (void *socket)
+{
+    int value = 1;
+    int rc =
+      zmq_setsockopt (socket, ZMQ_LOOPBACK_FASTPATH, &value, sizeof value);
```
bluca: won't this fail on non-windows?

src/tcp.hpp
```diff
@@ -66,6 +66,10 @@ int tcp_read (fd_t s_, void *data_, size_t size_);
 //  Asserts that an internal error did not occur.  Does not assert
 //  on network errors such as reset or aborted connections.
 void tcp_assert_tuning_error (fd_t s_, int rc_);
+
+#if defined ZMQ_HAVE_WINDOWS
```
bluca: perhaps it would be cleaner to move the ifdef inside the body of the function, and remove it from elsewhere, and just make it a no-op on non-windows?

tests/test_pair_tcp.cpp
sigiesec: Not like I implemented it at the moment. It just won't have an effect. This will also be the case if it is used on a Windows version that does not support it. This behaviour could be changed, and an error be returned on non-Windows systems (but the same option could be used for similar optimizations on other OS's if present), or also on Windows systems that don't support it. However, this would require error handling by the caller, e.g. when using with zmqpp, this would throw an exception, which I found suboptimal. With the current design, the option can always be set, regardless of the execution environment, as long as the caller ensures that both sides of the connection apply it. This must still be documented, of course!

Would you do this differently?

bluca: Gah, I'm just very tired and for some reasons I read setsockopt rather than zmq_setsockopt, sorry.

Although there are other options that are platform specific if you check in src/options.cpp, and they are compiled out which means they return EINVAL, so perhaps it's worth keeping it consistent?

bluca:
```
../tests/test_pair_tcp.cpp: In function ‘void set_sockopt_fastpath(void*)’:
../tests/test_pair_tcp.cpp:38:31: error: ‘ZMQ_LOOPBACK_FASTPATH’ was not declared in this scope
       zmq_setsockopt (socket, ZMQ_LOOPBACK_FASTPATH, &value, sizeof value);
```

It also needs to be ifdefd for drafts

sigiesec: Regarding EINVAL: I agree if we regard this as platform-specific. However, I somehow see this particular option as platform-agnostic in the following sense: Setting ZMQ_LOOPBACK_FASTPATH to 1 can be done on any platform to activate any optimizations for TCP loopback connections that platform provides which need to be activated explicitly on both connection ends. On platforms that do not provide such optimizations, or enable them by default, this will be a no-op.

This would mean that a cross-platform app could set this option on any platform. Otherwise, it would need to make some platform-specific distinction.

However, on Windows, I would return success regardless of whether the actual OS version supports it or not. The actual result of setting the underlying sockopt is not available at that point, and so we would need to somehow "guess" if it will be successful.

If you still think EINVAL is better on non-Windows, I am happy to change it :)

Regarding the DRAFT: Will fix that.

bluca: Ok that's fine, let's leave it

sigiesec: Fixed non-DRAFT

src/tcp.hpp
sigiesec: Fixed

src/tcp.cpp
```diff
@@ -332,3 +332,25 @@ void zmq::tcp_assert_tuning_error (zmq::fd_t s_, int rc_)
...
+    if (SOCKET_ERROR == rc) {
+        DWORD lastError = ::WSAGetLastError ();
+
+        if (WSAEOPNOTSUPP == lastError) {
+            // This system is not Windows 8 or Server 2012, and the call is not supported.
+        } else {
+            wsa_assert (false);
+        }
+    }
+#endif
```
bluca: I think you'll need an
```
#else
LIBZMQ_UNUSED(socket);
```

src/signaler.cpp
```diff
@@ -384,6 +385,18 @@ void zmq::signaler_t::forked ()
 }
 #endif
 
+#if defined ZMQ_HAVE_WINDOWS
```
bluca: won't the ifdef outside the function break the build on non-windows?

sigiesec: The function is only called within a Windows-specific block below. But I will check the travis builds again.

bluca: Ah right, sorry missed that. Should be fine then.

bluca:
Do you have an idea how to fix the failing windows tests with poll?

Also a while ago all appveyor builds basically doubled in time, do you see anything suspicious in the logs?

PR #2929, opened 2018-02-06
Add windows poll build on Appveyor

sigiesec:


sigiesec:
Some tests are currently failing, which was the case before, but not visible as this was never built on appveyor. 
I intend to add unit tests for poller_t to make it easier to identify and fix the causes.

sigiesec:
I added a first skeleton for unit tests in https://github.com/sigiesec/libzmq/commit/a0fdd66c222d9b579319f0ab512450febbdc7fd5

PR #2935, opened 2018-02-09
Problem: no test framework, no unit tests

sigiesec:
Solution: integrate unity, create basic unit test structure

- [x] adapt autotools build (@bluca?)
- [x] decide on directory structure: should API tests be parallel to unittests?

### DIFF
--- a/external/unity/version.txt
+++ b/external/unity/version.txt
@@ -1,0 +1,1 @@
+https://github.com/ThrowTheSwitch/Unity/commit/b4aca70fd9e0ddf0afbdafb1b826f5edcfc1049b
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -948,6 +948,11 @@ set (ZMQ_BUILD_TESTS ON CACHE BOOL "Build the tests for ZeroMQ")
 if (ZMQ_BUILD_TESTS)
   enable_testing () # Enable testing only works in root scope
   ADD_SUBDIRECTORY (tests)
+  if (BUILD_STATIC)
+    add_subdirectory (unittests)
+  else ()
+    message (WARNING "Not building unit tests, since BUILD_STATIC is not enabled")
+  endif ()
 endif ()
 
 #-----------------------------------------------------------------------------
--- a/Makefile.am
+++ b/Makefile.am
@@ -429,7 +429,9 @@ test_apps = \
 UNITY_CPPFLAGS = -I$(top_srcdir)/external/unity
 UNITY_LIBS = $(top_builddir)/external/unity/libunity.a
 noinst_LIBRARIES = external/unity/libunity.a
-external_unity_libunity_a_SOURCES = external/unity/unity.c
+external_unity_libunity_a_SOURCES = external/unity/unity.c \
+	external/unity/unity.h \
+	external/unity/unity_internals.h
 
 tests_test_ancillaries_SOURCES = tests/test_ancillaries.cpp
 tests_test_ancillaries_LDADD = src/libzmq.la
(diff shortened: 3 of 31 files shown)


external/unity/version.txt
```diff
@@ -0,0 +1 @@
+https://github.com/ThrowTheSwitch/Unity/commit/b4aca70fd9e0ddf0afbdafb1b826f5edcfc1049b
```
bluca: wouldn't it be better to sync from a release tag?

bluca:
I can take care of autotools, but over the weekend (I hope :-) )

tests/unittests/unittest_poller.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca: date range should be just 2018 - sorry but with these things it's better to be pedantic...

bluca:
I'm fine with having a unittest subdirectory of tests - it's better to keep them together

sigiesec:
I asked myself if the directory structure should be changed to have tests/unittests and tests/apitests (or some other name, containing the existing tests). The current structure is problematic with cmake: properties are inherited from the base directory.

bluca:
The problem is that moving all the files has the usual problems with git blame, and also there might be scripts and such that are broken and will annoy people, I'd rather leave them there if possible

bluca:
I've added autotools support for unity and the unittests.
For the latter, for now, it's linking with libzmq.a - the alternative would be to list individually all the required source files in each Makefile.am target.

sigiesec:
Cool, great :) Yes, that's what I am doing with cmake as well (link with libzmq-static). It would be quite hard to figure out which cpp files are really needed for each test, and I don't think that would provide a substantial benefit.

bluca:
Ok - still needs some work though, check works but distcheck doesn't

bluca:
distcheck should work fine now - also I rebased for the other changes to autotools, and added the copyright notice for debian packaging

bluca:
One of the problems is that the static library build might be disabled, and then I think the build will fail. Should probably check for it.

sigiesec:
Hm, yes... so maybe skip building the unit tests if the static library building is diabled? Maybe output a warning in that case. I think that is acceptable.

sigiesec:
Regarding the directory structure: I see the problem with moving the existing files. What about placing unittests not inside tests but parallel to tests? cmake doesn't really go well with nesting directories, which do not inherit all properties. I need to fix this within unittests/CMakeLists.txt at the moment.

bluca:
Fine for both - if you do that change, then I'll take care of the static thing in autotools

bluca:
@sigiesec  one of the windows tests fail:

```
c:\projects\libzmq\tests\testutil_security.hpp(558): warning C4550: expression evaluates to a function which is missing an argument list (C:\projects\libzmq\tests\test_security_curve.cpp) [C:\projects\build_libzmq\tests\test_security_curve.vcxproj]
C:\projects\libzmq\tests\test_security_curve.cpp(53): error C3861: 'snprintf': identifier not found [C:\projects\build_libzmq\tests\test_security_curve.vcxproj]
```

Could you please have a look?

sigiesec:
@bluca fixed the build failure with VS2013, hopefully

sigiesec:
@bluca ok, I changed the directory structure as proposed, and let the unit tests build only if BUILD_STATIC is enabled (in cmake)

bluca:
Great, done the same with autotools, also now ships the unittests/CMakeList.txt in the distributable tarball.

Also fixed the copyright dates in the new unit tests files

bluca:
Another TODO (but can be done separately) I think would be to split the security tests, as they are a bit hybrid right now and include the source files - move those to the unittests, and leave API-only stuff in tests/

bluca:
Also I've tested this branch with Unity on mips/armhf/arm64/hurd and it's working fine, later next week I'll do a few more architectures and Solaris 10

tests/unittests/unittest_poller.cpp
sigiesec: Ok, thanks for updating this!

external/unity/version.txt
sigiesec: Generally yes, I vaguely remember there was some reason not to use the latest release, but I will check again.

bluca:
So what's left to do for this to be merged?

sigiesec:
I think it can be merged :)

bluca:
Yay, great work!

What about the clang-format test? Should we apply that diff?

PR #2940, opened 2018-02-11
Problem: regression with "select" on *nix

bluca:
@sigiesec using select does not work anymore since recent changes, it works fine with 4.2.2 though.

There is still one weird linking problem left:

```
  CXXLD    tools/curve_keygen
src/.libs/libzmq.so: undefined reference to `zmq::select_t::family_entry_t::family_entry_t()'
collect2: error: ld returned 1 exit status
```

I think it's cause by this line: https://github.com/zeromq/libzmq/blob/master/src/select.hpp#L118

Is that perhaps a Windows compiler specific thing?

EDIT: I think the constructor was missing the body and GCC is not happy about that - tests still fail though

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ matrix:
   - env: BUILD_TYPE=default POLLER=select
     os: linux
 
-sudo: required
+sudo: false
 
 before_install:
 - if [ $TRAVIS_OS_NAME == "osx" -a $BUILD_TYPE == "android" ] ; then brew update; brew install binutils ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,6 +92,10 @@ matrix:
         - llvm-toolchain-trusty-5.0
       packages:
         - clang-5.0
+  - env: BUILD_TYPE=default POLLER=poll
+    os: linux
+  - env: BUILD_TYPE=default POLLER=select
+    os: linux
 
 sudo: required
 
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -48,6 +48,10 @@ if [ $BUILD_TYPE == "default" ]; then
         CONFIG_OPTS+=("--with-norm=yes")
     fi
 
+    if [ -n "$POLLER" ]; then
+        CONFIG_OPTS+=("--with-poller=${POLLER}")
+    fi
+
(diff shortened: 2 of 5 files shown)


bluca:
To be more precise, any test that creates a context gets stuck as the select call doesn't work anymore:

```
(gdb) bt
#0  0x00007ffff698f3c3 in select () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007ffff7b735bd in zmq::signaler_t::wait (this=0x555555769d28, timeout_=-1)
    at src/signaler.cpp:271
#2  0x00007ffff7b45f29 in zmq::mailbox_t::recv (this=0x555555769cc0, cmd_=0x7fffffffdc80, timeout_=-1)
    at src/mailbox.cpp:81
#3  0x00007ffff7b2900f in zmq::ctx_t::terminate (this=0x555555769c20) at src/ctx.cpp:186
#4  0x00007ffff7b9afae in zmq_ctx_term (ctx_=0x555555769c20) at src/zmq.cpp:179
#5  0x000055555555643c in main () at tests/test_pair_inproc.cpp:77
```

bluca:
I've added 2 jobs on Travis/Linux to test select and poll

sigiesec:
I will have a look at the test failures under Linux.

sigiesec:
I think I found the problem :) Will send a PR right away.

sigiesec:
Hm... I think there is more than one problem.

I am still confused about how the poller is intended to be used. Especially how should it be stopped?
stop/stopping is not synced at least in some of the implementations, so it looks like it should only be called from an event callback. However, it is also called when destructing the poller... If I remove that then it won't terminate at all in some tests. Or should a poller simply terminate when the last fd has been removed?
Since add_fd can also only be called from an event callback (correct?), there is no chance it does anything after the last has been removed.

bluca:
Sorry I'm not sure about the select implementation - but if you checkout v4.2.2 and build with it, all the tests appear to be working

sigiesec:
It's not about the select implementation, but about the use of any poller. I tried to write unit tests in my branch and am unsure what can be tested.

However I will try to analyse the differences with 4.2.2

sigiesec:
epoll_t::~epoll_t also writes to stopping, which is read by the worker thread. This is a data race as well.

Issue #2942, opened 2018-02-12
PUB crash when SUB exceeded SNDHWM


sublee:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description

When all of these conditions are satisfied, the assertion failure from `mtrie.cpp` occurs:

- A connection between a `PUB` socket and many `SUB` sockets.
- A `SUB` socket subscribe/unsubscribe many prefixes.
- Call `zmq_getsockopt()` with `ZMQ_EVENTS` for `SUB` sockets.

```
Assertion failed: erased == 1 (src/mtrie.cpp:297)
[1]    30266 abort (core dumped)  ./a.out
```

# Environment

* libzmq version (commit hash if unreleased): 4.2.0 and 4.2.3
* OS: Ubuntu 16.04 LTS

# Minimal test code / Steps to reproduce the issue

To reproduce this crash, we should prepare a `PUB` socket and many `SUB` sockets.

We will call this sequence (pseudo-code): `pub.connect(sub) or sub.connect(pub); pub.getsockopt(ZMQ_EVENTS); sub.subscribe(prefix); sub.getsockopt(ZMQ_EVENTS); sub.unsubscribe(prefix); sub.getsockopt(ZMQ_EVENTS)`.  There will be many prefixes to subscribe/unsubscribe.

Calling `getsockopt(ZMQ_EVENTS)` after `SUB`'s `SUBSCRIBE`/`UNSUBSCRIBE`, or `PUB`'s `zmq_connect()` will produce a crash due to the assertion failure in `mtrie_t::rm_helper`.

You can switch `PUB<->SUB` connection topology by the `pub_to_sub` variable.

```c
#include "zmq.h"
#include <stdio.h>

// Set 1 or 0 to switch the PUB<->SUB connection topology.
static int pub_to_sub = 1;

void gen_topic(int n, char* topic)
{
    // Simple hash function to generate a subscription prefix from a number.
    n = (n * 2654435761);
    sprintf(topic, "%08x", n);
}

void getsockopt_events_within_many_subscriptions(void* sub)
{
    char topic[8];
    char opt[256];
    size_t opt_len = 256;

    for (int j = 0; j < 10000; ++j)
    {
        gen_topic(j, topic);
        zmq_setsockopt(sub, ZMQ_SUBSCRIBE, &topic, 8);
        // CRASH: Get ZMQ_EVENTS from a SUB socket.
[...]

PR #2950, opened 2018-02-18
Problem: no unit tests for mtrie

sigiesec:
- [x] Inconsistent behavior/return values of rm (see TODOs in unittest_mtrie.cpp); @bluca please have a look

Probably the mentioned inconsistencies are somehow related to #2942


bluca:
May I ask to please rebase and add the unit tests first, and then the refactor? So that it can be checked that the tests pass before and after the refactoring for additional confidence

bluca:
I'll have a look as well tonight or tomorrow

sigiesec:
The refactoring is necessary to write the unit tests. Without the generalisation, the unit tests would need to instantiate pipe_t, which makes them much more complex.

bluca:
Ah I see - makes sense. I'll go through it tomorrow.

sigiesec:
@bluca Regarding the refactoring: I know/agree it is unfortunate to do this before adding unit tests. On the other hand, here and for several other classes some changes must be made to make the code unit-testable in the first place :(

All I changed is making the value type a template parameter, instead of using pipe_t directly. But of course this is hard to verify in the diff :(

sigiesec:
@bluca I added further tests, and also some TODO comments in the code. I am pretty sure now that the behaviour is inconsistent. Either it is completely illegal to call rm with a non-existing entry, but then it would be better to have assertions for all cases. However, I have some doubts that this is suitable for the uses within libzmq.

bluca:
Sorry for the delay, I've been swamped at work

sigiesec:
Never mind, I am AFK until Wednesday anyway.

src/generic_mtrie.hpp
```diff
@@ -0,0 +1,111 @@
...
+    typedef const unsigned char *prefix_t;
+
+    generic_mtrie_t ();
+    ~generic_mtrie_t ();
+
+    //  Add key to the trie. Returns true if it's a new subscription
+    //  rather than a duplicate (i.e. an entry with the same prefix
+    //  but a different pipe already exists).
+    //  TODO what if this is called with the same prefix AND pipe?
+    //  Is this legal? It is not checked anywhere.
```
bluca: To subscribe twice to the same topic on the same socket doesn't make much sense, but it is never specified that duplicated subscriptions are illegal so we can't raise errors, and I guess it kinda makes sense that these duplicates are just ignored.
As far as I can see internally it's a harmless no-op, given the pipes are stored in a set (which I suspect was intentional, to avoid these kind of dups)

bluca: In other words, from the point of view of the implementation, all that we care about when inserting a new sub is whether there were other sockets already subscribed or not (it's an XPUB feature) - and the code already does the right thing regardless of whether the other subscribers are the same or different sockets (pipes).

src/generic_mtrie_impl.hpp
```diff
@@ -0,0 +1,461 @@
...
+
+template <typename T>
+bool zmq::generic_mtrie_t<T>::rm_helper (prefix_t prefix_,
+                                         size_t size_,
+                                         value_t *pipe_)
+{
+    if (!size_) {
+        // TODO pipes can only be NULL here, if we are at the top level, i.e. 
+        // rm was already called with size_ == 0. This could be checked in rm, 
+        // and here we could just have an assertion or naught
```
bluca: If I understand correctly, this could also happen for duplicate unsubs - as above, not specified as illegal in the API docs and, I think, harmless as well

unittests/unittest_mtrie.cpp
```diff
@@ -0,0 +1,438 @@
...
+    mtrie.match (test_name, getlen (test_name), mtrie_count, &count);
+    TEST_ASSERT_EQUAL_INT (0, count);
+}
+
+void test_rm_nonexistent_0_size_empty ()
+{
+    int pipe;
+    zmq::generic_mtrie_t<int> mtrie;
+
+    // TODO why does this return true, but test_rm_nonexistent_empty returns false?
```
bluca: The single user of the return value of the mtrie is the verbose(r) XPUB option - unfortunately the documentation of that option is very confusing, and also there is no test for it at all.

I'm trying to understand exactly how the unsub verboser mode works and write tests for it, so that we can establish the boundaries (given by the existing public API behaviour) within which we can move and change things.

bluca: Ok I think I've covered the most common use cases - I'll push the test to this branch

src/generic_mtrie_impl.hpp
bluca: turns out duplicated unsubs for the same pipe can never happen, as XSUB filters them at the origin (it keeps a parallel trie)

src/generic_mtrie.hpp
bluca: Furthermore, XSUB does check for duplication but intentionally sends the duplicated subscription messages to the XPUB anyway: https://github.com/zeromq/libzmq/blob/master/src/xsub.cpp#L101

bluca:
Pushed the new test, updated docs, and a couple of fixups to be squashed later

src/generic_mtrie.hpp
sigiesec: I am not 100% sure how to map this to the mtrie_t. Are you saying that calling add should be idempotent, i.e. if you call it multiple times with the same arguments, it should behave as if called only once? I don't think this is the current behaviour, probably it would add multiple equivalent entries, but I need to check that, and add a test case.

src/generic_mtrie_impl.hpp
sigiesec: No, it could not happen even for duplicate unsubscriptions, since the tree is always reorganized after a removal (I have some doubts that this behaviour is optimal, but this is a separate issue).

src/generic_mtrie.hpp
bluca: Isn't that what the set data structure is for? If the same pipe is already in the set, adding it again won't do anything, right?

unittests/unittest_mtrie.cpp
sigiesec: Ok, that sounds great. I will look at the test cases, and see what can be done to mtrie based on them.

src/generic_mtrie.hpp
sigiesec: Yes, it is a set, forget about that. I added a test case, and it indeed behaves as you expected.

src/xpub.cpp
```diff
@@ -106,18 +107,23 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
...
-                //  passed to the user on next recv call unless verbose mode is enabled
-                //  which makes to pass always these messages.
-                if (options.type == ZMQ_XPUB
-                    && (unique || (*data == 1 && verbose_subs)
-                        || (*data == 0 && verbose_unsubs && verbose_subs))) {
+                bool notify;
+                if (*data == 0) {
+                    mtrie_t::rm_result rm_result =
+                      subscriptions.rm (data + 1, size - 1, pipe_);
+                    //  TODO reconsider what to do if rm_result == mtrie_t::not_found
```
bluca: This case will be hard to cover with the existing tests, given it can only happen if the sub socket thinks it has sent a subscription (since in the removal case xsub only sends a message if it can match it to its local table) but the xpub has not received it. Maybe fiddling around with HWM, but it could be flacky. Any ideas?

sigiesec:
I think I have now a consistent handling of the return value of mtrie_t::rm. 
One might have a look at zmq::xpub_t::xread_activated, but the tests are fine, including the newly added ones for xpub verbose mode.

src/xpub.cpp
sigiesec: This is probably also something that can better be tested with a kind of unit tests.

Can we defer this to after this PR?

bluca: Sure we can - just worried about external behaviour changes, that's all. Do you want to do a rebase and squash?

bluca: Can you push them as well? Can't see it

sigiesec: Oops... should be up now

bluca:
@sigiesec 32 bit builds are all failing:
https://build.opensuse.org/package/show/network:messaging:zeromq:git-stable/libzmq
Will try to have a look

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/2964

src/generic_mtrie.hpp
bluca: @sigiesec I was reading the RFCs and I remembered this discussion - turns out the spec says that subscribing twice to A means that you also need to unsubscribe twice:

> Subscriptions SHALL be additive and SHALL NOT be idempotent. That is, subscribing to "A" and "" is the same as subscribing to "" alone. Subscribing to "A" and "A" counts as two subscriptions, and would require two unsubscribe messages to undo.

So duplicated subs are really not illegal and there's an expected behaviour. The tests that were added here exercise that behaviour - but I'm worried that it's implemented in the client side, by the XSUB mtrie.

What did we say happens right now if a duplicated subscription is actually received by the PUB? I'm thinking about cross-implementation compatibility specifically

bluca: Basically I'm worried that on the client side we are fine as we just ignore them, but on the PUB side we are not doing the right thing given a set is used.

PR #2951, opened 2018-02-19
Problem: No support for addressing using TIPC Port Identity

eponsko:
Solution: 
ZeroMQ currently supports location independent addressing using TIPC Port Names with tipc://{type,instance}. This commits adds support for connecting and binding using TIPC Port Identity addresses. To connect using Port Identities the expected format is tipc://<Z.C.N:Ref>, e.g. "tipc://<1.2.3:123123>". To bind using TIPC Port Identities the expected format is "tipc://<*>".

### DIFF
--- a/src/tipc_address.cpp
+++ b/src/tipc_address.cpp
@@ -41,6 +41,7 @@
 zmq::tipc_address_t::tipc_address_t ()
 {
     memset (&address, 0, sizeof address);
+    _random = false;
 }
 
 zmq::tipc_address_t::tipc_address_t (const sockaddr *sa, socklen_t sa_len)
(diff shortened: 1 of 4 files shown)


bluca:
Could you please add some tests for this? There's a few tests/*tipc.cpp test files

src/tipc_address.cpp
```diff
@@ -93,6 +114,25 @@ int zmq::tipc_address_t::resolve (const char *name)
     address.addr.nameseq.upper = upper;
     address.scope = TIPC_ZONE_SCOPE;
     return 0;
+portid:
```
bluca: isn't this much simpler in the if (res == 0) block above? There's no reuse of this block

src/tipc_address.cpp
```diff
@@ -102,11 +142,28 @@ int zmq::tipc_address_t::to_string (std::string &addr_)
...
+        s << ", " << address.addr.nameseq.upper << "}";
+        addr_ = s.str ();
+    } else if (address.addrtype == TIPC_ADDR_ID) {
+        s << "tipc://"
+          << "<" << tipc_zone (address.addr.id.node);
+        s << "." << tipc_cluster (address.addr.id.node);
+        s << "." << tipc_node (address.addr.id.node);
+        s << ":" << address.addr.id.ref << ">";
+        addr_ = s.str ();
+    } else if (is_random ()) {
```
bluca: isn't the block below doing exactly the same operations as the one above?

PR #2956, opened 2018-02-20
Problem: Tests for different TIPC address types missing

eponsko:
Solution: 
Add tests for binding/connecting with different TIPC address types, and error checks misused addresses.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -773,7 +773,8 @@ test_apps += \
 	tests/test_router_mandatory_tipc \
 	tests/test_shutdown_stress_tipc \
 	tests/test_sub_forward_tipc \
-	tests/test_term_endpoint_tipc
+	tests/test_term_endpoint_tipc \
+	tests/test_address_tipc
 
 tests_test_connect_delay_tipc_SOURCES = tests/test_connect_delay_tipc.cpp
 tests_test_connect_delay_tipc_LDADD = src/libzmq.la
@@ -798,6 +799,10 @@ tests_test_sub_forward_tipc_LDADD = src/libzmq.la
 
 tests_test_term_endpoint_tipc_SOURCES = tests/test_term_endpoint_tipc.cpp
 tests_test_term_endpoint_tipc_LDADD = src/libzmq.la
+
+tests_test_address_tipc_SOURCES = tests/test_address_tipc.cpp
+tests_test_address_tipc_LDADD = src/libzmq.la ${UNITY_LIBS}
+tests_test_address_tipc_CPPFLAGS = ${UNITY_CPPFLAGS}
 
 endif
 
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -936,6 +936,13 @@ int zmq::socket_base_t::connect (const char *addr_)
             LIBZMQ_DELETE (paddr);
             return -1;
         }
+        sockaddr_tipc *saddr = (sockaddr_tipc *) paddr->resolved.tipc_addr->addr();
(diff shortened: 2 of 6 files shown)


eponsko:
Ooops forgot about automake, will have a look at test_hwm_pubsub

sigiesec:
zmq_init is deprecated, you should use zmq_ctx_new instead. I would suggest moving this into setUp and zmq_ctx_term into tearDown,so that every test case has a fresh context (without sockets with pending close e.g.).

Not sure if this resolves your problem on the go. I can have a closer look at the code when back at my desk on Wednesday.

sigiesec:
@eponsko Hm I tried it with setUp/tearDown in branch ZMQers/eponsko-pr2956 and it seems to work: https://travis-ci.org/ZMQers/libzmq/builds/347229430 

Maybe you want to merge the change I added, and then we can merge this PR?

eponsko:
Hi @sigiesec , sorry for lack of action here, been stuck in bed fighting some kind of virus.. Sounds like a good idea, I'll merge later today (hopefully) and the PR should be ready to go!

bluca:
@eponsko is there anything specific that one needs to be able to use TIPC on Linux? I'd like to enable the tests in Travis

sigiesec:
@bluca @eponsko Oh, they aren't enabled on travis? Then I did not test it ;) Forget what I wrote earlier.

eponsko:
@bluca To use TIPC you need to insert the tipc.ko kernel module, unless TIPC support has been compiled into the kernel (unlikely). So I guess automatic testing with Travis will be difficult..

bluca:
What's left to do here?

bluca:
@eponsko I can work on that at some point - there is sudo access on some configs so it might be possible. Anything else before we can merge this?

eponsko:
@bluca Sure that would be nice, it should be enough to just modprobe the tipc module. Normally one would set a node address and enable interfaces as well, but that should not be necessary for these tests. 
I'll update my branch with Simon's patch and add the test to the automake configuration, and then we should be good to go. Sorry again for the delay here, haven't been able to work :/

Makefile.am
```diff
@@ -792,6 +793,9 @@ tests_test_sub_forward_tipc_LDADD = src/libzmq.la
 tests_test_term_endpoint_tipc_SOURCES = tests/test_term_endpoint_tipc.cpp
 tests_test_term_endpoint_tipc_LDADD = src/libzmq.la
 
+tests_test_address_tipc_SOURCES = tests/test_term_endpoint_tipc.cpp
+tests_test_address_tipc_LDADD = src/libzmq.la
```
bluca: needs: ```${UNITY_LIBS}```
and below:
 ```tests_test_address_tipc_CPPFLAGS = ${UNITY_CPPFLAGS}```

eponsko:
@sigiesec I'm having the same issue as before with zmq_ctx_term() and Unity. Even with zmq_ctx_term()  placed in tearDown() the zmq_ctx_term() gets stuck in an infinite poll in signaler::wait(). I'll redo the test without Unity just to verify that it's actually Unity and not something else (I've had issues with the syscall poll() -> tipc_poll() in the kernel module before on certain kernel versions).

sigiesec:
@eponsko Ah great the problem resolved :) I couldn't really have imagined that it could be related to Unity, as it operates on a completely different level.

Could you apply the formatting fixes from https://travis-ci.org/zeromq/libzmq/jobs/349958124#L646 ? Then the PR can be merged :)

PR #2964, opened 2018-03-02
Problem: unittest_mtrie fails on 32bit *nix

bluca:
Solution: correctly dereference pointer in test.

@sigiesec - just a small error that was not visible on 64bit (I guess the pointer there evaluates as greater than 0, but on 32bit it's negative)


bluca:
Merging since it's a trivial fix and builds are broken

PR #2966, opened 2018-03-03
Pull request to merge porting to WindRiver VxWorks 6.x

mesegura:
Merging port to WindRiver VxWorks 6.x


bluca:
Sorry but this is breaking all other existing platforms as you can see in the CI, so it cannot be merged right now. Please keep changes to a bare minimum: no refactoring, no type changes, no extra whitespace changes, no logic changes - that way Linux Osx and Windows will not be affected.
Also check the formatting - there's a make clang-format-check target they will tell you what to fix.

bluca:
The platform.hpp is generated at runtime by autotools or cmake so it shouldn't need to be checked in -which toolchain is used on vxworks?

src/atomic_ptr.hpp
```diff
@@ -280,11 +286,9 @@ struct atomic_value_t
 #endif
 
 #if defined ZMQ_ATOMIC_PTR_MUTEX
-    mutex_t sync;
+    mutable mutex_t sync;
```
sigiesec: Why has this become necessary?

src/atomic_ptr.hpp
```diff
@@ -244,6 +244,12 @@ struct atomic_value_t
 
     atomic_value_t (const atomic_value_t &src) : value (src.load ()) {}
 
+    atomic_value_t &operator= (const atomic_value_t &src)
```
sigiesec: I think atomic_value_t was non-copyable by intention. Is this really necessary? If it is, probably the code location causing this should be fixed.

src/options.cpp
```diff
@@ -104,6 +104,62 @@ zmq::options_t::options_t () :
 #endif
 }
 
+zmq::options_t::options_t (const options_t &from) :
```
sigiesec: Why is this necessary? I think this should probably be avoided.

mesegura:
The toolchains used in VxWorks are the proprietary WindRiver Diab C/C++ compiler and a proprietary version of the GNU C/C++ compiler, so I don't think it CMake supports it without some customization. I'll address the problems and try again some other time.

src/atomic_ptr.hpp
mesegura: There's a const function somewhere in atomic_ptr.cpp that uses the lock/unlock functions of `sync`, and the implementation of mutex_t I made for VxWorks relies on VxWorks semaphores instead of POSIX mutexes; apparently the lock/unlock function changes the mutex_t object with VxWorks semaphores. I can make it a declaration only for VxWorks.

src/atomic_ptr.hpp
mesegura: I think the WindRiver Diab compiler needed me to declare and define the `operator=` because something needed it. I can try to see why it needed it.

src/options.cpp
mesegura: The ctx.hpp file declares the endpoint_t struct with an instance of type options_t, and I guess the WindRiver Diab compiler isn't smart enough to auto-define the copy constructor for the options_t class (it needed the copy constructor for some reason during compilation). I can make it a VxWorks only declaration and definition.
```
struct endpoint_t
{
    socket_base_t *socket;
    options_t options;
};
```

src/ipc_listener.cpp
```diff
@@ -193,7 +194,7 @@ void zmq::ipc_listener_t::in_event ()
 int zmq::ipc_listener_t::get_address (std::string &addr_)
 {
     struct sockaddr_storage ss;
-#ifdef ZMQ_HAVE_HPUX
+#if defined ZMQ_HAVE_HPUX
```
bluca: Was this supposed to check if _VXWORKS is defined as well? If not please remove the change

src/socks_connecter.cpp
```diff
@@ -342,16 +345,23 @@ int zmq::socks_connecter_t::connect_to_proxy ()
 
     // Set a source address for conversations
     if (tcp_addr->has_src_addr ()) {
-        rc = ::bind (s, tcp_addr->src_addr (), tcp_addr->src_addrlen ());
+#if defined ZMQ_HAVE_VXWORKS
+        rc = ::bind (s, (sockaddr *)tcp_addr->src_addr (), tcp_addr->src_addrlen ());
+#else
+		rc = ::bind(s, tcp_addr->src_addr(), tcp_addr->src_addrlen());
```
bluca: whitespace-only change

src/socks_connecter.cpp
```diff
@@ -342,16 +345,23 @@ int zmq::socks_connecter_t::connect_to_proxy ()
...
         }
     }
 
     //  Connect to the remote peer.
-    rc = ::connect (s, tcp_addr->addr (), tcp_addr->addrlen ());
-
+#if defined ZMQ_HAVE_VXWORKS
+    rc = ::connect (s, (sockaddr *)tcp_addr->addr (), tcp_addr->addrlen ());
+#else
+	rc = ::connect(s, tcp_addr->addr(), tcp_addr->addrlen());
```
bluca: w/s only change

src/socks_connecter.cpp
```diff
@@ -377,13 +387,13 @@ zmq::fd_t zmq::socks_connecter_t::check_proxy_connection ()
...
     int err = 0;
-#ifdef ZMQ_HAVE_HPUX
+#if defined ZMQ_HAVE_HPUX || defined ZMQ_HAVE_VXWORKS
     int len = sizeof err;
 #else
     socklen_t len = sizeof err;
 #endif
 
-    int rc = getsockopt (s, SOL_SOCKET, SO_ERROR, (char *) &err, &len);
+    int rc = getsockopt (s, SOL_SOCKET, SO_ERROR, (char*) &err, &len);
```
bluca: w/s only change

src/address.cpp
```diff
@@ -44,17 +44,18 @@
 #include <string>
 #include <sstream>
 
-zmq::address_t::address_t (const std::string &protocol_,
```
bluca: why was this change necessary?

bluca:
> The toolchains used in VxWorks are the proprietary WindRiver Diab C/C++ compiler and a proprietary version of the GNU C/C++ compiler, so I don't think it CMake supports them without some customization. I'll address the problems and try again.

Ok that's fine then, thanks for the explanation.

There are quite a few whitespace-only changes - given this diff is already quite large, please remove them to keep it smaller.

src/atomic_ptr.hpp
sigiesec: I think it's fine to have the mutex mutable, this is a recurring pattern. I just could not relate it to a change that causes it to be necessary.

src/address.cpp
mesegura: I undid the change and it still compiled, so it wasn't necessary.

mesegura:
I'm still in the process of checking/removing all the whitespace changes, while trying to un-break cross compilation.

bluca:
Ok, thanks. Let us know if you need any help.

mesegura:
Looks like the Clang-format-check failed. I have to address this too.

bluca:
Don't worry about that too much for now - see this PR: https://github.com/zeromq/libzmq/pull/2970
Once that is merged, then rebase and run make clang-format-diff to automatically format your code. But you can do that at the end.

src/generic_mtrie.hpp
```diff
@@ -0,0 +1,116 @@
+/*
```
sigiesec: Why are generic_mtrie.hpp and generic_mtrie_impl.hpp shown as new files in this commit? They already exist in the master. Interestingly, no conflict with the master is shown.

jimklimov:
Probably PRed branch is not based on current tip of target branch, but on
an older commit, so both evolutions grew the same code (maybe having clones
of same commit comtents with different hashes).

src/generic_mtrie.hpp
bluca: merge commit squashed I'd guess

bluca:
please rebase on the latest master, then run make sure you have at least clang-format 5.0 (with autotools: ./configure CLANG_FORMAT=clang-format-5.0 ) and then run ```make clang-format-diff``` so that your changes will be properly formatted

src/atomic_ptr.hpp
```diff
@@ -280,11 +280,12 @@ struct atomic_value_t
...
 
 #if defined ZMQ_ATOMIC_PTR_MUTEX
+#if defined ZMQ_HAVE_VXWORKS
+    mutable mutex_t sync;
+#else
     mutex_t sync;
 #endif
-
-  private:
-    atomic_value_t &operator= (const atomic_value_t &src);
```
bluca: Why is this being removed?

bluca:
Just one more question then we are ready I think

src/atomic_ptr.hpp
mesegura: That was an accident. I'll undo it.

bluca:
@mesegura thank you for your work! For the next improvements, maybe try to reduce the number of ifdefs, especially for casts around network APIs, and also another good improvement would be to come out with a native mutex-less way to do atomic operations. The mutex one works and is safe but will be very, very bad for performances.

chavadol:
@mesegura I am thinking of using zeroMQ with VxWorks 7. 
- Could you tell me, if your porting to VxWorks 6.X will also be working on VxWorks 7?
- Or do I need to specifically port the library for VxWorks 7?

bluca:
@mesegura could you please send another PR with a relicensing grant? See README here for details: https://github.com/zeromq/libzmq/tree/master/RELICENSE and this example: https://github.com/zeromq/libzmq/blob/master/RELICENSE/WenbinHou.md

mesegura:
I just noticed your message @bluca. I'm not sure why I wasn't prompted there was messages until @murat-sever mentioned me. I do that this weekend.

bluca:
thanks

mesegura:
@chavadol I never tried compiling for VxWorks 7, but I imagine it would only require minimal changes if any. @murat-sever it is meant to be compiled on VxWorks 6.x as a static user library, not a kernel library

mesegura:
@murat-sever @chavadol  There seems to have been some updates to the code since my last pull request that have broken my port to VxWorks 6.x. When I ported it back in March, ZMQ_IOTHREAD_POLLER_USE_* and ZMQ_POLL_BASED_ON_* did not exist in the code. I'll have to do another pull request to update the port.

mesegura:
@murat-sever @chavadol Until I fix it with a pull request, set the following in platform.hpp:
```
#define ZMQ_IOTHREAD_POLLER_USE_SELECT 1
#define ZMQ_POLL_BASED_ON_SELECT 1
```
And in zmq.cpp:
```
int zmq_has (const char *capability_)
{
#if !defined(ZMQ_HAVE_WINDOWS) && !defined(ZMQ_HAVE_OPENVMS) && !defined(ZMQ_HAVE_VXWORKS)
    if (strcmp (capability_, zmq::protocol_name::ipc) == 0)
        return true;
#endif
...
```

murat-sever:
Thanks, @mesegura. With your changes I am able build it as a user static library. But when I try to use it in an RTP app (simple hello world client c file) it doesn't build. It complains udp_address.cpp:107: undefined reference to if_nametoindex' 

Regards

PR #2969, opened 2018-03-05
Problem: enormous memory increase due to zero copy decoding

skaes:
The zero copy decoding strategy implemented for 4.2.0 can lead to a large
increase of main memory usage in some cases (I have seen one program go up to
40G from 10G after upgrading from 4.1.4). This commit adds a new option to
contexts, called ZMQ_ZERO_COPY_RECV, which allows one to switch to the old
decoding strategy.

### DIFF
--- a/doc/zmq_ctx_set.txt
+++ b/doc/zmq_ctx_set.txt
@@ -127,6 +127,18 @@ Default value:: INT_MAX
 Maximum value:: INT_MAX
 
 
+ZMQ_ZERO_COPY_RCV: Specify message decoding strategy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_ZERO_COPY_RCV' argument specifies whether the message decoder should
+use a zero copy strategy when receiving messages. The zero copy strategy can
+lead to increased memory usage in some cases. This option allows you to use the
+older copying strategy. You can query the value of this option with
+linkzmq:zmq_ctx_get[3] using the 'ZMQ_ZERO_COPY_RECV' option.
+
+[horizontal]
+Default value:: 1
+
+
 ZMQ_MAX_SOCKETS: Set maximum number of sockets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The 'ZMQ_MAX_SOCKETS' argument sets the maximum number of sockets allowed
--- a/tests/test_ctx_options.cpp
+++ b/tests/test_ctx_options.cpp
@@ -146,6 +146,56 @@ void test_ctx_thread_opts (void *ctx)
 #endif
 }
 
+void test_ctx_zero_copy (void *ctx)
+{
+#ifdef ZMQ_ZERO_COPY_RECV
+    int zero_copy;
+    // Default value is 1.
+    zero_copy = zmq_ctx_get (ctx, ZMQ_ZERO_COPY_RECV);
+    assert (zero_copy == 1);
+
+    // Test we can set it to 0.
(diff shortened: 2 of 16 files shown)


tests/test_ctx_options.cpp
```diff
@@ -146,6 +146,52 @@ void test_ctx_thread_opts (void *ctx)
 #endif
 }
 
+void test_ctx_zero_copy (void *ctx)
+{
+    int zero_copy;
```
bluca: @skaes thanks for adding a test, but the body needs a:
```
#ifdef ZMQ_ZERO_COPY_RECV
..
#else
(void)ctx;
#endif
```

as the API is DRAFT, see build failures:
https://travis-ci.org/zeromq/libzmq/jobs/349263452

skaes:
@bluca I have no idea why so many of the travis builds fail. Can you help?

bluca:
The new test always times out - on which platform are you running it?

skaes:
I'm running it on OS X. And those builds are all green.

bluca:
Yeah it's weird that it gets stuck only on Linux, I can't see platform-specific changes

bluca:
I'll fetch and check where it's stuck with gdb, hang on

bluca:
@skaes ok, the problem was apparently caused by using * instead of 127.0.0.1 to bind - I guess on some Travis configuration/VM/container that will pick a non-working interface. Forcing to the loopback address fixes the problem.

I've prepared a branch that you can fetch, merge --ff--only and then squash some commits:
https://github.com/bluca/libzmq/commits/optional-zero-copy-decoding

There's also a fix for the NORM engine, and a formatting fix (make clang-format-check) to squash, and an additional commit to increase the size of the endpoint string in the test since IPv6 is enabled so it will be longer (v4-in-v6 format).

doc/zmq_ctx_set.txt
```diff
@@ -127,6 +127,18 @@ Default value:: INT_MAX
 Maximum value:: INT_MAX
 
 
+ZMQ_ZERO_COPY_RCV: Specify message decoding strategy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_ZERO_COPY_RCV' argument specifies whether the message decoder should
+use a zero copy strategy when receiving messages. The zero copy strategy can
+lead to increased memory usage in some cases. This option allows you to use the
+older copying strategy. You can query the value of this option with
+linkzmq:zmq_ctx_get[3] using the 'ZMQ_ZERO_COPY_RECV' option.
```
sigiesec: There is an inconsistency in spelling: RCV vs. RECV. I am not completely sure how this is abbreviated at other places, but I think RECV would be more consistent.

bluca:
Merged, thanks for your work.

doc/zmq_ctx_set.txt
bluca: I'll fix it when doing the NEWS for the next release

skaes:
@bluca thanks for the support in getting this into master. I really appreciate how well you handled this.

bluca:
@skaes could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@skaes  another ping. Please consider helping us with the relicense. Thank you!

bluca:
> @skaes could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

@skaes another ping. Please consider helping us, only takes a minute. Thanks!

PR #2970, opened 2018-03-05
Problem: CI check for format fails

bluca:
Solution: apply fixes and add info to PR template


sigiesec:
@bluca Thanks for fixing this. I still don't understand why the formatting was different. Maybe clang-format behaves differently on Windows with Windows-style line endings. Hope this doesn't break again.

PR #2971, opened 2018-03-06
Problem: static initialization order fiasco

### LINKED ISSUE #2915: Global static mutex random_sync causes order of destruction bug
# Issue description

When using zmqpp with libzmq 4.2.3 there is an order of destruction bug in static builds if there are contexts defined globally/static. This is due to the global static mutex random_sync in random.cpp being destructed before the context is destructed.

I think the better approach would be to have client application control the initialization of sodium_init() via the zmq::random_open()/close() function. Client code would need to call this at start/end of main(). I think adding a couple functions like zmq_init() and zmq_destroy() to wrap these kind of init calls would do the trick.

# Environment

* libzmq version (commit hash if unreleased): 4.2.3
* OS: Ubuntu 16.04

```
Program received signal SIGABRT, Aborted.
0x00007ffff5534428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) list
49	in ../sysdeps/unix/sysv/linux/raise.c
(gdb) bt
#0  0x00007ffff5534428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff553602a in __GI_abort () at abort.c:89
#2  0x00000000012f39ee in zmq::zmq_abort (errmsg_=errmsg_@entry=0x7ffff568c10a "Invalid argument") at src/err.cpp:87
#3  0x00000000012fde65 in zmq::mutex_t::lock (this=0x4431200 <random_sync>) at src/mutex.hpp:122
#4  zmq::scoped_lock_t::scoped_lock_t (mutex_=..., this=<synthetic pointer>) at src/mutex.hpp:168
#5  zmq::random_close () at src/random.cpp:110
#6  0x00000000012ee546 in zmq::ctx_t::~ctx_t (this=0x61600000f980, __in_chrg=<optimized out>) at src/ctx.cpp:124
#7  0x00000000012eeb65 in zmq::ctx_t::terminate (this=0x61600000f980) at src/ctx.cpp:201
#8  0x00000000012dc711 in zmqpp::context::terminate() ()
#9  0x0000000000485eef in zmqpp::context::~context (this=0x4431188 <zmqpp::actor::actor_pipe_ctx_>, __in_chrg=<optimized out>) at ../../../../trunk/thirdparty/zmqpp4.2.0/include/zmqpp/context.hpp:100
#10 0x00007ffff5538ff8 in __run_exit_handlers (status=0, listp=0x7ffff58c35f8 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true) at exit.c:82
#11 0x00007ffff5539045 in __GI_exit (status=<optimized out>) at exit.c:104
#12 0x00007ffff551f837 in __libc_start_main (main=0x4792c0 <main(int, char**)>, argc=2, argv=0x7fffffffdb58, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdb48) at ../csu/libc-start.c:325
#13 0x00000000004463a9 in _start ()
```

# What's the expected result?

Correct order of destruction and no abort being thrown.

sigiesec:
Solution: use local statics when available in a thread-safe implementation

Fixes #2915 (at least on platforms with gcc >=4.3.0 and MSVC>=2015)

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -79,41 +79,76 @@ uint32_t zmq::generate_random ()
 //  The context class cannot be used with static variables as the curve
 //  utility APIs like zmq_curve_keypair also call into the crypto
 //  library.
-//  The safest solution for all use cases therefore is to have a global,
+//  The safest solution for all use cases therefore is to have a
 //  static lock to serialize calls into an initialiser and a finaliser,
 //  using refcounts to make sure that a thread does not close the library
-//  while another is still using it.
+//  while another is still using it. To avoid the static initialization
+//  order fiasco, this is done using function-local statics, if the
+//  compiler implementation supports thread-safe initialization of those.
+//  Otherwise, we fall back to global statics.
+
+//  TODO if there is some other user of libsodium besides libzmq, this must
+//    be synchronized by the application. This should probably also be
+//    configurable via config.h
+
+//  TODO this should probably be done via config.h
+#if __cplusplus >= 201103L                                                     \
+  || (defined(__cpp_threadsafe_static_init)                                    \
+      && __cpp_threadsafe_static_init >= 200806)                               \
+  || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#define ZMQ_HAVE_THREADSAFE_STATIC_LOCAL_INIT 1
+//  TODO this might probably also be set if a sufficiently recent gcc is used
+//  without -fno-threadsafe-statics, but this cannot be determined at
+//  compile-time, so it must be set via config.h
+#else
+#define ZMQ_HAVE_THREADSAFE_STATIC_LOCAL_INIT 0
+#endif
+
+#if !ZMQ_HAVE_THREADSAFE_STATIC_LOCAL_INIT                                     \
+  && (defined(ZMQ_USE_LIBSODIUM)                                               \
+      || (defined(ZMQ_USE_TWEETNACL) && !defined(ZMQ_HAVE_WINDOWS)             \
+          && !defined(ZMQ_HAVE_GETRANDOM)))
 static unsigned int random_refcount = 0;
 static zmq::mutex_t random_sync;
+#endif
 
-void zmq::random_open (void)
+static void manage_random (bool init)
 {
 #if defined(ZMQ_USE_LIBSODIUM)                                                 \
   || (defined(ZMQ_USE_TWEETNACL) && !defined(ZMQ_HAVE_WINDOWS)                 \
       && !defined(ZMQ_HAVE_GETRANDOM))
-    scoped_lock_t locker (random_sync);
 
-    if (random_refcount == 0) {
-        int rc = sodium_init ();
-        zmq_assert (rc != -1);
-    }
-
-    ++random_refcount;
-#else
-    LIBZMQ_UNUSED (random_refcount);
+#if ZMQ_HAVE_THREADSAFE_STATIC_LOCAL_INIT
+    static int random_refcount = 0;
+    static zmq::mutex_t random_sync;
 #endif
+
+    if (init) {
+        zmq::scoped_lock_t locker (random_sync);
+
+        if (random_refcount == 0) {
+            int rc = sodium_init ();
+            zmq_assert (rc != -1);
+        }
+
+        ++random_refcount;
+    } else {
+        zmq::scoped_lock_t locker (random_sync);
+        --random_refcount;
+
+        if (random_refcount == 0) {
+            randombytes_close ();
+        }
+    }
+#endif
+}
+
+void zmq::random_open (void)
+{
+    manage_random (true);
 }
 
 void zmq::random_close (void)
 {
-#if defined(ZMQ_USE_LIBSODIUM)                                                 \
-  || (defined(ZMQ_USE_TWEETNACL) && !defined(ZMQ_HAVE_WINDOWS)                 \
-      && !defined(ZMQ_HAVE_GETRANDOM))
-    scoped_lock_t locker (random_sync);
-    --random_refcount;
-
-    if (random_refcount == 0) {
-        randombytes_close ();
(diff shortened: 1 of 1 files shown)


bluca:
@sigiesec looks like there's a regression in CZMQ with these changes, here's a malamute selftest backtrace:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff732742a in __GI_abort () at abort.c:89
#2  0x00007ffff76bbaee in zmq::zmq_abort (errmsg_=errmsg_@entry=0x7ffff7454d8c "Invalid argument")
    at src/err.cpp:87
#3  0x00007ffff76dcd99 in zmq::mutex_t::lock (this=0x7ffff7929860 <manage_random(bool)::random_sync>)
    at src/mutex.hpp:103
#4  zmq::scoped_lock_t::scoped_lock_t (mutex_=..., this=<synthetic pointer>) at src/mutex.hpp:141
#5  manage_random (init=init@entry=false) at src/random.cpp:136
#6  0x00007ffff76dcfc7 in zmq::random_close () at src/random.cpp:153
#7  0x00007ffff76ae0a6 in zmq::ctx_t::~ctx_t (this=0x555555769890, __in_chrg=<optimized out>)
    at src/ctx.cpp:124
#8  0x00007ffff76ae765 in zmq::ctx_t::terminate (this=this@entry=0x555555769890) at src/ctx.cpp:206
#9  0x00007ffff7703b1f in zmq_ctx_term (ctx_=0x555555769890) at src/zmq.cpp:149
#10 0x00007ffff797200c in zsys_shutdown () at src/zsys.c:301
#11 0x00007ffff7328940 in __run_exit_handlers (status=0, listp=0x7ffff768c5d8 <__exit_funcs>, 
    run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:83
#12 0x00007ffff732899a in __GI_exit (status=<optimized out>) at exit.c:105
#13 0x00007ffff73132e8 in __libc_start_main (main=0x555555554d08 <main>, argc=1, argv=0x7fffffffdef8, 
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdee8)
    at ../csu/libc-start.c:325
#14 0x0000555555554a2a in _start ()
```

Any idea what's going on? pthread_mutex_lock is returning EINVAL which the manpage says it means the mutex was not initialised properly

bluca:
I don't think GCC already supports this feature:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81852

Same thread also says that it landed in clang only in 6.0. If I remove this from the ifdef: ```__cplusplus >= 201103L``` then it's no longer compiled in.

bluca:
Yeah the __cpp_threadsafe_static_init is only defined in GCC 8 - 7 and 6 do not define it.

bluca:
The immediate problem is fixed by https://github.com/zeromq/libzmq/pull/2974 but that will reintroduce the initialisation problem basically everywhere, since no distro uses gcc 8 or clang 6 as default today

sigiesec:
Hm... I don't think this is related to the gcc version. Maybe there is some general logic problem. How can I reproduce this? I will try to derive a libzmq test case from that.

bluca:
With GCC 8 it works. I think the key is to have another thread do the context init - build and install CZMQ and then run the malamute selftests and you'll see the failures

sigiesec:
I tested this with the malamute self test.

This has nothing to do with the thread-safe initialization of the static locals. In the test you mentioned, the initialization is done from the main thread before any other threads are started. And I am very sure that gcc implements thread-safe initialization of static locals for a long time.

However, there is a bug in czmq. 
Within zsys_init, atexit(zsys_shutdown) is called before zmq_init: https://github.com/zeromq/czmq/blob/086f1267b06cb4ec96e0d9a39210890ea0a25041/src/zsys.c#L208 This causes an erroneous destruction order. See http://en.cppreference.com/w/cpp/utility/program/atexit

I don't know why the behaviour changes with gcc 8, but I feel that the behaviour of gcc <8 is correct.

Of course, this is non-obvious for a caller of zmq_init. As said before, a clean solution will require explicit zmq library init and shutdown functions. Beside that, I don't think it is a good idea that czmq calls atexit. An application may do so, but not a library.

bluca:
I'll try with switching them around - but then why when disabling the thread local init the regression disappears?

bluca:
Note also that CZMQ is not C++ (fortunately...), but perhaps the same caveat applies

sigiesec:
If you disable it, it uses the implementation as it was before, with global statics. This of course affects the order of deinitialization, which is the reverse order of initialization, and the global statics are initialized when the library is loaded, while the local statics are initialized when manage_random is first called.

bluca:
Yep seems to work! Thanks for pointing out the right problem, will check the CI and then send the PR to czmq

bluca:
https://github.com/zeromq/czmq/pull/1876

sigiesec:
Ok great :) can we revert #2974 again then?

bluca:
Yes (after that one is merged) as long as you're 110% sure that the __cplusplus version is enough in all compilers

sigiesec:
Of course I don't know all compilers, but as specified, __cplusplus >= 201103L implies that thread-safe initialization of statics is implemented. In practice, maybe the condition with __cpp_threadsafe_static_init is redundant. In theory, there could be a compiler that supports the feature macro __cpp_threadsafe_static_init but not fully supports C++11. Visual C++ must be mentioned explicitly, as it still does not claim full C++11 support but does not support the feature macros either.

PR #2972, opened 2018-03-06
Problem: code duplication in getsockopt/setsockopt

sigiesec:
Solution: extracted common code into do_getsockopt/do_setsockopt functions

### DIFF
--- a/src/sub.cpp
+++ b/src/sub.cpp
@@ -59,10 +59,7 @@ int zmq::sub_t::xsetsockopt (int option_,
     int rc = msg.init_size (optvallen_ + 1);
     errno_assert (rc == 0);
     unsigned char *data = (unsigned char *) msg.data ();
-    if (option_ == ZMQ_SUBSCRIBE)
-        *data = 1;
-    else if (option_ == ZMQ_UNSUBSCRIBE)
-        *data = 0;
+    *data = (option_ == ZMQ_SUBSCRIBE);
     //  We explicitly allow a NULL subscription with size zero
     if (optvallen_) {
         assert (optval_);
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -44,6 +44,146 @@
 #define BINDDEVSIZ 16
 #endif
 
+static int sockopt_invalid ()
+{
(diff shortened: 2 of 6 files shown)


src/sub.cpp
```diff
@@ -59,10 +59,7 @@ int zmq::sub_t::xsetsockopt (int option_,
     int rc = msg.init_size (optvallen_ + 1);
     errno_assert (rc == 0);
     unsigned char *data = (unsigned char *) msg.data ();
-    if (option_ == ZMQ_SUBSCRIBE)
-        *data = 1;
-    else if (option_ == ZMQ_UNSUBSCRIBE)
-        *data = 0;
+    *data = (option_ == ZMQ_SUBSCRIBE);
```
bluca: Can't remember what the spec says right now - are you sure that on every platform, every architecture, every compiler and so on "true" is going to evaluate to the exactly 0x1? Otherwise things are going to break

sigiesec: Yes: section 4.5 paragraph 4 of the C++ standard says: "An rvalue of type bool can be converted to an rvalue of type int, with false becoming zero and true becoming one."

bluca: Grand, thanks for checking.

bluca:
@sigiesec this breaks the build with GCC 4.7 (Debian 7)

```
[  142s] src/options.cpp: In function 'int zmq::do_setsockopt_int_as_bool_strict(const void*, size_t, bool*)':
[  142s] src/options.cpp:121:5: error: 'value' may be used uninitialized in this function [-Werror=maybe-uninitialized]
[  142s] src/options.cpp: In function 'int zmq::do_setsockopt_int_as_bool_relaxed(const void*, size_t, bool*)':
[  142s] src/options.cpp:135:31: error: 'value' may be used uninitialized in this function [-Werror=maybe-uninitialized]
```

https://build.opensuse.org/package/live_build_log/network:messaging:zeromq:git-stable/libzmq/Debian_7.0/x86_64

bluca:
It's a trivial fix: https://github.com/zeromq/libzmq/pull/2973

PR #2973, opened 2018-03-06
Problem: build broken with gcc-4.7

bluca:
Solution: initialise variable in options.cpp to dummy value to fix
compiler complaint:

 src/options.cpp: In function
  'int zmq::do_setsockopt_int_as_bool_strict(const void*, size_t, bool*)':
 src/options.cpp:121:5: error: 'value' may be used uninitialized in
  this function [-Werror=maybe-uninitialized]
 src/options.cpp: In function
  'int zmq::do_setsockopt_int_as_bool_relaxed(const void*, size_t, bool*)':
 src/options.cpp:135:31: error: 'value' may be used uninitialized in
  this function [-Werror=maybe-uninitialized]

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -115,7 +115,7 @@ int zmq::do_setsockopt_int_as_bool_strict (const void *const optval_,
     // TODO handling of values other than 0 or 1 is not consistent,
     // here it is disallowed, but for other options such as
     // ZMQ_ROUTER_RAW any positive value is accepted
-    int value;
+    int value = -1;
     if (do_setsockopt (optval_, optvallen_, &value) == -1)
         return -1;
     if (value == 0 || value == 1) {
@@ -129,7 +129,7 @@ int zmq::do_setsockopt_int_as_bool_relaxed (const void *const optval_,
                                             const size_t optvallen_,
                                             bool *const out_value_)
 {
-    int value;
+    int value = -1;
     if (do_setsockopt (optval_, optvallen_, &value) == -1)
         return -1;
     *out_value_ = (value != 0);


src/options.cpp
```diff
@@ -115,7 +115,7 @@ int zmq::do_setsockopt_int_as_bool_strict (const void *const optval_,
     // TODO handling of values other than 0 or 1 is not consistent,
     // here it is disallowed, but for other options such as
     // ZMQ_ROUTER_RAW any positive value is accepted
-    int value;
+    int value = -1;
```
sigiesec: Thanks for the fix. However, I must say that I don't particularly like this, since this initialization is actually superfluous: if do_setsockopt returns 0, it guarantees that value is initialized. 
It is just necessary to satisfy the heuristics for gcc's maybe-uninitialized check. I would suggest not to use this is a "normal" build at all, and definitely not with -Werror. It makes sense to have this in a special static analysis build, but I regard this as a false positive.

sigiesec: In fact, a better static analysis that looks into do_setsockopt could issue a warning here that the -1 used to initialize value is never read...

bluca: It's definitely a false positive of 4.7, fixed in newer releases - but 4.7 is used by a couple LTS distros so we can't break it. I don't like it either but it's simple and harmless

sigiesec: The gcc manual says:
`For an automatic variable, if there exists a path from the function entry to a use of the variable that is initialized, but there exist some other paths for which the variable is not initialized, the compiler emits a warning if it cannot prove the uninitialized paths are not executed at run time. These warnings are made optional because GCC is not smart enough to see all the reasons why the code might be correct in spite of appearing to have an error.`

-Wmaybe-uninitialized is not to be confused with -Wuninitialized. That one makes sense.

sigiesec: I would rather suggest to add -Wno-maybe-uninitialized at least when building with a gcc version that is known to produce nasty false positives.

It does not make sense to let the build fail because of a broken heuristic.

The unnecessary initialization makes the code harder to read. A reader of the code will wonder under what circumstances the -1 value will be used.

bluca: Sorry, I would rather not - getting the heuristic right for the warning/no warning is much harder and prone to miss corner cases as you have to go with compiler versions, and distros (especially Red Hat) tend to do functional changes all the time that means a version actually might not tell you much, and on top of that they are specific to each build system, and so on and so forth. Been there, done that...

If you are worried about readability perhaps a comment would help just as well?

sigiesec: IMO it does not hurt if -Wno-maybe-uninitialized is disabled on too many builds, even if you feel we should not generally disable it. If it is enabled on the CI builds, it is sufficient. We won't release a version that triggers a warning on the CI. 

I still think it is not adequate to have it in a regular build at all. It is more like a static analysis, which will always potentially lead to false positives.

A comment would partly remedy the readability impact. However, I suspect that these are not the only two places where such an unnecessary initialization must be done to satisfy an older gcc.

PR #2974, opened 2018-03-06
Problem: mis-detection of threadsafe_static_init causes test failures

bluca:
Solution: do not rely __cplusplus >= 201103L to detect whether the
compiler supports thread safe static initialisation, but check only
the proper feature preprocessor macro.
GCC introduced it in version 8, and Clang in version 6.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -92,9 +92,8 @@ uint32_t zmq::generate_random ()
 //    configurable via config.h
 
 //  TODO this should probably be done via config.h
-#if __cplusplus >= 201103L                                                     \
-  || (defined(__cpp_threadsafe_static_init)                                    \
-      && __cpp_threadsafe_static_init >= 200806)                               \
+#if (defined(__cpp_threadsafe_static_init)                                     \
+     && __cpp_threadsafe_static_init >= 200806)                                \
   || (defined(_MSC_VER) && _MSC_VER >= 1900)
 #define ZMQ_HAVE_THREADSAFE_STATIC_LOCAL_INIT 1
 //  TODO this might probably also be set if a sufficiently recent gcc is used


src/random.cpp
```diff
@@ -92,9 +92,8 @@ uint32_t zmq::generate_random ()
 //    configurable via config.h
 
 //  TODO this should probably be done via config.h
-#if __cplusplus >= 201103L                                                     \
-  || (defined(__cpp_threadsafe_static_init)                                    \
-      && __cpp_threadsafe_static_init >= 200806)                               \
+#if (defined(__cpp_threadsafe_static_init)                                     \
```
sigiesec: @bluca Thanks for fixing this. However, probably this is now more restrictive than necessary. While __cpp_threadsafe_static_init is defined only from gcc 8, according to https://gcc.gnu.org/projects/cxx-status.html, thread-safe initialization is implemented already from gcc 4.3. Since gcc 8 is not widely used yet, I think it would be good to query the gcc version explicitly. Do you agree? Then I would send another PR.

bluca: The problem is that it doesn't work - it's easy to reproduce with malamute, the mlm_proto selftests (and others) always abort at shutdown as mutex lock fails with einval - see backtrace that I posted in your other PR :-/

bluca: Note that when building with GCC 8 the problem goes away - so not sure how reliable that page is...

bluca: the quickest test with malamute is:

```LD_LIBRARY_PATH=src/.libs src/.libs/mlm_selftest -t mlm_stream_simple```

Takes milliseconds to run and always fails when building libzmq with gcc < 8 and without this change

PR #2976, opened 2018-03-07
Revert "Problem: mis-detection of threadsafe_static_init causes test failures"

sigiesec:
Reverts zeromq/libzmq#2974

#2974 was based on wrong assumptions. The problem in czmq/malamute was unrelated to thread-safety; and while gcc/clang support the feature *macro* __cpp_threadsafe_static_init only in recent versions, they support the actual feature for a long time (gcc since at least 4.3.0). Probably, the condition could be even more relaxed based on the gcc version number.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -92,8 +92,9 @@ uint32_t zmq::generate_random ()
 //    configurable via config.h
 
 //  TODO this should probably be done via config.h
-#if (defined(__cpp_threadsafe_static_init)                                     \
-     && __cpp_threadsafe_static_init >= 200806)                                \
+#if __cplusplus >= 201103L                                                     \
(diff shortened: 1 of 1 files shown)


bluca:
I'll merge after this one is merged to avoid regressions: https://github.com/zeromq/czmq/pull/1876

BTW could you please push branches to your fork rather than to zeromq/libzmq ? Both CIs are rate-limited per-organisation, so having 2 runs per PR means other jobs will stay queued for twice as much :-/

sigiesec:
Usually I do. However, I reverted the PR via the GitHub Web UI, and it does not offer me to use my fork :(

bluca:
Ah yes - unfortunately the UI is a bit limited.

PR #2977, opened 2018-03-07
Problem: TIPC availability check is too strict

bluca:
Solution: at build time only check if the API is available. In the tests
do a first check and a skip if the functionality is not available.

TIPC needs an in-tree but not loaded by default kernel module, tipc.ko
to be loaded, which requires root, so it is unlikely to be available on
any build system by default.
This will allow most distributions to ship with TIPC support built in,
and to avoid tests failure if the module is not there.

Also add a CI job on Travis and have CMake mark tests as skipped if they return 77 (like autotools does).

@eponsko what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,9 @@ matrix:
         - libsodium-dev
         - asciidoc
         - xmlto
-  - env: BUILD_TYPE=default CURVE=libsodium DRAFT=enabled GSSAPI=enabled PGM=enabled NORM=enabled
+  - env: BUILD_TYPE=default CURVE=libsodium DRAFT=enabled GSSAPI=enabled PGM=enabled NORM=enabled TIPC=enabled
     os: linux
+    sudo: required
     addons:
       apt:
         sources:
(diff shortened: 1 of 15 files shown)


bluca:
Ping?

eponsko:
Oop sorry, missed this. Sounds like a good idea, I guess a portable way to test whether TIPC is available is to try to create a TIPC socket and check if it works. (Noticed now that this is the approach taken in ZMQSourceRunChecks.cmake:183). 

What would be needed is then just to add the check to all the TIPC tests and return -77 if it fails?

bluca:
Yes - that's what I've done if you check the diff. Rather than using directly the bsd socket api though I create a zmq socket, since the code snippet is much shorter. If you check the Travis run, it seems to be working fine (build 3745.13 is enabled, all the other Linux ones are skipped, the OSX/Windows ones are not built at all)

eponsko:
Ah yeah, I didn't see the commits, I thought that I was expected to do the work :-)
Great, I would accept the pull request if I could!

bluca:
@sigiesec ping

Issue #2284, opened 2016-12-31
Problem: no interoperability CI tests


bluca:
I think it would be quite helpful to have CI tests that check the build vs older stable releases, and maybe even other implementations of ZMTP like jeromq, netmq and gomq.

It can't be integrated in the make check/make test infrastructure obviously, but a possible implementation would be to add a builds/interop directory with a bash script per library to test, and to have Travis simply run all scripts in that directory.

This would give us much more confidence, when changing the internal of the library, that the protocol itself is not accidentally broken.

bluca:
This would be useful also to integrate with fuzzers, by adding a plain TCP side - https://github.com/zeromq/libzmq/issues/3867

bluca:
If we go overboards, we could implement this with GSL - add a backend for each ZMTP implementation, and then have as input high-level definition of tests (socket type, security, data to send back and forth)

PR #2981, opened 2018-03-09
Problem: The new threadsafe sockets does not support socket option ZMQ_ROUTING_ID

eponsko:
The ZMQ_ROUTING_ID socket option (previously known as ZMQ_IDENTITY) is a useful function for  per-connection metadata which can be obtained from received messages using zmq_msg_gets().
However, it is currently not supported on the threadsafe sockets in draft. 

This commit adds support for ZMQ_ROUTING_ID/ZMQ_IDENTITY on ZMQ_CLIENT, ZMQ_SERVER, and ZMQ_RADIO sockets. It also modifies existing tests for these sockets to verify that the option works.


bluca:
Please fix the formatting (see failing test)

somdoron:
This was actually intentionally, IMO routing id is kind of a hack and should be achieved through application protocol.

somdoron:
We started those new socket types to get rid of identity and routing id

eponsko:
Yeah I would agree, I'd rather not overload the existing IDENTITY/ROUTING_ID functionality. At the moment I'm using it as a way to pass some metadata about the connection that is useful sometimes on the receiving side (but not for each message or connection), so piggybacking it on the existing metadata exchange in the handshake I think makes sense. I also would prefer to not touch the payload at all, just to keep things simple and not add additional latency to the connection establishment.

I guess what I would prefer would be something like an arbitary string<->string mapping, like `zmq_setsockopt(socket, ZMQ_METADATA, "key:value" ..)`, that could be obtained with `value = zmq_msg_gets(msg, "key")`. I noticed now that the first approach seems to be in the ZMTP protocol (with the reservation that application metadata has to start with "X-"),  but I don't think it's implemented?

eponsko:
How about [something like this](/eponsko/libzmq/metadata) instead? Let's you specify whatever you like as long as the key is "X-*" and value is a string.

bluca:
Looks fine to me, and also looks compatible with ZMTP 3.1:

> Metadata names starting with "X-" SHALL be reserved for application use.

The unit test as well is fine, but the more the better given it's at the protocol level, so it really needs to be well covered.

Also I think our implementation of the protocol in old versions won't break when interoperating with a version with this change (if you could test it though it would be great). If we go ahead, we might want to give a heads-up to the maintainers of all the other full reimplementations of ZMTP to make sure nothing will break.

@somdoron what do you think?

somdoron:
Looks fine to me

eponsko:
I'm mostly familiar with pyzmq and jeromq, so can give those a try. Although since pyzmq is a binding I guess it won't be a too useful test. Is there some kind of implementation (language & version) interoperability test framework somewhere? If not, maybe some parts of the zguide examples could be taken and tested together on Travis, hm..

bluca:
Yeah pyzmq is a binding so it doesn't reimplement the engine - but jeromq will be useful. I guess @somdoron can help us with netmq which is the other one.

Regarding interop tests, that's something I'd really like to have, but -ENOTIME: https://github.com/zeromq/libzmq/issues/2284

eponsko:
Ah I see, would be nice to have indeed. I tested with JeroMQ now at least, additional metadata doesn't seem to bother it, although there's not mechanism to get it out so can't check if it is parsed correctly or not.

include/zmq.h
```diff
@@ -585,6 +585,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread);
 #define ZMQ_BINDTODEVICE 92
 #define ZMQ_ZAP_ENFORCE_DOMAIN 93
 #define ZMQ_LOOPBACK_FASTPATH 94
+#define ZMQ_METADATA 95
```
bluca: you need to add this in src/zmq_drafts.h as well

src/mechanism.cpp
```diff
@@ -170,13 +170,31 @@ size_t zmq::mechanism_t::add_basic_properties (unsigned char *buf,
                         options.routing_id, options.routing_id_size);
     }
 
+
+#ifdef ZMQ_BUILD_DRAFT_API
```
bluca: There's no need to wrap the implementation in ifdefs, it's enough that the symbol is not available - if one hacks it through, it's their problem if it changes

So you can remove all the ifdefs, but leave the test in the DRAFT section of the makefile (and also please add it to tests/CMakeList)

eponsko: Ah I see, will fix. I thought it was correctly added to tests/CMakeList, did I miss adding it somewhere else in there? 
https://github.com/eponsko/libzmq/blob/master/tests/CMakeLists.txt#L138

bluca:
@eponsko thanks! Could you please add documentation to docs/zmq_setsockopt.txt and docs/zmq_msg_get[s] ?

eponsko:
@bluca sure, will try to write something poetical

bluca:
asciiart or no merge

eponsko:
¯\_(ツ)_/¯

PR #2982, opened 2018-03-10
Problem: formatting issues in CI

bluca:
Solution: ignore tweetnacl.c and reformat the rest

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -641,9 +641,9 @@ int zmq::make_fdpair (fd_t *r_, fd_t *w_)
         *w_ = *r_ = -1;
         return -1;
     } else {
-        //  If there's no SOCK_CLOEXEC, let's try the second best option. Note that
-        //  race condition can cause socket not to be closed (if fork happens
-        //  between socket creation and this point).
+    //  If there's no SOCK_CLOEXEC, let's try the second best option. Note that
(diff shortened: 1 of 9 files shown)


sigiesec:
Unfortunately I had no time to investigate this further yet.

bluca:
Ok, no worries. Wanna merge this?

sigiesec:
@bluca I now found that it is not due to differences between OSs, but between versions. clang-format from LLVM 5.0.0 appears to have some bug when formatting comments preceding preprocessor directives, and I was using a more recent version under Windows. I will check if 5.0.1 fixes the issue already.

sigiesec:
5.0.1 still has the bug :(

sigiesec:
I think this is the fix of the bug: http://llvm.org/viewvc/llvm-project?view=revision&revision=308882

bluca:
Ah - so the version we have available on Travis is the wrong one? There is a 6.0 PPA that could be used:

http://apt.llvm.org/trusty/dists/

But then 6.0 is not really widely available yet:

https://packages.ubuntu.com/search?keywords=clang-format-6.0&searchon=names&suite=all&section=all
https://packages.debian.org/search?keywords=clang-format-6.0&searchon=names&suite=all&section=all

Only in unreleased versions of Debian and Ubuntu. What do we do? Wait until it's more widely available?

PR #2983, opened 2018-03-12
Fixes for mingw cross-compilation

tonytheodore:
Tested on both static/shared builds (untested runtime - no windows machine at hand).

`make test` on OSX native build has single failure `44 - test_many_sockets (Timeout)` both before and after changes.


CMakeLists.txt
```diff
@@ -949,20 +961,18 @@ else ()
...
+      add_library (libzmq-static STATIC $<TARGET_OBJECTS:objects> ${public_headers} ${html-docs} ${readme-docs} ${zmq-pkgconfig} ${CMAKE_CURRENT_BINARY_DIR}/version.rc)
+    endif ()
     set_target_properties (libzmq-static PROPERTIES
       PUBLIC_HEADER "${public_headers}"
       COMPILE_DEFINITIONS "ZMQ_STATIC"
       OUTPUT_NAME "zmq"
       PREFIX "lib")
   endif()
 
-  target_include_directories (objects
```
bluca: why is this being removed?

tests/CMakeLists.txt
```diff
@@ -192,6 +194,7 @@ foreach(test ${tests})
         link_directories(${test} PRIVATE "${CMAKE_SOURCE_DIR}/../lib")
     endif()
   endif()
+  set_target_properties(${test} PROPERTIES COMPILE_DEFINITIONS "${libzmq_DEFS}" )
```
sigiesec: I don't think this is the adequate way to propagate compile definitions to dependent targets.

target_compiler_definitions(libzmq PUBLIC ...) should be used somewhere above instead, then they should be propagated automatically.

See https://cmake.org/cmake/help/v3.0/command/target_compile_definitions.html

unittests/CMakeLists.txt
```diff
@@ -33,6 +34,7 @@ foreach(test ${unittests})
   
   target_link_libraries(${test} libzmq-static ${OPTIONAL_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} unity)
   
+  set_target_properties(${test} PROPERTIES COMPILE_DEFINITIONS "${libzmq_DEFS}" )
```
sigiesec: See comment above

CMakeLists.txt
tonytheodore: It's moved up to where the `add_library(objects ...` line is.

tests/CMakeLists.txt
tonytheodore: Thanks, I'll have a look at `PUBLIC`.

CMakeLists.txt
bluca: Ah sorry, missed that

tests/CMakeLists.txt
tonytheodore: Nice! `PUBLIC` is a very handy feature.

PR #2984, opened 2018-03-12
Partial fix for issue 2963

kachanovskiy:
Problem: issues found during static code analysis.
This PR addresses 8 similar issues where method's 2nd argument is of zmq:fd_t type, and value passed is of the same type, but for some reason is cast to int, causing static analysis to tag it as suspicious cast for 64bit mode. 

Solution: remove cast


sigiesec:
Great, thanks for fixing these!

PR #2987, opened 2018-03-13
Migrate further tests (test_hwm, test_router_mandatory) to unity

sigiesec:


eponsko:
Hi Simon, an unrelated comment (don't know how or even if its possible to send messages to a user on Github, so writing here). I noticed that if you place the context creation in setUp() and cleanup in tearDown(), it's easy to run into the problem that the context is not clean (sockets still open, etc) if a test fails halfway and tearDown is called, causing the test to hang in zmq_ctx_term(). 

So maybe it would be good to update the README.md in tests/ to say something about not doing that and any other best practices with Unity that should be taken into account when writing a test.

sigiesec:
Hmm... Yes that's true. The tearDown would need to call close_zero_linger on all sockets that are still open. This needs some support, I will try to design something for that. Thanks for the hint :)

bluca:
@sigiesec looks like there's a new compilation failure in windows: https://ci.appveyor.com/project/zeromq/libzmq/build/build-2121/job/s5byxnr1dnk28xf5 want to ahve a look before I merge?

bluca:
```
c:\projects\libzmq\tests\testutil_unity.hpp(45): error C3861: 'snprintf': identifier not found
```

sigiesec:
Ah, I didn't check the appveyor build, since I built with VS myself, but VS2013-- do not have snprintf...

bluca:
If you squash the commit then I'll merge

bluca:
@sigiesec I can do a squash-merge instead if you prefer?

sigiesec:
I squashed the commits now :)

PR #2988, opened 2018-03-13
[discussion] CMake: Linker PDB install rule does not work when Visual Studio generators are used

wittmeie:
Problem: When CMake Visual Studio generators are used, the install rule for the linker PDB does not work.

Solution: Update `cmake_minimum_required` to v3.1.3 and use generator expressions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # CMake build script for ZeroMQ
 
-cmake_minimum_required (VERSION 2.8.12)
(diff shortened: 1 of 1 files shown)


bluca:
Sorry, we can't break backward compatibility with Cmake << 3.0, as it is still used in many LTS distros, and older systems.

wittmeie:
That's what I expected! I will check whether there exists a workaround for this issue.

bluca:
Great, thanks for your work

PR #2990, opened 2018-03-14
Problem: test assertion failures do not clean up properly

sigiesec:
@eponsko Please have a look if this is what you expected.

### DIFF
--- a/tests/testutil_unity.hpp
+++ b/tests/testutil_unity.hpp
@@ -85,3 +85,102 @@ void recv_string_expect_success (void *socket, const char *str, int flags)
     if (str)
         TEST_ASSERT_EQUAL_STRING_LEN (str, buffer, len);
 }
+
+// do not call from tests directly, use setup_test_context, get_test_context and teardown_test_context only
+void *internal_manage_test_context (bool init, bool clear)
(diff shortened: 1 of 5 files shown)


eponsko:
Very nice, thanks!

tests/testutil_unity.hpp
```diff
@@ -85,3 +85,102 @@ void recv_string_expect_success (void *socket, const char *str, int flags)
     if (str)
         TEST_ASSERT_EQUAL_STRING_LEN (str, buffer, len);
 }
+
+// do not call from tests directly, use setup_test_context, get_test_context and teardown_test_context only
+void *internal_manage_test_context (bool init, bool clear)
```
eponsko: Could this function be declared as static? Then you wouldn't need the comment above, I guess the same comment is true for internal_manage_test_sockets()

sigiesec: You already mentioned the problem ;) Of course, this could be moved to a .cpp file, which would need to be compiled/linked with every test program.

bluca: I'd say: let's keep it simple :-)

Issue #2632, opened 2017-07-26
File descriptor leak in context constructor


hoditohod:
Hi libzmq guys!

## Issue
I have a test that spawns a few threads, creates a zmq context in each of them, does some boring stuff, then destroys the context and joins the threads. The test application leaks a few fd's to /dev/urandom in each iteration.

## Env
Ubuntu 16.04 64bit
libzmq 4.2.2 + cppzmq

## Root cause
The libzmq build I use has the ZMQ_USE_TWEETNACL undefined, so `randombytes()` directly reads /dev/urandom storing the fd in global static variable [here](https://github.com/zeromq/libzmq/blob/master/src/tweetnacl.c#L908) without any protection against concurrent execution. When calling the context constructor concurrently from multiple threads, the scoped_lock_t [here](https://github.com/zeromq/libzmq/blob/master/src/ctx.cpp#L94) is useless because the corresponding mutex is a class member (each instance has one mutex), and it doesn't prevent multiple concurrent context instances calling `randombytes()`.

As a result multiple threads open the file but only one close is issued.

bluca:
Thanks for the report, I can see the issue. We'll have to lock&refcount I guess.

In the meanwhile, you can use 4.2.2 built with libsodium with packages for 16.04 from OBS: http://download.opensuse.org/repositories/network:/messaging:/zeromq:/release-stable/xUbuntu_16.04/

bluca:
@hoditohod please try again from latest master

PR #3001, opened 2018-03-19
Problem:  global random init/deinit breaks existing applications

### LINKED ISSUE #2991: New invalid argument error in libzmq, mutex.hpp:142
# Issue description
One of the unit test suites we have finishes all of the test cases (which involve starting up and shutting down zmq, probably at least 50 times), but then there is an invalid argument error by zmq after boost reports the number of test cases that have passed. The first time observing this issue was on March 9th, but we cache builds of dependencies. The cached version on our develop branch was last updated February 27. Switching our builds to use the last tagged release (v4.2.3) makes the error go away.

The output for the zmq error: Invalid argument (/home/travis/build/GMLC-TDC/HELICS-src/libzmq/src/mutex.hpp:142)

# Environment
Travis CI Linux VMs using gcc-6, gcc-4.9, and clang-3.6

* libzmq version (commit hash if unreleased): One of the commits between February 27 and March 9.
* OS: Linux

1.  Checkout and build GMLC-TDC/HELICS-src develop branch
2. Run key-tests test suite. I haven't narrowed it down to a specific zeromq commit, or line in our test suite. If that changes I'll provide an update here.

https://travis-ci.org/GMLC-TDC/HELICS-src/jobs/353188544 - see error at the bottom

# What's the expected result?
https://travis-ci.org/GMLC-TDC/HELICS-src/jobs/353393510

bluca:
Solution: restrict it only to the original issue #2632, Tweetnacl on
*NIX when using /dev/urandom, ie: without the new Linux getrandom()
syscall.

Existing applications might use atexit to register cleanup functions
(like CZMQ does), and the current change as-is imposes an ordering
that did not exist before - the context MUST be created BEFORE
registering the cleanup with atexit. This is a backward incompatible
change that is reported to cause aborts in some applications.

Although libsodium's documentation says that its initialisation APIs
is not thread-safe, nobody has ever reported an issue with it, so
avoiding the global init/deinit in the libsodium case is the less
risky option we have.

Tweetnacl users on Windows and on Linux with getrandom (glibc 2.25 and
Linux kernel 3.17) are not affected by the original issue.

Fixes #2991

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -86,6 +86,14 @@ uint32_t zmq::generate_random ()
 //  order fiasco, this is done using function-local statics, if the
 //  compiler implementation supports thread-safe initialization of those.
 //  Otherwise, we fall back to global statics.
+//  HOWEVER, this initialisation code imposes ordering constraints, which
+//  are not obvious to users of libzmq, and may lead to problems if atexit
+//  or similar methods are used for cleanup.
+//  In that case, a strict ordering is imposed whereas the contexts MUST
+//  be initialised BEFORE registering the cleanup with atexit. CZMQ is an
+//  example. Hence we make the choice to restrict this global transition
+//  mechanism ONLY to Tweenacl + *NIX (when using /dev/urandom) as it is
+//  the less risky option.
 
 //  TODO if there is some other user of libsodium besides libzmq, this must
 //    be synchronized by the application. This should probably also be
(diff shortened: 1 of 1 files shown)


src/random.cpp
```diff
@@ -86,6 +86,13 @@ uint32_t zmq::generate_random ()
 //  order fiasco, this is done using function-local statics, if the
 //  compiler implementation supports thread-safe initialization of those.
 //  Otherwise, we fall back to global statics.
+//  HOWEVER, this initialisation code adds a race condition when an
```
sigiesec: It is not a race condition (this has nothing to do with concurrency). Maybe:

```
HOWEVER, this initialisation code imposes ordering constraints, which are not obvious to users of libzmq, and may lead to problems if atexit or similar methods are used for cleanup.
```

src/random.cpp
```diff
@@ -140,6 +145,14 @@ static void manage_random (bool init)
             randombytes_close ();
         }
     }
+
+#elif defined(ZMQ_USE_LIBSODIUM)
+    if (init) {
+        int rc = sodium_init ();
```
sigiesec: Since this does not use the refcount, this means that sodium_init and randombytes_close will be called for each context. Is that what you meant?

bluca: Yes - that is how it worked before my initial change. They were called inline in the Context class constructor and destructor.

In reality I think randombytes_close is mostly a no-op in libsodium, and nobody ever reported a problem with the multiple sodium_init calls either (with libsodium again).

src/random.cpp
bluca: Thanks, updated

src/random.cpp
sigiesec: Ok, just wanted to be sure.

minrk: Adding a clarifying correction for anyone who comes here in the future: this is only a no-op if getrandom is unavailable (already excluded in the above condition). If not, libsodium has all the same threadsafety issues as tweetnacl, so this PR causes crashes because libsodium has all the same requirements as tweetnacl here.

PR #3007, opened 2018-03-20
Problems: some APIs have met stable conditions

bluca:
Solutions: mark them as stable


bluca:
@sigiesec how's the new monitor events doing these days? Do you think it's good to go or is it likely that more changes will happen?

sigiesec:
I would still keep them as DRAFT.

bluca:
ACK

PR #3011, opened 2018-03-21
Migrate test_reqrep_tcp to unity

sigiesec:


tests/test_reqrep_tcp.cpp
```diff
@@ -28,129 +28,77 @@
...
+    void *sc = test_context_socket (ZMQ_REQ);
+    TEST_ASSERT_SUCCESS_ERRNO (
+      zmq_setsockopt (sc, ZMQ_IPV6, &ipv6, sizeof (int)));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (sc, my_endpoint));
 
     bounce (sb, sc);
 
-    rc = zmq_disconnect (sc, my_endpoint);
-    assert (rc == 0);
+    // TODO is explicit disconnect/unbind essential for the test? if not,
```
bluca: yes - it's checking that endpoints are saved correctly, especially for v6 there were a few regressions in the past were the strings didn't match

sigiesec: Ok, then I will change this to a comment that explains this.

PR #3018, opened 2018-03-22
Problem: backward incompatible change to NULL with ZAP

bluca:
Solution: fix it like for other mechanism, add test

### DIFF
--- a/src/null_mechanism.cpp
+++ b/src/null_mechanism.cpp
@@ -69,24 +69,27 @@ int zmq::null_mechanism_t::next_handshake_command (msg_t *msg_)
             errno = EAGAIN;
             return -1;
         }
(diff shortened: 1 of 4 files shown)


bluca:
@sigiesec when you have a sec could you please have a look and see if  I missed something?

sigiesec:
@bluca I gave it a quick look. Looks good from my point of view :)

bluca:
OK - I'll then do a revert of the recent changes to STABLE API, do a 4.2.5 bugfix only so that pyzmq can get unblocked, and then re-instate 4.3.0

sigiesec:
https://developercommunity.visualstudio.com/content/problem/220882/unexpected-c2610-error.html?childToView=221696#comment-221696

PR #3019, opened 2018-03-23
Problem: regression in 4.2.3 needs to be fixed before 4.3

bluca:
Solution: temporarily revert API changes, restore next version to 4.2.5 and compile changelog

After tagging 4.2.5 I'll restore the API changes and 4.3


bluca:
ping

PR #3032, opened 2018-03-28
Factorize poller

diorcety:
This is not the final PR, I will wait for feedbacks before formatting each commits

I start with this issue https://github.com/zeromq/czmq/issues/1881
TLDR; We need a way to interrupt the select poller on Windows. On Windows CTRL+C doesn't interrupt select like unix system. The best way to do that is to call QueueUserAPC on each thread which will interrupt some functions(https://msdn.microsoft.com/fr-fr/library/windows/desktop/ms684954(v=vs.85).aspx) with WAIT_IO_COMPLETION (that should work https://github.com/diorcety/czmq/commit/794a0e616519ed4ebb9c284ab9b14b0a7786aa0a).
For that we need to use WSA correctly.
The issue is that in multiple place in code where there is a select(or poll) implementations.
This PR goal is to factorize that code (using poller implementations) in order to allow in the future (soon) to use WSA in place of select on Windows.

I have tested it on Linux: epoll(only shutdown test doesn't pass), poll, select
and on Windows: select
There is certainly performance drawbacks. 
I'am opened to suggestions

### DIFF
--- a/src/i_poll_events.hpp
+++ b/src/i_poll_events.hpp
@@ -30,6 +30,8 @@
 #ifndef __ZMQ_I_POLL_EVENTS_HPP_INCLUDED__
 #define __ZMQ_I_POLL_EVENTS_HPP_INCLUDED__
 
+#include "fd.hpp"
+
 namespace zmq
 {
 // Virtual interface to be exposed by object that want to be notified
@@ -37,13 +39,21 @@ namespace zmq
 
 struct i_poll_events
 {
+    typedef void * handle_t;
+
     virtual ~i_poll_events () {}
 
     // Called by I/O thread when file descriptor is ready for reading.
-    virtual void in_event () = 0;
+    virtual void in_event (handle_t handle_) = 0;
 
     // Called by I/O thread when file descriptor is ready for writing.
-    virtual void out_event () = 0;
+    virtual void out_event (handle_t handle_) = 0;
+
+    // Called by I/O thread when file descriptor has error.
+    virtual void err_event (handle_t handle_) = 0;
+
+    // Called by I/O thread when file descriptor is ready (priority).
+    virtual void pri_event (handle_t handle_) = 0;
 
     // Called when timer expires.
     virtual void timer_event (int id_) = 0;
--- a/src/poller_base.hpp
+++ b/src/poller_base.hpp
@@ -130,6 +130,8 @@ class poller_base_t
     void add_timer (int timeout_, zmq::i_poll_events *sink_, int id_);
     void cancel_timer (zmq::i_poll_events *sink_, int id_);
 
+    virtual int wait (int timeout) = 0;
+
   protected:
     //  Called by individual poller implementations to manage the load.
     void adjust_load (int amount_);
@@ -160,7 +162,7 @@ class poller_base_t
 };
 
 //  Base class for a poller with a single worker thread.
-class worker_poller_base_t : public poller_base_t
+class worker_poller_base_t : public virtual poller_base_t
 {
   public:
     worker_poller_base_t (const thread_ctx_t &ctx_);
--- a/src/select.hpp
+++ b/src/select.hpp
@@ -58,13 +58,13 @@ struct i_poll_events;
 //  Implements socket polling mechanism using POSIX.1-2001 select()
 //  function.
 
-class select_t : public worker_poller_base_t
+class select_base_t : public virtual poller_base_t
 {
   public:
     typedef fd_t handle_t;
 
-    select_t (const thread_ctx_t &ctx_);
-    ~select_t ();
+    select_base_t ();
+    virtual ~select_base_t ();
 
     //  "poller" concept.
     handle_t add_fd (fd_t fd_, zmq::i_poll_events *events_);
@@ -73,13 +73,12 @@ class select_t : public worker_poller_base_t
     void reset_pollin (handle_t handle_);
     void set_pollout (handle_t handle_);
     void reset_pollout (handle_t handle_);
-    void stop ();
 
     static int max_fds ();
 
+    virtual int wait (int timeout);
   private:
-    //  Main event loop.
-    void loop ();
+    virtual void check_thread ();
 
     //  Internal state.
     struct fds_set_t
@@ -115,7 +114,7 @@ class select_t : public worker_poller_base_t
         bool has_retired;
     };
 
-    void select_family_entry (family_entry_t &family_entry_,
+    int select_family_entry (family_entry_t &family_entry_,
                               int max_fd_,
                               bool use_timeout_,
                               struct timeval &tv_);
@@ -161,11 +160,26 @@ class select_t : public worker_poller_base_t
     static fd_entries_t::iterator
     find_fd_entry_by_handle (fd_entries_t &fd_entries, handle_t handle_);
 
-    select_t (const select_t &);
-    const select_t &operator= (const select_t &);
+    select_base_t (const select_base_t &);
+    const select_base_t &operator= (const select_base_t &);
+};
+
+class select_t : public worker_poller_base_t, public select_base_t
+{
+  public:
+    select_t (const thread_ctx_t &ctx_);
+    virtual ~select_t ();
+
+    void stop();
+
+  private:
+    //  Main event loop.
+    virtual void loop ();
+    virtual void check_thread ();
 };
 
 typedef select_t poller_t;
+typedef select_base_t base_poller_t;
 }
 
 #endif
--- a/src/select.hpp
+++ b/src/select.hpp
@@ -61,7 +61,8 @@ struct i_poll_events;
 class select_base_t : public virtual poller_base_t
 {
   public:
-    typedef fd_t handle_t;
+    typedef void* handle_t;
+    static const handle_t handle_invalid;
 
     select_base_t ();
     virtual ~select_base_t ();
@@ -136,7 +137,7 @@ class select_base_t : public virtual poller_base_t
     family_entries_t::iterator current_family_entry_it;
 
     int try_retire_fd_entry (family_entries_t::iterator family_entry_it,
-                             zmq::fd_t &handle_);
+                             handle_t &handle_);
 
     static const size_t fd_family_cache_size = 8;
     std::pair<fd_t, u_short> fd_family_cache[fd_family_cache_size];
--- a/src/signaler.cpp
+++ b/src/signaler.cpp
@@ -239,8 +239,17 @@ void zmq::signaler_t::send ()
 #endif
 }
 
+int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_);
+
 int zmq::signaler_t::wait (int timeout_)
 {
+#if defined ZMQ_HAVE_POLLER
+    zmq_pollitem_t items[1];
+    items[0].socket = NULL;
+    items[0].fd = r;
+    items[0].events = ZMQ_POLLIN;
+    return zmq_poller_poll (items, 1, timeout_);
+#else
 #ifdef HAVE_FORK
     if (unlikely (pid != getpid ())) {
         // we have forked and the file descriptor is closed. Emulate an interrupt
@@ -305,6 +314,7 @@ int zmq::signaler_t::wait (int timeout_)
 
 #else
 #error
+#endif
 #endif
 }
 
--- a/src/vmci_connecter.cpp
+++ b/src/vmci_connecter.cpp
@@ -100,15 +100,15 @@ void zmq::vmci_connecter_t::process_term (int linger_)
     own_t::process_term (linger_);
 }
 
-void zmq::vmci_connecter_t::in_event ()
+void zmq::vmci_connecter_t::in_event (i_poll_events::handle_t handle_)
 {
     //  We are not polling for incoming data, so we are actually called
     //  because of error here. However, we can get error on out event as well
     //  on some platforms, so we'll simply handle both events in the same way.
-    out_event ();
+    out_event (i_poll_events::handle_t handle_);
 }
 
-void zmq::vmci_connecter_t::out_event ()
+void zmq::vmci_connecter_t::out_event (i_poll_events::handle_t handle_)
 {
     fd_t fd = connect ();
     rm_fd (handle);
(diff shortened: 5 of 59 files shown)


bluca:
What about building your project using poll on Windows, instead of select? This is a bit of a kitchen sink throw...

src/zmq.cpp
```diff
@@ -688,12 +688,12 @@ const char *zmq_msg_gets (const zmq_msg_t *msg_, const char *property_)
...
 
 #if defined ZMQ_HAVE_POLLER
-inline int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
+int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
 {
     // implement zmq_poll on top of zmq_poller
     int rc;
     zmq_poller_event_t *events;
-    zmq::socket_poller_t poller;
+    zmq::socket_poller_t *poller = new zmq::socket_poller_t;
```
bluca: IIRC this was on the stack intentionally, as it will murder performances otherwise

diorcety: This is crashing on my Windows computer otherwise (stack overflow)

diorcety:
Yes I understand. There is maybe a better way to do that. But could be acceptable on the other way: can't quit a libzmq console application on Linux using SIGINT?

src/zmq.cpp
bluca: This is documented in the INSTALL file, you need to increase the stack size: https://github.com/zeromq/libzmq/blob/master/INSTALL#L49

bluca:
What I mean is: each change should have a clear Problem..Solution statement, and the code change should reflect that. In this case, the problem appears to be that ctrl-c doesn't work on Windows in some cases - but it's not clear to me that the entire 1745 lines diffset is necessary to solve that problem?

diorcety:
The main issue is select(or poll) on Windows. Unlike on Linux these calls are not interruptable by a signal. These is easily repeatable try to run malamute on Windows, without any connecting client, and press ctrl+c. Nothing happens. The ctrl+c is correcly catched by czmq, the flags interruption flags of czmq are set, but all threads are stucks in select. So if there is no input on sockets, and no timeout, you have to kill the process hardly.
When I start to look for this issue, I try to only change the code in select.cpp (force to use WSAWaitForMultipleEvents with the alertable flag). But the select function is used at multiple places, and WSAWaitForMultipleEvents is badly used currently).
The goal is to factorize the select/poll code in one place. Then after to correctly use WSAWaitForMultipleEvents on Windows in order to have the same feature that Linux

src/i_poll_events.hpp
```diff
@@ -30,20 +30,30 @@
...
 
     // Called by I/O thread when file descriptor is ready for reading.
-    virtual void in_event () = 0;
+    virtual void in_event (handle_t handle_) = 0;
 
     // Called by I/O thread when file descriptor is ready for writing.
-    virtual void out_event () = 0;
+    virtual void out_event (handle_t handle_) = 0;
+
+    // Called by I/O thread when file descriptor has error.
```
sigiesec: I don't understand why err_event and pri_event were added. If I saw it correctly, all implementations call in_event.

src/i_poll_events.hpp
```diff
@@ -30,20 +30,30 @@
...
 
 struct i_poll_events
 {
+    typedef void * handle_t;
+
     virtual ~i_poll_events () {}
 
     // Called by I/O thread when file descriptor is ready for reading.
-    virtual void in_event () = 0;
+    virtual void in_event (handle_t handle_) = 0;
```
sigiesec: Why is it now necessary to pass the handle to the *_event methods?

src/select.hpp
```diff
@@ -61,7 +61,8 @@ struct i_poll_events;
 class select_base_t : public virtual poller_base_t
 {
   public:
-    typedef fd_t handle_t;
+    typedef void* handle_t;
```
sigiesec: Why was this done here? I don't think it is a good idea to use void* internally within libzmq, together with numerous casts that are prone to errors. I know that there are some other uses, but this should not be extended IMO.

In addition, due to the definition of i_poll_events::handle_t, this cannot be anything else than void*. So defining it in every poller implementation is redundant. If the void*-based definition is really added to i_poll_events, it should probably be removed in the poller_t implementations. But also see my comment in i_poll_events itself.

The external C libzmq API has them to make the data structures opaque, but the implementation of libzmq C++, where this should not be necessary.

src/signaler.cpp
```diff
@@ -239,8 +239,17 @@ void zmq::signaler_t::send ()
 #endif
 }
 
+int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_);
+
 int zmq::signaler_t::wait (int timeout_)
 {
+#if defined ZMQ_HAVE_POLLER
```
sigiesec: I am not sure, but it zmq_poller_poll (respectively poller_t) might have not been used here intentionally. If that was the case, then probably for performance reasons. The performance impact of this should be checked.

zmq_poller_poll is an external API method, and I am also not sure if they are intended to be called from within the library.

bluca:
Let's do one thing at a time. Mixing very large refactors and bug fixing very easily leads to regressions that are hard to bisect. Tests are already failing. Please fix the ctrl-c problem by itself, wherever it is, and then later we can think about refactoring (separately and with extensive unit tests added)

src/vmci_connecter.cpp
```diff
@@ -100,15 +100,15 @@ void zmq::vmci_connecter_t::process_term (int linger_)
...
 }
 
-void zmq::vmci_connecter_t::in_event ()
+void zmq::vmci_connecter_t::in_event (i_poll_events::handle_t handle_)
 {
     //  We are not polling for incoming data, so we are actually called
     //  because of error here. However, we can get error on out event as well
     //  on some platforms, so we'll simply handle both events in the same way.
-    out_event ();
+    out_event (i_poll_events::handle_t handle_);
```
sigiesec: This doesn't look as this was intended. Does this event compile?

Should probably be 
```
out_event (handle_);
```

src/i_poll_events.hpp
```diff
@@ -30,20 +30,30 @@
...
+#include "fd.hpp"
+
 namespace zmq
 {
 // Virtual interface to be exposed by object that want to be notified
 // about events on file descriptors.
 
 struct i_poll_events
 {
+    typedef void * handle_t;
```
sigiesec: if anything, this should probably be a typedef based on poller_t::handle_t (which needs to be moved somewhere else since this introduces an implicit cyclic dependency between i_poll_events and poller_t)

src/poller_base.hpp
```diff
@@ -160,7 +162,7 @@ class poller_base_t
 };
 
 //  Base class for a poller with a single worker thread.
-class worker_poller_base_t : public poller_base_t
+class worker_poller_base_t : public virtual poller_base_t
```
sigiesec: This did intentionally not use virtual inheritance, which increases the cost of virtual functions calls. Can you explain why using virtual inheritance has become necessary? Can it be avoided? This applies also to severals other classes that have been changed to use virtual inheritance in this commit.

src/poller_base.hpp
```diff
@@ -130,6 +130,8 @@ class poller_base_t
     void add_timer (int timeout_, zmq::i_poll_events *sink_, int id_);
     void cancel_timer (zmq::i_poll_events *sink_, int id_);
 
+    virtual int wait (int timeout) = 0;
```
sigiesec: This method deserves some documentation. What should it wait for? What do the return values mean?

sigiesec:
Apart from my detailed comments, I do not understand the Ctrl-C problem. What cannot be interrupted? Is it about the internal polling of the libzmq threads, which was using poller_t before? Probably not, why would you want these to be interrupted? If the application wants to shut down, it should properly shut down the zmq context, and the sockets. Alternatively, it could simply abort. In both cases, there is no need to interrupt any polling in the zmq threads. 

Therefore, it seems to be about the use of zmq_poll or zmq_poller_wait(_all). Wouldn't it be much easier to change the application either not to use an infinite timeout, or to add another socket to the poll items, which is used to interrupt polling (which might either be a zmq or a OS socket, similar suggestion is here: https://stackoverflow.com/a/384397/2019765), rather than to perform these extensive changes to libzmq?

The changes to libzmq might make sense for some other reason resp. to solve another problem, but I am with @bluca in that they seem to be very complex compared to the problem they are stated to solve.

src/i_poll_events.hpp
diorcety: For multiple reasons:
-There is 4 events defined in zmq: https://github.com/zeromq/libzmq/blob/c8a1c4542d13b6492949e7525f4fe8da266cac2b/include/zmq.h#L461
-For that: https://github.com/zeromq/libzmq/blob/master/src/ipc_connecter.cpp#L113
  Allow a more fine grain callback
-I was changing the definitions

src/i_poll_events.hpp
diorcety: I'm totally agree with you. I can put the handle definition in another file it will be nicer

src/poller_base.hpp
diorcety: The main issue is that worker_poller_base_t implements poller_base_t and *_poller class implements worker_poller_base_t and the *_base_poller which implements also poller_base_t.
We can avoid that by explicitly override the functions in *_poller. Can be done

src/poller_base.hpp
diorcety: Totally agree. This function behave like select/poll/...

diorcety:
Like I said this not for fun. Take malamute on Windows, run it without connection and try to do a CTRL+C it will not work, because the EINTR is handled on Linux inside libzmq, but not the equivalent "stuff" on Windows.
There is nothing related with CTRL+C on this PR. This is just the background.

Why this refactoring? Avoid to have multiple implementations of the poller? Maybe this is not the good way, and i'm open to all suggestions but actually there is multiple select stuff (specially for Windows) not implemented in the same way, without the same functionnality (cf https://github.com/zeromq/libzmq/blob/master/src/select.cpp#L386) which is not even correct: https://github.com/zeromq/libzmq/pull/3032/commits/e06c2cec62cf0ab1f634605dbf538267d9e36e36(was detected with this refactoring)

Maybe basing all on poller_t is not the right thing (performance issue), but for me it seems better to have a generic "select/poll" used everywhere doing the right things for each OS/methods

After that replace it one place the windows hack on select.cpp by not using select at all but use
WSAWaitForMultipleEvents and WSAEnumNetworkEvents

bluca:
Where is Malamute hanging exactly? Do you have a backtrace?

bluca:
As @sigiesec noted it's important to understad which select - if it's on the I/O thread it won't matter, if it's in an application thread (ie: via zmq_poll) then we should fix Malamute to avoid infinite waiting

diorcety:
Do the same on Linux, malamute will print "interrupted" and exit gracefully... This not a malamute issue at the end.

bluca:
It's this bit then: https://github.com/zeromq/malamute/blob/master/src/malamute.c#L125

```
    while (true) {
        char *message = zstr_recv (server);
        if (message) {
            puts (message);
            free (message);
        }
        else {
            puts ("interrupted");
            break;
        }
    }
```

As @sigiesec it's doing a blocking receive, without timeouts

src/i_poll_events.hpp
sigiesec: Ok I see. I suppose ipc_connecter_t should be changed then to make use of this finer distinction. In the end, the comment in in_event should be removed and the implementation be removed/replaced by an assertion.

src/signaler.cpp
sigiesec: Actually I noticed that zmq_poller_poll is not directly an external API method,but used to implement the deprecated zmq_poll. At some point, it will be removed, so using poller_base_t sounds better anyway :)

sigiesec:
@diorcety I totally agree that removing the duplication of code using select and poll should be removed. As you already suggested, I also have the gut feeling that this might better be done on a lower level than that of poller_t. Also, it might still make sense to use different implementations innthe io thread and for zmq_poller. But deciding on this requires some measurements.

bluca:
Again, please postpone any refactoring. Keep it simple - Problem:.. Solution:.. which in this case it's the EINTR handling.

diorcety:
Nevermind, I will keep that on my repo

bluca:
@diorcety please don't take this personally :-(
Just trying to follow the C4 process that we all have to abide to:

> 2.3.2 A patch SHOULD be a minimal and accurate answer to exactly one identified and agreed problem.
> 2.3.6 A patch MUST compile cleanly and pass project self-tests on at least the principle target platform.
> 2.4.1 Change on the project SHALL be governed by the pattern of accurately identifying problems and applying minimal, accurate solutions to these problems.

https://rfc.zeromq.org/spec:42/C4
This was written specifically to avoid mixing multiple big changes in one large diffset, which historically caused many problems.

If you could please send the single EINTR fix by itself (it's OK if it has to be fixed in a couple of different places separately) we can start by merging that and fixing the ctrl-c problem, and then separately the refactoring of the pollers can be tackled - as you have seen from the comments this is not a trivial problem and needs to be approached with care. It does not have to hold back the fix for the interrupt issue.

diorcety:
Hi,
I have no time currently to do the modification you want, (maybe the next month...)
I can't give you the EINTR fix, because I haven't it.
You will need for each select in libzmq to implement the WSAWaitForMultipleEvents hack (because this is really a hack) (and use it always) https://github.com/diorcety/libzmq/blob/52f14903ac1acbf529ca1fa79c49b8a2a55505b0/src/select.cpp#L379
(my version add a fix compare to the master version (unlink sockets and events)

sigiesec:
I don't think this part should be done. The use of WSAWaitForMultipleEvents is really weird, since select is called in addition afterwards. At the moment it is only used when there is a mix of different address families, but I don't think it should be done this way always.

The Ctrl-C problem must be addressed outside of libzmq IMO.

diorcety:
Yes that why I try to do a refactoring in order to do the replacement of this hack at one place.

If CTRL-c problem must be addressed outside of libzmq we can remove all the EINTR handling like this? https://github.com/zeromq/libzmq/blob/master/src/select.cpp#L434 

Yes I'm trolling a bit sorry.

Issue #2871, opened 2017-12-18
Publisher/Subscriber Sockets Not Honoring Send/Receive High Watermarks


9ee1:
# Issue description
I am having some trouble understanding the behavior I am seeing for Publisher/Subscriber sockets in respect to the send/receive high watermarks.

My basic understanding from the documentation is that the Publisher socket maintains 1 outbound queue for each connected subscriber and it is only impacted by the rate that publisher is able to send messages to that subscriber on the wire, and **not** on how fast the subscriber is processing the messages.

The Subscriber socket maintains 1 incoming queue for each publisher is is connected to and messages are simply lost if the subscriber does not process messages faster than they are received and the watermark is reached.

For the subscriber socket in particular, I  am not seeing this behavior. It seems the watermark is completely ignored. Can someone review this example and help me understand what I am doing wrong, if anything? 

Thanks!

# Environment
* libzmq version (commit hash if unreleased): 4.2.2, compiled as X64
* OS: Windows 10 X64

# Minimal test code / Steps to reproduce the issue
```c
int main()
{
	void* ctx = zmq_ctx_new();
	void* publisherSocket = zmq_socket(ctx, ZMQ_PUB);
	int sendHigh = 0;
	zmq_setsockopt(publisherSocket, ZMQ_SNDHWM, &sendHigh, sizeof(sendHigh));
	
	void* subscriberSocket = zmq_socket(ctx, ZMQ_SUB);
	int receiveHigh = 10;
	zmq_setsockopt(subscriberSocket, ZMQ_RCVHWM, &receiveHigh, sizeof(receiveHigh));	
	
        zmq_bind(publisherSocket, "tcp://localhost:50902");
	zmq_connect(subscriberSocket, "tcp://localhost:50902");
	zmq_setsockopt(subscriberSocket, ZMQ_SUBSCRIBE, "AAAAAA", 6);
	Sleep(5000);
	
	for (int i = 0; i < 1000; i++)
	{
		/* Create a new message, allocating 6 bytes for message content */
		zmq_msg_t msg;
		int rc = zmq_msg_init_size(&msg, 6);
		
		/* Fill in message content with 'AAAAAA' */
		memcpy(zmq_msg_data(&msg), "AAAAAA", 6);
		
		/* Send the message to the socket */
		rc = zmq_msg_send(&msg, publisherSocket, 0);
		printf("Sending message %d\n", i);
		if (rc == -1) {
			printf("%d\n", errno);
			break;
		}
	}

[...]

PR #3041, opened 2018-04-10
Improved Documentation for 'ZMQ_RCVHWM' and 'ZMQ_SNDHWM' #2871

9ee1:
Improved documentation for 'ZMQ_RCVHWM' and 'ZMQ_SNDHWM' options (#2871) to indicate they are logical extensions to one another.

### DIFF
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -551,6 +551,14 @@ blocking or dropping sent messages. Refer to the individual socket descriptions
 in linkzmq:zmq_socket[3] for details on the exact action taken for each socket
 type.
 
+The 'ZMQ_RCVHWM' and 'ZMQ_SNDHWM' options are logical extensions to one
+another. In general, when a socket's receive high water mark has reached its
+limit and enters an exceptional state, 1) the socket will stop picking up
+messages from the wire, 2) the wire buffer will get full and force the sending
+peer to stop sending messages to the socket, and 3) the sending peer will queue
+outbound messages until its send high water mark reaches its limit and the
+sending peer itself enters an exceptional state.
+
 [horizontal]
 Option value type:: int
 Option value unit:: messages
@@ -682,6 +690,14 @@ depending on the socket type, 0MQ shall take appropriate action such as
 blocking or dropping sent messages. Refer to the individual socket descriptions
 in linkzmq:zmq_socket[3] for details on the exact action taken for each socket
 type.
+
+The 'ZMQ_SNDHWM' and 'ZMQ_RCVHWM' options are logical extensions to one
+another. In general, when a receiving peer's receive high water mark has
+reached its limit and enters an exceptional state, 1) the receiving peer will
+stop picking up messages from the wire, 2) the wire buffer will get full and
+force the socket to stop sending messages to the peer, and 3) the socket will
+queue outbound messages until its send high water mark reaches its limit and
+the socket itself enters an an exceptional state.
 
 [horizontal]
 Option value type:: int
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -551,7 +551,7 @@ blocking or dropping sent messages. Refer to the individual socket descriptions
 in linkzmq:zmq_socket[3] for details on the exact action taken for each socket
 type.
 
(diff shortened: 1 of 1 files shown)


doc/zmq_getsockopt.txt
```diff
@@ -551,6 +551,14 @@ blocking or dropping sent messages. Refer to the individual socket descriptions
 in linkzmq:zmq_socket[3] for details on the exact action taken for each socket
 type.
 
+The 'ZMQ_RCVHWM' and 'ZMQ_SNDHWM' options are logical extensions to one
+another. In general, when a socket's receive high water mark has reached its
+limit and enters an exceptional state, 1) the socket will stop picking up
```
sigiesec: I don't think this is correct for all socket types and configurations. A general description would need to refer to "exceptional state" only. But I am unsure if this would lead to an understandable description. The other option were to provide individual explanations for each specific behaviour in exceptional state.

doc/zmq_getsockopt.txt
```diff
@@ -551,6 +551,14 @@ blocking or dropping sent messages. Refer to the individual socket descriptions
 in linkzmq:zmq_socket[3] for details on the exact action taken for each socket
 type.
 
+The 'ZMQ_RCVHWM' and 'ZMQ_SNDHWM' options are logical extensions to one
```
sigiesec: I am not sure what "logical extensions to one another" means. Maybe simply say "are closely related"?

doc/zmq_getsockopt.txt
9ee1: My understanding, and having confirmed through a ticket I had opened and other tickets I referenced, this behavior at least true for the TCP transport and I am fairly confident it would apply for all transports. My understanding is socket types and socket options impact only the actions taken once the high water marks are reached.

The intent here is to make clear that once a receive high watermark is reached, messages are not disregarded by the sender, but rather it "spills over" into the sender's outbound queue until it too is full. With the current description, it is easy to mistake, as you can see in the reference tickets, that messages might be disregarded if the receive high watermark is reached.

doc/zmq_getsockopt.txt
9ee1: "Closely related" might be a better term. I can use that term instead.

doc/zmq_getsockopt.txt
sigiesec: I was not thinking about the transport. However, as you mention it, at least inproc seems to be somewhat special with regard to HWM handling, or at least for the handling of the exact parameter values.

But there are socket types such as router, which will drop messages when the exceptional state is entered. In my understanding, this is not covered by your proposed explanation.

Which socket type did you have in mind when writing this?

Moreover, this description overlaps with the docs under zmq_socket. There should be no redundancy, especially for this topic which is admittedly quite tricky.

9ee1: Dropping messages or blocking when receiving/sending them is **what** action is taken **when** the socket enters the exceptional state. What I am trying to describe here is **how** the socket ends up entering the exceptional state to begin with. **What** happens is specific to the socket type and options like you point out, but that is not what I am trying to expand on here.

The documentation right now says the exceptional state is entered when the high water marks are reached. I am trying to expand on that explanation to make clear that the the high water marks are related.

Right now, it is easy to confuse that if Socket A has 1000 messages as a receive high water mark every message from 1001 onward attempted to be sent to it by Socket B as a connected peer will be dropped until at least one of those 1000 messages are read by the application. This doesn't happen.

Socket B will buffer on its end until either Socket A notifies it "I have room for you to end those messages now" or Socket B itself has no more room to buffer because of its send high water mark and itself enters an exceptional state, where the action is thus specific to its type and options applied to it. This "cooperative" buffering doesn't seem to be specific to a socket type.

I haven't seen this explanation on the zmq_socket docs page either. The explanations there simply describe **what** the action is when the exceptional state is entered. If I am mistaken, let me know what I missed and I can take a look again.

bluca:
what's the status here?

9ee1:
Code review is pending on me until I push another commit with a better explanation. My apologizes, I didn't get a chance to do that yet. I will do so ASAP.

9ee1:
@sigiesec, Yes I closed the PR temporarily because I didn't have a chance to revisit it for some time and I didn't want to leave it lingering. I plan on reopening it as soon as I push a new commit. Thank you for creating the issue to track it. I will reference the new issue when I reopen the PR.

PR #3044, opened 2018-04-13
Implement the wsa_error_no function with switch-case

asafkahlon:


src/err.cpp
```diff
@@ -100,128 +100,112 @@ const char *zmq::wsa_error_no (int no_, const char *wsae_wouldblock_string)
...
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  : "error not defined";
+    switch(no_) {
```
bluca: space before the bracket (see: https://travis-ci.org/zeromq/libzmq/jobs/366119755 )

asafkahlon: Just saw it and commited :)

Issue #3046, opened 2018-04-16
Aborts when iOS abuses EBADF to report a socket has been reclaimed


acsearle:
# Issue description

ZeroMQ intermittently calls abort() on iOS.

iOS abuses EBADF to report that a socket can no longer be used because its resources have been reclaimed while the app was in the background.  Apple documents this behaviour here:

https://developer.apple.com/library/content/technotes/tn2277/_index.html#//apple_ref/doc/uid/DTS40010841-CH1-SUBSECTION3

ZeroMQ asserts on the errno value EBADF in multiple locations, interpreting it as an impossible value indicating an internal error, so iOS apps with open sockets will intermittently crash after waking from background.

# Environment

* libzmq version (commit hash if unreleased): 4.2.3
* OS: iOS 11.2.6

# Minimal test code / Steps to reproduce the issue

Repeatedly sleep and wake an iOS device while running an app with an open ZeroMQ socket.

# What's the actual result? (include assertion message & call stack if applicable)

"Bad file descriptor (src/tcp_connecter.cpp:403)"
(aborts)

# What's the expected result?

On iOS only, treat EBADF like any other "networking problem", and recover by closing and attempting to reconnect.

sigiesec:
There is no libzmq version 4.3.2, you probably meant 4.2.3?

acsearle:
Sorry, corrected.

bluca:
Fixed by https://github.com/zeromq/libzmq/pull/3047

PR #3047, opened 2018-04-17
Allow EBADF on iOS #3046

acsearle:
Problem: Aborts when iOS returns EBADF to indicate a socket was reclaimed while the app slept.

Solution: When compiling for iOS, remove asserts on EBADF.

EBADF behavior documented here: https://developer.apple.com/library/content/technotes/tn2277/_index.html#//apple_ref/doc/uid/DTS40010841-CH1-SUBSECTION3

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -66,6 +66,10 @@
 #include <pgm/pgm.h>
 #endif
 
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#endif
+
 zmq::fd_t zmq::open_socket (int domain_, int type_, int protocol_)
 {
     int rc;
@@ -167,7 +171,11 @@ int zmq::get_peer_ip_address (fd_t sockfd_, std::string &ip_addr_)
     }
 #else
     if (rc == -1) {
+#if !defined(TARGET_OS_IPHONE) || !TARGET_OS_IPHONE
         errno_assert (errno != EBADF && errno != EFAULT && errno != ENOTSOCK);
+#else
+        errno_assert (errno != EFAULT && errno != ENOTSOCK);
+#endif
         return 0;
     }
 #endif
--- a/src/tcp.cpp
+++ b/src/tcp.cpp
@@ -48,6 +48,10 @@
 #include <ioctl.h>
 #endif
 
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#endif
+
 int zmq::tune_tcp_socket (fd_t s_)
 {
     //  Disable Nagle's algorithm. We are doing data batching on 0MQ level,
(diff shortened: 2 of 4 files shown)


sigiesec:
Please fix formatting according to https://travis-ci.org/zeromq/libzmq/jobs/367477422#L644

Unfortunately, there is no CI for iOS. Have you verified that the behaviour is correct now, under the relevant conditions?

src/ip.cpp
```diff
@@ -167,7 +167,11 @@ int zmq::get_peer_ip_address (fd_t sockfd_, std::string &ip_addr_)
     }
 #else
     if (rc == -1) {
+#if !defined(TARGET_IPHONE) || !TARGET_IPHONE
```
bluca: Where does TARGET_IPHONE come from?

acsearle: I believed it to be defined by the toolchain but that may not be true.  It's rather difficult to get the cross compiling on OSX to iOS to work and somebody else wrote the script that sets up the environment to enable it.  Looks like I can rely on the compiler defining __APPLE__ and __MACH__ but I should conditionally on them #include <TargetConditionals.h> to ensure we get TARGET_IPHONE.  Will do.

acsearle:
I have tested on iOS but can only trigger/observe the problem and confirm the fix for the change in tcp_connecter.cpp.  Will retest with the recommended changes above.

src/ip.cpp
sigiesec: If you manage to set up a build script, could you send a PR that adds a travis-ci job building for iOS?

bluca:
any update?

bluca:
@acsearle any news? Would be good to have this fixed

acsearle:
Uh, I think I can commit to sometime this week.

acsearle:
I think this is now testing for iOS targets in best-practice fashion, and it has been tested on iOS to fix crash.

acsearle:
Conflict looks trivial, how do you guys handle these?  Should I rebase and squash my PR?

bluca:
Yes please rebase and squash

sigiesec:
Isn't this ready to merge now? The CI failures are only the known failures.

bluca:
Sorry, I either didn't get the notification of the new push or I missed it

bluca:
@acsearle could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@acsearle  another ping. Please consider helping us with the relicense. Thank you!

PR #3068, opened 2018-04-29
More attempts to speed up Appveyor

bluca:
Small difference, but better than nothing. Compilation time is still the dominant factor, I wonder if the paralle builds are working - it doesn't seem like following the live logs.

@sigiesec do you know if the way to do parallel builds with the generated visual studio solutions by cmake changed with VS2015?
The VS2013 builds take just a couple of minutes, but the 2015 and 2017 one take 4 times as much, which would seem consistent with a sequential vs parallel build with 4 cores

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -494,6 +494,7 @@ endif ()
 if (MSVC)
   # Parallel make.
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /MP")
 
   # Compile the static lib with debug information included
   string (REGEX REPLACE "/Z." "/Z7" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -102,5 +102,5 @@ after_build:
 
 test_script:
   - cmd: cd "%LIBZMQ_BUILDDIR%"
-  - cmd: ctest -C "%Configuration%" -V
+  - cmd: ctest -C "%Configuration%" -V -j5
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -70,16 +70,26 @@ init:
   - msbuild /version
   - cmd: reg add "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp" /v UserAuthentication /t REG_DWORD /d 0 /f
 
+cache:
+    - C:\projects\libsodium
+
 install:
   - cmd: if "%Platform%"=="x64" set "CMAKE_GENERATOR=%CMAKE_GENERATOR% Win64"
   - cmd: echo "Generator='%CMAKE_GENERATOR%'"
   - cmd: echo "Platform='%Platform%'"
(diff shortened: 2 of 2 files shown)


sigiesec:
Unfortunately I am not really familiar with msbuild. @claasd Have you got an idea what changed with VS2015?

@bluca Are you sure the tests can be run in parallel reliably? Aren't there still some tests that use fixed ports?

bluca:
I fixed all the tests, the few fixed ports are unique, everywhere else has been running them in parallel for a while

claasd:
Hi,

I do not know what changed between those two MSBuild Versions. but the build script shows the multicpu switch (/m) on your build (https://ci.appveyor.com/project/zeromq/libzmq/build/build-2222/job/82p4cl7uybc4xid6):
```
msbuild "C:\projects\build_libzmq\ZeroMQ.sln" /m /verbosity:minimal /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
```
But it does in fact build only in one thread. This could have several reasons: the system only has one processor, the variable NUMBER_OF_PROCESSORS is wrong, or your solution has the dependencies in such a way that it cannot build in parallel.

bluca:
@claasd the thing is that the VS2013 build works in parallel just fine:
https://ci.appveyor.com/project/zeromq/libzmq/build/build-2222/job/lh42a742xqxdsdg5

The only difference between those builds are the VS versions parameters:
```
APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2013
CMAKE_GENERATOR=Visual Studio 12 2013
MSVCVERSION=v120
MSVCYEAR=vs2013
```

Hence why I am puzzled - if it was a problem with dependencies or other variables that wouldn't work either right? That's why I suspected something changed in VS/MSVC and    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP") might no longer be enough, or something like that...

sigiesec:
The /MP switch in CMAKE_CXX_FLAGS controls the parallelization within a single compiler run, while the /M or /MAXCPUCOUNT switch of msbuild.exe controls how many projects (?) are built in parallel. Since they are quite a number of independent test projects, it should be possible to build these in parallel. 

Both switches are documented here: https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes

MSBuild is called twice, once for libsodium, with /maxcpucount:%NUMBER_OF_PROCESSORS%, and then for libzmq itself with /m (without explicitly specifying the number of processors).

Maybe some time stamps should be output after each build step so one can see which steps really take

bluca:
msbuild is called explicitly only for libsodium though, appveyor generates the second call, perhaps that's the problem? the parallel appveyor config option is set though

bluca:
https://github.com/appveyor/ci/issues/216

> If you are building on Free plan there are "Small" instances with 1 virtual core, so /m switch doesn't make any difference.

Grrr. Could it be that the 2013 Appveyor instance doesn't have that limit? Perhaps we should switch most of the build to that image...

bluca:
Yep using the 2013 image now we get ~2 minutes build time instead of 8. I've left one 2015 and one 2017 build to test the newer compilers.

sigiesec:
Hm. Strange that the VS2013 image has more processor cores. Hopefully this doesn't change too soon.
The VS2017 job now has several test failures, I am not sure since when they are present. But they aren't related to your current changes.

Is this ready to merge from your point of view now?

bluca:
Yes it is - I think those are random failures

Issue #3069, opened 2018-04-30
Loose validation in address parsing code


simias:
# Issue description

The current TCP address parsing code (as currently implemented in `tcp_address.cpp`) is not very strict and will silently accept bogus address formats, sometimes with weird results that might be tricky to debug. In particular it allows ambiguous `IPv6:port` specifications and it does not validate that the port is actually valid.

# Environment

* libzmq version (commit hash if unreleased): v4.2.5
* OS: any

# Minimal test code / Steps to reproduce the issue

Consider this pyzmq code:

```python
import zmq

context = zmq.Context()

socket = context.socket(zmq.REQ)
socket.setsockopt(zmq.IPV6, 1)
socket.connect("tcp://[fd95:74a4:b46c:1000::1:100b]:5555")
```

This code works fine, the brackets are used to disambiguate between the IPv6 and the port specifier. However while ZMQ allows the brackets it merely ignores them and you can remove them without changing the meaning at all:

```python
socket.connect("tcp://fd95:74a4:b46c:1000::1:100b:5555")
```

It makes it pretty difficult to figure out what's going on for IPv6 address:port specifier, and it also means that if the user forgets to specify the port ZMQ will just assume that the final component of the IPv6 is actually the port specifier:

```python
socket.connect("tcp://fd95:74a4:b46c:1000::1:100b")
```

Here ZMQ will assume that `100b` is the port number, which you might argue is invalid but since the parser uses a simple `atoi` without further validation it'll actually attempt to connect to address `fd95:74a4:b46c:1000::1` on port `100` which doesn't exactly conform to the "principle of least astonishment".

# What's the expected result?

I think ZMQ should:

* Enforce that brackets be present around the hostname when a port specifier is expected *if* the hostname contains the ':' character

* Reject port numbers that contain trailing garbage

It's a very straightforward modification but it might cause regressions for people who rely on the current behavior to parse "weird" URLs.

bluca:
> It's a very straightforward modification but it might cause regressions for people who rely on the current behavior to parse "weird" URLs.

You figured out the problem yourself - so the first change you propose cannot be done, as avoiding regressions in existing, working applications is more important than guarding against silly mistakes.

Feel free to send a PR to fix the second one.

PR #3070, opened 2018-04-30
Problem: address parsing code is tied to the TCP code

simias:
Solution: Factor the code into a different file with a well defined API and add
unit tests.

My Debian doesn't have a recent-enough `clang-format` to run `make clang-format-check` so apologies in advance if I messed up the coding style here and there (which is almost certain given the size of the patch).

I've attempted to match the previous implementation perfectly, including the arguable "quirks" such as the ones I've reported in #3069 (and a few others that I'm not sure about, such as specifying a %scope for an IPv4 address).

This is a pretty huge patch so I tried to be very comprehensive in the unit tests. I think the only path of the code that's not tested is when specifying an interface name as a bind point since the implementations to find the IP address associated with a network interface vary wildly from platform to platform so it's not trivial to come up with generic test code. I've tested manually that it still worked on Linux at least and I haven't made significant modifications to the architecture-specific code so I have good hopes that I haven't irremediably broken it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,6 +83,8 @@ src_libzmq_la_SOURCES = \
 	src/io_thread.hpp \
 	src/ip.cpp \
 	src/ip.hpp \
+	src/ip_resolver.cpp \
+	src/ip_resolver.hpp \
 	src/ipc_address.cpp \
 	src/ipc_address.hpp \
 	src/ipc_connecter.cpp \
@@ -895,7 +897,8 @@ if ENABLE_STATIC
 test_apps += \
 	unittests/unittest_poller \
 	unittests/unittest_ypipe \
-	unittests/unittest_mtrie
+	unittests/unittest_mtrie \
+	unittests/unittest_ip_resolver
 
 unittests_unittest_poller_SOURCES = unittests/unittest_poller.cpp
 unittests_unittest_poller_CPPFLAGS = -I$(top_srcdir)/src ${UNITY_CPPFLAGS} $(CODE_COVERAGE_CPPFLAGS)
@@ -917,6 +920,14 @@ unittests_unittest_mtrie_SOURCES = unittests/unittest_mtrie.cpp
 unittests_unittest_mtrie_CPPFLAGS = -I$(top_srcdir)/src ${UNITY_CPPFLAGS} $(CODE_COVERAGE_CPPFLAGS)
 unittests_unittest_mtrie_CXXFLAGS = $(CODE_COVERAGE_CXXFLAGS)
 unittests_unittest_mtrie_LDADD = $(top_builddir)/src/.libs/libzmq.a \
+	${src_libzmq_la_LIBADD} \
+	${UNITY_LIBS} \
+	$(CODE_COVERAGE_LDFLAGS)
+
+unittests_unittest_ip_resolver_SOURCES = unittests/unittest_ip_resolver.cpp
+unittests_unittest_ip_resolver_CPPFLAGS = -I$(top_srcdir)/src ${UNITY_CPPFLAGS} $(CODE_COVERAGE_CPPFLAGS)
+unittests_unittest_ip_resolver_CXXFLAGS = $(CODE_COVERAGE_CXXFLAGS)
+unittests_unittest_ip_resolver_LDADD = $(top_builddir)/src/.libs/libzmq.a \
 	${src_libzmq_la_LIBADD} \
 	${UNITY_LIBS} \
 	$(CODE_COVERAGE_LDFLAGS)
--- a/src/ip_resolver.hpp
+++ b/src/ip_resolver.hpp
@@ -1,0 +1,102 @@
+/*
+    Copyright (c) 2007-2018 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef __ZMQ_IP_RESOLVER_HPP_INCLUDED__
+#define __ZMQ_IP_RESOLVER_HPP_INCLUDED__
+
+#if !defined ZMQ_HAVE_WINDOWS
+#include <sys/socket.h>
+#include <netinet/in.h>
+#endif
+
+namespace zmq
+{
+union ip_addr_t
+{
+    sockaddr generic;
+    sockaddr_in ipv4;
+    sockaddr_in6 ipv6;
+};
+
+class ip_resolver_options_t
+{
+  public:
+    ip_resolver_options_t ();
+
+    ip_resolver_options_t &bindable (bool bindable_);
+    ip_resolver_options_t &allow_nic_name (bool allow_);
+    ip_resolver_options_t &ipv6 (bool ipv6_);
+    ip_resolver_options_t &expect_port (bool expect_);
+    ip_resolver_options_t &allow_dns (bool allow_);
+
+    bool bindable ();
+    bool allow_nic_name ();
+    bool ipv6 ();
+    bool expect_port ();
+    bool allow_dns ();
+
+  private:
+    bool bindable_wanted;
+    bool nic_name_allowed;
+    bool ipv6_wanted;
+    bool port_expected;
+    bool dns_allowed;
+};
+
+class ip_resolver_t
+{
+  public:
+    ip_resolver_t (ip_resolver_options_t opts_);
+
+    int resolve (ip_addr_t *ip_addr_, const char *name_);
+
+  protected:
+    ip_resolver_options_t options;
+
+    int resolve_nic_name (ip_addr_t *ip_addr_, const char *nic_);
+    int resolve_getaddrinfo (ip_addr_t *ip_addr_, const char *addr_);
+
+#if defined ZMQ_HAVE_WINDOWS
+    int get_interface_name (unsigned long index, char **dest) const;
+    int wchar_to_utf8 (const WCHAR *src, char **dest) const;
+#endif
+
+    //  Virtual functions that are overriden in tests
+    virtual int do_getaddrinfo (const char *node_,
+                                const char *service_,
+                                const struct addrinfo *hints_,
+                                struct addrinfo **res_);
+
+    virtual void do_freeaddrinfo (struct addrinfo *res_);
+
+    virtual unsigned int do_if_nametoindex (const char *ifname_);
+};
+}
+
+#endif
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -5,6 +5,7 @@ set(unittests
   unittest_ypipe
   unittest_poller
   unittest_mtrie
+  unittest_ip_resolver
 )
 
 #IF (ENABLE_DRAFTS)
--- a/unittests/unittest_ip_resolver.cpp
+++ b/unittests/unittest_ip_resolver.cpp
@@ -1,0 +1,882 @@
+/*
+Copyright (c) 2018 Contributors as noted in the AUTHORS file
+
+This file is part of 0MQ.
+
+0MQ is free software; you can redistribute it and/or modify it under
+the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 3 of the License, or
+(at your option) any later version.
+
+0MQ is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <unity.h>
+#include "../tests/testutil.hpp"
+
+#include <ip_resolver.hpp>
+#include <ip.hpp>
+
+void setUp ()
+{
+}
+
+void tearDown ()
+{
+}
+
+class test_ip_resolver_t : public zmq::ip_resolver_t
+{
+  public:
+    test_ip_resolver_t (zmq::ip_resolver_options_t opts_) :
+        ip_resolver_t (opts_)
+    {
+    }
+
+  protected:
+    struct dns_lut_t
+    {
+        const char *hostname;
+        const char *ipv4;
+        const char *ipv6;
+    };
+
+    virtual int do_getaddrinfo (const char *node_,
+                                const char *service_,
+                                const struct addrinfo *hints_,
+                                struct addrinfo **res_)
+    {
+        static const struct dns_lut_t dns_lut[] = {
+          {"ip.zeromq.org", "10.100.0.1", "fdf5:d058:d656::1"},
+          {"ipv4only.zeromq.org", "10.100.0.2", "::ffff:10.100.0.2"},
+          {"ipv6only.zeromq.org", NULL, "fdf5:d058:d656::2"},
+        };
+        unsigned lut_len = sizeof (dns_lut) / sizeof (dns_lut[0]);
+        struct addrinfo ai;
+
+        assert (service_ == NULL);
+
+        bool ipv6 = (hints_->ai_family == AF_INET6);
+        bool no_dns = hints_->ai_flags & AI_NUMERICHOST;
+        const char *ip = NULL;
+
+        if (!no_dns) {
+            for (unsigned i = 0; i < lut_len; i++) {
+                if (strcmp (dns_lut[i].hostname, node_) == 0) {
+                    if (ipv6) {
+                        ip = dns_lut[i].ipv6;
+                    } else {
+                        ip = dns_lut[i].ipv4;
+
+                        if (ip == NULL) {
+                            //  No address associated with NAME
+                            return EAI_NODATA;
+                        }
+                    }
+                }
+            }
+        }
+
+        if (ip == NULL) {
+            //  No entry for 'node_' found in the LUT (or DNS is
+            //  forbidden), assume that it's a numeric IP address
+            ip = node_;
+        }
+
+        //  Call the real getaddrinfo implementation, making sure that it won't
+        //  attempt to resolve using DNS
+        ai = *hints_;
+        ai.ai_flags |= AI_NUMERICHOST;
+
+        return zmq::ip_resolver_t::do_getaddrinfo (ip, NULL, &ai, res_);
+    }
+
+    virtual unsigned int do_if_nametoindex (const char *ifname_)
+    {
+        static const char *dummy_interfaces[] = {
+          "lo0",
+          "eth0",
+          "eth1",
+        };
+        unsigned lut_len =
+          sizeof (dummy_interfaces) / sizeof (dummy_interfaces[0]);
+
+        for (unsigned i = 0; i < lut_len; i++) {
+            if (strcmp (dummy_interfaces[i], ifname_) == 0) {
+                //  The dummy index will be the position in the array + 1 (0 is
+                //  invalid)
+                return i + 1;
+            }
+        }
+
+        //  Not found
+        return 0;
+    }
+};
+
+//  Attempt a resolution and test the results. If 'expected_addr_' is NULL
+//  assume that the resolution is meant to fail.
+//
+//  On windows we can receive an IPv4 address even when an IPv6 is requested, if
+//  we're in this situation then we compare to 'expected_addr_v4_failover_'
+//  instead.
+static void test_resolve (zmq::ip_resolver_options_t opts_,
+                          const char *name_,
+                          const char *expected_addr_,
+                          uint16_t expected_port_ = 0,
+                          uint16_t expected_zone_ = 0,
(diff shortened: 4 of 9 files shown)


bluca:
On Debian 9 you can install the clang format tool by itself from stretch-backports:

```sudo apt install -t stretch-backports clang-format-5.0```

And then pass ```CLANG_FORMAT=clang-format-5.0``` to ```./configure``` to use it

**Comment by bluca** [comment]
@simias this is great work, thank you very much.
You need to add the new source and header in Makefile.am, right at the top, to fix the autotools build.

**Comment by bluca** [comment]
With some versions of g++ also there's another error:

```
/home/travis/build/zeromq/libzmq/unittests/unittest_ip_resolver_dns.cpp: In function ‘void freeaddrinfo(addrinfo*)’:
/home/travis/build/zeromq/libzmq/unittests/unittest_ip_resolver_dns.cpp:125:45: error: declaration of ‘void freeaddrinfo(addrinfo*)’ has a different exception specifier
     void freeaddrinfo (struct addrinfo *res_)
                                             ^
In file included from /home/travis/build/zeromq/libzmq/unittests/unittest_ip_resolver_dns.cpp:26:0:
/usr/include/netdb.h:668:13: error: from previous declaration ‘void freeaddrinfo(addrinfo*) throw ()’
 extern void freeaddrinfo (struct addrinfo *__ai) __THROW;
```

File: Makefile.am [hunk 1]
```diff
@@ -83,6 +83,7 @@ src_libzmq_la_SOURCES = \
 	src/io_thread.hpp \
 	src/ip.cpp \
 	src/ip.hpp \
+	src/ip_resolver.cpp \
```
bluca: @simias the header too

simias:
Ah, that's annoying. I "cheated" by putting the mock implementation in the C++ file in a bit extern "C" to avoid having to mess with how the autotools/cmake link the tests but I suppose that won't work if g++ is capricious. I guess I should put it in an external C file and do it right.

simias:
Do you prefer that I keep squashing my single commit or do you prefer that I add additionals fixup commits for the time being? I'm not too used to code reviews on github so I don't know what's more convenient.

bluca:
Please squash them, given they are fixing the first commit

Have you tried adding the ```__THROW``` to the mock as well?

simias:
Do you think that'll work with all compilers? is __THROW standard? I assumed it was "dialectal" but I'm not exactly up to date with the latest C++ standards.

Makefile.am
```diff
@@ -895,7 +897,9 @@ if ENABLE_STATIC
 test_apps += \
 	unittests/unittest_poller \
 	unittests/unittest_ypipe \
-	unittests/unittest_mtrie
+	unittests/unittest_mtrie \
+	unittests/unittest_ip_resolver \
+	unittests/unittest_ip_resolver_dns
 
 unittests_unittest_poller_SOURCES = unittests/unittest_poller.cpp
```
bluca: @simias for automake you also need an explicit section like these ones, just copy and paste
that's because each test is a program

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,383 @@
...
+    struct in_addr expected_addr;
+
+    resolver
+        .bindable (false)
+        .allow_dns (false)
+        .allow_nic_name (false)
+        .expect_port (false)
+        .ipv6 (false);
+
+    assert (inet_pton (AF_INET, "1.2.128.129", &expected_addr) == 1);
```
bluca: this will break Windows XP, you'll want a function that does something like:

```
#if defined(ZMQ_HAVE_WINDOWS) && (_WIN32_WINNT < 0x0600)
    ip4addr.sin_addr.s_addr = inet_addr ("127.0.0.1");
#else
    inet_pton (AF_INET, "127.0.0.1", &ip4addr.sin_addr);
#endif
```

bluca:
Windows:

```
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(25): fatal error C1083: Cannot open include file: 'sys/socket.h': No such file or directory
C:\projects\libzmq\unittests\unittest_ip_resolver.cpp(33): fatal error C1083: Cannot open include file: 'net/if.h': No such file or director
```

bluca:
The system headers are already included by testutil.hpp with the appropriate ifdefs, add them there if there's extra ones needed

bluca:
Various Linux builds from very old to very new and various architectures all look good:

https://build.opensuse.org/package/show/home:bluca:branches:network:messaging:zeromq:git-stable/libzmq

simias:
I'm not quite sure I understand the Windows build error there, is it a prototype mismatch or something else? I guess I should try to setup a Windows build environment and see for myself.

bluca:
From the looks of it, it would seem there is a prototype mismatch - I have no idea how this translates to, maybe @sigiesec can help:

```
NET_IFINDEX WINAPI if_nametoindex(
  _In_ PCSTR InterfaceName
);
```

https://msdn.microsoft.com/en-us/library/windows/desktop/bb408409(v=vs.85).aspx

bluca:
There's a few more:

```
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(43): error C2373: 'getaddrinfo' : redefinition; different type modifiers [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
          C:\Program Files (x86)\Windows Kits\8.1\Include\um\ws2tcpip.h(108) : see declaration of 'getaddrinfo'
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(43): error C2491: 'getaddrinfo' : definition of dllimport function not allowed [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(49): error C2065: 'EAI_ADDRFAMILY' : undeclared identifier [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(56): warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning) [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(127): error C2373: 'freeaddrinfo' : redefinition; different type modifiers [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
          C:\Program Files (x86)\Windows Kits\8.1\Include\um\ws2tcpip.h(386) : see declaration of 'freeaddrinfo'
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(127): error C2491: 'freeaddrinfo' : definition of dllimport function not allowed [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
```

simias:
Looks like it's not better and now clang is unhappy too. I wonder if there's a better way to do this.

bluca:
I mentioned it on IRC - I'd say go back to the previous version, and just don't build the DNS test on Windows - it's easy to do in cmake, just conditionally append it to the unittests list

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,369 @@
...
+static zmq::ip_addr_t test_bad_addr (void)
+{
+    zmq::ip_addr_t addr;
+
+    memset (&addr, 0xba, sizeof (addr));
+
+    return addr;
+}
+
+static void validade_ipv6_addr (const zmq::ip_addr_t &addr_,
```
sigiesec: This should probably be "validate".

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,369 @@
...
+    int neq = memcmp (&ip6_addr->sin6_addr,
+                      &expected_addr_,
+                      sizeof (expected_addr_));
+
+    TEST_ASSERT_EQUAL (0, neq);
+    TEST_ASSERT_EQUAL (htons (expected_port_), ip6_addr->sin6_port);
+    TEST_ASSERT_EQUAL (expected_zone_, ip6_addr->sin6_scope_id);
+}
+
+static void validade_ipv4_addr (const zmq::ip_addr_t &addr_,
```
sigiesec: See above, "validate"

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,369 @@
...
+                                uint16_t expected_port_ = 0)
+{
+    const sockaddr_in *ip4_addr = &addr_.ipv4;
+
+    TEST_ASSERT_EQUAL (AF_INET, addr_.generic.sa_family);
+    TEST_ASSERT_EQUAL (expected_addr_.s_addr, ip4_addr->sin_addr.s_addr);
+    TEST_ASSERT_EQUAL (htons (expected_port_), ip4_addr->sin_port);
+}
+
+void test_bind_any (int ipv6_)
```
sigiesec: This contains a lot of test cases, and should be split up, instead of chaining test cases, and reconfiguring the resolver in between.

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,369 @@
...
+    TEST_ASSERT_EQUAL (-1, resolver.resolve(&addr, "[1.2.128.129:*]"));
+
+    //  IPv6 should be rejected
+    TEST_ASSERT_EQUAL (-1, resolver.resolve(&addr, "::1"));
+
+    resolver.expect_port (true);
+    TEST_ASSERT_EQUAL (-1, resolver.resolve(&addr, "[::1]:1234"));
+}
+
+void test_parse_ipv6 () {
```
sigiesec: This also contains several test cases.

unittests/unittest_ip_resolver.cpp
```diff
@@ -0,0 +1,369 @@
...
+    TEST_ASSERT_EQUAL (0, resolver.resolve(&addr, "abcd:1234::1:0:234:123"));
+    validade_ipv6_addr (addr, expected_addr, 123);
+
+    resolver.expect_port (false);
+
+}
+
+//  Parsing IPv4 should also work if an IPv6 is requested, it
+//  returns an IPv6 with the IPv4 address embedded
+void test_parse_ipv4_in_ipv6 ()
```
sigiesec: This also contains several test cases.

src/ip_resolver.hpp
```diff
@@ -0,0 +1,77 @@
...
+    sockaddr_in ipv4;
+    sockaddr_in6 ipv6;
+};
+
+class ip_resolver_t
+{
+  public:
+    ip_resolver_t ();
+
+    ip_resolver_t &bindable (bool bindable_);
```
sigiesec: It would greatly reduce complexity if an ip_resolver_t were not reconfigurable. This is not necessary in the production code, and also not in the tests if the test cases are split up.

The class can be split up into a builder ip_resolver_builder_t (which contains the functions from lines 52-56, and an additional build function that creates an ip_resolver_t) and an ip_resolver_t (which contains the remaining functions, and gets a constructor that accepts all configuration parameters, and maybe an interface containing the functions that are platform-specific resp. intended to be mocked in the tests).

src/ip_resolver.hpp
```diff
@@ -0,0 +1,77 @@
...
+    ip_resolver_t &allow_nic_name (bool allow_);
+    ip_resolver_t &ipv6 (bool ipv6_);
+    ip_resolver_t &expect_port (bool expect_);
+    ip_resolver_t &allow_dns (bool allow_);
+
+    int resolve (ip_addr_t *ip_addr_, const char *name_);
+
+  protected:
```
sigiesec: This can and should be made private. No one derives from ip_resolver_t as far as I can see.

src/ip_resolver.hpp
simias: Ok, I'm going to split the builder code and the tests.

For the test "overrides" you prefer adding an API to change the implementation instead of a linked overload? @bluca and you seem to have a slightly different approach here, I'm not sure what to do.

sigiesec: Hm, from the comments here I cannot deduce what @bluca suggested, is it something you discussed on IRC?

I try to make my suggestion a bit more concrete, to ensure we talk about the same thing.
I imagine that ip_resolver_t gets a ctor such as
```
ip_resolver_t (..., i_dns_t);
```

with 
```
class i_dns_t
{
   virtual int getaddrinfo (const char *node_, const char *service_,
                     const struct addrinfo *hints_,
                     struct addrinfo **res_) = 0;
   virtual void freeaddrinfo (struct addrinfo *res_) = 0;

   virtual ~i_dns_t();
};
```
Maybe i_dns_t needs to contain other functions that need to be mock-injected by the tests. Maybe the signature of getaddrinfo can be abstracted somehow, such that it more closely fits the actual use in ip_resolver_t, and is somewhat more C++-style, to avoid the need for a separate freeaddrinfo function in the interface.

I can give it a closer look to make a more specific suggestion on this if you like.

simias:
I tried to split up the tests by "theme", let me know if you want me to split them even more.

sigiesec:
@simias Great :) Personally, I would split them up even more, such that there is only a single call to resolve in each test case. AFAIK, we do not have express guidelines on this for libzmq though, so this is certainly no "show-stopper". However, having them separate has several advantages: Every test case automatically has a name, so there is no need to add comments which might get out of sync. When running the tests e.g. after a production code change, they can fail independently, and one immediately see which ones work and which do not. When having several test cases in one, it will only run until the first sub test case fails, and you do not know anything about the remaining test cases.

simias:
Okay, I'll try to do that.

Do you have any idea what's happening with this test that times out: https://travis-ci.org/zeromq/libzmq/jobs/373845614 ?

unittests/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ set(unittests
   unittest_ypipe
   unittest_poller
   unittest_mtrie
+  unittest_ip_resolver
+  unittest_ip_resolver_dns
```
bluca: @simias move unittest_ip_resolver_dns below, like this:

```
IF (NOT WIN32)
    list(APPEND unittests
        unittest_ip_resolver_dns
    )
ENDIF (NOT WIN32)
```

so that building this is skipped with VS which doesn't like the mock

bluca:
@simias ignore that Travis failure, it has been happening for a while, I need to investigate it but it has nothing to do with your changes

simias:
So one more proposition for your consideration: I implement a getaddrinfo wrapper using virtual calls and inherit a custom test implementation in the tests. Functionally this is equivalent to the "weak wrapper" approach that you didn't like (and didn't work anyway) but without linker magic only using standard C++.

The main advantage is that it should be portable, it's easy to understand (if you grok virtual inheritance that is), the footprint on production code is small and the test code is free to call the original function if it wants which can simplify things a lot (I can cut my mock getaddrinfo implementation to the bare minimum if I can call the real getaddrinfo once I dealt with DNS).

The inconvenient is that it increases code size slightly and adds a small but unavoidable performance hit because of the indirection every time the function is called in production code.

bluca:
Windows is still having none of it:

```
C:\projects\libzmq\unittests\unittest_ip_resolver.cpp(41): error C2556: 'unsigned int if_nametoindex(const char *)' : overloaded function differs only by return type from 'NET_IFINDEX if_nametoindex(PCSTR)' [C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
          C:\Program Files (x86)\Windows Kits\8.1\Include\shared\netioapi.h(2526) : see declaration of 'if_nametoindex'
C:\projects\libzmq\unittests\unittest_ip_resolver.cpp(41): error C2373: 'if_nametoindex' : redefinition; different type modifiers [C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
          C:\Program Files (x86)\Windows Kits\8.1\Include\shared\netioapi.h(2526) : see declaration of 'if_nametoindex'
C:\projects\libzmq\unittests\unittest_ip_resolver_dns.cpp(55): error C2065: 'EAI_ADDRFAMILY' : undeclared identifier [C:\projects\build_libzmq\unittests\unittest_ip_resolver_dns.vcxproj]
```

Performance is not an issue here as this is setup code, and when doing DNS resolution the overhead of a function call is really not even a blip on the radar.

bluca:
perhaps the mock needs to be ifdefed and have a different prototype on win?

simias:
@bluca I haven't modified the if_nametoindex code, I just did the PoC on getaddrinfo to see if you were fine with this approach, if so I'll clean it up and resubmit it, if not and I won't have written pointless code. I also have an idea to split the tests into individual unitary functions without having an overwhelming amount of boilerplate that I want to try.

bluca:
Ah ok. IMHO it's fine to leave if_name like it is, and just ifdef the mock signature, should be much quicker to do, but up to you.

sigiesec:
@bluca The signature of if_nametoindex is different on Windows (the return type is NET_IFINDEX), yes, but that's not the point. You just cannot mock a function defined in a library (be it static or shared) like this on Windows.

bluca:
Great, looks good - except now windows has something else to complain about:

```
[C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
unity.lib(unity.obj) : error LNK2019: unresolved external symbol _setUp referenced in function _UnityDefaultTestRun [C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
unity.lib(unity.obj) : error LNK2019: unresolved external symbol _tearDown referenced in function _UnityDefaultTestRun [C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
C:\projects\build_libzmq\bin\Release\unittest_ip_resolver.exe : fatal error LNK1120: 2 unresolved externals [C:\projects\build_libzmq\unittests\unittest_ip_resolver.vcxproj]
```

sigiesec:
Just add 
```
void setUp ()
{
}

void tearDown ()
{
}
```
to unittest_ip_resolver.cpp

This is also some issue with missing weak linking support. Probably this could be overcome more elegantly, but this is the solution that is currently used in some other tests.

simias:
Sorry for the multiple pushes, I messed up.

sigiesec:
Hopefully this is only a minor remaining issue, and the tests not only build but also run successfully under Windows :) If not, I can give it a closer look again.

Apart from that: The PR looks really good to me now :) Great job, it's an important step in increasing unit test coverage in libzmq!

bluca:
> Hopefully this is only a minor remaining issue, and the tests not only build but also run successfully under Windows :) If not, I can give it a closer look again.

They build now which is great, unfortunately there are a few test failures now :-/

> Apart from that: The PR looks really good to me now :) Great job, it's an important step in increasing unit test coverage in libzmq!

Ditto - thanks for this great work

simias:
Looks like the calls to getaddrinfo all fail for some reason. That's very odd.

simias:
And it seems to work on Travis. Very strange indeed. Even `test_parse_ipv4_simple` fails on Windows and that's just:

```C++
test_resolve (resolver_opts, "1.2.128.129", "1.2.128.129");
```

My guess is either that I have some subtle memory corruption issue in the code that breaks the code or there's something different on Windows with regards to this API. At any rate I can't seem to reproduce it on Linux, even by tweaking the compilation flags.

sigiesec:
@simias I try running it on Windows on my local machine.

sigiesec:
@simias zmq::initialize_network must be called on Windows, see https://github.com/sigiesec/libzmq/commit/159534544b838989ba066633fed85398fd65e60c (this has quite a lot of changes since your file was not formatted with clang-format, but essentially only the calls to initialize_network and shutdown_network were added).

Now, only 4 test cases fail, which seem to be related to ipv6 handling, first failure is https://ci.appveyor.com/project/sigiesec/libzmq/build/build-488/job/abgf6r84i26q2t3h#L1458

sigiesec:
Ah, this is due to this piece of code:
https://github.com/sigiesec/libzmq/blob/159534544b838989ba066633fed85398fd65e60c/src/ip_resolver.cpp#L279

simias:
>this has quite a lot of changes since your file was not formatted with clang-format,

About that: "make clang-format-diff" does nothing on my machine for a reason I don't quite understand, it seemed suspicious given the amount of changes I've made but I thought I was just very lucky. I should've known better:

```
Formatting with clang-format (using clang-format) and showing differences with latest commit
Built target clang-format-diff
```

Regarding the new test error it seems to be that windows returns an AF_INET instead of an embedded IPV6 when asked to parse a numeric IPv4 IP:

```
test_parse_ipv4_in_ipv6:FAIL: Expected 23 Was 2
```

simias:
Ah, good catch. I was trying to figure out what was wrong with address mapping on Windows, I should've read the code. I guess I will tweak the testing code to work in this scenario.

simias:
Looks like window's getaddrinfo accepts brackets around IPv6 when Linux doesn't. Oh well. I'll just remove that test then, it's useless.

simias:
Since clang-format still refuses to do its job on my end I fear you'll have to do it yourself (or let me know how to fix it).

simias:
Actually I could tweak the parsing code so that brackets are accepted on Linux as well. I just need to add a copy of the bracket stripping code after the scope parsing code. This way you won't have the issue of some URLs being accepted on windows but not on linux. I can make a PR for that later if you want.

sigiesec:
> About that: "make clang-format-diff" does nothing on my machine for a reason I don't quite understand, it seemed suspicious given the amount of changes I've made but I thought I was just very lucky. I should've known better:

Hm, this should probably produce an error instead.

> Regarding the new test error it seems to be that windows returns an AF_INET instead of an embedded IPV6 when asked to parse a numeric IPv4 IP:

Yes, as I wrote in my comment before this is explicitly done on Windows at this point:
https://github.com/sigiesec/libzmq/blob/159534544b838989ba066633fed85398fd65e60c/src/ip_resolver.cpp#L279

simias:
Yeah I read the comment when I was porting the code but I didn't realize that meant that the code would behave differently.

Anyway, only one build fails on appveyor now and it seems unrelated to my patch. Is this a known problem or did I break something?

https://ci.appveyor.com/project/zeromq/libzmq/build/build-2249/job/uaei3wy47y8p7ai7

```
68: test_curve_security_with_valid_credentials (large routing id)
70/72 Test #68: test_security_curve ..............   Passed    5.70 sec
71/72 Test #15: test_immediate ...................***Timeout  30.01 sec
72/72 Test #44: test_many_sockets ................***Timeout  30.52 sec
97% tests passed, 2 tests failed out of 72
Total Test time (real) =  39.12 sec
The following tests FAILED:
	 15 - test_immediate (Timeout)
	 44 - test_many_sockets (Timeout)
Errors while running CTest
```

sigiesec:
@simias test_immediate and test_many_sockets are unfortunately failing for quite some time, this is not related to your changes!

sigiesec:
From my point of view, this is now ready to merge.
@bluca Do you want to have a final look?

bluca:
LGTM!

@simias thanks again. Given the amount of changes, please send another PR with 2 commits: one to add yourself to AUTHORS, and another one with a relicensing grant, see https://github.com/zeromq/libzmq/tree/master/RELICENSE for details

simias:
Will do, thank you for your help @bluca and @sigiesec !

PR #3075, opened 2018-05-02
Udp parser

simias:
This code contains two independent albeit related commits:

* The first adds unit tests for the `udp_address` code

* The 2nd replaces the current ad hoc parser with the new `ip_resolver` code

Let me know if you prefer to have two different PRs.

The new resolver code should not break any existing URLs, however it will allow things that didn't work previously such as connecting hostnames (`udp://zeromq.org:5555`), binding NIC names (`udp://eth0:1234`) and wildcard ports (`udp://127.0.0.1:*`).

### DIFF
--- a/unittests/unittest_udp_address.cpp
+++ b/unittests/unittest_udp_address.cpp
@@ -89,14 +89,12 @@ static void test_resolve_ipv4_bind ()
 
 static void test_resolve_ipv4_bind_any ()
 {
-    //  Wildcard port not supported
-    test_resolve ("*:*", NULL, 0, "0.0.0.0");
+    test_resolve ("*:*", "0.0.0.0", 0, "0.0.0.0");
 }
 
 static void test_resolve_ipv4_bind_anyport ()
 {
-    //  Wildcard port not supported
-    test_resolve ("127.0.0.1:*", NULL, 0, "127.0.0.1");
+    test_resolve ("127.0.0.1:*", "127.0.0.1", 0, "127.0.0.1");
 }
 
 static void test_resolve_ipv4_bind_any_port ()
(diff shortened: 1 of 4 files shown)


unittests/unittest_udp_address.cpp
```diff
@@ -0,0 +1,158 @@
...
+                          const char *bind_addr_ = NULL,
+                          bool multicast_ = false)
+{
+    zmq::udp_address_t addr;
+    bool bound = bind_addr_ != NULL;
+
+    int rc = addr.resolve (name_, bound);
+
+    if (dest_addr_ == NULL) {
+        TEST_ASSERT_EQUAL (-1, rc);
```
sigiesec: Maybe check for a specific errno here?

unittests/unittest_udp_address.cpp
```diff
@@ -111,7 +109,9 @@ static void test_resolve_ipv4_connect_any ()
 
 static void test_resolve_ipv4_connect_anyport ()
 {
-    test_resolve ("127.0.0.1:*", NULL);
+    //  Not sure if that's useful but the TCP code allows it so why
```
sigiesec: Hm. Does this have a defined behaviour on any platform? If not, it should be disallowed IMO.
The same applies to TCP.

simias:
@bluca if resolution can potentially happen much later maybe it's not a good idea to remove this code? Otherwise it will change the current behavior where you get an early error in case of an invalid URL. The port test could and probably should be added to the ip_resolver anyway, for good measure.

bluca:
Right, I thought by moving you mean refactoring into a function that is still called from the same place so that it can be reused.
In that case please leave it there, as an initial sanity check, however limited and flawed it can be, is useful.

simias:
Currently the UDP code doesn't delay the resolution (should it?) so there wouldn't be anything to reuse really. I'm about to make a PR to add the test to ip_resolver anyway, as it probably makes sense to have it here at any rate.

bluca:
> Currently the UDP code doesn't delay the resolution (should it?)

I think it should, yes, to get feature parity with TCP - but don't consider that blocking for the stuff you are working on. Maybe we can open an issue.

Issue #2212, opened 2016-11-15
zmq_udp: need ability to specify bind address separately from multicast address for multi-homed hosts


Bklyn:
The `ZMP_RADIO` and `ZMP_DISH` socket types look like great addition to ZMQ.  Thanks!

When receiving multicast data on multi-homed hosts, some users will require the ability to select which interface is used for a given multicast stream, as the data may not transit on the default interface.  

I'd suggest a syntax perhaps like `udp://mcaddr[@bindaddr]:port`  (where `[]` indicates the `@bindaddr` is optional not that it appears in square brackets)

So, in other words, this code on lines 114-116 from `udp_socket.cpp` would need to change:

```
    109     //  If a should bind and not a multicast, the dest address
    110     //  is actually the bind address
    111     if (bind_ && !is_mutlicast)
    112         bind_address = dest_address;
    113     else {
    114         bind_address.sin_family = AF_INET;
    115         bind_address.sin_port = htons (port);
    116         bind_address.sin_addr.s_addr = htons (INADDR_ANY);
    117     }
```
Also, I believe all of these `htons` should be `htonl` (all from udp_socket.cpp):
```
83:        dest_address.sin_addr.s_addr = htons (INADDR_ANY);
103:    iface.s_addr = htons (INADDR_ANY);
116:        bind_address.sin_addr.s_addr = htons (INADDR_ANY);
```

When configuring many hosts it is also useful to be able to specify the bind address by its interface name or using address/mask (i.e. 192.168.2.0/24).  These would require the use of additional OS-specific APIs (i.e. `getifaddrs(3)`), but this is very powerful and will make configuration much more manageable.

bluca:
Looks like a good idea, just a note: TCP sockets already support passing an address to bind to with the format: ```tcp://127.0.0.1:5564;127.0.0.1:5562```

This sounds similar, right? What about using the same format for consistency?

Bklyn:
I was not aware of that syntax.  Consistency makes perfect sense.

I think ultimately the code in `tcp_address::resolve_interface` and its delegate `resolve_nic_name` should be refactored into a separate class so they can be re-used in `udp_address`.  Then any additional magic like ADDR/MASK syntax can be the province of (I think) `resolve_nic_name`.

bluca:
I would suggest to do the same as the other multicast transport do w.r.t. default binding, so that there are the least surprises

simias:
Agreed, I need to figure out what PGM does.

[...]

PR #3078, opened 2018-05-04
Mcast bind

simias:
Rework the UDP code to factor as much as possible into 'ip_resolver' while adding support for an explicit bind address for multicast URLs (fixes issue #2212).

This also lays the groundwork for IPv6 support by replacing the hardcoded IPv4 handling code by the v4/v6 agnostic code from ip_resolver.


src/ip_resolver.cpp
```diff
@@ -20,6 +20,73 @@
...
+}
+
+bool zmq::ip_addr_t::is_multicast () const
+{
+    if (family () == AF_INET) {
+        uint32_t addr = ntohl (ipv4.sin_addr.s_addr);
+
+        //  IPv4 Multicast: address MSBs are 1110
+        //  Range: 224.0.0.0 - 239.255.255.255
+        return (addr & 0xf0000000) == 0xe0000000;
```
bluca: the standard library has functions and macros for this:

IN_IS_ADDR_MULTICAST
IN6_IS_ADDR_MULTICAST

If they work on Windows, perhaps it's best to use them rather than re-implement?

src/ip_resolver.cpp
```diff
@@ -20,6 +20,73 @@
...
+
+    if (family == AF_INET) {
+        sockaddr_in *ip4_addr = &addr.ipv4;
+        memset (ip4_addr, 0, sizeof (*ip4_addr));
+        ip4_addr->sin_family = AF_INET;
+        ip4_addr->sin_addr.s_addr = htonl (INADDR_ANY);
+    } else if (family == AF_INET6) {
+        sockaddr_in6 *ip6_addr = &addr.ipv6;
+
+        memset (ip6_addr, 0, sizeof (*ip6_addr));
```
bluca: Standard library has in6addr_any that can be assigned to sin6_addr, again if it works on Windows

bluca: ah never mind it's 2 lines below, duh

src/ip_resolver.cpp
```diff
@@ -20,6 +20,73 @@
...
+    if (family () == AF_INET6) {
+        return ntohs (ipv6.sin6_port);
+    } else {
+        return ntohs (ipv4.sin_port);
+    }
+}
+
+void zmq::ip_addr_t::set_port (uint16_t port)
+{
+    if (family () == AF_INET6) {
```
bluca: Valgrind reports ```Conditional jump or move depends on uninitialised value(s)```

```
==26101== Conditional jump or move depends on uninitialised value(s)
==26101==    at 0x4E68492: zmq::ip_addr_t::set_port(unsigned short) (ip_resolver.cpp:55)
==26101==    by 0x4EB1AA7: zmq::udp_address_t::resolve(char const*, bool) (udp_address.cpp:123)
==26101==    by 0x4E966F3: zmq::socket_base_t::bind(char const*) (socket_base.cpp:537)
==26101==    by 0x4EBEC03: zmq_bind (zmq.cpp:295)
==26101==    by 0x403430: test_radio_dish_udp() (test_radio_dish.cpp:207)
==26101==    by 0x4055AA: UnityDefaultTestRun (unity.c:1339)
==26101==    by 0x403574: main (test_radio_dish.cpp:232)
```

src/ip_resolver.cpp
simias: I considered it but I didn't know how portable it was and given that it was a one-liner I didn't risk it. I'll try the macros and we'll see if the buildbot is happy I suppose.

bluca: We use another one of those in CZMQ which builds more or less on the same platforms, so that's why I mentioned it:

https://github.com/zeromq/czmq/blob/master/src/zbeacon.c#L192

src/ip_resolver.cpp
simias: This is code I lifted directly from the original TCP resolver implementation:

https://github.com/zeromq/libzmq/blob/d062edd8c142384792955796329baf1e5a3377cd/src/tcp_address.cpp#L417-L427

I don't know if there's a good reason for not using the standard constants in this case. I can try I guess.

bluca: nah it's fine as is, in6addr_any is the standard constant

simias: So much for that:
```
../src/ip_resolver.cpp: In static member function ‘static zmq::ip_addr_t zmq::ip_addr_t::any(int)’:
../src/ip_resolver.cpp:73:73: error: extended initializer lists only available with -std=c++0x or -std=gnu++0x [-Werror]
../src/ip_resolver.cpp:73:73: error: extended initializer lists only available with -std=c++0x or -std=gnu++0x [-Werror]
```

bluca: ok, testing a build on a bunch of distros&archs here: https://build.opensuse.org/package/show/home:bluca:branches:network:messaging:zeromq:git-draft/libzmq

bluca: never mind then! thanks for trying

Issue #2309, opened 2017-01-19
Compiling for windows universal app failed: Error	C3861 'CoCreateInstanceFromApp': identifier not found


yuanhuihe:
I have compiled the newest zmq 4.2.1 with cmake:

> cmake -H. -Bbuild -G"Visual Studio 14 2015 Win64" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0

And then open the 'ZeroMQ.sln' in 'build' folder, build it with debug/x64/libzmq(Universal Windows), it complains:

`Severity	Code	Description	Project	File	Line	Suppression State
Error	C3861	'CoCreateInstanceFromApp': identifier not found	libzmq	C:\Program Files (x86)\Windows Kits\10\Include\10.0.14393.0\um\combaseapi.h	1189	
`

Anyone faced this issue?

bbdb68:
FYI I just managed to build an UWP version of the dll a load it from a Unity/UWP project. 
It is far to be PR ready but I will try to improve my hack.

basically you need to disasble security featreus (curve, tweetnacl)
I also defined WIN32_LEAN_AND_MEAN to get rid of COM stull.

then there is only a few hotsptos to fix.

I'll try to clean this an propose a PR

bluca:
Thanks, but why do the security need to be disabled? Look forward to see the PR

bbdb68:
for now this is the only way I found to get libzmq compile, I am discovering UWP...

bluca:
Ah I understand. Thanks for working on this!

bbdb68:
@bluca , how could I add a #define such as ZMQ_HAVE_WINDOWS_UWP so that I can isolate my changes ?

bluca:
Does the uwp stuff have its own build system? If so you can simply define it in there and nowhere else

bbdb68:
ok I have a first (very crude) version running on uwp, I was even able to perform a rep req from an uwp unity client using the uwp zeromq dll as a plugin  to a regular server. (I did not check if the bind can work but winsock is said to be available...)

you can check the changes here
https://github.com/bbdb68/libzmq/tree/uwp
I had to disable curve for some link issues, and look for ZMQ_HAVE_WINDOWS_UWP for finding the changes I had to do. This is far from being PR ready, any advice / tip welcome to re-enable the features I had to disable.

hyhrepo:
Great job @bbdb68 !
[...]

PR #3081, opened 2018-05-07
UDP Ipv6 support

simias:
Far from complete but I could use some feedback

### DIFF
--- a/src/udp_address.cpp
+++ b/src/udp_address.cpp
@@ -41,28 +41,26 @@
 #include <sys/types.h>
 #include <arpa/inet.h>
 #include <netdb.h>
+#include <net/if.h>
 #include <ctype.h>
 #endif
 
-#include "ip_resolver.hpp"
-
-zmq::udp_address_t::udp_address_t () : is_multicast (false)
+zmq::udp_address_t::udp_address_t () : bind_interface (-1), is_multicast (false)
 {
-    memset (&bind_address, 0, sizeof bind_address);
-    memset (&dest_address, 0, sizeof dest_address);
+    bind_address = ip_addr_t::any (AF_INET);
+    target_address = ip_addr_t::any (AF_INET);
 }
 
 zmq::udp_address_t::~udp_address_t ()
 {
 }
 
-int zmq::udp_address_t::resolve (const char *name_, bool bind_)
+int zmq::udp_address_t::resolve (const char *name_, bool bind_, bool ipv6_)
 {
     //  No IPv6 support yet
-    int family = AF_INET;
-    bool ipv6 = family == AF_INET6;
     bool has_interface = false;
-    ip_addr_t interface_addr;
+
+    address = name_;
 
     //  If we have a semicolon then we should have an interface specifier in the
     //  URL
@@ -79,22 +77,36 @@ int zmq::udp_address_t::resolve (const char *name_, bool bind_)
           //  indeterminate socktype.
           .allow_dns (false)
           .allow_nic_name (true)
-          .ipv6 (ipv6)
+          .ipv6 (ipv6_)
           .expect_port (false);
 
         ip_resolver_t src_resolver (src_resolver_opts);
 
-        const int rc =
-          src_resolver.resolve (&interface_addr, src_name.c_str ());
+        const int rc = src_resolver.resolve (&bind_address, src_name.c_str ());
 
         if (rc != 0) {
             return -1;
         }
 
-        if (interface_addr.is_multicast ()) {
+        if (bind_address.is_multicast ()) {
             //  It doesn't make sense to have a multicast address as a source
             errno = EINVAL;
             return -1;
+        }
+
+        //  This is a hack because we need the interface index when binding
+        //  multicast IPv6, we can't do it by address. Unfortunately for the
+        //  time being we don't have a generic platform-independent function to
+        //  resolve an interface index from an address, so we only support it
+        //  when an actual interface name is provided.
+        if (src_name == "*") {
+            bind_interface = 0;
+        } else {
+            bind_interface = if_nametoindex (src_name.c_str ());
+            if (bind_interface == 0) {
+                //  Error, probably not an interface name.
+                bind_interface = -1;
+            }
         }
 
         has_interface = true;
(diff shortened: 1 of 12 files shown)


src/udp_address.cpp
```diff
@@ -129,88 +123,55 @@ int zmq::udp_address_t::resolve (const char *name_, bool bind_)
...
-unsigned short zmq::udp_address_t::family () const
-#else
-sa_family_t zmq::udp_address_t::family () const
-#endif
+int zmq::udp_address_t::to_string (std::string &addr_)
 {
-    return AF_INET;
+    // XXX what do (factor TCP code?)
+    addr_ = address;
+    return 0;
```
simias: What exactly are the requirements for this `to_string` method? The current implementation just returns the string passed to the resolver but the TCP implementation actually bothers re-generating the string from the sockaddr values.

src/udp_engine.cpp
```diff
@@ -111,9 +111,11 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
...
         if (!options.raw_socket) {
-            out_address = address->resolved.udp_addr->dest_addr ();
-            out_addrlen = address->resolved.udp_addr->dest_addrlen ();
+            const ip_addr_t *out = address->resolved.udp_addr->target_addr ();
+            out_address = out->as_sockaddr ();
+            out_addrlen = out->sockaddr_len ();
         } else {
+            /// XXX fixme ?
             out_address = (sockaddr *) &raw_address;
             out_addrlen = sizeof (sockaddr_in);
```
simias: I don't quite understand how this `raw_address` part fits in, I see that the address is parsed within this very file in `out_event`, where does the string come from exactly? Why doesn't it use the `udp_address` parser?

src/udp_engine.cpp
```diff
@@ -145,11 +149,29 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
...
+                mreq.imr_multiaddr = mcast_addr->ipv4.sin_addr;
+                mreq.imr_interface = bind_addr->ipv4.sin_addr;
+
+                rc = setsockopt (fd, IPPROTO_IP, IP_ADD_MEMBERSHIP,
+                                 (char *) &mreq, sizeof (mreq));
+            } else if (mcast_addr->family () == AF_INET6) {
+                struct ipv6_mreq mreq;
+
+                mreq.ipv6mr_multiaddr = mcast_addr->ipv6.sin6_addr;
+                mreq.ipv6mr_interface = 0; // XXX fixme?
```
simias: So here I'm having a bit of a conundrum: unlike the IPv4 code `struct ipv6_mreq` doesn't let you specify the IP address of the interface to bind but rather the interface number. It's trivial to lookup if we have the interface name using `if_nametoindex` but of course at this point we don't have the interface name anymore and on top of that ZMQ URLs lets you specify an interface by IP: `udp://[abcd::1];[ff00::1]:5555`.

The easy solution would be to only support specifying a source interface by giving the interface name but that's obviously limiting and not very consistent. The hard solution would be to have a method to look up an interface index based on an IP (pretty much the opposite of `zmq::ip_resolver_t::resolve_nic_name`). Of course the tedious part is that this code is very platform-specific.

I think the latter solution is the way to go but I'd like to have some feedback on that.

unittests/unittest_resolver_common.hpp
```diff
@@ -0,0 +1,54 @@
+#ifndef __UNITTEST_RESOLVER_COMMON_INCLUDED__
```
bluca: a copyright header is needed here

src/udp_engine.cpp
bluca: If you want to start by only supporting the interface name, and then later on add support for the IP, that's fine given this socket type is a work in progress

src/udp_engine.cpp
bluca: not sure about the address, but thw raw_socket option is used to read/write UDP datagrams without any ZMTP headers

simias:
I cleaned up the code, added tests. IPv6 multicast works but only if you bind `*` or an interface name, binding by interface IP returns `ENODEV`. I haven't touched the raw address code so it still only supports IPv4, I suspect that it wouldn't be too hard to modify to use `ip_resolver` instead but it would probably be a good idea to add some tests first.

I want to add multicast tests but I'd need to add a socket option to allow looping the multicast packets on the same host using `setsockopt (IP[V6]_MULTICAST_LOOP)`. I think this ought to be added as a `zmq_setsockopt` parameter but I'll leave that for a future PR. Similarly it might be worth adding support for the `IP_MULTICAST_TTL` option.

simias:
I don't understand why some travis builds don't find my header in `unittests`, it works fine on my machine. I tried replacing `#include "unittest_resolver_common.hpp"` with `#include "../unittests/unittest_resolver_common.hpp"` to match the `testutil.hpp` include but it's still broken.

bluca:
Doesn't the header have to be added to the executable via CMakeLists.txt?

bluca:
The CI is already slow enough, I don't think we should remove the parallel builds - it's easy enough to do in a temporary commit when debugging is needed

bluca:
Like this:

https://github.com/zeromq/libzmq/blob/master/tests/CMakeLists.txt#L185

simias:
But why does it work when I build locally using cmake?

Also the unnittests CMakeLists.txt doesn't mention any include file, including `testutil.hpp`, should it?

https://github.com/zeromq/libzmq/blob/7aba6821ac56a15706896618ef35d4c9df5032a1/unittests/CMakeLists.txt#L29

src/udp_address.cpp
bluca: I have a feeling there was something related to disconnect/unbind - check the git log, there might be some clues

simias:
Uh, the plot thickens:

```
CMake Error at unittests/CMakeLists.txt:29 (add_executable):
  Cannot find source file:
    unittest_resolver_common.hpp
  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
```

Does it mean that the buildbot hasn't checked-out that file for some reason?

bluca:
gah - it's because the CI tests uses the distributable tarball. You need to add the new header to automake, then it will work

simias:
Looks better. If you don't have any objections maybe that's enough changes for this PR and the rest of the work (IPv6 raw sockets, to_string, LOOP support etc...) can be left for later.

simias:
Thanks

src/udp_address.cpp
```diff
@@ -79,24 +77,38 @@ int zmq::udp_address_t::resolve (const char *name_, bool bind_)
...
 
+        //  This is a hack because we need the interface index when binding
+        //  multicast IPv6, we can't do it by address. Unfortunately for the
+        //  time being we don't have a generic platform-independent function to
+        //  resolve an interface index from an address, so we only support it
+        //  when an actual interface name is provided.
+        if (src_name == "*") {
+            bind_interface = 0;
+        } else {
+            bind_interface = if_nametoindex (src_name.c_str ());
```
AntonioGMuriana: This breaks again compilation for Universal Windows App.
Reopen issues #2309

AntonioGMuriana: This breaks again compilation for Universal Windows App.
Reopen issue #2309

PR #3083, opened 2018-05-07
acinclude.m4: check if -latomic is needed.

asafkahlon:
On some cases, -latomic is needed for likning, and since the current
acinclude.m4 checks only compilation we can sometimes miss the need for -latomic
and the linking process will fail.
Therefore, the AC_CHECK_IFELSE was replaced with AC_LINK_IFELSE. If the first
try fails, we try to link again with -latomic and add LIBS="-latmoic" in case we
succeeded.

Signed-off-by: Asaf Kahlon <asafka7@gmail.com>


bluca:
@asafkahlon could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

bluca:
@asafkahlon  another ping. Please consider helping us with the relicense. Thank you!

Issue #3061, opened 2018-04-28
Problem: ZMTP 3.1 ping/pong context not implemented


bluca:
ZMTP 3.1 adds PING/PONG commands as heartbeat options that libzmq implements. The PING commands can add a context following the TTL (an arbitrary payload) which the PONG is supposed to echo back, but libzmq does not implement this.

> To invoke a single heartbeat, the peer MAY, at any point after the security handshake is complete, send a PING command:
> 
> ping = command-size %d4 "PING" ping-ttl ping-context
> ping-ttl = 2OCTET
> ping-context = *OCTET
>
> The ping-ttl is a 16-bit unsigned integer in network order that MAY be zero or MAY contain a time-to-live measured in tenths of seconds. The ping-ttl provides a strong hint to the other peer to disconnect if no further traffic is received after that time. The maximum TTL is 6553 seconds.
> 
> When a peer receives a PING command it SHALL respond with a PONG command that echoes the ping-context, which may be empty:
> 
> pong = command-size %d4 "PONG" ping-context
> ping-context = *OCTET

https://rfc.zeromq.org/spec:37/ZMTP

eponsko:
What's the purpose of echoing the TTL back? One thing I thought could be neat is to use it to provide an estimate of the latency between the peers, to use as a weight in or alternative to the RR behavior.

bluca:
It's not the TTL that's echoed back, it's an arbitrary content. I'm working on that, but I'll probably change the RFC to limit it - at the moment there's no size limit which might open it to abuse.

PR #3089, opened 2018-05-12
[WIP] Problem: ZMTP 3.1 PING Context not implemented

bluca:
Solution: add it

Fixes: https://github.com/zeromq/libzmq/issues/3061

@sigiesec do you have time to give me a hand with Windows, please?

I really don't understand what the problem is - sometimes, there is no data coming from the raw tcp read in the mocked test. Any clue?

### DIFF
--- a/src/stream_engine.cpp
+++ b/src/stream_engine.cpp
@@ -98,6 +98,8 @@ zmq::stream_engine_t::stream_engine_t (fd_t fd_,
 {
     int rc = tx_msg.init ();
     errno_assert (rc == 0);
+    rc = pong_msg.init ();
+    errno_assert (rc == 0);
 
     //  Put the socket into non-blocking mode.
     unblock_socket (s);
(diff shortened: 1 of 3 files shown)


sigiesec:
I will give it a look today... Provided my train happens to have a working internet connection ;)

sigiesec:
I think this is simply a timing issue. If the mock peer takes too long, the other peer has already closed the connection, which is a valid behaviour. By increasing the ZMQ_HEARTBEAT_IVL sufficiently, this should go away.

Note: in my branch https://github.com/sigiesec/libzmq/tree/ping-context I migrated the tests to unity.

bluca:
But in the mock test there is no heartbeat - and that's the one failing? I've also tried to have it loop forever, and it just gets stuck and never receives any data

sigiesec:
What do you mean by "in the mock test there is no heartbeat"? In the failing test case (with mock_ping == true), heartbeats are activated:
https://github.com/sigiesec/libzmq/blob/74fb2f433a74eb6dbef5f52ac3a10d03ce9ef90d/tests/test_heartbeats.cpp#L226

sigiesec:
Right. Hm. Will check again.

sigiesec:
@bluca I opened a new PR #3097 from my branch, which also contains the test framework migration. Ok to close this one? I also have found some trace of how this fails if it fails (but not yet why).

PR #3094, opened 2018-05-13
Problem: no way to specify source interface for UDP multicast sender …

simias:
…sockets

Solution: if a binding interface is provided in the sender URL we pass
it to IP[V6]_MULTICAST_IF setsockopt


simias:
I always forget about that

PR #3096, opened 2018-05-14
Add poller docs

### LINKED ISSUE #2188: Problem: no documentation for new zmq_poller_* APIs
It's still in DRAFT state, but it would be good to have at least an initial draft of the documentation.

@minrk @somdoron ping :-)

sigiesec:
Added initial documentation for zmq_poller_* functions.

Fixes #2188 

@bluca, @somdoron, @kurdybacha Maybe you can have a look?

### DIFF
--- a/doc/zmq_poller.txt
+++ b/doc/zmq_poller.txt
@@ -1,0 +1,269 @@
+zmq_poller(3)
+===========
+
+
+NAME
+----
+zmq_poller - input/output multiplexing
+
+
+SYNOPSIS
+--------
+
+*void *zmq_poller_new (void);*
+*int zmq_poller_destroy (void **'poller_p');*
+
+*int zmq_poller_add (void *'poller', void *'socket', void *'user_data', short 'events');
+*int zmq_poller_modify (void *'poller', void *'socket', short 'events');*
+*int zmq_poller_remove (void *'poller', void *'socket');*
+
+*int zmq_poller_add_fd (void *'poller', int 'fd', void *'user_data', short 'events');*
+*int zmq_poller_modify_fd (void *'poller', int 'fd', short 'events');*
+*int zmq_poller_remove_fd (void *'poller', int 'fd');*
+
+*int zmq_poller_wait_all (void *'poller',
+                          zmq_poller_event_t *'events',
+                          int 'n_events',
+                          long 'timeout');*
+
+DESCRIPTION
+-----------
+The _zmq_poller_*_ functions provide a mechanism for applications to multiplex
+input/output events in a level-triggered fashion over a set of sockets.
+
+_zmq_poller_new_ and _zmq_poller_destroy_ manage the lifetime of a poller
+instance. _zmq_poller_new_ creates and returns a new poller instance, while
+_zmq_poller_destroy_ destroys it. A pointer to a valid poller must be passed
+as the _poller_p_ argument of _zmq_poller_destroy_. In particular,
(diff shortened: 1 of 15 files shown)


doc/zmq_poller.txt
```diff
@@ -0,0 +1,274 @@
...
+
+All functions that return an int, return -1 in case of a failure. In that case,
+zmq_errno() can be used to query the type of the error as described below.
+
+_zmq_poller_wait_all_ returns the number of events signalled and returned in 
+the events array, 0 indicates a timeout.
+
+All other functions return 0 in case of a successful execution.
+
+Upon successful completion, the _zmq_poll()_ function shall return the number
```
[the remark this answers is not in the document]
sigiesec: Yes ;) I started from zmq_poll docs and this was left over.

sigiesec:
@bluca Is this correct? I tried to add symlinks with the script from here: https://coderwall.com/p/z86txw/make-symlink-on-windows-in-a-git-repo, but I am not sure if this worked correctly.

bluca:
it didn't, as the symlink contains the directory name, as if the symlink was placed in the root directory - it should just have the filename, since they are in the same dir

doc/zmq_poller.txt
```diff
@@ -0,0 +1,269 @@
...
+sockets registered with a poller.
+
+_zmq_poller_add_ registers a new _socket_ with a given _poller_. Both _poller_
+and _socket_ must point to valid 0MQ objects. The _events_ parameter specifies
+which event types the client wants to subscribe to. It is legal to specify no
+events (i.e. 0), and activate them later with _zmq_poller_modify_.
+In addition, _user_data_ may be specified, which is not used by the poller, but
+passed back to the caller when an event was signalled in a call to
+_zmq_poller_wait_ or _zmq_poller_wait_all_. _user_data_ may be NULL. If it is
+not NULL, it must be a valid pointer. Otherwise, behaviour is undefined.
```
kurdybacha: Just minor suggestion and feel free to ignore. Is it not up to the user to decide the content of a pointer and if this is undefined behavior or not? For example someone might decide to use it for passing index values etc..

doc/zmq_poller.txt
```diff
@@ -0,0 +1,269 @@
...
+registered objects. Otherwise, a livelock situation may result: If more than 
+'n_events' registered objects have an active event on each call to 
+_zmq_poller_wait_all_, it might happen that the same subset of registered 
+objects is always returned, and the caller never notices the events on the 
+others.
+
+_zmq_poller_wait_all_ returns the number of valid elements. The valid elements
+are placed in positions '0' to 'n_events - 1' in the 'events' array. All
+members of a valid element are set to valid values by _zmq_poller_wait_all_.
+The client does therefore not need to initialize the contents of the events
```
kurdybacha: :+1:

doc/zmq_poller.txt
sigiesec: That's the motivation for the remark. Someone might have this idea, but this explains that they should drop it ;) The argument type is a pointer, and you cannot legally use a pointer to transport a non-pointer.

One might argue that this remark is superfluous as this is illegal already due to type aliasing rules, but since it might be a typical idea of misuse, it is good to have it here.

sigiesec:
Removed the [WiP] after two reviews. I think this could be merged now as initial documentation, and maybe improved later on.

PR #3097, opened 2018-05-14
ZMTP 3.1 PING Context not implemented

### LINKED ISSUE #3061: Problem: ZMTP 3.1 ping/pong context not implemented

sigiesec:
Solution: if a PING message contains a context, echo it back in the
PONG message. In order to do so, create the PONG message when PING
is received and store it in the engine.
After the PING the engine goes straight to encoding and sending, so
there can always be at most one pending PING.
Add tests for various contexts.

Fixes #3061

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -651,7 +651,8 @@ tests_test_setsockopt_SOURCES = tests/test_setsockopt.cpp
 tests_test_setsockopt_LDADD = src/libzmq.la
 
 tests_test_heartbeats_SOURCES = tests/test_heartbeats.cpp
-tests_test_heartbeats_LDADD = src/libzmq.la
+tests_test_heartbeats_LDADD = src/libzmq.la ${UNITY_LIBS}
+tests_test_heartbeats_CPPFLAGS = ${UNITY_CPPFLAGS}
 
 tests_test_stream_exceeds_buffer_SOURCES = tests/test_stream_exceeds_buffer.cpp
 tests_test_stream_exceeds_buffer_LDADD = src/libzmq.la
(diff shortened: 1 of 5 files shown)


sigiesec:
The proper solution probably would be for the mock peer to wait for receiving the READY command before sending its PING.

bluca:
Ah I see! Want to add a commit to move the receives around?

sigiesec:
Will do :)

sigiesec:
Ok, let's wait for the CI jobs :)

sigiesec:
Looks good now... Some tests are failing, but not test_heartbeats. The failing ones are the usuals flaky/failing candidates :(

bluca:
Thanks for the help!

PR #3100, opened 2018-05-14
Problem: zmq_poller_[add/modify] accept invalid events arguments silently

### LINKED ISSUE #3088: zmq_poller_add/zmq_poller_modify should reject invalid events arguments

# Issue description

zmq_poller_add/zmq_poller_modify should reject invalid events arguments, rather than accepting them silently.

# Environment

* libzmq version (commit hash if unreleased): 4.2.5
* OS: any

```
void *ctx = zmq_ctx_new();
void *socket = zmq_socket(ctx, ZMQ_DEALER);
void *poller = zmq_poller_new();
int res = zmq_poller_add(poller, socket, nullptr, 42);
```

`res == 0`

# What's the expected result?

`res == -1 && zmq_errno() == EINVAL`

sigiesec:
Solution: check and return an error on invalid arguments. Fixes #3088

### DIFF
--- a/src/fd.hpp
+++ b/src/fd.hpp
@@ -38,6 +38,9 @@ namespace zmq
 {
 #ifdef ZMQ_HAVE_WINDOWS
 #if defined _MSC_VER && _MSC_VER <= 1400
+///< \todo zmq.h uses SOCKET unconditionally, so probably VS versions before
+/// VS2008 are unsupported anyway. Apart from that, this seems to depend on
+/// the Windows SDK version rather than the VS version.
 typedef UINT_PTR fd_t;
 enum
 {
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -1187,17 +1187,53 @@ int zmq_poller_destroy (void **poller_p_)
     return 0;
 }
 
-int zmq_poller_add (void *poller_, void *s_, void *user_data_, short events_)
+static int check_poller (void *const poller_)
 {
     if (!poller_ || !((zmq::socket_poller_t *) poller_)->check_tag ()) {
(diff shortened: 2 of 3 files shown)


include/zmq.h
```diff
@@ -463,14 +463,16 @@ ZMQ_EXPORT int zmq_socket_monitor (void *s, const char *addr, int events);
 #define ZMQ_POLLERR 4
 #define ZMQ_POLLPRI 8
 
-typedef struct zmq_pollitem_t
-{
-    void *socket;
 #if defined _WIN32
-    SOCKET fd;
+typedef SOCKET zmq_fd_t;
```
bluca: I'm not sure I like the idea of adding a new typedef to the ABI, especially changing existing ones... it's fine for internal usage ofc

sigiesec: Why should a typedef change the ABI?

bluca: I meant changing existing definitions - with all the broken compilers out there (*cough*oracle*cough*) I'd really rather avoid the risk of changing existing structs when there's no overwhelming benefits

sigiesec:
The test failure in the `BUILD_TYPE=default CURVE=libsodium DRAFT=enabled GSSAPI=enabled PGM=enabled NORM=enabled TIPC=enabled` build is strange. The tests that fail are deterministic. I will investigate that.

include/zmq.h
sigiesec: Hm. Ok. I will revert the changes to zmq.h

sigiesec:
Ah, I think I got it. Strange that the other test runs are successful, errno was not set at all.

PR #3111, opened 2018-05-15
Reenable DRAFT builds on Appveyor and fix test_radio_dish_udp_ipv6

### LINKED ISSUE #3110: test_radio_dish broken under Windows
# Issue description

test_radio_dish is broken under Windows. This was not visible in CI because Appveyor CI jobs do not build DRAFT API as of d81a041 (Maybe because of 9aa957b445d2cb53e5218cf7d4afdf7b521bdb2a, maybe already longer). There were syntax errors, which are fixed in  f832c0b, but after fixing them test execution fails. Maybe this is due to #3094, #3087 or #3081?

# Environment

* libzmq version (commit hash if unreleased):   f832c0b (branch)
* OS: Windows

1.  build with ENABLE_DRAFTS=ON
2.  run test_radio_dish

```
73: Assertion failed: Bad address [10014] (C:\projects\libzmq\src\udp_engine.cpp:436)
```
see
https://ci.appveyor.com/project/sigiesec/libzmq/build/build-542/job/0a4g6ek0r8hlb84n#L1645

# What's the expected result?

Successful test execution

sigiesec:
Fixes #3110


sigiesec:
This is blocked by #3110

sigiesec:
Looks good now. 

However, can someone please check if the assertion in https://github.com/zeromq/libzmq/pull/3111/files#diff-0fa1dc262fc2d43db20ec48c0b8c270fR462 is appropriate? @bluca @simias?

bluca:
I think the RAW support (ZMQ_DGRAM) was still WIP for v6, so AFAIK it should be fine

bluca:
Maybe it would be better to have half draft builds, half non-draft? To avoid missing errors

sigiesec:
Maybe a single non-draft build should be added.

We only have builds for Visual Studio 2013, 2015 and 2017. Are the earlier versions unsupported?

sigiesec:
> Appveyor is already quite slow, I'd rather just switch one of the existing builds tbh...

Hm. Ok, I think we could switch the x86 Debug job, e.g.

> No idea - can Appveyor do anything else?

Yes, Appveyor also has Visual Studio 2008, 2010 and 2012:
https://www.appveyor.com/docs/build-environment/#pre-installed-software

I think we should add at least one build for each version that is supported. Even if it further slows the CI down.

We could configure Appveyor with fast_failure, to stop a job after the first build failure.

A premium account with 2 instead of 1 concurrent jobs would be ca. 500 US$ per year.

PR #3116, opened 2018-05-18
Fix some more code style issues

sigiesec:
See individual commits

### DIFF
--- a/src/blob.hpp
+++ b/src/blob.hpp
@@ -75,7 +75,7 @@ struct blob_t
 
     //  Creates a blob_t of a given size, with uninitialized content.
     explicit blob_t (const size_t size) :
-        data_ ((unsigned char *) malloc (size)),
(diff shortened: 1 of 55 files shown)


bluca:
I don't think I have seen this fail before:

```
Expected monitor event 400 but received 200
lt-test_monitor: ../tests/testutil_security.hpp:407: void expect_monitor_event(void*, int): Assertion `event == expected_event' failed.
```

Could you please double check?

sigiesec:
@bluca Where did you see this?

sigiesec:
Ah, got it, here: https://travis-ci.org/zeromq/libzmq/jobs/380740484#L1587

sigiesec:
Hm, 200 would be ZMQ_EVENT_DISCONNECTED, and there are already some checks in test_monitor that handle this as optional on the server side, but not on the client side. I think this optional handling is only missing on the client side.

PR #3117, opened 2018-05-19
prevent duplicate connections from PUB sockets also (see https://gith…

WallStProg:
This pull request adds doc changes and a test program to go along with https://github.com/zeromq/libzmq/pull/2879


bluca:
1) please don't have +X flag to the txt file
2) pleae check why this breaks the CMake build:

```
CMake Error at tests/CMakeLists.txt:186 (add_executable):
  Cannot find source file:
    test_mult_connect.cpp
  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
CMake Error: CMake can not determine linker language for target: test_mult_connect
CMake Error: Cannot determine link language for target "test_mult_connect".
```

https://travis-ci.org/zeromq/libzmq/jobs/381149258

bluca:
Also please add the test to Makefile.am

WallStProg:
Fixed permissions on zmq_connect.txt

Also added test_mult_connect to Makefile.am -- looks like that was breaking cmake build because cmake build is untarring result of autotools build?

I'll check the build, but pls let me know if there are still problems.

tests/test_mult_connect.cpp
```diff
@@ -0,0 +1,109 @@
...
+*/
+
+/*
+   This program tests that multiple connections to the same endpoint are (silently) ignored
+   when SUB binds and PUB connects, as per https://github.com/zeromq/libzmq/pull/2879
+
+*/
+
+#include <zmq.h>
+#include <assert.h>
```
bluca: These includes break the Windows build as you can see in the CI, remove them and leave only testutil to fix it

WallStProg: Sorry -- new to this ...

bluca:
Have you tried running the test when building with draft apis? It seems to fail in that case

bluca:
Not sure, sorry - maybe try with ```msleep (SETTLE_TIME);``` after the connects, or with inproc instead of tcp

Also please fix the formatting: https://travis-ci.org/zeromq/libzmq/jobs/381640804

WallStProg:
FWIW, I tried running the test in a loop, and it hung on iteration 248 (this on my CentOS 6 machine).

It looks like the workaround for SUB bind/PUB connect is not entirely reliable.  That may be an issue in itself, so in the meantime it may be best to revert this change as  I'm not going to be able to deal with this for a few days.  

Let me know what you'd like me to do, if anything.  I can always submit another PR once we've got this sorted out.

Thanks!

bluca:
In that case, I would recommend to remove the test from this PR and just send the DOC update, then if you have time to look again at the test we can take it from there.

WallStProg:
Done -- pls let me know if anything else.  

Sorry for the mess :-(

Makefile.am
```diff
@@ -644,6 +644,9 @@ tests_test_sodium_LDADD = src/libzmq.la
 tests_test_socket_null_SOURCES = tests/test_socket_null.cpp
 tests_test_socket_null_LDADD = src/libzmq.la
 
+tests_test_mult_connect_SOURCES = tests/test_mult_connect.cpp
```
bluca: this change got left in

WallStProg:
Oops -- sorry, fixed.

PR #3129, opened 2018-05-23
Problem: prepared Visual Studio solution files are hard to maintain and redundant with CMake builds

sigiesec:
Solution: declare them as deprecated


bluca:
@sigiesec unfortunately this breaks all the autoconf builds, as a file is not available anymore:

```
[  163s] Makefile.am:5: error: required directory ./builds/msvc does not exist
[  163s] parallel-tests: installing 'config/test-driver'
[  163s] autoreconf: automake failed with exit status: 1
```

sigiesec:
Oops. Will fix that tomorrow

PR #3135, opened 2018-05-25
Fix ZMQ_HEARTBEAT_TTL maximum value check

### LINKED ISSUE #3136: Cannot set ZMQ_HEARTBEAT_TTL to more than 655.3 seconds
# Issue description

The maximum value for ZMQ_HEARTBEAT_TTL is not correctly checked.

# Environment

* libzmq version (commit hash if unreleased): 50374bf
* OS: any

```
int value = 655400;
zmq_setsockopt (socket, ZMQ_HEARTBEAT_TTL, &value, sizeof (value))
```

-1 / EINVAL

# What's the expected result?

0

sigiesec:
Fixes https://sonarcloud.io/project/issues?id=zeromq-libzmq&issues=AWOS_NBc_Tl7jhO8m4MM&open=AWOS_NBc_Tl7jhO8m4MM en passant

Fixes #3136

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -282,6 +282,8 @@ int zmq::options_t::set_curve_key (uint8_t *destination,
     return -1;
 }
 
+const int deciseconds_per_millisecond = 100;
+
 int zmq::options_t::setsockopt (int option_,
                                 const void *optval_,
                                 size_t optvallen_)
@@ -665,8 +667,8 @@ int zmq::options_t::setsockopt (int option_,
 
         case ZMQ_HEARTBEAT_TTL:
             // Convert this to deciseconds from milliseconds
-            value = value / 100;
-            if (is_int && value >= 0 && value <= 6553) {
+            value = value / deciseconds_per_millisecond;
+            if (is_int && value >= 0 && value <= UINT16_MAX) {
                 heartbeat_ttl = static_cast<uint16_t> (value);
                 return 0;
             }
--- a/tests/test_heartbeats.cpp
+++ b/tests/test_heartbeats.cpp
@@ -29,6 +29,14 @@ typedef SOCKET raw_socket;
 typedef int raw_socket;
 #endif
 
+#include <limits.h>
+
+// TODO remove this here, either ensure that UINT16_MAX is always properly
+// defined or handle this at a more central location
+#ifndef UINT16_MAX
+#define UINT16_MAX 65535
+#endif
+
 #include "testutil_unity.hpp"
 
 void setUp ()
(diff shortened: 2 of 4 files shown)


src/options.cpp
```diff
@@ -665,8 +667,8 @@ int zmq::options_t::setsockopt (int option_,
 
         case ZMQ_HEARTBEAT_TTL:
             // Convert this to deciseconds from milliseconds
-            value = value / 100;
-            if (is_int && value >= 0 && value <= 6553) {
```
bluca: I'm not sure this was a typo? Right now the max TTL is one hour and a half, this would make it 18 hours which is a bit much

sigiesec: Ok... that's really in the spec... But it is a very strange value, isn't it? ;)

Anyway, I will change it back to the previous value, but it makes sense to add the tests, so I keep the PR open and fix.

bluca: It's very very strange :-)

The spec is in DRAFT form, so feel free to amend it to something less strange if you want to

sigiesec: Oh, I checked it again... and now I understand, but it is even a bit weirder:

First, the spec says:
> The ping-ttl is a 16-bit unsigned integer in network order that MAY be zero or MAY contain a time-to-live *measured in tenths of seconds*

This would allow values up to UINT16_MAX==65535 deciseconds, as my PR implements.

Then it says:
> The maximum TTL is 6553 seconds

This would allow values up to 65530 deciseconds

So UINT16_MAX is almost correct according to the spec, in fact it were UINT16_MAX - UINT16_MAX % 10.

However, I think this second sentence is not meant to restrict the value, but simply explain the consequence of having a 16 bit unsigned integer specifying deciseconds.

Maybe the wording in the spec should be changed to:
> As a consequence, the maximum TTL is 6553.5 seconds

bluca: right, that makes even more sense, feel free to send a pr to the rfc and I'll merge that too

sigiesec: Ok, I will also add this to doc/zmq_setsockopt.txt

sigiesec: Done

tests/test_heartbeats.cpp
```diff
@@ -366,6 +374,49 @@ DEFINE_TESTS (pull, push, ZMQ_PULL, ZMQ_PUSH)
 DEFINE_TESTS (sub, pub, ZMQ_SUB, ZMQ_PUB)
 DEFINE_TESTS (pair, pair, ZMQ_PAIR, ZMQ_PAIR)
 
+const int deciseconds_per_millisecond = 100;
+// TODO the maximum value is not documented, should it?
```
bluca: the TODO can go now?

sigiesec: Right, removed it

Issue #3123, opened 2018-05-23
Test failures on Appveyor Visual Studio 2017 image


sigiesec:
# Issue description

The following test cases are (almost?) always failing on the Visual Studio 2017 build on Appveyor, but are (almost?) always fine on all Visual Studio 2013/2015 builds:
```
	  8 - test_hwm_pubsub (Failed)
	 47 - test_xpub_nodrop (Failed)
	 50 - test_sockopt_hwm (Failed)
```
See, e.g. https://ci.appveyor.com/project/zeromq/libzmq/build/build-2398/job/xo0f0of2bnn5amei

There is one test case failing in each of these test programs:
```
8: c:\projects\libzmq\tests\test_hwm_pubsub.cpp:75:test_defaults_1000:FAIL: Expected 1000 Was 0
47: c:\projects\libzmq\tests\test_xpub_nodrop.cpp:87:test:FAIL: Expected 1999 Was 0
50: c:\projects\libzmq\tests\test_sockopt_hwm.cpp:133:test_decrease_when_full:FAIL: Expected 0 to be greater than 10
```

It is interesting to note that all of these tests use specific SNDHWM/RCVHWM settings.

Apart from the Visual Studio version, the image is different from the others in that it uses Windows Server 2016 rather than Windows Server 2012 R2.

However, I couldn't easily reproduce this with Visual Studio 2017 on another Windows Server 2016 machine.

# Environment

* libzmq version (commit hash if unreleased): several master versions for an unknown time
* OS: Windows Server 2016

# Minimal test code / Steps to reproduce the issue

Run the Appveyor CI build.

# What's the actual result? (include assertion message & call stack if applicable)

The Visual Studio 2017 job is failing with the errors mentioned above.

# What's the expected result?

The Visual Studio 2017 is good.

sigiesec:
It seems that the test failures occur only when the tests are run in parallel.

One possible explanation might be that the test_many_sockets test, which takes very long on Windows, consumes some global resources and therefore interferes with the other tests.

PR #3137, opened 2018-05-25
Problem: VS2008/2010/2017 tests fail if run in parallel

sigiesec:
Solution: disabled parallel testing for these VS versions

Related to #3123 but it does not fix it, it must be investigated why the tests fail when run in parallel.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,20 +31,12 @@ environment:
       ENABLE_CURVE: ON
     - platform: Win32
       configuration: Release
+      POLLER: epoll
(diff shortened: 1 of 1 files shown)


bluca:
Could you please double check that the config switch is working? It looks like all jobs are ran sequentially, test runtime is back to 110 seconds from 25 everywhere

sigiesec:
Will give it a look. 
I think we could restrict some builds to be run only on master, what do you think?

bluca:
Yeah that sounds good as well, thanks

PR #3142, opened 2018-05-29
Remove duplication between stream_t and router_t

sigiesec:
- [x] Please have a look at the comment in stream_t::xpipe_terminated, resolution would further reduce duplication

### DIFF
--- a/src/stream.cpp
+++ b/src/stream.cpp
@@ -53,7 +53,7 @@ zmq::stream_t::stream_t (class ctx_t *parent_, uint32_t tid_, int sid_) :
 
 zmq::stream_t::~stream_t ()
 {
-    zmq_assert (_outpipes.empty ());
+    zmq_assert (_out_pipes.empty ());
     _prefetched_routing_id.close ();
     _prefetched_msg.close ();
 }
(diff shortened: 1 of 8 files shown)


src/stream.cpp
```diff
@@ -70,10 +69,10 @@ void zmq::stream_t::xattach_pipe (pipe_t *pipe_, bool subscribe_to_all_)
 
 void zmq::stream_t::xpipe_terminated (pipe_t *pipe_)
 {
-    outpipes_t::iterator it = _outpipes.find (pipe_->get_routing_id ());
-    zmq_assert (it != _outpipes.end ());
-    _outpipes.erase (it);
+    erase_out_pipe (pipe_);
     _fq.pipe_terminated (pipe_);
+    // TODO router_t calls pipe_->rollback() here; should this be done here as
```
bluca: the rollback was added here: https://github.com/zeromq/libzmq/pull/2577
I don't think there should be cases where the stream socket has unfinished messages left in the pipe, but it should be harmless to call it.
On the other hand I'm not that confident, given the very small amount of tests for the stream socket...

sigiesec: Ok, then maybe we leave it as it is, until more stream tests are added.

PR #3143, opened 2018-05-30
Problem: temporary files in repo

sigiesec:
Solution: remove them


pavel-pimenov:
Add *.tmp *.cpp.orig .gitignore ?

PR #3145, opened 2018-05-30
More code style fixes and improvements to static analysis configuration

sigiesec:
See individual commits

Ca. 56 clang-tidy warnings remaining

### DIFF
--- a/src/encoder.hpp
+++ b/src/encoder.hpp
@@ -54,14 +54,14 @@ namespace zmq
 template <typename T> class encoder_base_t : public i_encoder
 {
   public:
-    inline encoder_base_t (size_t bufsize_) :
+    inline explicit encoder_base_t (size_t bufsize_) :
         _write_pos (0),
         _to_write (0),
         _next (NULL),
         _new_msg_flag (false),
         _buf_size (bufsize_),
         _buf (static_cast<unsigned char *> (malloc (bufsize_))),
-        in_progress (NULL)
+        _in_progress (NULL)
     {
         alloc_assert (_buf);
     }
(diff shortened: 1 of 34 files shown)


src/encoder.hpp
```diff
@@ -152,6 +152,8 @@ template <typename T> class encoder_base_t : public i_encoder
         _new_msg_flag = new_msg_flag_;
     }
 
+    msg_t *in_progress () { return _in_progress; }
```
bluca: should this be marked with "inline" to give the compiler an hint? given it will be called many times per message

sigiesec: Member functions defined inside a class definition are implicitly inline.

Do you think of some implementation-specific forceinline attribute? I suspect that won't be necessary here since the function is trivial.

bluca: It was just to be sure the compiler gets the idea

sigiesec: AFAIK the inline declaration has no effect on whether a function is inlined at all, at least for MSVC, gcc and clang. As described on the cppreference, the keyword name is misleading, as controlling inlining was the original intention, but this is no longer true,now it only controls whether a funimay have multiple definitions. 
Msvc, gccand clang consider any function as inlinable as long as they know there definition. (With link time code generation resp. link time optimization, even if the compiler does not know the definition).

Nevertheless, I am fine if we mark all header-defined functions as inline by convention. Should we?

sigiesec: Here, the inline declaration is indeed implicit, as also stated in the gcc documentation:
> As required by ISO C++, GCC considers member functions defined within the body of a class to be marked inline even if they are not explicitly declared with the inline keyword. 

However, I wrongly assumed (not relevant for this case) that gcc by default consider any function for inlining when optimizations are enabled. This behaviour would need to be explicitly requested:
> You can also direct GCC to try to integrate all “simple enough” functions into their callers with the option -finline-functions.

(both quotes from: https://gcc.gnu.org/onlinedocs/gcc/Inline.html)

PR #3151, opened 2018-06-02
fix computation of winsock fd_set size

pijyoi:
Problem: optimized_fd_set_t computes required bytes wrongly

Solution: fix it
- vector type change from char to SOCKET
- take into account alignment padding bytes in struct fd_set


sigiesec:
Oops, thanks for fixing this. We are probably missing some test case :( Can you apply the formatting fixes: https://travis-ci.org/zeromq/libzmq/jobs/387046195#L654 ?
Thanks :)

Issue #3031, opened 2018-03-28
Add mingw CI jobs on Appveyor


sigiesec:
Currently, there are no mingw CI jobs on Appveyor, but it should be supported. To ensure that support does not break unnoticed, CI jobs should be added.

sigiesec:
Fixed by #3195

PR #3153, opened 2018-06-03
Problem: not able to use wepoll in configure script

pijyoi:
Solution: add wepoll as a polling system

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -257,6 +257,11 @@ src_libzmq_la_SOURCES = \
 	src/zap_client.hpp \
 	src/zmq_draft.h
 
+if USE_WEPOLL
+src_libzmq_la_SOURCES += \
+	external/wepoll/wepoll.c
+endif
+
 if USE_TWEETNACL
 src_libzmq_la_SOURCES += \
 	src/tweetnacl.c \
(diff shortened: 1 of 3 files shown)


sigiesec:
I wasn't aware that the autotoola build can be used on Windows. With what toolset are you using it? Should we have a CI job for that maybe?

pijyoi:
I was cross compiling from Fedora using distro provided mingw64 tool chain.

It's also possible to use the configure script in Windows under msys
environment. Although my preference then would be to use cmake.

sigiesec:
Ah ok... That makes sense. I was not thinking about MinGW when adding wepoll. 

We should probably have a CI job for MinGW/MSys, at least using either AutoTools or CMake, maybe both, or declare the other unsupported. I made up a ticket some time ago: https://github.com/zeromq/libzmq/issues/3031
If you wanted to do that, to ensure it doesn't break, that would be great :)

I started work on doing the same for Cygwin64: https://github.com/sigiesec/libzmq/tree/add-cygwin64-ci, but did not finish that yet. Some tests are failing, and I need to investigate why.

Makefile.am
```diff
@@ -257,6 +257,11 @@ src_libzmq_la_SOURCES = \
 	src/zap_client.hpp \
 	src/zmq_draft.h
 
+if USE_WEPOLL
+src_libzmq_la_SOURCES += \
+	external/wepoll/wepoll.c
```
bluca: the header needs to be included too, or it won't be part of the dist tarball

bluca:
If it needs to be used with autotools, the other files in the wepoll directory (especially the license) need to be distributed too (EXTRA_DIST in Makefile.am)

PR #3156, opened 2018-06-04
Add cygwin64 CI on Appveyor

### LINKED ISSUE #3030: Add cygwin CI jobs on Appveyor
Currently, there are no cygwin CI jobs on Appveyor, but it should be supported. To ensure that support does not break unnoticed, CI jobs should be added.

sigiesec:
Note: some tests are currently failing, which need to be fixed in a separate effort. these are ignored for the overall result of the appveyor

Fixes #3030

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,7 +227,11 @@ include (CMakeDependentOption)
 include (CheckCXXSymbolExists)
 include (CheckSymbolExists)
 
-check_include_files (windows.h ZMQ_HAVE_WINDOWS)
+if (NOT CYGWIN)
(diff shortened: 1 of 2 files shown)


bluca:
A lot of IPC tests are failing - is AF_UNIX supposed to work?

sigiesec:
Apparently not at the moment, but I am no Cygwin expert. Maybe it is better to disable it at the moment.

sigiesec:
Hm, I had a look at the implementation. There is an implementation of AF_LOCAL/AF_UNIX, but it is hard for to judge its state. socketpair, e.g., does something similar as what zmq::mk_fdpair does on Win32, i.e. open a TCP loopback connection.

There is also ongoing support to add support being based on the native Windows 10 AF_UNIX sockets.

PR #2944, opened 2018-02-13
Problem: insufficient unit tests for poller concept and ypipe

sigiesec:
Solution: add some unit tests, fix resulting test failures, add documentation on poller concept

Resolve several minor issues including formatting, test timeouts, test output, ...

PR #3159, opened 2018-06-07
Problem: epoll_t employs an unnecessary mutex

sigiesec:
Solution: remove the mutex

Note: the mutex is unnecessary as all uses must come from the same thread anyway, and this is already checked by check_thread


bluca:
This was added by: https://github.com/zeromq/libzmq/pull/2239
which was not very clear - but I assume it fixed a problem for the submitter

sigiesec:
I don't know if there was a problem at the time when it was introduced, but it is a relic.

None of the other poller_t implementations have such a mutex, but they all have a comparable handling of retired fds.

There were some race conditions w.r.t. the stopping bool member, which were resolved.

By #2944, it is now both documented and by the calls to check_thread also checked at runtime that the data structure is not used from multiple threads. An assertion would be triggered otherwise.

PR #3168, opened 2018-06-16
Problem: can't process ZMTP 3.1 cancel/subscribe commands

bluca:
There's a bit of a hack to manage to use the existing flag field in msg - it works since a msg can't be more than one command at any time.

This is just to process on receive - to send, it's a bit more complicated so I'll send separately


src/msg.cpp
```diff
@@ -444,6 +444,43 @@ bool zmq::msg_t::is_leave () const
...
+    if (this->is_ping () || this->is_pong ())
+        data =
+          static_cast<unsigned char *> (this->data ()) + ping_cmd_name_size;
+    if (this->is_subscribe ())
+        data = static_cast<unsigned char *> (this->data ()) + sub_cmd_name_size;
+    if (this->is_cancel ())
+        data =
+          static_cast<unsigned char *> (this->data ()) + cancel_cmd_name_size;
+
+    return static_cast<void *> (data);
```
sigiesec: I think this cast is redundant.

src/msg.hpp
```diff
@@ -39,6 +39,9 @@
 #include "atomic_counter.hpp"
 #include "metadata.hpp"
 
+//  bits 2-5
+#define CMD_TYPE_MASK 28
```
sigiesec: maybe define this as a hex literal instead to make the relationship to the bits more obvious?

In c++ style, a constant would be preferable to a preprocessor definition.

src/xpub.cpp
```diff
@@ -88,57 +88,89 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
...
+            if (!subscribe)
+                _manual_subscriptions.rm (data, size, pipe_);
+            else
+                _manual_subscriptions.add (data, size, pipe_);
+
+            _pending_pipes.push_back (pipe_);
+
+            //  ZMTP 3.1 hack: we need to support sub/cancel commands, but
+            //  we can't give them back to userspace as it would be an API
+            //  breackage since the payload of the message is completely
```
sigiesec: breakage?

sigiesec:
So you intend to add some tests for this when the send side is also added?

I must admit that I cannot really judge the additions. Should someone else have a look before merging? Otherwise, I will merge it.

bluca:
> So you intend to add some tests for this when the send side is also added?

Yeah it's a bit hard to do without the sending side - I can try to mock it

sigiesec:
I think the best were to have tests for the receiver side only as unittests. However, we do not have any comparable tests until now, so probably this would require quite significant changes to the affected classes to make them unit-testable.

So if you intend to add support for the send side in the near future, and then add API-level tests, this should be fine. Then it is at the same level of tests as comparable existing features.

laplaceyang:
I am confused, since the rfc document defined:

> Subscriptions SHALL be additive and SHALL NOT be idempotent. That is, subscribing to "A" and "" is the same as subscribing to "" alone. Subscribing to "A" and "A" counts as two subscriptions, and would **require** two CANCEL commands to undo.

If the sub side send "SUB A" to pub side twice, we store the pipe of sub into same node of struct trie, and save into a struct of set inner the node. 
If sub side send "UNSUB A" once, the pipe is just removed. No need for the second "UNSUB A" command.

The behavior is different from the definition of rfc doc.

I am not sure whether I'm wrong or not.

bluca:
That is enforced on the sub side - if you send "subscribe A" "subscribe A" and then "cancel A" "cancel A", the sub will wait for the second cancel before sending

bluca:
Which I guess was done to reduce traffic - but it does breach the protocol. But at this point other implementations are probably doing the same thing, so we can't really change that. Perhaps the RFC should be changed instead to reflect the actual implementations...

sigiesec:
Sorry I was AFK for the last few days. Thanks for the test. So is this ready to merge in spite of the discussion on the protocol?

bluca:
No worries!

Yeah I'm not changing the way the PUB-SUB subscription behaviour works here - it was already like that. This just adds the new on-the-wire format.

PR #3170, opened 2018-06-21
update pair doc

jamesdillonharvey:
Problem: ZMQ_PAIR not suitable all transports

Solution: Add documentation to discourage using ZMQ_PAIR for reconnecting transports

### DIFF
--- a/doc/zmq_socket.txt
+++ b/doc/zmq_socket.txt
@@ -358,6 +358,11 @@ high water mark for the connected peer, or if no peer is connected, then
 any linkzmq:zmq_send[3] operations on the socket shall block until the peer
 becomes available for sending; messages are not discarded.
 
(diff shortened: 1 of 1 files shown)


doc/zmq_socket.txt
```diff
@@ -358,6 +358,11 @@ high water mark for the connected peer, or if no peer is connected, then
 any linkzmq:zmq_send[3] operations on the socket shall block until the peer
 becomes available for sending; messages are not discarded.
 
+While 'ZMQ_PAIR' sockets can be used over transports other than linkzmq:zmq_inproc[7],
+their inability to auto-reconnect coupled with the fact new incoming connections will
+be terminated while any previous connections (including ones in a closing state)
+exist makes them unsuitable in most cases.
```
bluca: it's actually only TCP that is a problem - IPC obviously doesn't have problems with connections dropping

jamesdillonharvey: I changed the text

PR #3177, opened 2018-06-27
[RFC][WIP] Problem: libzmq does not send ZMTP 3.1 sub/cancel commands

bluca:
This one is tricky, and the workaround to be backward-compatible means that subscription messages with a topic > 32 bytes get duplicated for each ZMTP < 3.1 peer, which might be a huge performance hog. The details of the hack are commented inline.
On the other hand, I really don't want to make the choice manual - it's a protocol detail, users should not care one bit.

Any ideas?

### DIFF
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -212,6 +212,40 @@ int zmq::msg_t::init_leave ()
     return 0;
 }
 
+int zmq::msg_t::init_subscribe (const size_t size_, const unsigned char *topic)
+{
+    int rc = init_size (size_ + sub_cmd_name_size);
+    if (rc == 0) {
+        set_flags (zmq::msg_t::command);
+        set_flags (zmq::msg_t::subscribe);
+        memcpy (data (), "\x9SUBSCRIBE", sub_cmd_name_size);
+
+        //  We explicitly allow a NULL subscription with size zero
+        if (size_) {
+            assert (topic);
+            memcpy (command_body (), topic, size_);
+        }
+    }
+    return rc;
+}
+
+int zmq::msg_t::init_cancel (const size_t size_, const unsigned char *topic)
+{
+    int rc = init_size (size_ + cancel_cmd_name_size);
+    if (rc == 0) {
+        set_flags (zmq::msg_t::command);
+        set_flags (zmq::msg_t::cancel);
+        memcpy (data (), "\6CANCEL", cancel_cmd_name_size);
+
+        //  We explicitly allow a NULL subscription with size zero
+        if (size_) {
+            assert (topic);
+            memcpy (command_body (), topic, size_);
+        }
+    }
+    return rc;
+}
+
 int zmq::msg_t::close ()
 {
     //  Check the validity of the message.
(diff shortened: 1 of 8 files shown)


jamesdillonharvey:
Hi, 

I am late to the table here, but I use PUB/SUB extensively with varying size topics so I am really interested.

 I am reading the code/rfc but trying to catch up so please excuse naive questions while I do.

**Questions:**

 - ZMTP 3.1 has a new sub communication method that is stored in the command rather than the payload. Is that correct?
 - What is the 32 byte limit? Is that related to a limit on the command size?
 - What happens in ZMTP 3.1 if the topic is over 32 in size? 
 - The rfc says that filtering should happen at pub side, does this rule out pgm/mcast from ZMTP 3.1?
 - What are the advantages of the new mechanism? 

Thanks

James

bluca:
> ZMTP 3.1 has a new sub communication method that is stored in the command rather than the payload. Is that correct?

Yes

> What is the 32 byte limit? Is that related to a limit on the command size?

Internal detail - payloads below 33 bytes are on the stack, so there's no need to copy the message to modify it, as it's already a different copy

> What happens in ZMTP 3.1 if the topic is over 32 in size?

If both peers are 3.1, nothing

> The rfc says that filtering should happen at pub side, does this rule out pgm/mcast from ZMTP 3.1?

I'm not sure I understand? It's not about filtering, it's about sending the subscription message from the client

> What are the advantages of the new mechanism?

I didn't write the spec, but I can see that having a structure command is "cleaner", and more importantly it allows more messages to be sent from the xsub to the xpub - right now if a message payload starts with a 0 or a 1 (byte), then it's interpreted as a sub/cancel - so if you send any type of control data up from xsub to xpub you _must_ ensure that it doesn't start with either value

bluca:
> One question though is will ZMTP 3.1 break existing application that read the first byte (0/1) in the message to determine what's a sub or unsub. I have plenty of those systems that use a XPUB with ZMQ_XPUB_VERBOSE(R) to determine when initial and repeat sub/unsubs arrive. They use this to push out a snapshot of the current state of the cache in band with the published data. This is common pattern to allow reverse communication channel pub/sub in market data.

For backward compatibility the messages returned to userspace always have the older format, regardless of what happened on the wire

bluca:
The reason for copying is that if there are 2 peers, one that is 3.1 and one 3.0, the latter needs the modified message, but not the former

jamesdillonharvey:
I get that a copy needs to be made, I am referring to the shrink and memmove.

Line 959 - 962 

Just an idea but maybe making msg data() point past the command size will be more efficient than the memmove.

I don't know if it possible it's just an idea as moving memory is expensive.

bluca:
Ah I see what you mean - to be honest I'm more worried about having to do the de-couple at all, and if there's no way to avoid it I'm quite sure it's better to not support this at all...

jamesdillonharvey:
I just looked through this again and I still have no good ideas on better/cleaner ways to achieve it.

I would also say the advantages of the new subscription commands are not particularly compelling given the extra complexity required in the code.

If it were up to me I would just not support this feature.

bluca:
One possible idea: add an atomic pointer in content_t, so that the conversion, if necessary, has to be done only once and then it's shared as usual. Still not ideal, but a long way from one copy per subscriber.

jamesdillonharvey:
That sounds like a more efficient and simpler solution to me.

I assume for simplicity once the conversion has occurred we will not try and revert if the <3.1 component leaves,

bluca:
> Add new member to msg_t, something like size_t data_offset_ which is zero by default.

The class parameters need to fit into 64 bytes, so there's no room for a full size_t, and even a small int would have the side effect of making the largest VSM (which fits onto the stack and thus doesn't need malloc/copy) smaller, so it would most likely actually make things worse :-/

Issue #2876, opened 2017-12-25
stack over flow on windows x64


crinai:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description
socket on windows is different from it on linux, the struct is a little bigger, needs more stack size to work
normal.
Change the link option to increase the stack size can solve it ,but it's not working if the program it desined as a DLL to be loaded by anthoer EXE

# Environment
Microsoft Visual Studio 2010
* libzmq version (commit hash if unreleased): 
v4.2.3
* OS: 
Windows 7 x64 , Windows 10 x64
# Minimal test code / Steps to reproduce the issue

1.  build the example "inproc_thr" in x64
2.  set any args run the program
3. run,and it will crash 

# What's the actual result? (include assertion message & call stack if applicable)
 	inproc_thr.exe!__chkstk()	未知
>	inproc_thr.exe!std::map<unsigned short,zmq::select_t::family_entry_t,std::less<unsigned short>,std::allocator<std::pair<unsigned short const ,zmq::select_t::family_entry_t> > >::operator[]() 行 210	C++
 	inproc_thr.exe!zmq::select_t::add_fd() 行 80	C++
 	inproc_thr.exe!zmq::socket_base_t::start_reaping() 行 1330	C++
 	inproc_thr.exe!zmq::reaper_t::process_reap() 行 128	C++
 	inproc_thr.exe!zmq::object_t::process_command() 行 143	C++
 	inproc_thr.exe!zmq::reaper_t::in_event() 行 98	C++
 	inproc_thr.exe!zmq::select_t::trigger_events() 行 122	C++
 	inproc_thr.exe!zmq::select_t::select_family_entry() 行 404	C++
 	inproc_thr.exe!zmq::select_t::loop() 行 372	C++
 	inproc_thr.exe!zmq::select_t::worker_routine() 行 416	C++
 	inproc_thr.exe!thread_routine() 行 47	C++
 	inproc_thr.exe!_callthreadstartex() 行 314	C
 	inproc_thr.exe!_threadstartex(void * ptd) 行 297	C
 	kernel32.dll!00007ffa41982774()	未知
 	ntdll.dll!00007ffa44150d51()	未知

# What's the expected result?
increase the stack size,there is two way:
1.roperties - Configuration Properties - Linker - System - Stack Reserve Size
2.edit the src/thread.cpp file(line 59):
from: 
 descriptor = (HANDLE) _beginthreadex (NULL, 0,
[...]

PR #3179, opened 2018-06-27
Problem: stack overflow on win64 #2876

masariello:


src/signaler.cpp
```diff
@@ -270,19 +270,18 @@ int zmq::signaler_t::wait (int timeout_)
 
 #elif defined ZMQ_POLL_BASED_ON_SELECT
 
-    fd_set fds;
-    FD_ZERO (&fds);
```
sigiesec: I think the FD_ZERO call is still needed with optimized_fd_set_t

sigiesec:
The formatting is broken: https://travis-ci.org/zeromq/libzmq/jobs/397707175#L651 
You can apply the patch from there without having clang-format locally.

If you then squash everything into a single commit with the original commit message, this can be merged :)

Thanks for fixing this, it was an omission when I added the use of optimized_fd_set at other places.

sigiesec:
Thanks again for the contribution :)

Issue #3191, opened 2018-07-25
Race condition with received message causes ZMQ_CONNECT_ROUTING_ID to be assigned to wrong socket


mvilim:
# Issue description

There is an problem when a ROUTER socket is trying to connect to another socket using ZMQ_CONNECT_ROUTING_ID while receiving messages from a third socket. If a message is received (in the io thread) between setting the socket option and connecting to the socket, the routing id can be applied to the socket from which we received the message instead of the socket to which we are connecting.

I believe this is caused by socket_base_t::connect [where pending commands are processed](https://github.com/zeromq/libzmq/blob/25f47ccbf8775534878f37eedee4a74fe9166d36/src/socket_base.cpp#L672). If we process a command where a received message causes us to bind to new socket, we will use the routing id for that socket rather than the target of the connect call. The same behavior can be triggered by setting ZMQ_CONNECT_ROUTING_ID and then sending or receiving messages on the same socket (as that will also trigger process_commands).

I believe a possible solution is to change identify_peer in the stream and router socket implementations so that they can differentiate between peers being connected to from a local connect call (possibly by passing another argument in) and peers being connected to because of a received message.

I'm open to trying to write a pull request for this, but I first wanted to validate that those more familiar with the library see it as a bug. According to the docs, the option should only apply to sockets connected to via zmq_connect: "The ZMQ_CONNECT_RID option sets the peer id of the next host connected via the zmq_connect() call, and immediately readies that connection for data transfer with the named id". With the current behavior, there is no way to use ZMQ_CONNECT_ROUTING_ID reliably on a socket that is also receiving connections.

# Environment

* libzmq version: commit: 25f47ccb 
* OS: Linux version 4.17.4-1-ARCH

# Minimal test code / Steps to reproduce the issue

```
#include "testutil.hpp"

void test_router_2_router_while_receiving ()
{
    void *xbind, *zbind, *yconn;
    int ret;
    char buff[256];
    char msg[] = "hi 1";
    const char *xbindip = "tcp://127.0.0.1:5555";
    const char *zbindip = "tcp://127.0.0.1:5556";
    int zero = 0;
    size_t len = MAX_SOCKET_STRING;
    char x_endpoint[MAX_SOCKET_STRING];
    char z_endpoint[MAX_SOCKET_STRING];
    void *ctx = zmq_ctx_new ();

    //  Create xbind socket.
    xbind = zmq_socket (ctx, ZMQ_ROUTER);
    assert (xbind);
[...]

PR #3193, opened 2018-07-25
Race condition with received message causes ZMQ_CONNECT_ROUTING_ID to be assigned to wrong socket

mvilim:
Problem: ZMQ_CONNECT_ROUTING_ID can be assigned to incoming socket connection (Issue #3191)

Solution: Add an identifier parameter for local attach to zmq::socket_base_t::attach_pipe

### DIFF
--- a/tests/test_connect_rid.cpp
+++ b/tests/test_connect_rid.cpp
@@ -189,6 +189,106 @@ void test_router_2_router (bool named_)
     zmq_ctx_destroy (ctx);
 }
 
+void test_router_2_router_while_receiving ()
+{
+    void *xbind, *zbind, *yconn;
+    int ret;
+    char buff[256];
+    char msg[] = "hi 1";
+    const char *wildcard_bind = "tcp://127.0.0.1:*";
+    int zero = 0;
+    size_t len = MAX_SOCKET_STRING;
+    char x_endpoint[MAX_SOCKET_STRING];
+    char z_endpoint[MAX_SOCKET_STRING];
+    void *ctx = zmq_ctx_new ();
+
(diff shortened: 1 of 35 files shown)


mvilim:
I followed the pattern of "subscribe_to_all_" and added another parameter to attach_pipe. However, this is a bit ugly because it appears in a lot of socket types where it has no impact (and it makes the default values problematic because we only care about one of the optional parameters in any one place). Let me know if you'd like to try to come up with a different approach.

tests/test_connect_rid.cpp
```diff
@@ -189,13 +189,115 @@ void test_router_2_router (bool named_)
     zmq_ctx_destroy (ctx);
 }
 
+void test_router_2_router_while_receiving ()
+{
+    void *xbind, *zbind, *yconn;
+    int ret;
+    char buff[256];
+    char msg[] = "hi 1";
+    const char *xbindip = "tcp://127.0.0.1:5555";
```
bluca: don't hard-code the bind ports, use a wildcard

mvilim:
Fixed. I force-pushed a commit with the fix squashed in. I wasn't sure if that was preferred to having commits with fixes in the pull request.

bluca:
That's fine, thanks

bluca:
@mvilim could you please send a new pull request or an email with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE we are trying to relicense libzmq under MPL-2. Thank you!

PR #3205, opened 2018-08-09
Code style improvements

sigiesec:
See individual commits

### DIFF
--- a/src/timers.cpp
+++ b/src/timers.cpp
@@ -144,7 +144,7 @@ long zmq::timers_t::timeout ()
     for (; it != end; ++it) {
         if (0 == _cancelled_timers.erase (it->second.timer_id)) {
             //  Live timer, lets return the timeout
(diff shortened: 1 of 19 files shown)


src/timers.cpp
```diff
@@ -144,7 +144,7 @@ long zmq::timers_t::timeout ()
     for (; it != end; ++it) {
         if (0 == _cancelled_timers.erase (it->second.timer_id)) {
             //  Live timer, lets return the timeout
-            res = it->first > now ? static_cast<long> (it->first - now) : 0;
+            res = std::max (static_cast<long> (it->first - now), 0l);
```
bluca: 01 ?

sigiesec: That's an l (for long), not a 1 ;)

bluca: lol, misleading font

PR #3206, opened 2018-08-09
Problem: stream_engine_t instance may access its fields after it deleted itself

sigiesec:
Solution: prevent access to data if the object was deleted

Fixes issue detected by clang static analyzer-: https://travis-ci.org/zeromq/libzmq/jobs/411364097#L2126

### DIFF
--- a/src/i_engine.hpp
+++ b/src/i_engine.hpp
@@ -50,7 +50,10 @@ struct i_engine
 
     //  This method is called by the session to signalise that more
     //  messages can be written to the pipe.
-    virtual void restart_input () = 0;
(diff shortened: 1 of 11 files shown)


bluca:
norm and pgm need to be updated too (see travis)

sigiesec:
Should be fine now.

bluca:
there's a test failing on OSX that wasn't failing before, could you please have a look?

sigiesec:
@bluca I had a look, but I can't see a connection to the changes. The only change in behaviour is when ZAP is used, which is not the case with test_many_sockets.

bluca:
Right, it was probably an overloaded OSX VM

PR #3211, opened 2018-08-10
Migrate and extend probe router test

sigiesec:


sigiesec:
Can someone please have a look at the TODO here: https://github.com/zeromq/libzmq/pull/3211/files#diff-6fedb156a28fa5404fc7711229c27e42R72

I would have expected that "Hello" can be received here, but only garbage (?) of length 5 is received. This wasn't checked by the original test case either, so there might be a bug present for some time. Or I don't understand the behaviour of ZMQ_PROBE_ROUTER correctly (or am confusing the behaviour of a ZMQ_ROUTER in general).

bluca:
The message is received correctly, but the router first returns the routing ID and something goes wrong there:

```
(gdb) p routing_id
$26 = (const zmq::blob_t &) @0x5555557796c0: {_data = 0x555555779980 "", _size = 5, _owned = true}
```

I'll have another look later today

bluca:
It's because the other ROUTER socket does not set an ID, so get_server_socket_routing_id returns garbage. If you set a routing id on the server ROUTER too, it will work fine.

sigiesec:
@bluca Thanks for the hint in the right direction. I have adapted the test case, the PR should be ready to merge now.

Issue #3201, opened 2018-08-02
New socket option ZMQ_ROUTER_NOTIFY


hoditohod:
**This issue is more of a debate than a concrete issue**

ZeroMQ's socket abstraction is really powerful that it handles multiple connections in the background and the library goes far to hide the complexity of managing these connections. This is beneficial as it simplifies application code. Usually knowing that messages are round-robin out and fair-queued in is enough. And where it is not: the ROUTER socket is used which identifies its peers by routing-id and allows targeted responses.

Though the routing-id's are exposed to the application by the socket, much of the connection management still remains hidden. This is in accordance with other zmq socket types, but I think it imposes limitations on the user code. 

The specific case I would like to discuss is: when a peer disconnects from the ROUTER. Most ROUTER based applications maintain state related to the routing-id (at least they store the id itself), and need to know when the state can be released. There are ways to achieve this (MANDATORY routing, heartbeats, timeouts, get_peer_state), but none of these give the "if libzmq releases the routing-id, then let me know".

This topic comes up from time to time on the mailing list and the github issues, and there are still uses-cases not covered by the above listed solutions, so it's worth revisiting.

**Proposal**

What I propose is a ZMQ_ROUTER_NOTIFY socket option that (as the STREAM counterpart) sends a routing-id only message when the connection is established or closed. 

I'm completely unfamiliar with libzmq internals (though I'm a user for years), I would do the PR, but I need expert guidance. The first question before I start to dig myself into the code: is it feasible at all, or there're some design obstacles that are very hard to overcome?

What do you think?

bluca:
There is already machinery for this with mandatory routing - and I'm not convinced this would be a good addition. There's a reason these details are hidden - it's because they get abused. From day one zeromq abstracted away the low level details, to encourage better designed application protocols for distributed systems, as explained with examples in the zguide.
If one really wants to micro-manage all the TCP connections, then the system socket APIs can simply be used.

Having said that, in compliance with C4 if you want to implement this (with tests and documentation, pretty please) nobody's going to stop you, and we'll merge it.

bluca:
> I don't want to abuse C4, if you think this against the core principles of the library, but as I mentioned I've seen similar requests before. (Now jumping into the code to see if I can even do this or not...)

It's not an abuse - it's working as intended :-) All that is requested is that there are tests and documentation, and no backward-incompatible changes. Besides that, go for it.

Issue #2306, opened 2017-01-13
Heartbeating: no way to check which peer timed out


Asmod4n:
It would be nice if zmq_socket_monitor could also tell you which peer (socket identity?) has timed out.

bluca:
Yes I can see how this could be useful. PRs welcome :-)

Asmod4n:
The only thing i could think of is to add zap and zmtp metadata to one of the event messages.

sigiesec:
This is similar to #2653

PR #3213, opened 2018-08-13
Add ZMQ_ROUTER_NOTIFY draft socket option

hoditohod:
# New feature: ZMQ_ROUTER_NOTIFY socket option
This PR proposes a new feature for ROUTER sockets on stream oriented transports: configurable connect/disconnect notifications. Related issue/discussion: #3201 

The implementation follows the already existing ZMQ_STREAM_NOTIFY option, notifications are sent close  to the `handshake_succeded` and `disconnected` monitor events.

### DIFF
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -904,6 +904,23 @@ Option value unit:: 0, 1
 Default value:: 1
 Applicable socket types:: ZMQ_RADIO, when using UDP multicast transport
 
+
+ZMQ_ROUTER_NOTIFY: Retrieve router socket notification settings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Retrieve the current notification settings of a router socket. The returned
+value is a bitmask composed of ZMQ_NOTIFY_CONNECT and ZMQ_NOTIFY_DISCONNECT
+flags, meaning connect and disconnect notifications are enabled, respectively.
+A value of '0' means the notifications are off.
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: 0, ZMQ_NOTIFY_CONNECT, ZMQ_NOTIFY_DISCONNECT, ZMQ_NOTIFY_CONNECT|ZMQ_NOTIFY_DISCONNECT
+Default value:: 0
+Applicable socket types:: ZMQ_ROUTER
+
+
 RETURN VALUE
 ------------
 The _zmq_getsockopt()_ function shall return zero if successful. Otherwise it
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -906,7 +906,7 @@ Applicable socket types:: ZMQ_RADIO, when using UDP multicast transport
 
 
 ZMQ_ROUTER_NOTIFY: Retrieve router socket notification settings
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Retrieve the current notification settings of a router socket. The returned
 value is a bitmask composed of ZMQ_NOTIFY_CONNECT and ZMQ_NOTIFY_DISCONNECT
 flags, meaning connect and disconnect notifications are enabled, respectively.
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -1302,8 +1302,8 @@ NOTE: in DRAFT state, not yet available in stable releases.
 
 [horizontal]
 Option value type:: int
-Option value unit:: 0, ZMQ_NOTIFY_CONNECT, ZMQ_NOTIFY_DISCONNECT
-Default value:: 1
+Option value unit:: 0, ZMQ_NOTIFY_CONNECT, ZMQ_NOTIFY_DISCONNECT, ZMQ_NOTIFY_CONNECT|ZMQ_NOTIFY_DISCONNECT
+Default value:: 0
 Applicable socket types:: ZMQ_ROUTER
 
 
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -1289,6 +1289,24 @@ Option value unit:: 0, 1
 Default value:: 1
 Applicable socket types:: ZMQ_RADIO, when using UDP multicast transport
 
+
+ZMQ_ROUTER_NOTIFY: Send connect and disconnect notifications
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Enable connect and disconnect notifications on a ROUTER socket.
+When enabled, the socket delivers a zero-length message (with routing-id
+as first frame) when a peer connects or disconnects. It's possible
+to notify both events for a peer by OR-ing the flag values. This option
+only applies to stream oriented (tcp, ipc) transports.
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: 0, ZMQ_NOTIFY_CONNECT, ZMQ_NOTIFY_DISCONNECT
+Default value:: 1
+Applicable socket types:: ZMQ_ROUTER
+
+
 RETURN VALUE
 ------------
 The _zmq_setsockopt()_ function shall return zero if successful. Otherwise it
--- a/src/options.hpp
+++ b/src/options.hpp
@@ -261,7 +261,7 @@ struct options_t
     // Use zero copy strategy for storing message content when decoding.
     bool zero_copy;
 
-    // Router socket connect(1)/disconnect(2) notifications
+    // Router socket ZMQ_NOTIFY_CONNECT/ZMQ_NOTIFY_DISCONNECT notifications
     int router_notify;
 
     // Application metadata
--- a/src/options.hpp
+++ b/src/options.hpp
@@ -261,6 +261,9 @@ struct options_t
     // Use zero copy strategy for storing message content when decoding.
     bool zero_copy;
 
+    // Router socket connect(1)/disconnect(2) notifications
+    int router_notify;
+
     // Application metadata
     std::map<std::string, std::string> app_metadata;
 };
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -141,7 +141,8 @@ int zmq::router_t::xsetsockopt (int option_,
 
 #ifdef ZMQ_BUILD_DRAFT_API
         case ZMQ_ROUTER_NOTIFY:
-            if (is_int && value >= 0 && value <= (ZMQ_NOTIFY_CONNECT|ZMQ_NOTIFY_DISCONNECT)) {
+            if (is_int && value >= 0
+                && value <= (ZMQ_NOTIFY_CONNECT | ZMQ_NOTIFY_DISCONNECT)) {
                 options.router_notify = value;
                 return 0;
             }
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -139,6 +139,15 @@ int zmq::router_t::xsetsockopt (int option_,
             }
             break;
 
+#ifdef ZMQ_BUILD_DRAFT_API
+        case ZMQ_ROUTER_NOTIFY:
+            if (is_int && value >= 0 && value <= 3) {
+                options.router_notify = value;
+                return 0;
+            }
+            break;
+#endif
+
         default:
             return routing_socket_base_t::xsetsockopt (option_, optval_,
                                                        optvallen_);
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -141,7 +141,7 @@ int zmq::router_t::xsetsockopt (int option_,
 
 #ifdef ZMQ_BUILD_DRAFT_API
         case ZMQ_ROUTER_NOTIFY:
-            if (is_int && value >= 0 && value <= 3) {
+            if (is_int && value >= 0 && value <= (ZMQ_NOTIFY_CONNECT|ZMQ_NOTIFY_DISCONNECT)) {
                 options.router_notify = value;
                 return 0;
             }
--- a/src/stream_engine.cpp
+++ b/src/stream_engine.cpp
@@ -1048,9 +1048,8 @@ int zmq::stream_engine_t::push_one_then_decode_and_push (msg_t *msg_)
 
 void zmq::stream_engine_t::error (error_reason_t reason_)
 {
-    if ((_options.raw_socket && _options.raw_notify)
-        || ((_options.router_notify & ZMQ_NOTIFY_DISCONNECT) && !_handshaking)) {
-        //  For raw or router sockets, send a final 0-length message to the application
+    if (_options.raw_socket && _options.raw_notify) {
+        //  For raw sockets, send a final 0-length message to the application
         //  so that it knows the peer has been disconnected.
         msg_t terminator;
         terminator.init ();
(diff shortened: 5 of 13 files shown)


bluca:
Two TODOs: 
1) You need to add the new definitions in src/zmq_draft.h as well - as you can see, the non-draft builds are failing
2) Given the option value is stored, it needs a zmq_getsockopt as well

doc/zmq_setsockopt.txt
```diff
@@ -1289,6 +1289,24 @@ Option value unit:: 0, 1
 Default value:: 1
 Applicable socket types:: ZMQ_RADIO, when using UDP multicast transport
 
+
+ZMQ_ROUTER_NOTIFY: Send connect and disconnect notifications
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Enable connect and disconnect notifications on a ROUTER socket.
+When enabled, the socket delivers a zero-length message (with routing-id
+as first frame) when a peer connects or disconnects. It's possible
```
sigiesec: Is it ensured that connect/disconnect messages are always correctly paired?
Isn't it possible that either might be lost? Then it wouldn't be possible to correctly tell whether a notification is a connect or disconnect notification. In that case, it would be necessary not to deliver a zero-length message, but a longer message that allows to differentiate this.

doc/zmq_setsockopt.txt
```diff
@@ -1289,6 +1289,24 @@ Option value unit:: 0, 1
...
+to notify both events for a peer by OR-ing the flag values. This option
+only applies to stream oriented (tcp, ipc) transports.
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: 0, ZMQ_NOTIFY_CONNECT, ZMQ_NOTIFY_DISCONNECT
```
sigiesec: Are these flags, which can be OR-ed?

doc/zmq_setsockopt.txt
```diff
@@ -1289,6 +1289,24 @@ Option value unit:: 0, 1
...
+Default value:: 1
```
sigiesec: What is "1"? ZMQ_NOTIFY_CONNECT as defined in zmq.h, but since the line before refers to the named constants, this line should be consistent with that.

sigiesec: However, the default is actually 0, not 1, as implemented in options.cpp

src/options.hpp
```diff
@@ -261,6 +261,9 @@ struct options_t
     // Use zero copy strategy for storing message content when decoding.
     bool zero_copy;
 
+    // Router socket connect(1)/disconnect(2) notifications
```
sigiesec: Refer to the named constants here, rather than their literal values

src/router.cpp
```diff
@@ -139,6 +139,15 @@ int zmq::router_t::xsetsockopt (int option_,
             }
             break;
 
+#ifdef ZMQ_BUILD_DRAFT_API
+        case ZMQ_ROUTER_NOTIFY:
+            if (is_int && value >= 0 && value <= 3) {
```
sigiesec: Use ZMQ_NOTIFY_CONNECT | ZMQ_NOTIFY_DISCONNECT rather than the "magic" literal 3 here.

src/stream_engine.cpp
```diff
@@ -908,9 +910,25 @@ void zmq::stream_engine_t::mechanism_ready ()
...
+            return;
+        }
+        errno_assert (rc == 0);
+        flush_session = true;
+    }
+
+    if (flush_session)
+        _session->flush ();
+
+
```
sigiesec: Extra empty line

src/stream_engine.cpp
```diff
@@ -908,9 +910,25 @@ void zmq::stream_engine_t::mechanism_ready ()
             return;
         }
         errno_assert (rc == 0);
-        _session->flush ();
+        flush_session = true;
     }
 
+    if (_options.router_notify & ZMQ_NOTIFY_CONNECT) {
+        msg_t connector;
```
sigiesec: Maybe rename to connect_notification?

src/stream_engine.cpp
```diff
@@ -908,9 +910,25 @@ void zmq::stream_engine_t::mechanism_ready ()
...
+        connector.init ();
+        const int rc = _session->push_msg (&connector);
+        if (rc == -1 && errno == EAGAIN) {
+            // See above comment
```
sigiesec: Unclear which comment this refers to (in particular when the code is modified in the future)

tests/test_router_notify.cpp
```diff
@@ -0,0 +1,219 @@
...
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_disconnect (dealer, ENDPOINT_0));
+
+    // need one more process_commands() (???)
+    msleep (SETTLE_TIME);
+    zmq_getsockopt (dealer, ZMQ_EVENTS, &opt_events, &opt_events_length);
+
+    // connection notification msg
+    if (notify_opt & ZMQ_NOTIFY_DISCONNECT) {
+        // routing-id only message of the connect
+        printf ("ehe\n");
```
sigiesec: Remove this? (what does "ehe" mean anyway?)

sigiesec:
I really like the intent of this extension. However, please give my review comments a look. In particular, the one on the specification of the socket option. I have the impression that the design must be changed somehow to ensure this feature can be used reliably.

doc/zmq_setsockopt.txt
hoditohod: Well, I'm afraid my knowledge of libzmq internals doesn't let me confidently answer that. :( I tried to understand the stream_engine and follow the already existing STREAM_NOTIFY implementation that seemed to be stable. 

The connect notification is emitted right after the routing id is sent from the stream_engine to the router_socket, the disconnect is emitted on error only if the handshake is complete. These notifications are close to the corresponding monitor events, which (as I understand) are correctly paired. Apart from this, I can't provide more solid guarantees. :(

Regarding your proposal, yes, the notification messages could be longer the disambiguate between the two. This way the application could detect a malfunction in the library, but it won't prevent emitting incorrectly paired notifications, only improve detectability. I can add one byte with the ZMQ_NOTIFY_* flag values to the notification message if that looks better.

My main problem with the current design is that a malicious peer can craft messages identical to the notifications, but I couldn't come up with anything better, and tried to stick to what already exists (STREAM_NOTIFY).

doc/zmq_setsockopt.txt
hoditohod: Yes, should I add a 4th `ZMQ_NOTIFY_CONNECT|ZMQ_NOTIFY_DISCONNECT` value?

doc/zmq_setsockopt.txt
hoditohod: Typo in zmq_setsockopt.txt, the default value is of course 'off'.

src/stream_engine.cpp
hoditohod: Duplicated the comment for clarification.

tests/test_router_notify.cpp
hoditohod: Uhm, yeah... it's a high quality debug event, it doesn't mean anything... :) Removed.

doc/zmq_setsockopt.txt
sigiesec: I think I was mistaken. The connect/disconnect messages are only sent locally, not via the underlying socket, right? Then they cannot be lost, of course.

However, this has another consequence, and you note that the disconnect message is only sent in case of an error. What happens in case of a normal connection termination? Or does at connection always end with an stream_engine_t::error?

I originally thought (not based on the implementation, but on this documentation), that the peer sends these messages over the network, in particular when it closes its socket. This would be useful IMO, because otherwise a disconnecting client cannot be reliably detected in time.

The same seems to be true for ZMQ_STREAM_NOTIFY, which may be stable, but I don't think it is widely used.

src/stream_engine.cpp
sigiesec: I now saw that this is copied from the ZMQ_STREAM_NOTIFY handling, so better leave it as it is.

doc/zmq_getsockopt.txt
```diff
@@ -904,6 +904,23 @@ Option value unit:: 0, 1
 Default value:: 1
 Applicable socket types:: ZMQ_RADIO, when using UDP multicast transport
 
+
+ZMQ_ROUTER_NOTIFY: Retrieve router socket notification settings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
bluca: For the markdown to work, this line must be the same length as the above one

bluca:
Definitions mirroring zmq.h are still missing from src/zmq_draft.h which breaks all the builds - I'm afraid this needs to be fixed before it can be merged

hoditohod:
@bluca I know. I bumped into another issue, and held back the zmq_draft commit on purpose, to avoid being merged as is. I'm marking the PR WIP to make it clear.

doc/zmq_setsockopt.txt
hoditohod: The notification messages are created in the stream_engine and sent to the corresponding socket internally, these messages never appear "on the wire", contrary to ROUTER_PROBE messages.

AFAICT, all connection terminations end up in zmq::stream_engine_t::error(), as this is the single point in code where the 'disconnected' monitor event is emitted.

In my use-case I would like to be notified of the termination of a routing-id when a peer abruptly goes away, so my application can release the corresponding state (on a connection oriented transport). Disconnect messages are already present in the application layer protocol, but these are not sent when the app crashes, or forcefully killed.

hoditohod:
Simon had a comment about the ambiguity of the notifications. Should I update the code to contain a single char or a string constant in the notification message? Or is it ok like this for a draft API?

sigiesec:
@hoditohod I think it is ok without additional content in the notification message. After our discussion, I think the messages should be paired correctly. For DRAFT status this is definitely ok.

doc/zmq_setsockopt.txt
sigiesec: Ok, thanks for elaborating on this. I think there should be some comments in the code that clarify this, but this is beyond the scope of this PR.

hoditohod:
I have a few more technical questions regarding the stream_engine:
- do I need to close the connect/disconnect_notification messages (msg_t) in the stream engine? The stream notification code does it (line 220, 1051), but when the routing-id is sent on 902, it is not closed.
- which method is the right one for sending messages to the socket? I used `_session->push_msg()`, but the stream notification code uses the `_process_msg` methodptr and the `push_raw_message_to_session()` method. It's not clear which one should I use...
- should I check the return value from `_session->push_msg()`? When the routing-id is sent on line 905, it's checked for EAGAIN only (and the msg_t is not closed even on error), but the stream notify code lacks any return value check. Is that ok?

sigiesec:
> do I need to close the connect/disconnect_notification messages (msg_t) in the stream engine? The stream notification code does it (line 220, 1051), but when the routing-id is sent on 902, it is not closed.

https://github.com/zeromq/libzmq/blob/dbe289509e2554fd56383067fff9e02d9b379d0b/src/session_base.hpp#L72

However, this comment is not really accurate. After a successful call, msg_ is empty, not closed (which is a different state). So probably it is more consistent to close it afterwards, but closing an empty msg_t instance essentially has no effect. Handling of initialization/closing of msg_t instances is somewhat confusing across libzmq IMO.

> which method is the right one for sending messages to the socket? I used _session->push_msg(), but the stream notification code uses the _process_msg methodptr and the push_raw_message_to_session() method. It's not clear which one should I use...

I cannot answer this one. @bluca?

> should I check the return value from _session->push_msg()? When the routing-id is sent on line 905, it's checked for EAGAIN only (and the msg_t is not closed even on error), but the stream notify code lacks any return value check. Is that ok?

I think I have to revert my comment on the possibility of losing a message once again. For the connector message, it can probably never fail in practice (but I am not sure). For the terminator message, it might fail however due to the HWM being exceeded. In that case, the terminator message is lost.

bluca:
In a separate PR, could you please send your re-licensing grant? Follow the example in this directory: https://github.com/zeromq/libzmq/tree/master/RELICENSE

bluca:
> I cannot answer this one. @bluca?

Don't know, sorry

hoditohod:
Thanks for the answers!

I looked around in the code and messages initialized with `msg_t::init()` (VSM) are generally not closed, so I leave it this way.

Regarding the loss of the disconnect notification when HWM is reached, this concerns me a bit, but currently I see no way to delay this message (until the HWM condition is over), or ignore the HWM for this single message. I'll think about this...

For now, I consider this PR done from my side.

smalls12:
I see that that starting point for this change was following ZMQ_STREAM_NOTIFY, which was already implemented.

Would another option be to just use the zmq_socket_monitor?

The socket monitor can already detect when a DEALER connects/disconnects from the ROUTER. However, the id is unknown so the ROUTER has to figure out which DEALER triggered the event.

Just wondering :)

smalls12:
#2306

sigiesec:
@smalls12 As you write yourself, with a monitoring event it is not possible to know which peer disconnected. Unfortunately, the structure of the monitoring events is fixed at the moment, so this information cannot be easily added. I ran into this problem also when trying to emit information on problems in the CURVE handshake. Maybe the whole monitoring event system should be reworked, but this is a larger effort.

smalls12:
@sigiesec thanks for the insight.

My expectation now is that it's up to my program to track these blank messages from clients. Assuming the first blank message that comes in is the CONNECT while any subsequent blank messages just flip the status back and forth.

One concern that I have, is it not possible for a DEALER to send a blank message to the ROUTER? If so, how would a ROUTER know this is a connection event or just an actual message from a DEALER.

Thanks :)

sigiesec:
@smalls12 Yes, that's true. This is a drawback in the current design of ZMQ_ROUTER_NOTIFY. If you have a situation where a peer might send an empty message, it cannot be used reliably.

smalls12:
@sigiesec I was perusing through the `router.cpp` and `tcp_connecter.cpp`.

The tcp_connector (among other files) are the files that actually send the event.

I can override the event handler from `socket_base.cpp` in `router.cpp` in order to append the routing_id onto the inproc socket of the socket monitor, but when that event comes in the router has to figure out who actually sent that event.

This just leads into what you were saying about overhauling the event system.

What could the event contain that would allow the `router.cpp` to correlate it to a routing_id that it assigned?

smalls12:
Maybe this is more appropriate to move to another thread.

hoditohod:
Hi!
Sorry for joining so late... [the usual excuses apply].

I wasn't aware of the feature request about extending the monitor API when implemented this, tough, I looked around both on the mailing list and the github issues. 

My problem with the monitor approach is the event ordering. Can the Monitor API guarantee that the monitor events and data (on the monitored socket) are delivered in the same order as the actual events (provided they are registered in the same poller)? If so, then the Monitor API is much cleaner design to deliver the routing id with the connect/disconnect events. But if not, it looks more or less useless to me: getting the last message on a socket after the disconnect event defies the whole purpose of the feature.
What do you think?

As for the malicious peer sending empty messages, yes this is an issue with the current implementation. Maybe the existence (or the lack of) a metadata property could be used to disambiguate?

Is the new issue already open?

sigiesec:
As you assume, the relative order of events on different sockets is not guaranteed. If you need that, then the current implementation must be used.

But there are different use cases I think. For our use case, this is probably not an issue.

Disambiguation could probably be done via metadata.

As for empty messages sent by the peer: I don't think you can regard this as malicious in general.

lzbgt:
@hoditohod
hi, this draft is not working on raspbian  (raspberry pi 3b+), any clue?
```
root@raspberrypi:~# uname -a
Linux raspberrypi 4.19.57-v7+ #1244 SMP Thu Jul 4 18:45:25 BST 2019 armv7l GNU/Linux
root@raspberrypi:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

lzbgt:
can't get connection/disconnection notification when router is deployed in docker and exposed via port-mapping?

hoditohod:
I'm not aware of any such issue. We're using this option to get disconnect notification on linux/docker (tcp/ipc), windows/native (tcp); without problem.

Issue #802, opened 2014-01-02
Authentication with curve


drewcrawford:
This is the issue opened for the mailing list discussion of the same name.  The canonical problem is to figure out which publickey is the responsible sender for zeromq packets.

@hintjens's  description of the issue:
- We cannot send the metadata before authentication and final key
  exchange as that would make it vulnerable to attacks.
- The canonical identification for the client is its long term key,
  and any metadata associated with it (as in a certificate), rather than
  anything sent/claimed by the client on its connection.
- The ZAP handler, as part of authentication, will also retrieve
  metadata for the client long term key, and make that available to the
  server-side application.

This moves away from using client identities, which mix two layers
(routing and identity), and thus are rather fragile. E.g. limited to
one connection per client, difficulty of recovery when a connection
dies, etc.

@hintjens  latest effort suggests:

> Hmm... this isn't going to be that easy. The connection doesn't have
> an identity until after it's authenticated. This means the only
> plausible solution is to provide authentication metadata to the
> zmq_msg API.

PR #3215, opened 2018-08-13
Problem: unused get_credential methods and associated fields

sigiesec:
Solution: remove them

These methods were added by #819, and subsequently in new socket types, since they were required by socket_base, but they are never called. They also use some fields, which are not used (read) anywhere else, so they can be removed as well.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -96,11 +96,6 @@ bool zmq::client_t::xhas_out ()
     return _lb.has_out ();
 }
 
-const zmq::blob_t &zmq::client_t::get_credential () const
-{
-    return _fq.get_credential ();
-}
-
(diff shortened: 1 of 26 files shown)


bluca:
Mmh it's hard to work out exactly what happened, since it was scattered around on various mediums (issues, emails, irc, wiki). Maybe this was a work in progress that was never completed? Or a superseded attempt that was never backed out?

I'll try to scour the git log for more info one of these days

sigiesec:
In #802, there is some discussion around a zmq_msg_who method, which was never implemented. Maybe that was supposed to use get_credential. However, over the past 4 years, no one seems to have missed it.

bluca:
Yeah, I think it's fine to remove this.

PR #3217, opened 2018-08-15
Update unittests/CMakeLists.txt: use ZeroMQ_SOURCE_DIR, ZeroMQ_BINARY_DIR

WenbinHou:
# Problem
When placing libzmq in a subdirectory, and use cmake `add_subdirectory` command to build libzmq as a sub-project (static build, with ZMQ_BUILD_TESTS ON), some header files are not found.
This is due to usage of `CMAKE_SOURCE_DIR`, `CMAKE_BINARY_DIR` in **unittests/CMakeLists.txt**, which does not refers to libzmq source directory now.

# Solution:
Use `ZeroMQ_SOURCE_DIR`, `ZeroMQ_BINARY_DIR` instead of `CMAKE_SOURCE_DIR`, `CMAKE_BINARY_DIR` in **unittests/CMakeLists.txt**, which allows libzmq to become an "add_subdirectory" target (with static build).

Currently minimum required CMake version (in unittests/CMakeLists.txt: 2.8.1) is enough to support this. See [https://cmake.org/cmake/help/v2.8.1/cmake.html](https://cmake.org/cmake/help/v2.8.1/cmake.html)

> project: Set a name for the entire project.
> Sets the name of the project. Additionally this sets the variables \<projectName\>_BINARY_DIR and
> \<projectName\>_SOURCE_DIR to the respective values.

### DIFF
--- a/RELICENSE/WenbinHou.md
+++ b/RELICENSE/WenbinHou.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Wenbin Hou
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other 
+Open Source Initiative approved license chosen by the current ZeroMQ 
+BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle "Wenbin Hou", with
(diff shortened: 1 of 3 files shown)


bluca:
Hi, could you please add a separate commit with your relicensing grant? See https://github.com/zeromq/libzmq/commit/69d69f21a875de06e8e473df373ce7b2b9253be1 and the readme in the relicense directory for details

WenbinHou:
Hi, relicensing grant has been added.

sigiesec:
Thanks!

Issue #3236, opened 2018-08-29
UDP dish socket can't bind to a multicast port already in use


Mystfit:
# Issue description

If I bind a UDP dish socket to a multicast address that is already bound in another running program instance I get hit an "address in use" error. Tracing down the initialisation of the socket in udp_engine.cpp I found that the socket has SO_REUSEADDR set by default which sounds correct but still results in this issue.

The socket will bind if I add a SO_REUSEPORT flag immediately after like the following (this is from [udp_engine.cpp](https://github.com/zeromq/libzmq/blob/50dbd80c515e31afda2193bd0b875aa31b7c9059/src/udp_engine.cpp#L194))

```
int rc = setsockopt (_fd, SOL_SOCKET, SO_REUSEADDR, reinterpret_cast<char *> (&on), sizeof (on));
#ifdef ZMQ_HAVE_WINDOWS
        wsa_assert (rc != SOCKET_ERROR);
#else
        errno_assert (rc == 0);
#endif

rc = setsockopt (fd, SOL_SOCKET, SO_REUSEPORT, reinterpret_cast<char *> &on, sizeof (on));
#ifdef ZMQ_HAVE_WINDOWS
        wsa_assert (rc != SOCKET_ERROR);
#else
        errno_assert (rc == 0);
#endif
```

I'd like to submit a PR for this, but I'm uncertain if SO_REUSEPORT should be a flag set by default. Is there a chance that it would not be supported on some platforms?

# Environment

* libzmq version: 4.2.5
* OS: OSX 10.13.6

# Minimal test code / Steps to reproduce the issue

zsock_t* first = zsock_new_dish("@udp://239.0.0.1:40000");      //Works

//In another program instance
zsock_t* second = zsock_new_dish("@udp://239.0.0.1:40000");   //Address already in use error

# What's the actual result? (include assertion message & call stack if applicable)

Binding the UDP dish socket aborts with the message "Address already in use".

# What's the expected result?

Dish socket binds with a result of 0.

bluca:
Yes please send a PR. Just have an ```#ifdef SO_REUSEPORT``` for other platforms, I think that's enough.

jamesdillonharvey:
Just as a heads up (I fixed a similar issue in openpgm) on linux >=3.9
SO_REUSEPORT
limits the reuse to the same user ID I think. Also I remember something
about load balancing (basically its not the same as the old BSD SO_REUSEPORT
option).

bluca:
[...]

PR #3237, opened 2018-08-29
Added SO_REUSEPORT socket option for UDP recv ports

Mystfit:
Problem: Binding UDP multicast ports more than once throws "Address already in use errors" [(referenced in this issue).](https://github.com/zeromq/libzmq/issues/3236)

Solution: Add SO_REUSEPORT socket option when binding UDP receive ports on platforms that support it.

### DIFF
--- a/src/udp_engine.cpp
+++ b/src/udp_engine.cpp
@@ -199,12 +199,6 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
         errno_assert (rc == 0);
 #endif
 
-#ifdef SO_REUSEPORT
-        rc = setsockopt (_fd, SOL_SOCKET, SO_REUSEPORT,
-                         reinterpret_cast<char *> (&on), sizeof (on));
-        errno_assert (rc == 0);
-#endif
-
         const ip_addr_t *bind_addr = udp_addr->bind_addr ();
         ip_addr_t any = ip_addr_t::any (bind_addr->family ());
         const ip_addr_t *real_bind_addr;
@@ -212,10 +206,17 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
         bool multicast = udp_addr->is_mcast ();
 
         if (multicast) {
+            //  Multicast addresses should be allowed to bind to more than
+            //  one port as all ports should receive the message
+#ifdef SO_REUSEPORT
+            rc = setsockopt (_fd, SOL_SOCKET, SO_REUSEPORT,
+                             reinterpret_cast<char *> (&on), sizeof (on));
+            errno_assert (rc == 0);
+#endif
+
             //  In multicast we should bind ANY and use the mreq struct to
             //  specify the interface
             any.set_port (bind_addr->port ());
-
             real_bind_addr = &any;
         } else {
             real_bind_addr = bind_addr;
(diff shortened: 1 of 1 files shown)


src/udp_engine.cpp
```diff
@@ -199,6 +199,12 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
         errno_assert (rc == 0);
 #endif
 
+#ifdef SO_REUSEPORT
```
sigiesec: Is this intentionally done for both unicast and multicast addresses? #3236 was referring specifically to multicast addresses.

src/udp_engine.cpp
```diff
@@ -199,6 +199,12 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
...
+        rc = setsockopt (_fd, SOL_SOCKET, SO_REUSEPORT,
+                         reinterpret_cast<char *> (&on), sizeof (on));
+        errno_assert (rc == 0);
```
bluca: Just in the off chance Windows adds support for this in the future, please add the ifdef ZMQ_HAVE_WINDOWS bit

src/udp_engine.cpp
Mystfit: I've left this in for both address types as I was having trouble with both multicast and wildcard addresses. This solved my issue when using a non-wildcard address (in this case an IP in the multicast address range) but still caused an issue when using a wildcard address like udp://*:port but I think it might be an unrelated issue. 

My apologies for the multicast confusion, I can try using a fixed IP instead of one from the 239.*.*.* and see if that affects anything.

src/udp_engine.cpp
Mystfit: Done. Here's hoping Windows to add this in the future as well.

src/udp_engine.cpp
jamesdillonharvey: I think if you set SO_REUSEADDR on multiple unicast sockets the data arriving on the system will be load balanced between them.  That is why the linux limits processes using SO_REUSEADDR to the same user ID.

sigiesec: The behaviour depends on the OS (in particular between Linux and BSD systems), and somewhat differs between multicast and unicast addresses. https://stackoverflow.com/a/14388707/2019765 has quite comprehensive information on this.

Mystfit: @sigiesec I've written a [gist](https://gist.github.com/Mystfit/6c015257b637ae31bcb63130da67627c) to test this. Multicast works as expected but using udp://127.0.0.1:40000 results in the first bound dish to hog all messages, even though a second program is bound to the same address:port. Killing the first bound program lets the second waiting program receive the messages instead.

src/udp_engine.cpp
```diff
@@ -206,10 +206,21 @@ void zmq::udp_engine_t::plug (io_thread_t *io_thread_, session_base_t *session_)
...
+            wsa_assert (rc != SOCKET_ERROR);
+#else
+            errno_assert (rc == 0);
+#endif
+#endif
+
             //  In multicast we should bind ANY and use the mreq struct to
             //  specify the interface
             any.set_port (bind_addr->port ());
-
```
bluca: Please remove this white space change

bluca: nvm I can do it myself

bluca:
@Mystfit thanks for your PR. Since this is your first contribution, please send another PR with a relicensing grant. See README here for details: https://github.com/zeromq/libzmq/tree/master/RELICENSE and this example: https://github.com/zeromq/libzmq/blob/master/RELICENSE/WenbinHou.md

PR #3255, opened 2018-09-20
Fix for test_hwm_pubsub

f18m:
This PR changes test_hwm_pubsub to use XPUB and wait for subscription before start to sending messages out

### DIFF
--- a/tests/test_hwm_pubsub.cpp
+++ b/tests/test_hwm_pubsub.cpp
@@ -30,6 +30,8 @@
 #include "testutil.hpp"
 #include "testutil_unity.hpp"
 
+// NOTE: on OSX the endpoint returned by ZMQ_LAST_ENDPOINT may be quite long,
+//       ensure we have extra space for that:
 #define SOCKET_STRING_LEN (MAX_SOCKET_STRING * 4)
 
 void setUp ()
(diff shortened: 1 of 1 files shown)


bluca:
This is a good chance, but unfortunately it still sometimes fails in valgrind:

```
WARNING: Forced closure of 2 sockets, this is an implementation error unless the test case failed
tests/test_hwm_pubsub.cpp:157:test_tcp:FAIL: Expected 6000 Was 5000
tests/test_hwm_pubsub.cpp:271:test_inproc:PASS
tests/test_hwm_pubsub.cpp:273:test_ipc:PASS
tests/test_hwm_pubsub.cpp:275:test_reset_hwm:PASS
```

f18m:
Hi @bluca,
two possibilities come to my mind:
1) simple, dumb fix: increase RX timeout from 10ms to 100ms, to give more time to valgrind test to receive all messages before the assert() check is done
2) when reaching msg_cnt_ == send_count, we send on the XPUB socket a termination message like the string "end". In the SUB we remove the msleep(2*SETTLE_TIME) and we instead loop till we get that end message...

bluca:
Option 2 sounds like it should work best - on architectures like arm it might be even slower so timers are not very reliable

PR #3257, opened 2018-09-21
Remove race condition from XPUB/SUB test in test_blocking()

f18m:
Title says pretty much all. This PR removes the need for msleep() from test_blocking().

Question is whether we should change also test_defaults() and test_reset_hwm() functions, which still employ msleep() calls for synching pub/sub.


bluca:
Great thanks, we can see how the CI acts up and add more as needed

PR #3262, opened 2018-09-26
Problem: low-likelyhood memory leaks

TomMD:
There are three memory leaks:

* `curve_client_tools` is dealing with NaCL's inaccurate API.  That is, the encryption function can never fail but returns a success/error code anyway which is dutifully checked.  In that branch, which can never be taken, we should free the allocated buffer.

* `msg.cpp` has an unlikely issue where if one allocation attempt fails then a prior allocation won't be freed before to returning an error code.   It is possible I freed this memory incorrectly - look at this pessimistically.

* `curve_client` has the most interesting case where a server sending an invalid READY message can cause a leak in `process_ready()`.  The implication of this leak depends on the application.

### DIFF
--- a/src/curve_client.cpp
+++ b/src/curve_client.cpp
@@ -240,6 +240,7 @@ int zmq::curve_client_t::process_ready (const uint8_t *msg_data_,
         session->get_socket ()->event_handshake_failed_protocol (
           session->get_endpoint (), ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC);
         errno = EPROTO;
+        free(ready_plaintext);
         return -1;
     }
 
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -169,6 +169,7 @@ int zmq::msg_t::init_data (void *data_,
         _u.lmsg.content =
           static_cast<content_t *> (malloc (sizeof (content_t)));
         if (!_u.lmsg.content) {
+            ffn_(data_,hint_);
             errno = ENOMEM;
             return -1;
         }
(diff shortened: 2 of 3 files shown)


src/msg.cpp
```diff
@@ -169,6 +169,7 @@ int zmq::msg_t::init_data (void *data_,
         _u.lmsg.content =
           static_cast<content_t *> (malloc (sizeof (content_t)));
         if (!_u.lmsg.content) {
+            ffn_(data_,hint_);
```
bluca: In case of error we leave the user data alone - it's their responsibility to take care of it - or in most cases, to simply try again without redoing it

src/curve_client.cpp
```diff
@@ -240,6 +240,7 @@ int zmq::curve_client_t::process_ready (const uint8_t *msg_data_,
         session->get_socket ()->event_handshake_failed_protocol (
           session->get_endpoint (), ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC);
         errno = EPROTO;
+        free(ready_plaintext);
```
bluca: to fix the clang format error, add a space between the function name and the opening bracket, ditto below

src/msg.cpp
bluca: More importantly, this would be an API change so we can't really merge as-is, sorry, needs to be fixed first

bluca: @TomMD ping

bluca:
@TomMD ping. We also need a relicense grant (as mentioned in the PR template), please.

Issue #3268, opened 2018-10-08
Problem: HWM in ZMQ_DGRAM socket does not respect multipart message


gabm:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description

When sending a lot of messages over a ZMQ_DGRAM socket, I hit an assertion in: `Assertion failed: check () (/home/gabm/zeromq/src/msg.cpp:347)`.

# Environment

* libzmq version (commit hash if unreleased): 
* OS: 

# Minimal test code / Steps to reproduce the issue

```
    auto ctx = zmq_ctx_new();
    auto sender = zmq_socket(ctx, ZMQ_DGRAM);

    int hwm = 3;
    zmq_setsockopt(sender, ZMQ_SNDHWM, &hwm, sizeof(hwm));
    zmq_bind(sender, "udp://127.0.0.1:5556");

    std::string remote = "127.0.0.1:5555";
    for (int i = 0; i < 100000; i++) {
        zmq_send(sender, remote.c_str(), remote.length(), ZMQ_SNDMORE);

        zmq_msg_t msg;
        zmq_msg_init_size(&msg, 8000);
        zmq_msg_send(&msg, sender, 0);
        zmq_msg_close(&msg);
    }

    zmq_close(sender);
    zmq_ctx_destroy(ctx);
```

# What's the actual result? (include assertion message & call stack if applicable)

The callstack is:

```
1  raise                                                         0x7ffff7705d7f 
2  abort                                                         0x7ffff76f0672 
3  zmq::zmq_abort                            err.cpp         88  0x7ffff7d2170e 
4  zmq::msg_t::size                          msg.cpp         347 0x7ffff7d2ce6e 
5  zmq::udp_engine_t::out_event              udp_engine.cpp  380 0x7ffff7d80db6 
6  zmq::epoll_t::loop                        epoll.cpp       202 0x7ffff7d2085c 
7  zmq::worker_poller_base_t::worker_routine poller_base.cpp 139 0x7ffff7d3d50d 
8  thread_routine                            thread.cpp      182 0x7ffff7d6736e 
9  start_thread                                                  0x7ffff7bd3a9d 
10 clone                                                         0x7ffff77c9a43 
```

So it happens in `udp_engine.cpp` at line 380. The rc value of the second `pull_msg` call is `-1`. 

# What's the expected result?

that it works :smile: 

# Cause

[...]

PR #3270, opened 2018-10-08
Problem: The flag that indicates the next expected message gets set even if the send fails

gabm:
In the dgram socket we have a flag that indicates the next expected message type to ensure that always a pair of "address" + "body" messages gets sent. The first one MUST have the sendmore flag, the second MUST NOT.

In case the message does not get sent because of HWM full, then the function returns EAGAIN as it should. But unfortunately the next expected message type-flag gets flipped as well. When the socket_base::send function now tries to resend the message, it became the _wrong_ message type... If you don't stop sending pairs of messages here (like me) then the next message that gets through will be of the wrong type, which in turn crashes the udp_engine function as described in #3268 

This PR fixes the flag behavior and in turn all of the described problems...

### DIFF
--- a/src/dgram.cpp
+++ b/src/dgram.cpp
@@ -106,18 +106,12 @@ int zmq::dgram_t::xsend (msg_t *msg_)
             errno = EINVAL;
             return -1;
         }
-
-        //  Expect one more message frame.
-        _more_out = true;
     } else {
         //  dgram messages are two part only, reject part if more is set
         if (msg_->flags () & msg_t::more) {
             errno = EINVAL;
             return -1;
         }
-
-        //  This is the last part of the message.
-        _more_out = false;
     }
 
     // Push the message into the pipe.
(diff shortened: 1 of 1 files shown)


gabm:
I'll send a relicense PR as well

src/dgram.cpp
```diff
@@ -129,6 +123,9 @@ int zmq::dgram_t::xsend (msg_t *msg_)
     if (!(msg_->flags () & msg_t::more))
         _pipe->flush ();
 
+    // flip the more flag
+    _more_out = !_more_out; 
+
```
bluca: trailing whitespace: https://travis-ci.org/zeromq/libzmq/jobs/438634993

bluca:
clang is still complaining, I suggest to run ```make clang-format-diff``` as the PR template suggests

gabm:
I did, but there are other unrelated changes.. thats why I thought it wouldnt be necessary.. I'll add all changes..

gabm:
(i didnt touch all the other files...)

bluca:
that's weird, @sigiesec any idea why clang would do that?

bluca:
@gabm maybe it's a different in clang version, I'd suggest to drop that last commit and instead simply apply the diff from: https://travis-ci.org/zeromq/libzmq/jobs/438643567

gabm:
lets see if this works, should be fine now

sigiesec:
I don't know what happened exactly, but I think the clang-format version used on travis has some strange bugs. I will try to update it, and adapt the existing formatting.

bluca:
On Debian with clang-format-5.0 I get the same result as Travis

PR #3272, opened 2018-10-09
fix for android building

hpsaturn:
# CMake validation for Android building issue

Problem: CMake stop configuration on:

`zmq_check_noexcept()`

Solution: I added validation on CMake for Android

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,7 +417,9 @@ if(CMAKE_SYSTEM_NAME MATCHES "SunOS" OR CMAKE_SYSTEM_NAME MATCHES "NetBSD")
   endif()
 endif()
 
-zmq_check_noexcept()
+if(NOT ANDROID)
+    zmq_check_noexcept()
+endif()
 
 #-----------------------------------------------------------------------------
 if(NOT CMAKE_CROSSCOMPILING AND NOT MSVC)


CMakeLists.txt
```diff
@@ -417,7 +417,9 @@ if(CMAKE_SYSTEM_NAME MATCHES "SunOS" OR CMAKE_SYSTEM_NAME MATCHES "NetBSD")
   endif()
 endif()
 
-zmq_check_noexcept()
+if(NOT ANDROID)
+    zmq_check_noexcept()
```
sigiesec: Wouldn't the condition better be placed inside zmq_check_noexcept?

Also, I have the impression that it is somewhat too general. There are Android builds on travis-ci that have been building fine

hpsaturn: Ok, with Docker and Ubuntu 16.04 we have this issue. But let me check again our configuration. Thanks

bluca: We use autoconf for the android builds here, so I'll merge this anyway as I think you are the first to use CMake for Android

bluca: Actually I can't yet, just noticed it's your first PR - @hpsaturn please add another commit with a relicense grant as the PR template indicates, thanks

bluca: See https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

hpsaturn: I'm ready. I have the PR with relicense grant. Interesting to know that I'm the first one with build it with CMake, for now I still migrating more libs for our project, libzmq is compiling and linking well for now, but I heven't tested it yet, the idea is testing it next week.

PR #3278, opened 2018-10-22
Always include ZMQSourceRunChecks.cmake

ojura:
I could not cross-compile ZMQ for QNX without this.
https://github.com/zeromq/libzmq/blob/cfc9d5f5198e611d8d3ad4482d57cf6bb5b6bf59/CMakeLists.txt#L420


bluca:
This was added specifically for crosscompiling, so can't undo it sorry or it would cause a regression: 
https://github.com/zeromq/libzmq/pull/2493

If it's possible to check for the actual platform and you can add an "OR QNX" or something like that, please send another PR doing so

ojura:
How can this work if `CMAKE_CROSSCOMPILING` is true? `zmq_check_noexcept()` is called always. It is not related to QNX.

bluca:
Sorry, I don't use Cmake much so I don't know. @herbrechtsmeier any idea?

ojura:
Including `ZMQSourceRunChecks.cmake` is needed to execute `zmq_check_noexcept()` because it is defined there. We are conditionally including `ZMQSourceRunChecks.cmake` based on `CMAKE_CROSSCOMPILING`, but always (unconditionally) executing `zmq_check_noexcept()`. This is obviously wrong.

bluca:
See https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

ojura:
Done (#3281).

PR #3284, opened 2018-10-30
Add support for UDP multicast ttl (ZMQ_MULTICAST_HOPS)

kfish:
This adds support for setting the multicast TTL for UDP, allowing multicast traffic to be routed. This feature was already supported for PGM.


bluca:
Thanks

Issue #1400, opened 2015-05-16
Migrate trie implementation to patricia trie to increase possible subscription count on a single box


jabberfest:
In:
https://github.com/zeromq/libzmq/blob/master/src/trie.cpp
https://github.com/zeromq/libzmq/blob/master/src/mtrie.cpp

a trie is used. When subscriptions for the pub/sub protocol near several million the memory footprint on a single box becomes fairly large. The instance of ZeroMQ can become memory bound if you have a large amount of subscriptions, and not IO bound. I'm open to discussion and thoughts! nanomsg appears to have this done already.

See for additional reasoning:
http://250bpm.com/blog:19

hintjens:
Feel free to provide a pull request, and a test case if possible.

farias-r7:
My c++ is rusty, but I'll take a stab at in a couple of weeks. 

@sustrik am I right in thinking that just writing a radix tree (patricia trie) implementation and swapping out the calls to trie and mtrie over to the patricia trie implementation is all I need to do?

ssbl:
I've written a radix tree implementation with the same API as that in `trie.hpp`; it's fairly well-tested and seems to work. It could potentially be a more efficient alternative to the trie libzmq currently uses.

The repository is [here](https://github.com/ssbl/radix-tree); just wanted to confirm before proposing a somewhat big change. It won't be the exact same code, I'll adapt it to libzmq's guidelines.

bluca:
Have you done some performance measurement to see if there are improvements, and in which conditions? Also it needs to be C++ 98 as there are users on very old compilers

ssbl:
I don't have any measurements yet, I'll collect some and see if there's actually an improvement in performance. Porting it to C++98 shouldn't be a problem, I think.

ssbl:
I've added the radix tree and a simple benchmarking script to my fork of libzmq ([link to commit](https://github.com/ssbl/libzmq/commit/1f1cd56c34200d23699fa36b24c18319f384452a)). The script initializes both data structures with the same data, and runs a common set of queries on each. Here are the timings for running 100K queries on a data set of 10K randomly-generated keys of size 20 bytes each (times are in ms):

```
~/code/radix-tree/bench/libzmq/build $ bin/unittest_trie
[trie]
Starting warmup...
Collecting samples...
34.00
32.00
32.00
32.00
32.00
32.00
32.00
32.00
32.00
32.00
[radix_tree]
Starting warmup...
Collecting samples...
12.00
[...]

PR #3286, opened 2018-10-30
Problem: potentially large memory footprint of trie as number of subscriptions increases (issue #1400)

ssbl:
Solution: use a radix tree instead of a trie to track subscriptions

This PR also adds unit tests and benchmarks. Based on the discussion
in #1400, the radix tree uses 1/10th of the memory that the trie uses
for the same set of subscriptions. The benchmarks also show upto 3x
speedups in inserting, deleting and looking up subscriptions. These
results are suspect until further code review, although I've tried to
make sure the benchmarks are valid and correct.

Currently the code is a bit inconsistent formatting and style-wise,
but I'm pretty sure that will be easy to fix once we've ironed out any
correctness/design issues! Also, I've excluded the benchmarks from the
build since they use C++11 and add an external dependency.

cc @bluca @sigiesec @jamesdillonharvey

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -661,6 +661,7 @@ set(cxx-sources
   tcp_listener.cpp
   thread.cpp
   trie.cpp
+  radix_tree.cpp
   v1_decoder.cpp
   v1_encoder.cpp
   v2_decoder.cpp
--- a/src/radix_tree.cpp
+++ b/src/radix_tree.cpp
@@ -1,0 +1,578 @@
+/*
+    Copyright (c) 2018 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "precompiled.hpp"
+#include "macros.hpp"
+#include "err.hpp"
+#include "radix_tree.hpp"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+node::node (unsigned char *data) : data_ (data)
+{
+}
+
+uint32_t node::refcount ()
+{
+    uint32_t u32;
+    memcpy (&u32, data_, sizeof (u32));
+    return u32;
+}
+
+void node::set_refcount (uint32_t value)
+{
+    memcpy (data_, &value, sizeof (value));
+}
+
+uint32_t node::prefix_length ()
+{
+    uint32_t u32;
+    memcpy (&u32, data_ + sizeof (uint32_t), sizeof (u32));
+    return u32;
+}
+
+void node::set_prefix_length (uint32_t value)
+{
+    memcpy (data_ + sizeof (value), &value, sizeof (value));
+}
+
+uint32_t node::edgecount ()
+{
+    uint32_t u32;
+    memcpy (&u32, data_ + 2 * sizeof (uint32_t), sizeof (u32));
+    return u32;
+}
+
+void node::set_edgecount (uint32_t value)
+{
+    memcpy (data_ + 2 * sizeof (value), &value, sizeof (value));
+}
+
+unsigned char *node::prefix ()
+{
+    return data_ + 3 * sizeof (uint32_t);
+}
+
+void node::set_prefix (unsigned char const *bytes)
+{
+    memcpy (prefix (), bytes, prefix_length ());
+}
+
+unsigned char *node::first_bytes ()
+{
+    return prefix () + prefix_length ();
+}
+
+void node::set_first_bytes (unsigned char const *bytes)
+{
+    memcpy (first_bytes (), bytes, edgecount ());
+}
+
+unsigned char node::first_byte_at (size_t i)
+{
+    zmq_assert (i < edgecount ());
+    return first_bytes ()[i];
+}
+
+void node::set_first_byte_at (size_t i, unsigned char byte)
+{
+    zmq_assert (i < edgecount ());
+    first_bytes ()[i] = byte;
+}
+
+unsigned char *node::node_ptrs ()
+{
+    return prefix () + prefix_length () + edgecount ();
+}
+
+void node::set_node_ptrs (unsigned char const *ptrs)
+{
+    memcpy (node_ptrs (), ptrs, edgecount () * sizeof (void *));
+}
+
+node node::node_at (size_t i)
+{
+    zmq_assert (i < edgecount ());
+
+    unsigned char *data;
+    memcpy (&data, node_ptrs () + i * sizeof (void *), sizeof (data));
+    return node (data);
+}
+
+void node::set_node_at (size_t i, node n)
+{
+    zmq_assert (i < edgecount ());
+    memcpy (node_ptrs () + i * sizeof (void *), &n.data_, sizeof (n.data_));
+}
+
+void node::set_edge_at (size_t i, unsigned char byte, node n)
+{
+    set_first_byte_at (i, byte);
+    set_node_at (i, n);
+}
+
+bool node::operator== (node other) const
+{
+    return data_ == other.data_;
+}
+
+bool node::operator!= (node other) const
+{
+    return !(*this == other);
+}
+
+void node::resize (size_t prefix_length, size_t edgecount)
+{
+    size_t sz =
+      3 * sizeof (uint32_t) + prefix_length + edgecount * (1 + sizeof (void *));
+    unsigned char *new_data =
+      static_cast<unsigned char *> (realloc (data_, sz));
+    zmq_assert (new_data);
+    data_ = new_data;
+    set_prefix_length (static_cast<uint32_t> (prefix_length));
+    set_edgecount (static_cast<uint32_t> (edgecount));
+}
+
+node make_node (size_t refs, size_t bytes, size_t edges)
+{
+    size_t size = 3 * sizeof (uint32_t) + bytes + edges * (1 + sizeof (void *));
+
+    unsigned char *data = static_cast<unsigned char *> (malloc (size));
+    zmq_assert (data);
+
+    node n (data);
+    n.set_refcount (static_cast<uint32_t> (refs));
+    n.set_prefix_length (static_cast<uint32_t> (bytes));
+    n.set_edgecount (static_cast<uint32_t> (edges));
+    return n;
+}
+
+// ----------------------------------------------------------------------
+
+zmq::radix_tree::radix_tree () : root_ (make_node (0, 0, 0)), size_ (0)
+{
+}
+
+static void free_nodes (node n)
+{
+    for (size_t i = 0; i < n.edgecount (); ++i)
+        free_nodes (n.node_at (i));
+    free (n.data_);
+}
+
+zmq::radix_tree::~radix_tree ()
+{
+    free_nodes (root_);
+}
+
+match_result::match_result (size_t i,
+                            size_t j,
+                            size_t edge_index,
+                            size_t gp_edge_index,
+                            node current,
+                            node parent,
+                            node grandparent) :
+    nkey (i),
+    nprefix (j),
+    edge_index (edge_index),
+    gp_edge_index (gp_edge_index),
+    current_node (current),
+    parent_node (parent),
+    grandparent_node (grandparent)
+{
+}
+
+inline match_result zmq::radix_tree::match (const unsigned char *key,
+                                            size_t size,
+                                            bool check = false) const
+{
+    zmq_assert (key);
+
+    size_t i = 0;           // Number of characters matched in key.
+    size_t j = 0;           // Number of characters matched in current node.
+    size_t edge_idx = 0;    // Index of outgoing edge from the parent node.
+    size_t gp_edge_idx = 0; // Index of outgoing edge from grandparent.
+    node current_node = root_;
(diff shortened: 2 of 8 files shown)


unittests/unittest_radix_tree.cpp
```diff
@@ -0,0 +1,324 @@
...
+    for (size_t i = 0; i < size; ++i)
+        key.push_back (static_cast<char> (data[i]));
+    vec->push_back (key);
+}
+
+void test_apply ()
+{
+    zmq::radix_tree tree;
+
+    std::unordered_set<std::string> keys;
```
sigiesec: This is only available in C++11. But I think this can simply be changed to std::set

unittests/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ set(unittests
   unittest_mtrie
   unittest_ip_resolver
   unittest_udp_address
+  unittest_radix_tree
+  #  unittest_trie
+  #  unittest_trie_gb
```
sigiesec: These are no unit tests at all, but performance measurements, they should be renamed. 

It is not desirable to have code that is never built. I think it would be better to conditionally build them if the toolset is c++11 compliant. 

But probably this is not the right place, and they should not be run along with the unittests.

src/xsub.hpp
```diff
@@ -78,7 +79,8 @@ class xsub_t : public socket_base_t
     dist_t _dist;
 
     //  The repository of subscriptions.
-    trie_t _subscriptions;
+    // trie_t _subscriptions;
```
sigiesec: The commented out lines should be removed.

CMakeLists.txt
```diff
@@ -661,6 +661,7 @@ set(cxx-sources
   tcp_listener.cpp
   thread.cpp
   trie.cpp
+  radix_tree.cpp
```
sigiesec: Is trie still used elsewhere? If not, it should be removed.

unittests/unittest_radix_tree.cpp
```diff
@@ -0,0 +1,324 @@
...
+void setUp ()
+{
+}
+void tearDown ()
+{
+}
+
+bool tree_add (zmq::radix_tree &tree, const std::string &key)
+{
+    unsigned char *data = new unsigned char[key.size ()];
```
jamesdillonharvey: I know this is only test code but cant we just memcpy key.data() rather than doing  new/loop/delete

src/radix_tree.cpp
```diff
@@ -0,0 +1,602 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+node::node(unsigned char* data)
+    : data_(data)
+{}
+
+inline
```
sigiesec: It doesn't make sense to declare the methods inline in the cpp file. Either move them to the header file, or remove the inline declaration.

src/radix_tree.cpp
```diff
@@ -0,0 +1,602 @@
...
+        printf(" [*]");
+    printf("\n");
+    for (uint32_t i = 0; i < child_node.edgecount(); ++i)
+        visit_child(child_node.node_at(i), level + 1);
+}
+
+void zmq::radix_tree::print()
+{
+    puts("[root]");
+    for (uint32_t i = 0; i < root_.edgecount(); ++i)
```
sigiesec: This is probably only for testing purposes. Can this be moved to the tests? The library code should not write to stdout

CMakeLists.txt
bluca: no, keep it - I'll add a build flag for now, I don't want to do a new release with untested code

src/radix_tree.cpp
```diff
@@ -0,0 +1,602 @@
...
+    zmq_assert(new_data);
+    data_ = new_data;
+    set_prefix_length(static_cast<uint32_t>(prefix_length));
+    set_edgecount(static_cast<uint32_t>(edgecount));
+}
+
+node make_node(size_t refs, size_t bytes, size_t edges)
+{
+    size_t size = 3 * sizeof(uint32_t) + bytes
+        + edges * (1 + sizeof(void*));
```
sigiesec: Maybe define a struct for the "header" part? This would significantly simplify size calculations and field accesses.

src/radix_tree.cpp
```diff
@@ -0,0 +1,602 @@
...
+#include <string.h>
+
+node::node(unsigned char* data)
+    : data_(data)
+{}
+
+inline
+uint32_t node::refcount()
+{
+    uint32_t u32;
```
jamesdillonharvey: You could drop the mempy to the local and just cast here I think

`*((uint32_t*)data_)`

or whatever the correct c++ cast is.

CMakeLists.txt
ssbl: It seems the only files that use the trie now are the benchmarking scripts (`unittests/unittest_trie.cpp` and `unittests/unittest_trie_gb.cpp`). If this code isn't needed, I can move it out-of-tree elsewhere, a separate branch in my libzmq fork maybe. Does that sound like a good idea?

unittests/unittest_radix_tree.cpp
ssbl: Thanks, fixed!

unittests/unittest_radix_tree.cpp
ssbl: So something like `memcpy(data, key.data(), key.size())` instead of the loop? I think that could work as well.

jamesdillonharvey: not sure why you would want overhead of allocating memory just to delete in the same function.

src/radix_tree.cpp
ssbl: Thanks, moved the inline specifiers to the header file. GCC doesn't inline these functions if we remove the specifier; Clang doesn't have this problem, though.

unittests/unittest_radix_tree.cpp
ssbl: Oh I see the issue now, yes this would work if we pass the pointer as a const value. If we change the functions to use const and this should work; also we don't modify the data anyway so this looks like the right thing to do.

ssbl: Thanks, we're using `key.data()` directly now and the test functions are much cleaner.

src/radix_tree.cpp
bjovke: And probably most or all of the `memcpy`s in this file should be removed.

CMakeLists.txt
ssbl: Alright - we could try to port the benchmarks to C++98, so we'll just need to think about renaming and moving them elsewhere (this was suggested by @sigiesec in another comment). It seems the only significant C++11 features being used are the `<chrono>` and `<random>` headers; maybe writing a custom timer class and using another RNG should be enough to make it C++98-only.

src/radix_tree.cpp
```diff
@@ -0,0 +1,583 @@
...
+    memcpy(&u32, data_, sizeof(u32));
+    return u32;
+}
+
+void node::set_refcount(uint32_t value)
+{
+    memcpy(data_, &value, sizeof(value));
```
bjovke: No need to complicate things.
Calling `memcpy()` for this kind of thing is really too much:

`*reinterpret_cast<uint32_t *>(data_) = value;`

src/radix_tree.cpp
```diff
@@ -0,0 +1,583 @@
...
+}
+
+uint32_t node::prefix_length()
+{
+    uint32_t u32;
+    memcpy(&u32, data_ + sizeof(uint32_t), sizeof(u32));
+    return u32;
```
bjovke: Again, unnecessary too complicated.
The whole code in this function can be replaced by:

`return *(reinterpret_cast<uint32_t *>(data_) + 1);`

But, then again, as @sigiesec noted, the cleaner way would be to create a struct where `data_` points to.

src/radix_tree.cpp
```diff
@@ -0,0 +1,583 @@
...
+}
+
+void node::set_prefix_length(uint32_t value)
+{
+    memcpy(data_ + sizeof(value), &value, sizeof(value));
```
bjovke: Too complicated. Function call is overkill.
The code could be this instead:

`*(reinterpret_cast<uint32_t *>(data_) + 1) = value;`

src/radix_tree.cpp
bjovke: @bluca 
Yes, you're right, there are more important stuff.
But there's a lot of code like this one and it might be challenging to change it later.
`memcpy()` for small sizes is optimized if compiler implements it as intrinsic function or is very good at inlining.
Even then there's some excess machine code to check for length and alignment of copied block.
So there's **no way** that this could be as optimal as cast.
Another thing. This `memcpy()` is hiding the cast which is still happening. I generally prefer casts to be explicit because they generally mean that you have a mismatched design.

But I think that the real solution is what @sigiesec said, to make a struct and/or union so everything would be much cleaner.

src/radix_tree.cpp
ssbl: Correct me if I got this wrong, would it look something like this?

```
struct header { /* 3 uint32_ts here */ };

struct node
{
    header hdr;
    unsigned char *data;
};
```

I'm not sure how this would affect the data layout. AFAIK the cleaner way would be to use C99's flexible array members, but that isn't standard C++. Using flexible array members will change `node` to:

```
struct node
{
    uint32_t refcount;
    uint32_t prefix_length;
    uint32_t edgecount;
    unsigned char data[];
};
```

This would eliminate the getter/setter functions for the first 3 fields with no change in the data layout.

cc @bjovke

bluca:
@ssbl two things:

1) remove both benchmarks completely for now
2) fix the formatting, you can use the cmake/make targets as the pull request template suggests
3) rebase and squash all commits into one apart from the relicensing one (keep that separate)

Then we can merge, and do any more fixups and take care of the benchmarks in a different PR, in the spirit of C4

ssbl:
@bluca - Done, but it looks like one of the tests timed out in the travis build. Formatting seems to be fixed.

src/radix_tree.cpp
bjovke: @ssbl Flexible array members are not standard in C++.
One of the ways to do this is to preserve the `data_` member as it is and cast it to `struct node` when needed.
You might consider making `data_` void pointer since it's already being casted many times. It would be consistent with `malloc()`/`free()`.
This would make the code more verbose but also clearer and less error prone.
Also, whenever pointer arithmetic is being done it's a good idea to put some comments.

For example:

```
struct node
{
    // Header is located at the beginning of data_ block.
    struct header
    {
        uint32_t refcount;
        uint32_t prefix_length;
        uint32_t edgecount;
    }

    void *data_;
};
```

`make_node()` might look like this:

```
node make_node (size_t refs, size_t bytes, size_t edges)
{
    node n (malloc (sizeof (node::header) + bytes + edges * (1 + sizeof (void *))));
    zmq_assert (n.data_);

    node::header &hdr = *reinterpret_cast<node::header *>(n.data_);
    hdr.refcount = refs;
    hdr.prefix_length = bytes;
    hdr.edgecount = edges;

    return n;
}
```
prefix_length() might look like this:

```
uint32_t node::prefix_length ()
{
    return reinterpret_cast<node::header *>(data_)->prefix_length;
}
```

bluca:
Don't worry about that test, it doesn't even use pub/sub and every now and then it fails, it's a known issue.
There are a couple of pub-sub tests failing on Windows though, so please have a look at the appveyor log.

In the meanwhile, I'll  merge this and then send another PR to make the new implementation used only in draft builds. Please send another PR to follow up on the other changes and benchmark and so on.

src/radix_tree.cpp
ssbl: @bjovke Thanks, that makes sense!

ssbl:
Thanks @bluca ! I've identified the failing tests from the logs, I'll look into fixing this and adding the above changes now.

src/radix_tree.cpp
bjovke: @bluca You're right, `memcpy ()` is really optimized out in this case (checked on MSVC 15, release, optimization on), compiler produces the same machine code as if it was a typecast.
But for me it doesn't look readable and the intent of the coder is not clear at first sight.
Anyway, this is minor thing so it can be as it is if you ask me.

sigiesec: When the data structure is accessed this way (i.e. without an underlying struct), it should be done with memcpy. While this is probably ok when all of the the header fields are of the same type (it is essentially an array of uint32_t with size 3), this may easily break if some of the header fields is changed to a different type. In that case, using reinterpret_cast may still work and produce the same code on some/most compilers/architectures, but is not portable, since it violates strict aliasing rules. See, e.g. https://en.cppreference.com/w/cpp/language/reinterpret_cast

src/radix_tree.cpp
jamesdillonharvey: Leave as is, as pointed out by @sigiesec the memcpy is the correct way to go due to aliasing rules.

src/radix_tree.cpp
bjovke: @sigiesec There's no type safety here whatever is used, `memcpy()` or `reinterpret_cast`. `memcpy()` is anyway doing the same casting as `reinterpret_cast` since no types are involved. It's just that `reinterpret_cast` is explicit of what's going on and `memcpy()` is not.
I don't think that aliasing rules are violated in any way since since `uint32_t` is stored in that memory and read again like `uint32_t`, the type is the same. If they are violated then they are violated for both `memcpy()` and `reinterpret_cast`.
What **is** violated is memory alignment of pointers:

```
            // Make room for the new edge. We need to shift the chunk
            // of node pointers one byte to the right. Since resize()
            // increments the edgecount by 1, node_ptrs() tells us the
            // destination address. The chunk of node pointers starts
            // at one byte to the left of this destination.
            //
            // Since the regions can overlap, we use memmove.
            memmove (current_node.node_ptrs (), current_node.node_ptrs () - 1,
                     (current_node.edgecount () - 1) * sizeof (void *));
```
If `std::vector` was used, which is the case for some other implementations of this trie, life would be much easier.

ssbl: My understanding was that using memmove and memcpy would be the safest choice when dealing with raw memory. `node_ptrs` returns an `unsigned char *` so the call to memmove should work as expected, or maybe I'm missing something here?

bjovke: @ssbl 
_My understanding was that using memmove and memcpy would be the safest choice when dealing with raw memory._

You're not dealing with raw memory in a sense that you're processing some data stream from the network or file. The data stored is well known and formed. So the logical choice is to use C++ container objects (`std::vector`, `std::string`) and structs. But since you started it this way then it doesn't matter.
When using `memxxxx()` functions there's also no bounds checking: [SL.con.3: Avoid bounds errors](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#slcon3-avoid-bounds-errors)

Yes, the call to `memmove()` works, I wasn't speaking about that. The memory block where pointers are stored is moved by 1 byte:
  - `current_node.node_ptrs () - 1`

And then data in this block is used as pointers to nodes.
My remark is that the pointers themselves are not aligned in memory to a 4 or 8 byte boundary and are being accessed. This causes high performance penalty on x86 and on some CPUs might not work at all. There's also a case when it doesn't work on x86 when MMX/SSE optimizations are switched on.

PR #3287, opened 2018-10-31
Problem: new radix tree implementation is experimental

bluca:
Solution: use it only for DRAFT builds for now, to leave it time to be
tested and stabilised.

### DIFF
--- a/src/xsub.hpp
+++ b/src/xsub.hpp
@@ -34,7 +34,11 @@
 #include "session_base.hpp"
 #include "dist.hpp"
 #include "fq.hpp"
+#ifdef ZMQ_BUILD_DRAFT_API
 #include "radix_tree.hpp"
+#else
+#include "trie.hpp"
+#endif
 
 namespace zmq
 {
@@ -78,7 +82,11 @@ class xsub_t : public socket_base_t
     dist_t _dist;
 
     //  The repository of subscriptions.
+#ifdef ZMQ_BUILD_DRAFT_API
     radix_tree _subscriptions;
+#else
+    trie_t _subscriptions;
+#endif
 
     //  If true, 'message' contains a matching message to return on the
     //  next recv call.


bluca:
@sigiesec ping

sigiesec:
It's not really a DRAFT API, but a DRAFT implementation facet, so controlling it via ZMQ_BUILD_DRAFT_API is a bit confusing, but probably not too much ;)

bluca:
Yeah I know, but better than a specific build option that nobody would see :-) We need real usage to test it properly

WallStProg:
I would suggest that it's not such a good idea to lump the new radix tree implementation in with the draft APIs in general.  Some of us (umm, like me), are happily using draft functionality after having been reassured that doing so was reasonable (https://lists.zeromq.org/pipermail/zeromq-dev/2017-October/031960.html).

OTOH, the radix tree implementation is just that -- an implementation -- and as such one cannot choose to use it or not -- if drafts are enabled, then that's what you get (apparently).  I'm uncomfortable with that given that this feature is relatively new and untested. 

I would suggest a separate flag to enable the radix tree, distinct from ZMQ_BUILD_DRAFT_API.  I'm happy to do a PR (for CMake) if that would help.

Thanks!

bluca:
The problem, as I mentioned, is that a separate flag means it's not going to be used by anybody. We need real world tests for it before it's enabled by default.

If you want, you can add an additional flag, but that if not specified takes the value of _DRAFT. That way it's enabled by default in draft builds, but can be overridden.

bluca:
or to put it another way: at _some point_ it's going to get enabled one way or the other. It's better if before that there was real testing.

WallStProg:
I understand your rationale, but I'm preparing a build that will be deployed to prod, and there is no way I'm going to let that go out with the radix tree implementation at this point.  

So I need a way to disable radix trees without disabling ALL of draft functionality (some of which has been in draft state literally for years, AND which must be deliberately chosen by the developer).

I'm fine with a flag that defaults to the value of _DRAFT but than can be overridden.  I'll put something together and submit a PR.

PR #3288, opened 2018-11-01
Problems: wrong version and outdated NEWS

bluca:
The NEWS is only half complete, I'll finish it in the next few days.

Yes, the version is going backwards, but such are the risks of using development versions! Also the ABI does not change so the filename is the same, and the packages built on OBS use the previous version (+date) so they are not affected.


bluca:
@sigiesec ping

PR #3290, opened 2018-11-02
Problem: radix tree needs benchmarks and improvements

ssbl:
Solution: add a benchmark test and make suggested improvements

This PR adds a script for comparing lookup performance and memory
consumption of the trie and radix tree implementations. The benchmark
shows close to 2x improvement in lookup performance (13-15ms vs
30-35ms). Running heaptrack on the benchmark shows more than 10x
reduction in memory consumption (peak 438.1 kB vs 5.2 MB for 10K
subscriptions).

This also adds changes suggested in #3286:

- removed unused print function
- moved an inline specifier into the header file for that function

Reviews for changes from #3286 also welcome! Any thoughts on using an
external dependency like google benchmark appreciated.

cc @bluca @sigiesec @jamesdillonharvey @bjovke 

[heaptrack.test_benchmark_radix_tree.11015.gz](https://github.com/zeromq/libzmq/files/2542885/heaptrack.test_benchmark_radix_tree.11015.gz)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1209,6 +1209,15 @@ if(BUILD_SHARED)
           COMPONENT PerfTools)
       endif()
     endforeach()
+
+    if(BUILD_STATIC)
+      add_executable(benchmark_radix_tree perf/benchmark_radix_tree.cpp)
+      target_link_libraries(benchmark_radix_tree libzmq-static)
+      target_include_directories(benchmark_radix_tree
+        PUBLIC
+        "${CMAKE_SOURCE_DIR}/src")
+    endif()
+
   endif()
 elseif(WITH_PERF_TOOL)
   message(FATAL_ERROR "Shared library disabled - perf-tools unavailable.")
--- a/Makefile.am
+++ b/Makefile.am
@@ -377,7 +377,9 @@ if ENABLE_STATIC
 noinst_PROGRAMS += \
 	perf/benchmark_radix_tree
 
-perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
+perf_benchmark_radix_tree_CPPFLAGS = -I$(top_srcdir)/src
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a \
+	${src_libzmq_la_LIBADD}
 perf_benchmark_radix_tree_SOURCES = perf/benchmark_radix_tree.cpp
 endif
 endif
--- a/Makefile.am
+++ b/Makefile.am
@@ -372,6 +372,14 @@ perf_inproc_lat_SOURCES = perf/inproc_lat.cpp
 
 perf_inproc_thr_LDADD = src/libzmq.la
 perf_inproc_thr_SOURCES = perf/inproc_thr.cpp
+
+if ENABLE_STATIC
+noinst_PROGRAMS += \
+	perf/benchmark_radix_tree
+
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
+perf_benchmark_radix_tree_SOURCES = perf/benchmark_radix_tree.cpp
+endif
 endif
 
 if ENABLE_CURVE_KEYGEN
@@ -913,8 +921,7 @@ test_apps += tests/test_poller \
 	tests/test_scatter_gather \
 	tests/test_dgram \
 	tests/test_app_meta \
-	tests/test_router_notify \
-	tests/test_benchmark_radix_tree
+	tests/test_router_notify
 
 tests_test_poller_SOURCES = tests/test_poller.cpp
 tests_test_poller_LDADD = src/libzmq.la ${UNITY_LIBS}
@@ -949,10 +956,6 @@ tests_test_app_meta_CPPFLAGS = ${UNITY_CPPFLAGS}
 tests_test_router_notify_SOURCES = tests/test_router_notify.cpp
 tests_test_router_notify_LDADD = src/libzmq.la ${UNITY_LIBS}
 tests_test_router_notify_CPPFLAGS = ${UNITY_CPPFLAGS}
-
-tests_test_benchmark_radix_tree_SOURCES = tests/test_benchmark_radix_tree.cpp
-tests_test_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
-tests_test_benchmark_radix_tree_CPPFLAGS = -I$(top_srcdir)/src
 endif
 
 if ENABLE_STATIC
--- a/Makefile.am
+++ b/Makefile.am
@@ -377,9 +377,9 @@ if ENABLE_STATIC
 noinst_PROGRAMS += \
 	perf/benchmark_radix_tree
 
+perf_benchmark_radix_tree_DEPENDENCIES = src/libzmq.la
 perf_benchmark_radix_tree_CPPFLAGS = -I$(top_srcdir)/src
-perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a \
-	${src_libzmq_la_LIBADD}
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
 perf_benchmark_radix_tree_SOURCES = perf/benchmark_radix_tree.cpp
 endif
 endif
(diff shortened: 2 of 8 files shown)


ssbl:
It seems I need to add coverage flags to the new file's CPPFLAGS, but I'd like to work on the reviews before triggering another build.

bluca:
For the external dependency, it's fine to add it as long as it's optional and disabled by default. Then we can choose to turn it on in specific CI jobs.

bluca:
Rather than a unit test, what about a new program in the perf directory? I think it would make more sense

ssbl:
Sounds good to me! We could look into making the benchmark output less verbose as well.

CMakeLists.txt
```diff
@@ -1209,11 +1209,24 @@ if(BUILD_SHARED)
           COMPONENT PerfTools)
       endif()
     endforeach()
+
+    if(ENABLE_DRAFTS)
```
bluca: the objects are always built, so you don't need the enable_drafts here or elsewhere

Makefile.am
```diff
@@ -948,6 +952,9 @@ tests_test_app_meta_CPPFLAGS = ${UNITY_CPPFLAGS}
 tests_test_router_notify_SOURCES = tests/test_router_notify.cpp
 tests_test_router_notify_LDADD = src/libzmq.la ${UNITY_LIBS}
 tests_test_router_notify_CPPFLAGS = ${UNITY_CPPFLAGS}
+
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
```
bluca: this is duplicated and can be removed

Makefile.am
```diff
@@ -913,7 +917,7 @@ test_apps += tests/test_poller \
 	tests/test_scatter_gather \
 	tests/test_dgram \
 	tests/test_app_meta \
-	tests/test_router_notify
+	tests/test_router_notify \
```
bluca: this will likely break the build and is not necessary anymore

ssbl: Sorry, I pushed to master before checking everything, fixing this now.

Makefile.am
```diff
@@ -372,6 +372,14 @@ perf_inproc_lat_SOURCES = perf/inproc_lat.cpp
 
 perf_inproc_thr_LDADD = src/libzmq.la
 perf_inproc_thr_SOURCES = perf/inproc_thr.cpp
+
+if ENABLE_STATIC
+noinst_PROGRAMS += \
+	perf/benchmark_radix_tree.cpp
```
bluca: you need to remove .cpp and that should fix the build

ssbl: I tried this, but the build failed with: `no rule to make target 'src/.libs/libzmq.a'`. The cmake build seems to be working though.

bluca:
Also the Windows build fails with:

```
93>..\libzmq\perf\benchmark_radix_tree.cpp(33) : fatal error C1083: Cannot open include file: 'chrono': No such file or directory
```

if the libraries are not available you can just exclude it for now from windows

Makefile.am
```diff
@@ -372,6 +372,14 @@ perf_inproc_lat_SOURCES = perf/inproc_lat.cpp
 
 perf_inproc_thr_LDADD = src/libzmq.la
 perf_inproc_thr_SOURCES = perf/inproc_thr.cpp
+
+if ENABLE_STATIC
+noinst_PROGRAMS += \
+	perf/benchmark_radix_tree
+
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a
```
bluca: try with:

```
perf_benchmark_radix_tree_CPPFLAGS = -I$(top_srcdir)/src
perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a \
	${src_libzmq_la_LIBADD}
```

sigiesec:
> Also the Windows build fails with:
> 
> ```
> 93>..\libzmq\perf\benchmark_radix_tree.cpp(33) : fatal error C1083: Cannot open include file: 'chrono': No such file or directory
> ```
> if the libraries are not available you can just exclude it for now from windows

VC2010 does not have <chrono>, but it is available from VC2012 upwards

perf/benchmark_radix_tree.cpp
```diff
@@ -0,0 +1,113 @@
...
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <trie.hpp>
+#include <radix_tree.hpp>
```
bluca: both of these are not system includes, they need to be quoted:

```
#include "trie.hpp"
#include "radix_tree.hpp"
```

otherwise the build fails

ssbl: I tried this but it fails with the same error. It looks like a dependency issue; the `src/libzmq.la` target builds `libzmq.a`, so adding this line seems to fix it:

```
perf_benchmark_radix_tree_DEPENDENCIES = src/libzmq.la
```

I'll add this if everything works.

Makefile.am
```diff
@@ -372,6 +372,16 @@ perf_inproc_lat_SOURCES = perf/inproc_lat.cpp
...
 perf_inproc_thr_SOURCES = perf/inproc_thr.cpp
+
+if ENABLE_STATIC
+noinst_PROGRAMS += \
+	perf/benchmark_radix_tree
+
+perf_benchmark_radix_tree_CPPFLAGS = -I$(top_srcdir)/src
+perf_benchmark_radix_tree_LDADD = $(top_builddir)/src/.libs/libzmq.a \
+	${src_libzmq_la_LIBADD}
+perf_benchmark_radix_tree_SOURCES = perf/benchmark_radix_tree.cpp
```
bluca: to finally fix the build (i've tested it), add:

```
perf_benchmark_radix_tree_DEPENDENCIES = src/libzmq.la
```

For some reasons automake is not picking up the dependency automatically, so let's just add it manually.

perf/benchmark_radix_tree.cpp
bluca: lol, I just suggested exactly the same thing. You need this too though.

ssbl: Yup, also i tested it without the `${src_libzmq_la_LIBADD}` line and it works.

bluca:
> VC2010 does not have , but it is available from VC2012 upwards

@sigiesec can this be worked around in CMake?

Makefile.am
ssbl: The unit tests also have this dependency though, shouldn't there be a problem with those too? Maybe because they are built after `tests`, which needs `libzmq.la`?

bluca: Yes, probably the tests target is always built after the programs/libraries targets, I'd assume (but I really really don't want to go and look into automake to confirm...)

sigiesec:
> > VC2010 does not have , but it is available from VC2012 upwards
> 
> @sigiesec can this be worked around in CMake?

A check such as
```
check_include_files(chrono ZMQ_HAVE_CHRONO)
```
can be done in the CMakeLists.txt, and then ZMQ_HAVE_CHRONO can be used in the source file.

That's no workaround, but a check for availability of certain features, if that's what you mean?

bluca:
Yeah, that's what I meant, thanks (hadn't really looked at what was missing)

ssbl:
I've fixed this by using `#if HAVE_CXX11` and falling back to an empty main function if this check fails. Is there a better or preferred way to do it?

bluca:
Is C++ 11 required?

ssbl:
I'm not sure. I couldn't find a high precision timer in the standard library, so I used chrono which was added in C++11. If the benchmark logic looks fine, we can probably make it C++98-only.

bluca:
No need to overcomplicate it, use the ifdef if that works, the perf programs are not critical so they can be restricted to newer compilers of needed

CMakeLists.txt
```diff
@@ -1209,6 +1209,15 @@ if(BUILD_SHARED)
           COMPONENT PerfTools)
       endif()
     endforeach()
+
+    if(BUILD_STATIC)
```
bluca: Here you can do:

```
check_include_files(chrono ZMQ_HAVE_CHRONO)
if(BUILD_STATIC AND ZMQ_HAVE_CHRONO)
```

And it should fix the Windows build

ssbl: I don't think there should be a problem on Windows now since we're using an ifdef.

bluca:
Looks good now, thanks!

ssbl:
@bluca - Thanks!

PR #3291, opened 2018-11-02
Some changes for CMakeLists.txt:

WallStProg:
Problem: not all options are available using CMake build
Solution: add options to CMakeLists.txt

- MacOS version requires CMake version 3.0.2 (because of policy CMP0042)
- Add option to build instrumented binaries with Address Sanitizer
- Add option to select compiler intrinsics for atomic ops
- Only build docs on request (saves build time)


bluca:
@WallStProg have you already sent a relicense grant? https://github.com/zeromq/libzmq/tree/master/RELICENSE

Issue #3263, opened 2018-09-28
zmq raise exception in I/O thread


vibk:
when we send bad data into zmq socket, I/O thread raises exception and process gets killed. 

Call stack of crash is as below

00 KERNELBASE!RaiseException+0x68
01 zmq::zmq_abort(char * errmsg_ = <Value unavailable error>)+0x21 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\err.cpp @ 89]
02 zmq::tcp_read(unsigned int64 s_ = <Value unavailable error>, void * data_ = <Value unavailable error>, unsigned int64 size_ = <Value unavailable error>)+0xa9 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\tcp.cpp @ 276]
03 zmq::stream_engine_t::in_event(void)+0x132 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\stream_engine.cpp @ 318]
04 zmq::select_t::trigger_events(class std::vector<zmq::select_t::fd_entry_t,std::allocator<zmq::select_t::fd_entry_t> > * fd_entries_ = 0x0000019d`323400c8, struct zmq::select_t::fds_set_t * local_fds_set_ = 0x0000006f`e3f9fb90, int event_count_ = 0n2)+0x74 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\select.cpp @ 122]
05 zmq::select_t::select_family_entry(struct zmq::select_t::family_entry_t * family_entry_ = 0x0000019d`323400c8, int max_fd_ = 0n0, bool use_timeout_ = <Value unavailable error>, struct timeval * tv_ = 0x0000006f`e3fffc98)+0xe9 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\select.cpp @ 404]
06 zmq::select_t::loop(void)+0x352 [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\select.cpp @ 360]
07 thread_routine(void * arg_ = <Value unavailable error>)+0xd [objdir-windows-msvc-14.0.x86_64\build\libzmq\src\thread.cpp @ 47]
08 invoke_thread_procedure+0xe (Inline Function @ 00007ff7`50b99cf9) [d:\th\minkernel\crts\ucrt\src\appcrt\startup\thread.cpp @ 91]

Can anyone please help how to catch this exception ?

bluca:
What do you mean by "bad data"? Can you provide an example program to reproduce it?

[...]

PR #3297, opened 2018-11-05
Windows apparently returns `WSAENOBUFS` from `recv`

toonetown:
If the buffer size is huge, `WSAENOBUFS` appears to be returned from `recv` - this is despite the fact that it's not explicitly listed at https://support.microsoft.com/en-us/kb/201213.  See https://github.com/zeromq/libzmq/issues/3263

This PR attempts to address that issue by expecting `WSAENOBUFS` as a valid error that may be returned.


bluca:
Please follow the guidelines: https://github.com/zeromq/libzmq/blob/master/.github/PULL_REQUEST_TEMPLATE.md

bluca:
Thanks @toonetown

PR #3301, opened 2018-11-09
Problems: some tests still use hard-coded TCP ports, wrong usage of CPPFLAGS, missing NEWS, no way to force C++98 builds and CI for it

bluca:
Solutions: fix them all

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,7 @@ test_reqrep_device_tipc
 test_reqrep_tipc
 test_router_handover
 test_router_mandatory_tipc
+test_router_notify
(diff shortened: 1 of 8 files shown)


bluca:
@sigiesec can we make the new monitoring events STABLE now? I think they've been committed for more than a year, and more than one release

sigiesec:
@bluca I think they can be declared STABLE. They are of limited use without extending the event data structure, but that is a different issue.

bluca:
Ok thanks, I'll send a separate pr later tonight. Can you merge this in the meanwhile, please?

Issue #3302, opened 2018-11-10
libzmq project doesn't have reference to C4.1


crocket:
# Issue description

It is one of the core projects of zeromq. Yet, I do not see reference to C4.1 in this project.

Did developers decide to not observe or stick to C4.1?

C4.1 is the best contribution policy that I have ever seen so far. If there is a better alternative, let me know.

bluca:
It's linked in the pr template, feel free to send a pr to mention it elsewhere too https://github.com/zeromq/libzmq/blob/master/.github/PULL_REQUEST_TEMPLATE.md

PR #3303, opened 2018-11-10
Problem: C4 process is not easily discoverable.

crocket:
Solution: Add 'Contributing' section to README.md

Do I need a relicensing grant? I vaguely remember that I consented to MPLv2 in a github issue, but I'm not totally sure.

It fixes https://github.com/zeromq/libzmq/issues/3302


bluca:
> Do I need a relicensing grant? I vaguely remember that I consented to MPLv2 in a github issue, but I'm not totally sure.

I don't see it in the directory, so let's take it just in case. Thanks.

PR #3312, opened 2018-11-21
Android Things 1.06 with Termux changes

brubakerjeff:
Please consider the following changes to the configuration scripts. After making the changes below ( macro to check for libatomic and enhanced verification for tipc[I don't have a tipc unit to verify with but inserted the line that was causing an error on this configuration into configure.ac]) I was able to compile with termux under Android 1.06 and succesfully use test cases 

[still had to manually change configure script to use bash in this context but think that may normally be encapsulated in termux build scripts]
./configure --prefix=/data/data/com.termux/files/usr

This effort was necessitated to recompile libzmq from scratch in support of addressing this issue https://github.com/zeromq/pyzmq/issues/1240

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -89,7 +89,6 @@ fi
 # Check whether to build a with debug symbols
 LIBZMQ_CHECK_ENABLE_DEBUG
 
-AC_CHECK_LIB([atomic],[__atomic_load_4],[LIBS="${LIBS} -latomic"]) 
 
 # Check whether to enable code coverage
 LIBZMQ_WITH_GCOV
@@ -196,7 +195,8 @@ case "${host_os}" in
         case "${host_os}" in
             *android*)
                 AC_DEFINE(ZMQ_HAVE_ANDROID, 1, [Have Android OS])
-                libzmq_on_android="yes"
+                AC_CHECK_LIB([atomic],[__atomic_load_4],[LIBS="${LIBS} -latomic"]) 
+		libzmq_on_android="yes"
             ;;
         esac
         ;;
--- a/configure.ac
+++ b/configure.ac
@@ -89,6 +89,8 @@ fi
 # Check whether to build a with debug symbols
 LIBZMQ_CHECK_ENABLE_DEBUG
 
+AC_CHECK_LIB([atomic],[__atomic_load_4],[LIBS="${LIBS} -latomic"]) 
+
 # Check whether to enable code coverage
 LIBZMQ_WITH_GCOV
 
--- a/configure.ac
+++ b/configure.ac
@@ -110,6 +110,7 @@ AC_RUN_IFELSE(
             memset(&topsrv, 0, sizeof(topsrv));
             topsrv.family = AF_TIPC;
             topsrv.addrtype = TIPC_ADDR_NAME;
+            topsrv.addr.name.domain = tipc_addr (10, 10, 10); 
             topsrv.addr.name.name.type = TIPC_TOP_SRV;
             topsrv.addr.name.name.instance = TIPC_TOP_SRV;
             fcntl(sd, F_SETFL, O_NONBLOCK);
--- a/RELICENSE/jeffbrubaker.md
+++ b/RELICENSE/jeffbrubaker.md
@@ -1,0 +1,14 @@
+
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+This is a statement by Simon Giesecke that grants permission to
+relicense its copyrights in the libzmq C++ library (ZeroMQ) under the
+Mozilla Public License v2 (MPLv2) or any other Open Source Initiative
+approved license chosen by the current ZeroMQ BDFL (Benevolent
+Dictator for Life).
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -89,6 +89,8 @@ fi
 # Check whether to build a with debug symbols
 LIBZMQ_CHECK_ENABLE_DEBUG
 
+AC_CHECK_LIB([atomic],[__atomic_load_4],[LIBS="${LIBS} -latomic"])
```
bluca: This will always enable this linking, which is not right. There's a section that checks the host os, and looks for android, you should add it there instead with the appropriate checks for "android things"

bluca:
thanks - please send a separate PR with a relicensing grant, see bottom of PR template: https://github.com/zeromq/libzmq/blob/master/.github/PULL_REQUEST_TEMPLATE.md

bluca:
actually I double checked, and there's already an m4 test to see if linking to latomic is needed, so check the log instead and see why it is not triggering: https://github.com/zeromq/libzmq/pull/3083/files

brubakerjeff:
Can I reply in this email? Is there a stakeholder for the tipc change? I
don’t see a way to test the positive case (when tipc is present)

bluca:
the tipc change is fine, it's the latomic one that is not as it overwrites the work done by the linked pr in acinclude.m4

brubakerjeff:
from configure
"checking for __atomic_load_4 in -latomic... yes"

which means this should be triggered.

[AC_MSG_RESULT(yes) ; libzmq_cv_has_atomic_instrisics="yes" LIBS="-latomic" ; $1], 

My understanding is in the current version LIBS= was factored into the test case in the new change?

https://github.com/PIDriverAssist/libzmq/blob/4f48c5615ecbbd70423e163151a3448e75570f99/acinclude.m4#L694
 [AC_MSG_RESULT(yes) ; libzmq_cv_has_atomic_instrisics="yes" ; $1],

changed at 
https://github.com/PIDriverAssist/libzmq/commit/6de5f18be3baeccd2c7c4b095a1c09dbecb779a4#diff-a5522b90382504dc530177f63ced5209
Note is "with -latomic if needed however using LDFLAGS doesn't work when statically linking because LDFLAGS is added before LIBS"

If you can confirm my perception, a thought is to reinstate setting LIBS under condition that linking is not static?

bluca:
LIBS is already been set (or rather, it's not reset back if it's needed), so the problem must be something else

bluca:
if you build with make V=1 what's the actual content of *FLAGS?

brubakerjeff:
from configure..
checking whether compiler supports __atomic_Xxx intrinsics... no

running make
$ make V=1
Making all in doc
make[1]: Entering directory '/data/data/com.termux/files/home/libzmq-3/libzmq/doc'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/data/data/com.termux/files/home/libzmq-3/libzmq/doc'
make[1]: Entering directory '/data/data/com.termux/files/home/libzmq-3/libzmq'
/bin/sh ./libtool  --tag=CXX   --mode=link g++  -pedantic -Werror -Wall -Wno-long-long -g -O2 -Wno-tautological-constant-compare   -o tools/curve_keygen tools/curve_keygen.o src/libzmq.la -lrt -lpthread 
libtool: link: g++ -pedantic -Werror -Wall -Wno-long-long -g -O2 -Wno-tautological-constant-compare -o tools/.libs/curve_keygen tools/curve_keygen.o  src/.libs/libzmq.so -lrt -lpthread -Wl,-rpath -Wl,/usr/local/lib
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_fetch_sub_4'
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_load_4'
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_exchange_4'
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_compare_exchange_4'
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_fetch_add_4'
/data/data/com.termux/files/usr/bin/arm-linux-androideabi-ld: src/.libs/libzmq.so: undefined reference to `__atomic_store_4'

bluca:
are you passing all the cross compilation flags for autoconf?

brubakerjeff:
I'm compiling on device. Would they be relevant?

brubakerjeff:
Thank you. I think we will need another iteration as we understand this more. I'm thinking the branches should condition upon presence of atomic intrinsic and whether or not libzmq is statically compiled [based upon the notes] (none of this logic is encapsulated by lower level tools?) but I will read more to confirm this.

bluca:
Ok, feel free to send more pull requests

PR #3314, opened 2018-11-26
Consolidation of libatomic checks

brubakerjeff:
Per discussions, lets merge the two checks for atomic @bluca  @ffontaine 

This from my perspective accomplishes the goal of the test, which is to show the linker has to include "latomic"

On compilation the same error is produced. (CXXFLAGS is reset so the result below is produced) 

In file included from ./src/ctx.hpp:44:
./src/atomic_counter.hpp:110:21: error: misaligned or large atomic operation may incur significant performance penalty
      [-Werror,-Watomic-alignment]
        old_value = __atomic_fetch_add (&_value, increment_, __ATOMIC_ACQ_REL);"

Ultimately I removed the save CXX flags citing the bug report previously mentioned.
https://bugs.llvm.org/show_bug.cgi?id=38593

Out of curiosity, what is the basis for using an atomic add operation in C++? Is the built in + not atomic?

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -681,7 +681,6 @@ int main (int, char **)
 
     if test "x$GCC_ATOMIC_BUILTINS_SUPPORTED" != x1; then
         save_LIBS=$LIBS
-        CXXFLAGS="$CXXFLAGS -Wno-atomic-alignment"
         LIBS="$LIBS -latomic"
         AC_LINK_IFELSE([AC_LANG_SOURCE([
         /* atomic intrinsics test */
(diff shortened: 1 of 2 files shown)


bluca:
That flag cannot just be added, as the compiler might not support it. Instead please do the same as it's already done for the tautological comparison warning in configure.ac

brubakerjeff:
Updated per feedback.

bluca:
Thanks

brubakerjeff:
Thank you

PR #3326, opened 2018-12-08
Switching or desactivate wifi can crash Android application

hgourvest:
Problem: Switching or desactivate wifi can lead to have EINVAL error code on Android
Solution: do not abort when this error happen


bluca:
I don't think this is right. The socket is dead if accept fails with that error. At the very least ifdef it for Android app that it doesn't cause issues on other systems.

hgourvest:
It seems reasonable, I pushed a correction to limit this behavior to Android.

bluca:
I think the whole expression has to be ifdeffed, it breaks the build:

../src/tcp_listener.cpp:311:1: error: embedding a directive within macro arguments is not portable [-Werror]

Also please rebase and squash the fixups to avoid breaking bisect

Issue #3329, opened 2018-12-12
Using zmq_msg_data/zmq_msg_size after zmq_msg_close results in SIGABRT


joelfrederico:
# Issue description

Calling zmq_msg_size/zmq_msg_data on a closed zmq_msg_t results in a ZMQ crash. It seems to me that we could remove the crash without changing the API. Both crashes are triggered off of a zmq_assert.

https://github.com/zeromq/libzmq/blob/b70d31c432f1dced4e65a83959787e7f4a3ba0d7/src/msg.cpp#L327 and https://github.com/zeromq/libzmq/blob/b70d31c432f1dced4e65a83959787e7f4a3ba0d7/src/msg.cpp#L347.)

Both of these are triggered off of a check() call. If we are going to check(), why not prevent the crash?

- [zmq_msg_size](http://api.zeromq.org/master:zmq-msg-size) says nothing about a closed message. We could return 0 and still be within the documented behavior (although this isn't quite ideal).
- [zmq_msg_data](http://api.zeromq.org/master:zmq-msg-data) says nothing about a closed message either. We could return NULL and still be within the documented behavior. We could also set errno to EFAULT, a la http://api.zeromq.org/4-2:zmq-msg-close.

We could also update the documentation with whatever behavior we decide to use. We could also deprecate zmq_msg_size and release a new zmq_msg_size:

```
int zmq_msg_size(zmq_msg_t *msg, size_t *size);
```

This version would return 0 on success and -1 on error, and set errno to EFAULT, a la http://api.zeromq.org/4-2:zmq-msg-close.

# Environment

* libzmq version (commit hash if unreleased): 4.3.0
* OS: macOS Sierra

# Minimal test code / Steps to reproduce the issue

```
extern "C" {
#include <zmq.h>
}

#include <iostream>

void print_error(int line) {
  std::cout << "Error (line " << line << "): " << zmq_strerror(zmq_errno()) << std::endl;
}

void check_error(bool condition, int line) {
  if (!condition) {
    print_error(line);
  }
}

int main() {
  // Create context
  void* ctx = zmq_ctx_new();
  check_error(ctx != NULL, __LINE__);

  // Init and close a message
  zmq_msg_t msg;
  int status = zmq_msg_init(&msg);
  check_error(status == 0, __LINE__);
  status = zmq_msg_close(&msg);
  check_error(status == 0, __LINE__);

  // A repeated message is idempotent with an error. This makes sense!
  status = zmq_msg_close(&msg);
[...]

PR #3330, opened 2018-12-12
Prevent crashes related to zmq_msg_data, zmq_msg_size. Introduce zmq_msg_bytes

joelfrederico:
Solution:

- Change asserts to unlikely tests.
- `zmq_msg_data(...)` returns NULL on invalid `zmq_msg_t` and sets `errno` to `EFAULT`.
- `zmq_msg_size(...)` returns 0 on invalid `zmq_msg_t` and sets `errno` to `EFAULT`. Deprecated due to ambiguity of return result, legitimate messages may be size 0.
- Introduce error code oriented `int zmq_msg_bytes(zmq_msg_t *msg, size_t *bytes)`. `zmq_msg_bytes(...)` sets `size_t *bytes` to the size in bytes of the message. It returns zero on success and -1 on error. It sets `errno` to `EFAULT` on error. The only error condition is an invalid `zmq_msg_t`.
- Add tests and update docs.

See #3329.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,6 +65,7 @@ Ilya Kulakov
 Ivo Danihelka
 Jacob Rideout
 Joe Thornber
+Joel Frederico
 Jon Dyte
 Kamil Shakirov
 Ken Steele
--- a/Makefile.am
+++ b/Makefile.am
@@ -387,6 +387,7 @@ test_apps = \
 	tests/test_reqrep_device \
 	tests/test_sub_forward \
 	tests/test_invalid_rep \
+	tests/test_msg_access \
 	tests/test_msg_flags \
 	tests/test_msg_ffn \
 	tests/test_connect_resolve \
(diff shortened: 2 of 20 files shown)


joelfrederico:
Ah I forgot to mention that I deprecated zmq_msg_size. I just thought I'd do the full solution to start. If it's not wanted, needs approvals, should be split, whatever, I'm okay with that.

bluca:
Sorry, but we cannot change existing API. Adding new APIs is fine, but please add them as DRAFT as per C4 process (see sec/zmq_draft.h and ifdef ..DRAFT section in include/zmq.h)

joelfrederico:
Ah okay, so don't deprecate? Is it possible to change the API behavior if we only specify currently undocumented behavior? If the documentation doesn't change and the only side effects are things API-adhering code doesn't do, it seems like that would be a valid solution?

joelfrederico:
Oh the whole object could be accessed in other threads. So it's not thread-safe... Actually, checking whether it's valid (`msg_t::check ()`) may not be thread-safe, it isn't in debug where it accesses twice and has a data race. If we make `check ()` thread-safe by copying `u_.base.type` and testing that, then we can do a thread-safe check to see if `msg_t` is valid. If it isn't, we can return 0/NULL safely which would fix the problem of unsafe access, right? I'm not actually accessing internal memory when I return from an invalid `msg_t`, so that should be fine, right?

joelfrederico:
I think I've come to the opinion that this shouldn't be implemented. We should keep assert-crashing when we detect an error. See https://github.com/zeromq/libzmq/issues/3329#issuecomment-446774851.

@bluca, close if appropriate!

bluca:
Ok, thanks for the analysis, I'll close this in the meanwhile, please do send a new one to clarify the documentation if you can

PR #3322, opened 2018-12-07
Portable solution for library symbols visibility

jcourreges:
Using symbol visibility tricks is a common practice to hide library internals from consumers.

Right now `-Wl,--version-script=...` is used on a few selected architectures to hide symbols not already handled by `-fvisibility=hidden`. This leads to clutter in configure.ac and an unmaintainable if/else/endif dance in Makefile.am.

libzmq uses libtool, which provides tools to address symbol visibility needs portably.  This merge request implements that and should improve the situation on all platforms not already using `-Wl,--version-script=...` (Windows, *BSD, Solaris, etc).  It has only been tested on OpenBSD, however.

This merge request does not remove now superfluous OS-specific code in Makefile.am and configure.ac.

PR #3335, opened 2018-12-14
Go back to ld --version-script to hide private symbols

jcourreges:
libtool -export-symbols-regexp used ld(1) --retain-symbols-file under
the hood, the latter lets some C++ weak symbols make their way into the
dynamic symbols table, along with the zmq_* interface.  This was discussed in #3322. The reason for such behavior is unknown to me.

However, this is a regression for platforms which were previously using -Wl,--version-script.

This pull request goes back to using --version-script or equivalent, extending its use on more platforms such as OpenBSD.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -298,7 +298,9 @@ endif
 endif
 endif
 
-src_libzmq_la_LDFLAGS += -export-symbols-regex '^zmq_.*'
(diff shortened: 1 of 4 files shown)


bluca:
Thanks for the followup, if you ever find out how to solve the problem with the regex please do open a new PR

Issue #3351, opened 2019-01-08
Remote code execution vulnerability


guidovranken:
# Bug details

In ```src/v2_decoder.cpp zmq::v2_decoder_t::eight_byte_size_ready()```, the attacker can provide an ```uint64_t``` of his choosing:

```cpp
 85 int zmq::v2_decoder_t::eight_byte_size_ready (unsigned char const *read_from_)
 86 {
 87     //  The payload size is encoded as 64-bit unsigned integer.
 88     //  The most significant byte comes first.
 89     const uint64_t msg_size = get_uint64 (_tmpbuf);
 90 
 91     return size_ready (msg_size, read_from_);
 92 }
```

Then, in ```src/v2_decoder.cpp zmq::v2_decoder_t::size_ready()```, a comparison is performed to check if this peer-supplied ```msg_size_``` is within the bounds of the currently allocated block of memory:

```cpp
117     if (unlikely (!_zero_copy
118                   || ((unsigned char *) read_pos_ + msg_size_
119                       > (allocator.data () + allocator.size ())))) {
```

This is inadequate because a very large ```msg_size_``` will overflow the pointer (```read_pos_```).
In other words, the comparison will compute as 'false' even though ```msg_size_``` bytes don't fit in the currently allocated block.

# Exploit details

Now that ```msg_size_``` has been set to a very high value, the attacker is allowed to send this amount of bytes, and libzmq will copy it to its internal buffer without any further checks.

This means that it's possible to write beyond the bounds of the allocated space.

However, for the exploit this is not necessary to corrupt memory beyond the buffer proper.

As it turns out, the space the attacker is writing to is immediately followed by a ```struct content_t``` block:

```cpp
 67     struct content_t
 68     {
 69         void *data;
 70         size_t size;
 71         msg_free_fn *ffn;
 72         void *hint;
 73         zmq::atomic_counter_t refcnt;
 74     };
```

So the memory layout is such that the receive buffer is immediately followed by ```data```, then ```size```, then ```ffn```, then ```hint```, then ```refcnt```.
Note that the receive buffer + the ```struct content_t``` is a single, solid block of memory; by overwriting beyond the designated receive buffer's bounds, no dlmalloc state variables in memory (like ```bk```, ```fd```) are corrupted (or, in other words, it wouldn't trigger AddressSanitizer).

This means that the attacker can overwrite all these members with arbitrary values.

[...]

PR #3353, opened 2019-01-08
V2 decoder overflow RCE fix

guidovranken:


guidovranken:
Per this rule in the spec

```
When two pointers are subtracted, both shall point to elements of the same array object, or one past the last element of the array object;
the result is the difference of the subscripts of the two array elements. 
```

the following ought to be legal pointer arithmetic:

```msg_size_ > allocator.data () + allocator.size () - read_pos_ ```

because

```allocator.data () + allocator.size ()``` constitutes a valid pointer, and subtracting the valid pointer ```read_pos_``` from the result, yields the "difference of the subscripts of the two array elements", which is the distance in bytes for the type ```unsigned char```.

To guarantee this in all circumstances, we might want to add a ```static_assert``` (or similar) that the pointer types of ```allocator.data ()``` and ```read_pos_``` are equivalent, and moreover point to a type of size ```1```.

Furthermore, this assumes ```read_pos_``` >= ```allocator.data ()```, for which we might want to add an ```assert()```, as crashing via ```assert()``` is less severe than potential remote code execution resulting from an invalid pointer computation.

sigiesec:
Thanks for the fix! Unfortunately, this fails to compile with some builds, e.g. here: https://travis-ci.org/zeromq/libzmq/jobs/477075295#L1215
A cast to ptrdiff_t might be added to resolve this.

sigiesec:
Fixes #3351

bluca:
@guidovranken thank you very much for finding the bug and sending the fix. There's a formatting error but I'll take care of that, as you've done enough already.

I'll try and do a new release soon-ish, as soon as I can do the validation and changelog.

guidovranken:
> @guidovranken thank you very much for finding the bug and sending the fix. There's a formatting error but I'll take care of that, as you've done enough already.
> 
> I'll try and do a new release soon-ish, as soon as I can do the validation and changelog.

Oh, I just pushed the fix for formatting before I saw your message.

PR #3355, opened 2019-01-09
Problem: small formatting issue highlighted by clang

bluca:
Solution: fix it with make clang-format-diff

@sigiesec would appreciate a speedy merge :-)


bluca:
Thanks!

PR #3364, opened 2019-01-15
Problem: ZMQ_SOCKET_LIMIT and ZMQ_THREAD_PRIORITY have the same value

### LINKED ISSUE #3362: Problem: ZMQ_SOCKET_LIMIT and ZMQ_THREAD_PRIORITY have the same value
# Issue description
`test_ctx_options` fails and I'm unable to identify the reasons why.
In particular, the part of the test that fails, located in `test_ctx_options.cpp`, is the second assert in:
```cpp
    if (is_allowed_to_raise_priority ()) {
        rc = zmq_ctx_set (
          ctx_, ZMQ_THREAD_PRIORITY,
          1 /* any positive value different than the default will be ok */);
        assert (rc == 0);
        rc = zmq_ctx_get (ctx_, ZMQ_THREAD_PRIORITY);
        assert (rc == 1);
    }
```
I manually checked and the call to `zmq_ctx_set ( ctx_, ZMQ_THREAD_PRIORITY, 1)` is successful, returning `1`. However, the call to `rc = zmq_ctx_get (ctx_, ZMQ_THREAD_PRIORITY)` returns the value `65,535`. In the beginning I thought this value was a `-1`, indicating that the call to `zmq_ctx_get` failed.

After investigating the returned value, it seems that it is not a `-1` as `zmq_ctx_get` returns an `int` and not a `short int`.

How could I debug this problem further? It may seem that the problem is in the way that priority is acquired? 

# Environment

* libzmq version (commit hash if unreleased): 4.3.0
* OS: Arch GNU/Linux
The command `ulimit -a` shows:
```
-t: cpu time (seconds)              unlimited
-f: file size (blocks)              unlimited
-d: data seg size (kbytes)          unlimited
-s: stack size (kbytes)             8192
-c: core file size (blocks)         unlimited
-m: resident set size (kbytes)      unlimited
-u: processes                       63741
-n: file descriptors                1048576
-l: locked-in-memory size (kbytes)  65536
-v: address space (kbytes)          unlimited
-x: file locks                      unlimited
-i: pending signals                 63741
-q: bytes in POSIX msg queues       819200
-e: max nice                        40
-r: max rt priority                 0
-N 15:                              unlimited
```

# What's the actual result?

## Command line output of the test:
```
PASS: tests/test_router_handover
./config/test-driver: line 107: 21636 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: tests/test_ctx_options
PASS: tests/test_unbind_wildcard
```

## Contents of the test log:
```
lt-test_ctx_options: tests/test_ctx_options.cpp:116: void test_ctx_thread_opts(void*): Assertion `rc == 1' failed.
```

bluca:
Solution: remove documents and tests for ZMQ_THREAD_PRIORITY getter. It
never worked and can never work as it has the same value as a get-only
option ZMQ_SOCKET_LIMIT. It cannot be changed without breaking ABI.
Note that the setter works fine as ZMQ_SOCKET_LIMIT is get-only.

Fixes #3362

### DIFF
--- a/doc/zmq_ctx_get.txt
+++ b/doc/zmq_ctx_get.txt
@@ -70,12 +70,6 @@ The 'ZMQ_THREAD_SCHED_POLICY' argument returns the scheduling policy for
 internal context's thread pool.
 
 
-ZMQ_THREAD_PRIORITY: Get scheduling priority for I/O threads
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The 'ZMQ_THREAD_PRIORITY' argument returns the scheduling priority for
-internal context's thread pool.
-
-
 ZMQ_THREAD_NAME_PREFIX: Get name prefix for I/O threads
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The 'ZMQ_THREAD_NAME_PREFIX' argument gets the numeric prefix of each thread
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -469,10 +469,7 @@ int zmq::thread_ctx_t::set (int option_, int optval_)
 int zmq::thread_ctx_t::get (int option_)
 {
     int rc = 0;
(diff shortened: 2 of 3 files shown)


c-rack:
@somdoron you are too fast for me :)

PR #3369, opened 2019-01-20
Problem: have to check ABI compatibility manually

bluca:
Solution: automate it on Travis

### DIFF
--- a/builds/abi-compliance-checker/ci_build.sh
+++ b/builds/abi-compliance-checker/ci_build.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -x
+
+cd ../../
+
+mkdir tmp
+BUILD_PREFIX=$PWD/tmp
+
+CONFIG_OPTS=()
+CONFIG_OPTS+=("CFLAGS=-I${BUILD_PREFIX}/include -g -Og")
+CONFIG_OPTS+=("CPPFLAGS=-I${BUILD_PREFIX}/include")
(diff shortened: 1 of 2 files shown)


builds/abi-compliance-checker/ci_build.sh
```diff
@@ -0,0 +1,40 @@
...
+   cat compat_reports/libzmq/${LATEST_VERSION}_to_HEAD/src_affected.txt | c++filt
+   exit 1
+}
+
+./autogen.sh
+./configure "${CONFIG_OPTS[@]}"
+make VERBOSE=1 -j5
+abi-dumper src/.libs/libzmq.so -o ${BUILD_PREFIX}/libzmq.head.dump -lver HEAD
+
+git clone --depth 1 -b latest_release https://github.com/zeromq/libzmq.git latest_release
```
QuLogic: Can you not do this from the existing clone?

bluca: Travis does a depth 50 clone which means the last tag might not be there

QuLogic: You can do: `git fetch --depth 1 origin latest_release:latest_release` even on a shallow clone, but now I'm not so sure it saves much. The `latest_release` branch is not really old enough to make a suitable test.

bluca: sure, feel free to send a PR to update it

PR #3375, opened 2019-01-30
An alternative approach to threads naming

opensource-kymetacorp:
# Pull Request Notice

Before sending a pull request make sure each commit solves one clear, minimal,
plausible problem. Further each commit should have the following format:

```
Problem: X is broken

Solution: do Y and Z to fix X
```

Please try to have the code changes conform to our coding style. For your
convenience, you can install clang-format (at least version 5.0) and then
run ```make clang-format-check```. Don't fix existing issues, if any - just
make sure your changes are compliant. ```make clang-format-diff``` will
automatically apply the required changes.
To set a specific clang-format binary with autotools, you can for example
run: ```./configure CLANG_FORMAT=clang-format-5.0```

```
git fetch upstream
git rebase upstream/master
git push -f
```

```
git rebase -i HEAD~10
```

# FIRST TIME CONTRIBUTORS PLEASE NOTE

Please add an additional commit with a relicensing grant.

[Example](https://github.com/zeromq/libzmq/commit/fecbd42dbe45455fff3b6456350ceca047b82050)

[More information on RELICENSING effort](https://github.com/zeromq/libzmq/tree/master/RELICENSE/README.md)

### DIFF
--- a/RELICENSE/Kymetacorp.md
+++ b/RELICENSE/Kymetacorp.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Kymeta Corporation
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other
+Open Source Initiative approved license chosen by the current ZeroMQ
+BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle "opensource-kymetacorp", with
+commit author "opensource-kymetacorp", are copyright of Kymeta Corporation.
+This document hereby grants the libzmq project team to relicense libzmq,
+including all past, present and future contributions of the author listed above.
+
+Kymeta Corporation
+2018/11/21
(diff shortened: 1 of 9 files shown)


opensource-kymetacorp:
An alternative approach to threads naming.
Allows to set thread name in the "start" method.
Threads started through the context have a default name even if name wasn't explicitly specified.
Gives epoll the ability to name its thread.
io_thread and reaper name their threads automatically.

bluca:
Thanks for the PR. pthread_set_name_np, as the name says, it's not part of posix so this will break all platforms other than Linux. There's already a method to set the thread name in a platform compatible way that can be used

bluca:
I think this will break the existing ZMQ_THREAD_NAME_PREFIX API, as it overwrites the name, won't it?

bluca:
I'll fix those after merging in https://github.com/zeromq/libzmq/pull/3411

opensource-kymetacorp:
Luca
Thanks.
We are aware that an OS independent solution for this issue has been implemented upstream from the version Kymeta is using.  This patch was submitted as an alternative approach to the issue.

Kymeta Client Software

bluca:
I have adapted and fixed up your patch via https://github.com/zeromq/libzmq/pull/3411 so it now works everywhere

PR #3379, opened 2019-02-02
Add v2 monitoring

sigiesec:
V2 monitoring adds a new API (zmq_socket_monitor_versioned) and a new version of the monitor protocol, which, as a first step, extends the size of the event type and value fields to 64 bits, and provides both local and remote addresses in monitor event, which allows to identify individual connections.


bluca:
the new ipv6 test in the monitor test also hangs

sigiesec:
I will give it a look in my Linux VM tomorrow, under MinGW I have no ipv6/tipc

bluca:
The ipv6 is simple, it's just not setting the ipv6 flag on the client

bluca:
also with the v2 monitor, doing the stats test, both endpoints are still empty

src/socket_base.cpp
```diff
@@ -970,55 +979,56 @@ int zmq::socket_base_t::connect (const char *endpoint_uri_)
     //  Save last endpoint URI
     paddr->to_string (_last_endpoint);
 
-    add_endpoint (endpoint_uri_, static_cast<own_t *> (session), newpipe);
+    add_endpoint (make_unconnected_connect_endpoint_pair (_last_endpoint),
```
bluca: @sigiesec this is what causes the TIPC issue - changing from endpoint_uri_ to _last_endpoint means that the user calling disconnect with endpoint_uri_ will not match anymore

bluca: https://github.com/zeromq/libzmq/pull/3382

sigiesec: Ah... I changed this somewhat by mistake. I think I noticed that in some contexts the original endpoint is used, but in some cases the resolved one. I think some uses need to be changed to provide meaningful monitor events, but apparently this is not consistent right now.

What I was thinking about: shouldn't it be possible to perform multiple wildcard binds on the same socket? Then each would bind to a different port/name. But then it doesn't make sense to store the original wildcard address for identifying an endpoint.

bluca: yes it should be possible to do multiple wildcard binds - but this is in the connect call

PR #3383, opened 2019-02-03
Problem: cannot monitor state of queues at runtime

### LINKED ISSUE #3403: error building w/-DENABLE_DRAFTS=On  -DWITH_DOCS=On
# error building w/-DENABLE_DRAFTS=On  -DWITH_DOCS=On

# Environment

* libzmq version (commit hash if unreleased): 4bdb05df78d6806daa8f0b84d2901e37d2ea5981
* OS: CentOS 6.9, gcc 4.8.2

1. Build with -DENABLE_DRAFTS=On  -DWITH_DOCS=On

```
[ 35%] Generating zmq_socket_monitor_versioned.html
ERROR: manpage document title is mandatory
ERROR: zmq_socket_monitor_versioned.txt: line 4: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 10: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 15: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 64: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 67: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 74: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 78: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 84: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 90: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 95: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 101: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 106: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 111: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 116: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 121: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 125: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 130: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 135: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 163: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 171: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 178: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 190: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 302: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 307: section title not permitted in delimited block
ERROR: zmq_socket_monitor_versioned.txt: line 1: [blockdef-example] missing closing delimiter
make[2]: *** [doc/zmq_socket_monitor_versioned.html] Error 1
make[1]: *** [CMakeFiles/libzmq.dir/all] Error 2
make: *** [all] Error 2
+ rc=2
+ [[ 2 != 0 ]]
+ exit 2
```

bluca:
Solution: add API and event

Fixes #3403


src/session_base.cpp
```diff
@@ -409,6 +409,11 @@ void zmq::session_base_t::process_attach (i_engine *engine_)
         zmq_assert (!_pipe);
         _pipe = pipes[0];
 
+        //  The endpoints strings are not set on bind, set them here so that
+        //  events can use them.
+        pipes[0]->set_endpoint_pair (engine_->get_endpoint());
```
bluca: @sigiesec - this does the trick to make the endpoint available. On bind, the pipe is not created for tcp/ipc, that's why it was not set before and we didn't see the address.

src/socket_base.cpp
```diff
@@ -1421,6 +1421,32 @@ void zmq::socket_base_t::process_term_endpoint (std::string *endpoint_)
     delete endpoint_;
 }
 
+void zmq::socket_base_t::process_pipe_stats_publish (
+  uint64_t outbound_queue_count_,
+  uint64_t inbound_queue_count_,
+  endpoint_uri_pair_t *endpoint_pair_)
+{
+    event (*endpoint_pair_, outbound_queue_count_, ZMQ_EVENT_PIPES_STATS);
```
bluca: TODO: refactor the event function to allow multiple values, so that a single event can be used

tests/test_hwm_pubsub.cpp
```diff
@@ -44,7 +45,62 @@ void tearDown ()
     teardown_test_context ();
 }
 
-int test_defaults (int send_hwm_, int msg_cnt_, const char *endpoint)
```
bluca: TODO: move to test_monitor.cpp

src/socket_base.cpp
bluca: Another question here: by having the event raised from the application thread, we run into the issue that a dummy getsockopt or poll has to be done on the monitored socket for the event to be raised. This would be solved by doing this in the pipe class from the I/O thread, but the issue then becomes that the I/O thread, rather than just processing a command and forwarding it, has to acquire the monitor mutex, which is most likely way worse for performances.

I'd be inclined to leave this to be raised on the application thread. Thoughts? @somdoron @sigiesec

somdoron: How often does the mutex happen? Is it happening if monitoring is not enabled as well?

bluca: No, only when the request is triggered and being processed - so once per probe

somdoron: how often a probe? is it by interval or user requesting it?

bluca: user requesting it, although i'm considering allowing an interval as well, as it might be useful

sigiesec: The monitoring needs a mutex, since the stored monitor socket is accessed from both the application (through zmq_socket_monitor*) and the I/O thread. Probably this could be solved without a mutex, but it has been like this for a long time.

bluca: I considered waiting for the command, but especially without unit tests for the commands pipe it would get tricky avoiding regressions.

src/session_base.cpp
sigiesec: Ah :) And on connect, the endpoint_pair is already set, and it is now set to the same value again?

bluca: On the connect side the pipe is created immediately, so this branch is never ran

bluca:
Currently working on improving the tests, then I'll squash the double events into one

src/socket_base.cpp
```diff
@@ -1658,101 +1685,115 @@ int zmq::socket_base_t::monitor (const char *endpoint_,
 void zmq::socket_base_t::event_connected (
   const endpoint_uri_pair_t &endpoint_uri_pair_, zmq::fd_t fd_)
 {
-    event (endpoint_uri_pair_, fd_, ZMQ_EVENT_CONNECTED);
+    event (endpoint_uri_pair_, std::list<uint64_t>{(uint64_t) fd_},
```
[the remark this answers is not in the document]
bluca: well never mind, this does not work with C++ 98
```
error: extended initializer lists only available with -std=c++0x or -std=gnu++0x [-Werror]
```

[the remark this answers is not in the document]
sigiesec: I would suggest to pass a plain C array with the values.

src/socket_base.cpp
```diff
@@ -1792,18 +1834,22 @@ void zmq::socket_base_t::monitor_event (
                 memcpy (zmq_msg_data (&msg), &event_, sizeof event_);
                 zmq_msg_send (&msg, _monitor_socket, ZMQ_SNDMORE);
 
-                //  Send value in second frame (64bit unsigned)
-                zmq_msg_init_size (&msg, sizeof value_);
-                memcpy (zmq_msg_data (&msg), &value_, sizeof value_);
-                zmq_msg_send (&msg, _monitor_socket, ZMQ_SNDMORE);
+                //  Send values in second-Nth frames (64bit unsigned)
```
sigiesec: I would suggest to first send a message part containing the number of values that will follow.

bluca: yeah, makes sense

src/socket_base.cpp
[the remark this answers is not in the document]
bluca: Yeah, sounds like the easiest way

bluca:
The valgrind build is still failing, locally too, most likely a race condition. Once that is solved, this should be ready

sigiesec:
I ran with valgrind locally and did not manage to let it fail. But I saw that already the basic v2 monitor test is failing sometimes on travis with valgrind, where an unexpected event is received. Not yet sure if this is something to expect in the test.

bluca:
yeah there's 2 extra events, a disconnect and a retried, when running in valgrind - not sure why, probably the monitor terminates later due to the slowness

bluca:
the remaining valgrind failure is weird, it's in the new tcp ipv4 test but it doesn't say anywhere how it's failing

bluca:
and I can't reproduce it

sigiesec:
I think the cause for the monitor test failure is the following:
* while we initiate the closing of the client socket before the server socket, this is asynchronous, so the server socket might close first, causing the client to try to reconnect before being closed
* this needs to be synchronized by first waiting for the ZMQ_EVENT_CLOSED on the client before proceeding to close the server socket (or generalizing the assertions by expecting this flow of events)

The same race condition probably already exists in the old test case, but happens less often.

sigiesec:
Eh sorry, we don't get a ZMQ_EVENT_CLOSED on the client side at all. This is only emitted by the connecter at the moment, if the socket is closed before a connection is established. Not sure if this is really consistent.

We only get a ZMQ_EVENT_MONITOR_STOPPED at the moment.

bluca:
The problem is that the test only enables the _STATS event, nothing else - so it can't be an issue of wrong event expected :-/

sigiesec:
Ok, I think we are talking about two different things here. I will try to give the stats test a look later on.

bluca:
Yeah I fixed the other tests in a commit here - as you said the events didn't match every now and then. Right now it's the ipv4 stats test that fails, with no discernible log message whatsoever, and can't be reproduced locally :-/

bluca:
it looks like it's timing out when doing the send in a loop, reducing the HWM seems to help, but now breaks on osx...

bluca:
Ok, I finally got it - it was a leftover from when the test initially was pub-sub, and was busy-looping with _DONTWAIT. It was too much for valgrind, which just timed out. Changing to blocking send, which is fine for push-pull, fixes the issue.

So, as far as I can see, this is _finally_ ready

bluca:
@sigiesec ping :-) does it look good now?

src/session_base.cpp
```diff
@@ -409,6 +409,11 @@ void zmq::session_base_t::process_attach (i_engine *engine_)
         zmq_assert (!_pipe);
         _pipe = pipes[0];
 
+        //  The endpoints strings are not set on bind, set them here so that
+        //  events can use them.
+        pipes[0]->set_endpoint_pair (engine_->get_endpoint ());
+        pipes[1]->set_endpoint_pair (engine_->get_endpoint ());
```
sigiesec: I am not sure, somehow it feels that the endpoint pair should be reversed on one of the pipes. But I don't oversee where this might have an effect.

src/zmq.cpp
```diff
@@ -1452,3 +1452,10 @@ int zmq_has (const char *capability_)
     //  Whatever the application asked for, we don't have
     return false;
 }
+
+void zmq_socket_monitor_pipes_stats (void *s_)
```
sigiesec: For uniformity, shouldn't this return an int, and return -1 on error (ENOTSOCK/EINVAL e.g.)? What if there is no socket monitor on the socket?

sigiesec:
Yes, looks good to me. I added two minor comments, but these can be addressed later on :)
Hope I will be able to test this soon on our side :)

src/zmq.cpp
bluca: yes I forgot to deal with that - I'm adding a few more error checks and values

src/session_base.cpp
bluca: I guess it depends on how one interprets the local and remote attributes - if the reference point is the application thread then yes, but if the reference point is the socket itself then no

sigiesec: Well yes, local and remote are the same wrt the socket. Still the direction is different. Hm. Let's see, when analysing the data from the monitoring :)

bluca:
great - so can we merge now?

sigiesec:
@bluca Merged it now, thanks a lot for working on this :)

bluca:
thanks!

viciious:
Sorry for the necro-post, but could someone please clarify the following: the docs says that "the pipe stats on the sender will not get updated until the receiver has processed at least lwm ((hwm + 1) / 2) messages". What happens if the receiver has been N/A the whole time: e.g. the network has been down? Will the stat message report 0 queue length?

bluca:
yes

viciious:
> yes

Oh well, this is quite useless for our monitoring purposes then...

What is "lwm ((hwm + 1) / 2)" anyway? Is it lwm multipled by ((hwm + 1) / 2) or something else?

viciious:
> yes

Sorry, but looking at the code, it appears not to be the case. 

```
    send_pipe_peer_stats (_peer, _msgs_written - _peers_msgs_read, socket_base_,
                          ep);
```

We've been using a custom patched zeromq internally for quite a while, with similar logic. From I can remember, _peers_msgs_read on the sender does indeed only get updated when the receiver has reached its lwm, but _msgs_written is incremented whenever the sender has queued another message. In case the receiver is offline, `send_pipe_peer_stats` will simply write _msgs_written to the stats message. Am I missing something here?

PR #3384, opened 2019-02-04
Remove socks/tcp code duplication

sigiesec:


src/tcp.hpp
```diff
@@ -34,6 +34,9 @@
 
 namespace zmq
 {
+struct tcp_address_t;
```
bluca: the C98 checker says:

```
In file included from ../src/signaler.cpp:64:
../src/tcp.hpp:37:1: error: struct 'tcp_address_t' was previously declared as a
      class [-Werror,-Wmismatched-tags]
struct tcp_address_t;
^
../src/tcp_address.hpp:42:7: note: previous use is here
class tcp_address_t
      ^
../src/tcp.hpp:37:1: note: did you mean class here?
struct tcp_address_t;
^~~~~~
class
```

PR #3386, opened 2019-02-05
Problem: tcp address strings are dependent on locale

### LINKED ISSUE #3385: TCP port in ZMQ_LAST_ENDPOINT depends on global locale
# Issue description

The TCP port formatting in `ZMQ_LAST_ENDPOINT` depends on the locale set by `std::locale::global`. With thousands grouping enabled, the port number contains commas. UDP and other ports are likely affected as well.

# Environment

* libzmq version 4.3.1
* OS: Arch Linux

```
#include <zmq.h>
#include <assert.h>
#include <iostream>
#include <sstream>

int main(void) {
    struct comma : std::numpunct<char> {
        char do_thousands_sep() const override
        {
            return ',';
        } // comma
        std::string do_grouping() const { return "\3"; } // groups of 3 digits
    };
    std::locale::global(std::locale(std::locale("C"), new comma));
    void *context = zmq_ctx_new ();
    void *sock = zmq_socket (context, ZMQ_REP);
    int rc = zmq_bind (sock, "tcp://127.0.0.1:*");
    assert(rc == 0);
    char ep[1024];
    size_t sz = sizeof(ep);
    zmq_getsockopt(sock, ZMQ_LAST_ENDPOINT, ep, &sz);
    std::cout << "EP " << ep << std::endl;
}
````

# What's the actual result?
`EP tcp://127.0.0.1:39,375`

# What's the expected result?
`EP tcp://127.0.0.1:39375`

# Possible fix
In https://github.com/zeromq/libzmq/blob/master/src/tcp_address.cpp#L134, use `s.imbue(std::locale("C"));`

sigiesec:
Solution: use sprintf instead of std::stringstream
Fixes #3385

### DIFF
--- a/src/tcp_address.cpp
+++ b/src/tcp_address.cpp
@@ -29,7 +29,6 @@
 
 #include "precompiled.hpp"
 #include <string>
-#include <sstream>
 
 #include "macros.hpp"
 #include "tcp_address.hpp"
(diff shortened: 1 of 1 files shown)


sigiesec:
Android build is failing, i will give that a look

PR #3400, opened 2019-02-07
Add test for ZMQ_BINDTODEVICE

### LINKED ISSUE #3397: No test case for ZMQ_BINDTODEVICE
There is no test case setting the ZMQ_BINDTODEVICE socket option.

sigiesec:
Fixes #3397

### DIFF
--- a/tests/test_pair_tcp_cap_net_admin.cpp
+++ b/tests/test_pair_tcp_cap_net_admin.cpp
@@ -1,0 +1,89 @@
+/*
+    Copyright (c) 2007-2017 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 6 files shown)


tests/test_pair_tcp_cap_net_admin.cpp
```diff
@@ -0,0 +1,89 @@
...
+    teardown_test_context ();
+}
+
+typedef void (*extra_func_t) (void *socket_);
+
+void set_sockopt_bind_to_device (void *socket)
+{
+    const char device[] = "lo";
+    int rc =
+      zmq_setsockopt (socket, ZMQ_BINDTODEVICE, &device, sizeof device - 1);
```
bluca: I've had problems in the past with some compilers not being happy about sizeof without () (I think it was Sun studio)

sigiesec: Ok, changed this.

PR #3406, opened 2019-02-13
Problem: trie benchmarks count loop times instead of lookup calls

ssbl:
Solution: measure the time taken by each lookup call instead

Other changes:
- show average time taken and reduce other output
- show number of keys, queries and key size before running the benchmark
- use steady_clock as high_resolution_clock might not be steady
- increase size of query set to 1M

The output now looks like this:
```
keys = 10000, queries = 1000000, key size = 20
[trie]
Average lookup time = 405.6 ns
[radix_tree]
Average lookup time = 183.4 ns
```

### DIFF
--- a/perf/benchmark_radix_tree.cpp
+++ b/perf/benchmark_radix_tree.cpp
@@ -109,7 +109,7 @@ int main ()
     }
 
     // Create a benchmark.
-    std::printf ("keys = %ld, queries = %ld, key size = %ld\n", nkeys, nqueries,
+    std::printf ("keys = %lu, queries = %lu, key size = %lu\n", nkeys, nqueries,
                  key_length);
     std::puts ("[trie]");
     benchmark_lookup (trie, input_set, queries);
--- a/perf/benchmark_radix_tree.cpp
+++ b/perf/benchmark_radix_tree.cpp
@@ -36,10 +36,11 @@
 #include <cstddef>
 #include <cstdio>
 #include <random>
+#include <ratio>
 #include <vector>
 
 const std::size_t nkeys = 10000;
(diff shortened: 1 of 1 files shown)


perf/benchmark_radix_tree.cpp
```diff
@@ -104,6 +109,8 @@ int main ()
     }
 
     // Create a benchmark.
+    std::printf ("keys = %ld, queries = %ld, key size = %ld\n", nkeys, nqueries,
+                 key_length);
```
bluca: This fails to compile with some combinations of gcc/libc:

```
perf/benchmark_radix_tree.cpp:113:28: error: format '%ld' expects argument of type 'long int', but argument 2 has type 'std::size_t {aka unsigned int}' [-Werror=format=]
                  key_length);
```

ssbl: Used wrong format specifier, using `%lu` now

bluca: still not happy:

```
perf/benchmark_radix_tree.cpp:113:28: error: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'std::size_t {aka unsigned int}' [-Werror=format=]
```

You probably want to use llu and cast the variables to (unsigned long long) for portability

ssbl: Looks like `%zu` is the correct specifier but it might not be supported in some cases. I'll just change the variable type to unsigned long since the build still fails.

ssbl: Using a cast sounds like a better idea

ssbl:
thanks!

PR #3408, opened 2019-02-16
Problem: missing relicensing statement from bluca

bluca:
Solution: add it - I realised only now that I only submitted a
statement for the copyright owned by a previous employer, and not
my own.


bluca:
@somdoron yours is missing too :-)

PR #3411, opened 2019-02-17
Problem: non-linux build broken by last PR

bluca:
Solution: fix it


bluca:
@somdoron @sigiesec can we get a merge please? To fix osx and windows

bluca:
Thanks!

PR #3428, opened 2019-02-23
Don't crash if memory allocation in socket_poller_t::rebuild fails.

### LINKED ISSUE #3427: Memory allocation failure in socket_poller_t::rebuild is unnecessarily fatal
Consider the following line in `zmq::socket_poller_t::rebuild`:
https://github.com/zeromq/libzmq/blob/d814f370672afa2e06eba34d3e7ba1723a897b96/src/socket_poller.cpp#L293
This makes the program crash if the malloc on the preceding line failed to allocate memory (for example due to a ulimit).

This is probably unnecessarily fatal. Instead of crashing the program, I think the error should be propagated through to `socket_poller_t::wait`, which should report it as ENOMEM.

Eelis:
Fixes #3427.

### DIFF
--- a/doc/zmq_poller.txt
+++ b/doc/zmq_poller.txt
@@ -223,6 +223,8 @@ On _zmq_poller_add_fd_, _zmq_poller_modify_fd_ and _zmq_poller_remove_fd_:
 The _fd_ specified was the retired fd.
 
 On _zmq_poller_wait_ and _zmq_poller_wait_all_:
+*ENOMEM*::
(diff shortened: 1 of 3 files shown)


bluca:
Doesn't this leave the poller in an undefined, broken state? The previous set is gone and nothing was built to replace it

Eelis:
@bluca: Yes, but that fact is remembered by leaving `_need_rebuild` set to true, so that the next time `wait` is called, it will retry building `_pollfds`.

PR #3438, opened 2019-03-01
Fix appveyor test execution

sigiesec:
Finally, after a long time of always-red appveyor builds, this PR allows to get an overall green status on appveyor again, at least on my branch it was green once.

This is not as great as I hoped, since for now I disabled some test cases which were (almost) always failing. But I also fixed some tests which suffered from race conditions.

While the remaining tests should also be fixed, at least we are now able to see again is some PR breaks something additional.

The cygwin and mingw tests are still ignored completely. In another step towards fixing, this should probably also be reduced to ignore only the individual tests that are currently failing.


somdoron:
Wow, all green :)

bluca:
Fantastic work!

tests/test_xpub_nodrop.cpp
```diff
@@ -54,14 +54,18 @@ void test ()
...
     void *sub = test_context_socket (ZMQ_SUB);
     TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (sub, "inproc://soname"));
 
     //  Subscribe for all messages.
     TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (sub, ZMQ_SUBSCRIBE, "", 0));
 
+    //  we must wait for the subscription to be processed here, so we just
+    //  hope SETTLE_TIME suffices; otherwise some or all published messages
+    //  might be lost
+    msleep (SETTLE_TIME);
```
bluca: In these cases it's better to just switch to xpub and wait for the sub message directly. That way it's deterministic and cannot fail on slow platforms.

sigiesec: Ah, that's even better, I wasn't aware of that possibility. :)

sigiesec: Hm, the test_xpub_nodrop test case is still not running reliably indeed.

If we implement your suggestion, we are actually testing something different, since the publisher socket is the main test subject here. We would no longer test ZMQ_PUB but ZMQ_XPUB. (A test for ZMQ_XPUB with ZMQ_XPUB_NODROP set is actually missing at the moment).

bluca: I don't think it's too much of an issue, pub is basically xpub with xrecv disabled

sigiesec: ok, if it stays implemented that way. I will add a comment that this should be considered when making future changes

bluca: yes I don't see that changing at any point, it's always been like that

Issue #3248, opened 2018-09-10
ZMQ_CLIENT send not blocking


Cross22:
# Issue description
Invoking zmq_send() with a ZMQ_CLIENT socket which is NOT connected to a peer causes the message to be queued. Once a peer connects the queued message gets sent. This is in contrast to the documentation which states that send should be blocking when there is no peer.

# Sample code
void *context = zmq_ctx_new ();
void *sock = zmq_socket (context, ZMQ_CLIENT);
zmq_connect (sock, "tcp://127.0.0.1:12345");
zmq_send(sock, "HELLO", 6, 0);
printf("Why am I here when no data was sent?");

# Environment

* libzmq version (commit hash if unreleased):  4.2.5
* OS: OSX

# What's the actual result? (include assertion message & call stack if applicable)

Function call returns indicating number of bytes sent.

# What's the expected result?

Function blocks until it is able to send the data.

trya:
This behavior can also be encountered with ZMQ_DEALER, ZMQ_REQ and ZMQ_PUSH sockets (all sockets that are supposed to block when connected to no peer). Apparently, the culprit is the `socket_base_t::connect` method: https://github.com/zeromq/libzmq/blob/2cb1240db64ce1ea299e00474c646a2453a8435b/src/socket_base.cpp#L949-L963
If the ZMQ_IMMEDIATE option is not set, a pipe is forcibly attached to the socket, thus making the socket accepting messages in its sending queue even when it's not connected to a peer. Setting the ZMQ_IMMEDIATE to true will restore the expected behavior when `send` is called. I don't know if the documentation should change to reflect the new behavior or if ZMQ_IMMEDIATE set to true should be the default for these socket types.

bluca:
Behaviour of sockets cannot be changed (apart from the client one as it's still draft), so it's the documentation that needs updating. Feel free to send a PR.

trya:
As per RFCs 28/REQREP and 30/PIPELINE:
>For processing outgoing messages:
>* SHALL block on sending, or return a suitable error, when it has no connected peers.
As of now, the documentation sticks with the RFCs, but libzmq sockets do not comply. Are you really sure that updating the documentation is the right course of action?

bluca:
Yes - it's mostly an internal detail with no consequences for networked peers, and changing it is a breackage of backward compatibility

sigiesec:
@bluca Do you imply that the statement should be removed from the RFC?

bluca:
[...]

PR #3447, opened 2019-03-15
doc: Reflect real behavior of sockets blocking on send operations

trya:
Sockets that block in mute state only block when no peer is available if
the transport is connection-oriented and the ZMQ_IMMEDIATE socket option
is set.

May provide clarification for #3248.

### DIFF
--- a/doc/zmq_msg_send.txt
+++ b/doc/zmq_msg_send.txt
@@ -23,10 +23,11 @@ argument to be sent to the socket referenced by the 'socket' argument.  The
 'flags' argument is a combination of the flags defined below:
 
 *ZMQ_DONTWAIT*::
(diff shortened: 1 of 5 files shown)


bluca:
Please follow the pr template and add a relicensing grant https://github.com/zeromq/libzmq/blob/master/.github/PULL_REQUEST_TEMPLATE.md

trya:
Well, that won't do it, since I'm not able to commit under my real name, sorry.

bluca:
It can be just GitHub username or any other preferred form of identification, it's just to confirm in writing that your contributions can be moved from one licence to the new. Sorry but we cannot accept me contributions without grants, or we'll never be able to complete the relicensing

PR #3449, opened 2019-03-16
Migrate security tests for NULL and PLAIN mechanisms to Unity

sigiesec:


bluca:
looks like there are new segfaults on windows?

sigiesec:
Yes on MSVC2013 Debug, similar to the ones on some other tests, showing heap corruption. Strange. I will attempt to reproduce it again on my windows machine.

sigiesec:
@bluca I think I found the cause of the problem, which also caused the heap corruption errors in the other test cases. There were extra flags set in the appveyor.yml controlling the runtime library used, which led to strdup being used from the release library, allocating on its heap, and then using free from the debug library, causing the assertion failure. I now reenabled all of those tests, and hope they are fine now :)

bluca:
Nice find!

sigiesec:
test_xpub_nodrop still failed in one build on appveyor but all others are fine now :)

bluca:
Great - so there's still some timing issue with that test, beside the subscription reception?

bluca:
I guess one possible improvement could be to do one blocking recv before entering the loop at https://github.com/zeromq/libzmq/blob/master/tests/test_xpub_nodrop.cpp#L112

Issue #3238, opened 2018-08-31
descriptor leak on ZMQ_ROUTER if ZMQ_ROUTER_HANDOVER is set


rouming:
Hi guys,

# Description
Recently I discovered enormous number of opened sockets on a router which eventually becomes  unresponsive when number of file descriptors reach the limit.  Further investigations shows that this happens because of ZMQ_ROUTER_HANDOVER flag set.

# Environment
linux- 4.18.5-arch1-1-ARCH
glibc 2.28-4

* libzmq version (commit hash if unreleased): 
recent master, 
e0ec6a9b364c Problem: formatting errors introduced by PR #3237

# Minimal test code / Steps to reproduce the issue

Source file to reproduce in attach, which almost repeats test_router_handover and shows
number of descriptors: 
$ gcc -o descriptors-leak descriptors-leak.c -Wall -lzmq
$ ./descriptors-leak 
118
218
318
418
518
618
718
818
918
1018
descriptors-leak: descriptors-leak.c:65: main: Assertion `dealer_two' failed.
Aborted (core dumped)

if comment out #define REPRODUCE_A_BUG line everything  works fine.

--
Roman
[descriptors-leak.c.txt](https://github.com/zeromq/libzmq/files/2341405/descriptors-leak.c.txt)

PR #3450, opened 2019-03-17
session_base: fix fd leak if ZMQ_ROUTER_HANDOVER is set

rouming:
This fixes the issue #3238
"decriptor leak on ZMQ_ROUTER if ZMQ_ROUTER_HANDOVER is set"

When ZMQ_ROUTER_HANDOVER is set router_t::identify_peer() closes
old pipe and uses the new one.  When old_pipe->terminate() is called
eventually session_base_t::pipe_terminated() is invoked, but the
following path:

  if (_pending && !_pipe && !_zap_pipe && _terminating_pipes.empty()) {
	  own_t::process_term(0);
  }

is never taken, because _pending == 0, thus session_base_t object leaks
and fd is never closed.

In order to free session_base_t correctly terminate() has to be called.

Signed-off-by: Roman Penyaev <r.peniaev@gmail.com>

### DIFF
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -137,6 +137,7 @@ int zmq::router_t::xsetsockopt (int option_,
                 if (!options.connected) {
                     // We expect nothing is binded yet
                     _handover = (value != 0);
+                    options.handover = _handover;
                     return 0;
                 }
             }
--- a/src/router.cpp
+++ b/src/router.cpp
@@ -134,8 +134,11 @@ int zmq::router_t::xsetsockopt (int option_,
 
         case ZMQ_ROUTER_HANDOVER:
             if (is_int && value >= 0) {
-                _handover = (value != 0);
-                return 0;
+                if (!options.connected) {
+                    // We expect nothing is binded yet
+                    _handover = (value != 0);
+                    return 0;
+                }
             }
             break;
 
--- a/tests/test_router_handover.cpp
+++ b/tests/test_router_handover.cpp
@@ -49,15 +49,15 @@ void test_with_handover ()
     char my_endpoint[MAX_SOCKET_STRING];
     void *router = test_context_socket (ZMQ_ROUTER);
 
+	// Enable the handover flag
+    int handover = 1;
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (router, ZMQ_ROUTER_HANDOVER,
+                                               &handover, sizeof (handover)));
+
     TEST_ASSERT_SUCCESS_ERRNO (zmq_bind (router, "tcp://127.0.0.1:*"));
 
     TEST_ASSERT_SUCCESS_ERRNO (
       zmq_getsockopt (router, ZMQ_LAST_ENDPOINT, my_endpoint, &len));
-
-    // Enable the handover flag
-    int handover = 1;
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (router, ZMQ_ROUTER_HANDOVER,
-                                               &handover, sizeof (handover)));
 
     //  Create dealer called "X" and connect it to our router
     void *dealer_one = test_context_socket (ZMQ_DEALER);
--- a/tests/test_router_handover_leak.cpp
+++ b/tests/test_router_handover_leak.cpp
@@ -1,0 +1,164 @@
+/*
+    Copyright (c) 2007-2019 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <string.h>
+#include <assert.h>
+#include <unistd.h>
+
+#include "testutil.hpp"
+
+int main(int argc, char **argv)
+{
+	char my_endpoint[1024];
+	char buffer[255];
+	int flag, rc, i;
+	size_t len;
+
+	void *dealer_one, *dealer_two;
+	void *router;
+	void *ctx;
+
+	len = sizeof(my_endpoint);
+
+	ctx = zmq_ctx_new();
+	assert(ctx);
+
+	router = zmq_socket(ctx, ZMQ_ROUTER);
+	assert(router);
+
+	flag = 0;
+	rc = zmq_setsockopt(router, ZMQ_LINGER, &flag, sizeof(flag));
+	assert(rc == 0);
+
+	// Enable the handover flag
+	flag = 1;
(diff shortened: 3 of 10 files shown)


sigiesec:
@rouming Thanks for the PR! Unfortunately, it seems to break the test_immediate and test_security_zap tests on all platforms. Could you give these a look?

sigiesec:
@rouming Another thing: Could you please add a RELICENSING agreement based on one of the templates in https://github.com/zeromq/libzmq/tree/master/RELICENSE/templates?

rouming:
@sigiesec  
> Unfortunately, it seems to break the test_immediate and test_security_zap tests on all platforms. Could you give these a look?

Yes, my bad. Eventually I've just added new option.handover and check it inside session_base not to break other tests.  I am not very much happy with that, probably there is a cleaner way.

BTW on my latest master unittests/unittest_udp_address fails.

>  Could you please add a RELICENSING agreement based on one of the 

I do not understand what I have to add and where .

Shortly I will submit updated PR.

bluca:
Please correctly rebase on the latest master and force push, you are re-adding previous commits from other PRs

bluca:
This looks like a backward incompatible change to me?

rouming:
> This looks like a backward incompatible change to me?

You mean the following patch "router: forbid changing of ZMQ_ROUTER_HANDOVER on already binded socket" ?  I would say that should be done from the very beginning, otherwise I do not understand how the following race is solved:

  1. socket
  2. bind
           << new request/connection comes
  3. setsockopt(HADOVER)
 
Am I wrong that if you have HANDOVER set on step 3 you can't guarantee that your request which has come in-between will be really handovered?

To avoid such incompatibility change session_base_t has to know somehow that on pipe_terminated() extra terminate() has to be called exactly for this handover case.  Because of my shallow code understanding I did not find any good way. So please suggest something.

sigiesec:
Indeed, ZMQ_ROUTER_HANDOVER is specifically documented as one of the socket options that affect binds/connects that have already been performed (see doc/zmq_setsockopt.txt), so this is breaking.

But I agree with you, that this is inherently racy. Even if the leak is fixed, the behaviour for a connecting happening concurrently to the zmq_setsockopt call is not be predictable, which is not good.

A backwards compatible solution might be to ensure that setting the socket option only affects subsequent binds. Since one cannot make assumptions on *when* it takes effect on prior binds changing this to *never* take effect for those would not be breaking in a strict sense. Still, it might be surprising, and even somewhat more so than your current solution of reporting an error (but it might still remain unnoticed when the client code fails to check the result of zmq_setsockopt). However, I think there are no other socket options that fail depending on the current state, so this would also not be consistent within zmq_setsockopt.

So, overall I would suggest to change ZMQ_ROUTER_HANDOVER not to affect prior binds.

In that case, your PR must be modified however, to reflect this, and there shouldn't be a failure on zmq_setsockopt if the socket was already bound.

The documentation should also be updated to reflect this, and make it clear that ZMQ_ROUTER_HANDOVER does NOT have an effect on prior binds.

@bluca What do you think?

tests/test_router_handover_leak.cpp
```diff
@@ -0,0 +1,164 @@
+/*
```
sigiesec: I am not sure if this test makes sense. It demonstrates the leak, but it doesn't assert anything related to the sockets, so it probably won't fail if the leak is present (or it will fail because resources are exhausted). It probably consumes quite a lot of time compared to other tests, provided it does so many iterations. 

Unfortunately, leaking file descriptors cannot be detected in the same way as leaking memory with valgrind. valgrind can show open file descriptors at exit with the --track-fds option, but it will only report them. The output of valgrind would need to be checked for that explicitly.

If this test case is really added, please use the Unity test framework, and the assertion macros like in test_router_handover. You can add it to the existing test_router_handover.cpp file.

src/router.cpp
```diff
@@ -134,8 +134,12 @@ int zmq::router_t::xsetsockopt (int option_,
 
         case ZMQ_ROUTER_HANDOVER:
             if (is_int && value >= 0) {
-                _handover = (value != 0);
-                return 0;
+                if (!options.connected) {
+                    // We expect nothing is binded yet
```
sigiesec: binded -> bound

src/router.cpp
```diff
@@ -134,8 +134,12 @@ int zmq::router_t::xsetsockopt (int option_,
...
+                    _handover = (value != 0);
```
sigiesec: Not sure, but if handover is added to options, then maybe it should be only there, and not duplicated as a field of router_t? This opens up to inconsistent states, and may be confusing.

tests/test_router_handover.cpp
```diff
@@ -49,16 +49,16 @@ void test_with_handover ()
     char my_endpoint[MAX_SOCKET_STRING];
     void *router = test_context_socket (ZMQ_ROUTER);
 
+	// Enable the handover flag
```
sigiesec: Wrong indentation, please format all source files with clang-format, or apply the patch from the travis-ci clang-format job

tests/test_router_handover_leak.cpp
rouming: > so it probably won't fail if the leak is present (or it will fail because resources are exhausted).

Exactly, it will fail because resources are exhausted. And why this is bad?

> It probably consumes quite a lot of time compared to other tests, provided it does so many iterations.

Good pass:
```
$ time ./tests/test_router_handover_leak 
Total fds: 18
Total fds: 18
Total fds: 18
Total fds: 18
OK

real    0m0.948s
user    0m0.496s
sys     0m0.719s
```

Bad pass:
```
$ time ./tests/test_router_handover_leak 
Total fds: 1018
lt-test_router_handover_leak: tests/test_router_handover_leak.cpp:103: int main(int, char**): Assertion `dealer_two' failed.
Aborted (core dumped)

real    0m0.767s
user    0m0.133s
sys     0m0.340s
```

Less than 1 second for both runs.

And that's how bad pass looks when I run 'make check':

```
./config/test-driver: line 107:  5975 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: tests/test_router_handover_leak
```

>Unfortunately, leaking file descriptors cannot be detected in the same way as leaking memory with valgrind.

I do not need valgrind for that, why? The test simply crashes if we reach soft kernel limit of fds (I am not sure what happens on windows, but there should be something similar, right? Otherwise you can exhaust kernel resources from user app).

The other simple way is to check number of descriptors and expect they won't grow up to, say, 50. I did not want to do that, because I do not know how to do that on windows, so I wanted to keep code less os specific, thus test just crashes.

> please use the Unity test framework, and the assertion macros like in test_router_handover.

sure

> You can add it to the existing test_router_handover.cpp file.

I do not want to mix leak testing with generic handover test.

src/router.cpp
rouming: Oops. Thanks.

src/router.cpp
rouming: True. I can change that, but frankly I do not like this options.handover. We can pass handover flag just in old_pipe->terminate() call. Then receiver will know how to free the session inside session_base_t::pipe_terminated() callback. No incompatibilities.

tests/test_router_handover.cpp
rouming: Ok.

tests/test_router_handover_leak.cpp
sigiesec: > Exactly, it will fail because resources are exhausted. And why this is bad?

It is maybe not bad but less than optimal, because it is dependent on the environment (OS defaults and custom user limits) when it fails. I suspect that once the build is fine it will take much longer under Windows, but I am not sure.

If there is no viable alternative, it can be done this way, and maybe disable it on Windows if necessary.

I don't know if there is a documented way to check the number of open sockets in Windows.

sigiesec:
Another option might be to deprecate ZMQ_ROUTER_HANDOVER and add a new socket option with more well defined behavior

rouming:
@sigiesec  Frankly, thinking again about my fix I still can't answer one question which seems crucial: if peer side closes the socket (it does!) why ROUTER does not detect that and does not go through normal terminate stream_engine procedure? What I see on router side is that inside 'stream_engine_t::in_event' execution takes the following path when peer closes the socket:

```
    //  Tear down the connection if we have failed to decode input data
    //  or the session has rejected the message.
    if (rc == -1) {
        if (errno != EAGAIN) {
            error (protocol_error);
            return;
        }
        _input_stopped = true;
        reset_pollin (_handle);   <<< we end here
    }
```
errno == EAGAIN and we eventually execute epoll_ctl(EPOLL_CTL_MOD) passing 0 as pe->ev.events.  What I want to emphasize is that eventually we stop listening to descriptor events and never gracefully terminate the stream_engine regardless what has happened to peer side. I would assume that this is another problem, which can be successfully hidden by my fix.

bluca:
No activity for more than a year, closing.

PR #3451, opened 2019-03-18
Restore /dev/poll support

jacquesg:
Restore support for ```/dev/poll``` on Solaris, remove one unused class variable from ```epoll_t```.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ case of a released version.
 
 OS and version | Architecture | Compiler and version | Build system | Last report | Remarks
 -------------- | ------------ | -------------------- | ------------ | ----------- | -------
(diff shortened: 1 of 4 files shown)


sigiesec:
Thanks a lot for fixing this, and sorry it broke without notice! Unfortunately, we don't have any Solaris CI. 

Recently, I added some information on supported platforms here: https://github.com/zeromq/libzmq/blob/master/README.md#supported-platforms-with-known-active-users

As you seem to be a "known user" on Solaris, could you perhaps add an entry here with your exact platform data?

jacquesg:
@sigiesec Added as requested.

sigiesec:
@jacquesg Thanks a lot :)

PR #3452, opened 2019-03-18
Problem: strnlen may not be available

jacquesg:
Solution: Provide an implementation

I wasn't quite sure where to put the implementation, there is only 1 caller at this stage.


bluca:
Please add the same check to configure.ac

jacquesg:
@bluca Done.

Issue #1462, opened 2015-07-01
Test failure in test_filter_ipc


tkoeppe:
Tracking issue for test failure of `test_filter_ipc` on Linux x86-64, with GCC 4.8.4 and Clang 3.5:

    libzmq/tests/testutil.hpp:78: void bounce(void *, void *): Assertion `rc == 32' failed.

I haven't investigated, but I wanted to make a note of it.

tkoeppe:
I think the problem is that the test suite (in `testutil.h`) calls `zmq_send` and `zmq_recv` just once and doesn't admit failure like `EINTR` and `EAGAIN`. These functions should be called in a loop. I have a proof of concept fix that I might propose for discussion.

tkoeppe:
Indeed the test failure is accompanied by return value `-1` and `errno == EAGAIN`. Unfortunately, if I just call send/recv in a loop while this condition holds, the program gets stuck in an infinite loop.

jakirkham:
Also, seeing this test failure on Linux. Did anyone get any closer to finding a solution to this?

bdangit:
I recently was not able to pass this specific test within `test_filter_ipc`: https://github.com/zeromq/libzmq/blob/master/tests/test_filter_ipc.cpp#L144

In order to get it to pass, I had to make sure that my user was actually in another group.  `gpasswd -a someuser anothergroup`.  ref: https://github.com/habitat-sh/core-plans/blob/master/zeromq/plan.sh#L22-L25

bluca:
It's probably not a good idea to build things as root... Perhaps the the test could skip if you'd is 0?

panlinux:
I believe I'm hitting a similar issue in Ubuntu, but just in launchpad builds:

https://launchpadlibrarian.net/415262191/buildlog_ubuntu-disco-amd64.zeromq3_4.3.1-3ubuntu1~ppa2_BUILDING.txt.gz

But it does show a core dump earlier:
```PASS: tests/test_zmq_poll_fd
PASS: tests/test_timeo
./config/test-driver: line 107:   709 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: tests/test_filter_ipc
SKIP: tests/test_connect_delay_tipc
SKIP: tests/test_pair_tipc
(...)
FAIL: tests/test_filter_ipc
===========================

test_filter_ipc: tests/testutil.hpp:142: void bounce(void*, void*): Assertion `rc == 32' failed.
[...]

PR #3453, opened 2019-03-18
Validate group before using it

panlinux:
Validate the group before using it

This makes the test a bit more robust in scenarios where, probably due  to a misconfiguration in the environment, a group present in the list of supplementary groups is not really a valid choice. This was seen in the Launchpad builders (where `getgrgid(supgroup)` returned `NULL`), and possibly elsewhere (see https://github.com/habitat-sh/core-plans/blob/master/zeromq/plan.sh#L22-L25)

This fixes issue #1462

### DIFF
--- a/tests/testutil.hpp
+++ b/tests/testutil.hpp
@@ -78,8 +78,10 @@
 #include <unistd.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <grp.h>
 #include <sys/wait.h>
 #include <sys/socket.h>
+#include <sys/types.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <net/if.h>
--- a/RELICENSE/panlinux.md
+++ b/RELICENSE/panlinux.md
@@ -1,0 +1,17 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Andreas Hasenack that grants permission to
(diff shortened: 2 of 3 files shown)


tests/testutil.hpp
```diff
@@ -62,6 +62,8 @@
 #include <stdarg.h>
 #include <string>
 #include <string.h>
+#include <sys/types.h>
```
bluca: Appveyor seems to be broken at the moment unfortunately - are these headers available with MSVC? I think it's safer to move them below in the !_WIN32 section

panlinux: I have no idea, but should I then put the `getgrgid()` also inside such an `#ifdef`?

bluca: The IPC tests are already all disabled on Windows, so that's not necessary

panlinux:
Can the failed test run be retried? I don't get such an error locally.

sigiesec:
I retriggered the failed travis-ci test, it is indeed flaky and not related to your changes. The Windows builds are still in the queue.

panlinux:
I'm not sure my `sys/types.h` include is necessary, I added it because it was listed in the `getgrid()` manpage, but I see that a few lines down it's pulled in if `ZMQ_HAVE_AIX` is defined.

bluca:
It's probably not required with glibc, but it's definitely required with other c libraries like musl, so it's good to have it

PR #3462, opened 2019-03-27
Problem: Send with don't wait or zero timeout failed with eagain when pipe is not full

somdoron:
Because don't wait or timeout zero rarely process commands, the send operation might fail even if the pipe is not full.
Recv method is already protecting again such a behavior. The following example will fail, although it is clear that the pipe is not full:

```C
    void *ctx = zmq_ctx_new ();
    int timeo = 0;
    int hwm = 1;

    void *sb = zmq_socket (ctx, ZMQ_DEALER);
    zmq_setsockopt (sb, ZMQ_RCVHWM, &hwm, sizeof (int));
    zmq_bind (sb, "inproc://a");

    void *sc = zmq_socket (ctx, ZMQ_DEALER);
    zmq_setsockopt (sc, ZMQ_SNDHWM, &hwm, sizeof (int));
    zmq_setsockopt (sc, ZMQ_SNDTIMEO, &timeo, sizeof (int));
    zmq_connect (sc, "inproc://a");

    zmq_msg_t msg;
    zmq_msg_init (&msg);

    for (int i = 0; i < 10000; ++i) {
        int rc = zmq_msg_send (&msg, sc, 0);
        assert (rc >= 0);

        zmq_msg_recv (&msg, sb, 0);
    }

    zmq_msg_close (&msg);

    zmq_close (sb);
    zmq_close (sc);
    zmq_ctx_term (ctx);
```

Suggested solution: When send operation return EAGAIN with don't wait or zero timeout, process commands and try sending again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,6 +152,7 @@ test_socket_null
 test_xpub_verbose
 test_mock_pub_sub
 test_proxy_hwm
+test_sndtimeo_zero
 unittest_ip_resolver
 unittest_mtrie
 unittest_poller
--- a/Makefile.am
+++ b/Makefile.am
@@ -459,7 +459,8 @@ test_apps = \
 	tests/test_sodium \
 	tests/test_reconnect_ivl \
 	tests/test_mock_pub_sub \
-	tests/test_socket_null
+	tests/test_socket_null \
+	tests/test_sndtimeo_zero
 
 UNITY_CPPFLAGS = -I$(top_srcdir)/external/unity -DUNITY_USE_COMMAND_LINE_ARGS -DUNITY_EXCLUDE_FLOAT
 UNITY_LIBS = $(top_builddir)/external/unity/libunity.a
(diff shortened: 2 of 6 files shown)


somdoron:
test_hwm_pubsub is failing on windows, I guess setting the TCP buffers to zero is not enforced by windows.

@bluca @sigiesec any idea how to fix the test?

bluca:
I think 0 is not accepted anywhere, the minimum on Linux is 4096 I think - with a quick Google search I can't find what's the minimum size on Windows. Try with 4096 and see what happens.

Also there's a formatting problem: https://travis-ci.org/zeromq/libzmq/jobs/511829321

somdoron:
> Also there's a formatting problem: https://travis-ci.org/zeromq/libzmq/jobs/511829321

Yes, I saw that, but the error is actually not consistent with how the tests are written right now. All of the tests don't start a new line.

So the question is, should I be consistent with all tests or with the CI?

bluca:
We like to see green, so with the CI ;-)

sigiesec:
Actually there are some Windows documents that refer to setting SO_SNDBUF to 0, but I suspect that this might refer to a rather old Windows version, and the effect appears to be different than what might be expected: https://support.microsoft.com/en-us/help/214397/design-issues-sending-small-data-segments-over-tcp-with-winsock

somdoron:
I think I found another way to fix the test, waiting for the CI

somdoron:
Windows still failing, will investigate...

bluca:
@somdoron there's a merge conflict in the test

somdoron:
I need to investigate first why it doesn't work on Windows, I can make a separate pr for the test fix

PR #3473, opened 2019-04-05
Problem: CI format check fails

bluca:
Solution: fix it

Yes it looks weird, but it's a known problem in clang-format


sigiesec:
We can probably update the clang version used, probably that will fix this issue

PR #3476, opened 2019-04-11
Request handling leaked memory.  Depending on the application's use o…

TomMD:
Problem: xsend() can leak memory

Solution: free memory in case of error return from sendpipe().

### DIFF
--- a/src/req.cpp
+++ b/src/req.cpp
@@ -36,14 +36,6 @@
 #include "random.hpp"
 #include "likely.hpp"
 
-extern "C" {
-static void free_id (void *data_, void *hint_)
-{
-    LIBZMQ_UNUSED (hint_);
-    free (data_);
-}
-}
-
 zmq::req_t::req_t (class ctx_t *parent_, uint32_t tid_, int sid_) :
     dealer_t (parent_, tid_, sid_),
     _receiving_reply (false),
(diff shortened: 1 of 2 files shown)


src/req.cpp
```diff
@@ -95,8 +95,10 @@ int zmq::req_t::xsend (msg_t *msg_)
             id.set_flags (msg_t::more);
 
             rc = dealer_t::sendpipe (&id, &_reply_pipe);
-            if (rc != 0)
+            if (rc != 0) {
+                free (request_id_copy);
```
bluca: thanks for the PR, but it's actually completely pointless to do a malloc manually here - it's just 4 bytes. Could you please instead change from init_data to init_size and then memcpy &_request_id into id.data() ? That way it will be only on the stack

src/req.cpp
```diff
@@ -82,21 +82,20 @@ int zmq::req_t::xsend (msg_t *msg_)
...
-            zmq_assert (request_id_copy);
+            uint32_t request_id_copy;
 
-            *request_id_copy = _request_id;
+            request_id_copy = _request_id;
 
             msg_t id;
             int rc =
-              id.init_data (request_id_copy, sizeof (uint32_t), free_id, NULL);
+              id.init_data (&request_id_copy, sizeof (uint32_t), free_id, NULL);
```
bluca: As mentioned in the comment, this should change to:

```
rc = id.init_size (sizeof (uint32_t));
...
memcpy (id.data (), &_request_id, sizeof (uint32_t));
```
request_id_copy can be removed entirely.

That's because init_data takes a pointer to user-supplied data buffer, which is useful for large buffers but actually detriment for small buffers (here it's 4 bytes) and as you have found out it can lead to memory leaks if the message is not closed.
Using init_size with a small message on the other hand means it's entirely on the stack, so there's no dynamic memory allocation involved at all.

bluca: @methodmissing ping - do you have time to look at this? The current version of the PR breaks a few tests. Thanks!

TomMD: Yes, I'll get back to it soon.

bluca: @TomMD did you have time to look at this? Otherwise I'm happy to pull your branch and do the required fixup. The commit will still be yours for full credit.

TomMD: @bluca I believe I fixed the leak in the manner you desired but please to verify. I just code-monkey'ed in your lines without investing in understanding the `id.*` functions.

If you're curious: this and a prior PR (which I'll revive and see through) was motivated by an analysis tool, MuseDev.  We could turn it on for libzmq and get comments on PRs which are either worthy of discussion or easily ignored.

Typical Muse results are [summaries of project analysis results not necessarily related to the PR](https://github.com/TomMD/libzmq/pull/73#issuecomment-508338184) or [line comments for bugs potentially introduced by the PR](https://github.com/TomMD/libzmq/pull/9).  I've been happy running this for libzmq without any special configuration while other projects benefit from a few rules (like ignoring dead stores) - it's all configurable depending on developer preferences.

bluca: Thank you, this looks great, merged. Having more analysis tools is a good idea, what do I need to do? Is it free for oss projects?

bluca: I have enabled the muse app on the repository, let's see how it works

PR #3484, opened 2019-04-28
 Problem: Cannot get thread safe socket fd 

### LINKED ISSUE #2941: Problem: not all sockets support ZMQ_FD
Related to: https://github.com/zeromq/libzmq/issues/1434

To integrate well with existing eventloops, zmq sockets must be pollable via a system select/poll mechanism *without* using `zmq_poll`, because forcing the replacement the inner-loop implementation of an eventloop in order to use zmq sockets is generally not advisable, and not always available. ZMQ_FD is the mechanism for this, but it seems that some sockets, such as the new threadsafe variants and inproc connections, lack this feature (https://github.com/zeromq/pyzmq/issues/1139). This means that those sockets cannot be integrated into any application with an eventloop without resorting to periodic polling. At the most basic level, all sockets and all transports should support the existing ZMQ_FD.

An alternate solution that would solve longstanding issued with exposing the edge-triggered FD would be a new, traditional level-triggered FD API, with as many of the following features as achievable:

1. it's level-triggered
2. it works for all sockets and all transports
3. it's not an internal implementation detail that is exposed to users, but an explicit user API
4. either support read/write signaling on one FD or separate read-only FDs for read and write
5. closing the FD results in closing the socket, not crashing the process (I don't know how feasible that is)

```python
import zmq
s = zmq.Context().socket(zmq.DISH)
s.FD
```

EINVAL

# What's the expected result?

an FD I can use to integrate with select.

jean-airoldie:
This is based on [somdoron's comment](https://github.com/zeromq/libzmq/issues/2941#issuecomment-366983640).

This partially fixes #2941 by adding a way to retrieve the fd from the zmq::socket_poller_t signaler, if it exists. This allows integration of thread-safe socket types into event loops by creating a poller instance for each thread safe socket and polling the poller's fd.

### DIFF
--- a/RELICENSE/jean-airoldie.md
+++ b/RELICENSE/jean-airoldie.md
@@ -1,0 +1,7 @@
+Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Jean Airoldie that grants permission to relicense its copyrights in the libzmq C++ library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other Open Source Initiative approved license chosen by the current ZeroMQ BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle "jean-airoldie", with commit author "jean-airoldie <25088801+jean-airoldie@users.noreply.github.com>", are copyright of Jean Airoldie. This document hereby grants the libzmq project team to relicense libzmq, including all past, present and future contributions of the author listed above.
+
+Jean Airoldie 2019/04/27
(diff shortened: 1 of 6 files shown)


bluca:
Thanks - could you please send another PR adding the new function to the documentation https://github.com/zeromq/libzmq/blob/master/doc/zmq_poller.txt and unit tests for it?

jean-airoldie:
@bluca Sure thing!

jean-airoldie:
@bluca Note that merging this PR automatically closed #2941, but it does not fully resolve the issue. I'll be careful about using the word "fixes" next time.

PR #3490, opened 2019-05-02
Problem: Missing doc & unit tests for zmq_poller_fd

jean-airoldie:
Solution: Add doc & unit tests

This adds unit tests and doc for #3484. I also added `zmq_poller_wait` signature to the doc since it was missing.

### DIFF
--- a/doc/zmq_poller.txt
+++ b/doc/zmq_poller.txt
@@ -249,6 +249,8 @@ No registered event was signalled before the timeout was reached.
 On _zmq_poller_fd:
 *EINVAL*::
 The poller has no associated file descriptor.
+*EFAULT*::
+The provided 'poller' did not point to a valid poller.
 
 EXAMPLE
 -------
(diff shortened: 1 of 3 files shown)


bluca:
There's a couple of issues that fail unit tests could you please fix them?

- the poller needs to be destroyed at the end of the test, valgrind reports a memory leak and fails the test
- there's a formatting problem that fails the clang test, could you please fix it:

```
--- a/src/zmq.cpp
+++ b/src/zmq.cpp
@@ -1280,7 +1280,8 @@ int zmq_poller_wait_all (void *poller_,
 
 int zmq_poller_fd (void *poller_)
 {
-    if (!poller_ || !(static_cast<zmq::socket_poller_t *> (poller_)->check_tag())) {
+    if (!poller_
+        || !(static_cast<zmq::socket_poller_t *> (poller_)->check_tag ())) {
         errno = EFAULT;
         return -1;
     } else {
```

jean-airoldie:
Should be good now.

bluca:
thanks!

PR #3499, opened 2019-05-09
Problem: signature of zmq_poller_fd does is incompatible with regular…

sigiesec:
… error handling

Solution: change return type to int (again) and return fd via an output parameter


somdoron:
Czmq might be broken, as ot is using zmq poller

sigiesec:
This is only changing zmq_poller_fd, which was added just a few days ago

somdoron:
Cool

PR #3506, opened 2019-05-13
Problem: Cannot retrieve socket events concurrently

### LINKED ISSUE #3505: Problem: outgoing messages are still queued when auth fails
This is related to #882 but specifically for thread-safe socket types.

# Issue description
When `ZAP` is enabled and the connection succeeds but the authentication fails, the socket will still queue outgoing messages. This means that the application code will never be able to tell that the authentication failed via the `send` call (until the hwm is reached). This is a huge footgun in my opinion.

Currently this issue can be mitigated by making `connect` calls blocking by creating a `PAIR` socket that would wait for a  handshake event (success or failure) via `zmq_socket_monitor`. However this adds overhead to every `connect` call.

But this mitigation cannot work on thread safe types since multiple thread since making multiple concurrent connect calls would not guarantee that their `PAIR` socket would receive the appropriate event for their endpoint (e.g. first connect call receives the event for the endpoint of the second call and vice versa).

# Mitigation Ideas
* ~~Make the an API similar to `zmq_socket_monitor` but with a `PUB` socket instead of a `PAIR`. The message sent would the same as the first frame of the current API `zmq_socket_monitor` but the message topic would be the endpoint string. This way multiple concurrent connect calls would be guaranteed to received the events associated with their endpoint. Note that this cannot work with a `RADIO` socket since the maximum group length is of 15 chars. This works but its not ideal.~~
* Add a socket option / API flag so that the socket stays in mute state until a `ZMQ_EVENT_HANDSHAKE_SUCCEEDED` event. This would cause the `send` call to fail with `EAGAIN` if the autentication failed (similar to if the connection failed). Probably too hard to implement.
* ~~Add a zmq function call that blocks and returns only when the authentication succeeded or failed. Maybe a methods that calls connect underneat then wait until the event if received. This seems pretty ideal.~~

Edit1: Mitigation ideas that cannot work were crossed-out.

jean-airoldie:
Solution: Toggling the high bit of the 'event_version' field of the
'zmq_socket_monitor_versioned' method makes the monitoring socket
a `ZMQ_PUB` instead of a `ZMQ_PAIR`. Thus multiple subscribers can
retrieve events from the socket monitor concurrently.

This would fix #3505 by allowing concurrent calls to `zmq_connect` and `zmq_bind` that would block until authentication succeed or fails. Each call would need to create and connect a `SUB` socket to the monitor endpoint so that events can be independently retrieved. It is not ideal but it should work.

I realize that toggling the `event_version`'s high bit to modify behavior is a travesty but it probably won't ever be used anyway. This way, this is not a breaking change.

What do you think? Is this too messy?

### DIFF
--- a/doc/zmq_socket_monitor_versioned.txt
+++ b/doc/zmq_socket_monitor_versioned.txt
@@ -11,6 +11,8 @@ zmq_socket_monitor_versioned - monitor socket events
 SYNOPSIS
 --------
 *int zmq_socket_monitor_versioned (void '*socket', char '*endpoint', uint64_t 'events', int 'event_version');*
+*int zmq_socket_monitor_versioned_typed (
+    void '*socket', char '*endpoint', uint64_t 'events', int 'event_version', int 'type');*
 
 *int zmq_socket_monitor_pipes_stats (void '*socket');*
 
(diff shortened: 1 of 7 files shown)


bluca:
What about simply adding a new function call? The socket type used is anyway tied semantically to the function, as you can't really just pass the new flag and be done, but the client needs to be changed to sub as well.

Also, please add new APIs in the DRAFT section (and in src/zmq_draft.h concurrently)

jean-airoldie:
Yeah, makes sense. I'll do that tomorrow.

bluca:
thanks!

jean-airoldie:
Alright, this should be cleaner. However, I'm not a big fan of the method name I came up with.

jean-airoldie:
Also I ran `make clang-format-diff` which formatted unrelated code.

jean-airoldie:
After further thoughts, I realized that this PR [doesn't solve my issue](https://github.com/zeromq/libzmq/issues/3505#issuecomment-492125107).

jean-airoldie:
I'm reopening this PR but for a different reason than the original.

Problem: Socket monitoring requires a new ZMQ_PAIR for each socket. This means that if someone wants to log socket events in real-time, a new thread must be used per socket.

Solution: Allow `ZMQ_PUB` and `ZMQ_PUSH` sockets types for the monitoring. This way someone could create a ZMQ_PULL socket connected to multiple monitoring sockets at the same time.

jean-airoldie:
I'll make a new PR to be clearer.

PR #3508, opened 2019-05-15
Problem: Socket monitoring only allows ZMQ_PAIR

jean-airoldie:
Solution: Allow ZMQ_PUB and ZMQ_PUSH sockets types for the monitoring.
This way someone could create a ZMQ_PULL socket connected to multiple monitoring
sockets at the same time. Or event better, create a ZMQ_PUB monitor socket with `ZMQ_EVENT_ALL` specified and then let the subscribers decide what events they want to receive.

This is especially helpful if someone wants to dedicate only one thread to globally log events.

### DIFF
--- a/tests/test_monitor.cpp
+++ b/tests/test_monitor.cpp
@@ -126,8 +126,21 @@ void test_monitor_basic ()
 #if (defined ZMQ_CURRENT_EVENT_VERSION && ZMQ_CURRENT_EVENT_VERSION >= 2)      \
   || (defined ZMQ_CURRENT_EVENT_VERSION                                        \
       && ZMQ_CURRENT_EVENT_VERSION_DRAFT >= 2)
-void test_monitor_versioned_basic (bind_function_t bind_function_,
-                                   const char *expected_prefix_)
+void test_monitor_versioned_typed_invalid_socket_type ()
+{
+    void *client = test_context_socket (ZMQ_DEALER);
+
+    //  Socket monitoring only works with ZMQ_PAIR, ZMQ_PUB and ZMQ_PUSH.
+    TEST_ASSERT_FAILURE_ERRNO (
+      EINVAL, zmq_socket_monitor_versioned_typed (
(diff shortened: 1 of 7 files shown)


bluca:
zmq_socket_monitor_versioned is still draft, so it could be merged perhaps and just get the new parameter there?

sigiesec:
Yes, I think so too. Will remove the versioned but non-typed function

tests/test_monitor.cpp
```diff
@@ -136,14 +149,36 @@ void test_monitor_versioned_basic (bind_function_t bind_function_,
...
-      client, "inproc://monitor-client", ZMQ_EVENT_ALL_V2, 2));
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_socket_monitor_versioned (
-      server, "inproc://monitor-server", ZMQ_EVENT_ALL_V2, 2));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_socket_monitor_versioned_typed (
+      client, "inproc://monitor-client", ZMQ_EVENT_ALL_V2, 2, type_));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_socket_monitor_versioned_typed (
+      server, "inproc://monitor-server", ZMQ_EVENT_ALL_V2, 2, type_));
+
+    //  Choose the appropriate consumer socket type.
+    int mon_type;
```
bluca: this breaks all builds:

```
[  211s] tests/test_monitor.cpp: In function 'void test_monitor_versioned_typed_basic(bind_function_t, const char*, int)':
[  211s] tests/test_monitor.cpp:173:44: error: 'mon_type' may be used uninitialized in this function [-Werror=maybe-uninitialized]
[  211s]      void *server_mon = test_context_socket (mon_type);
```

https://build.opensuse.org/package/show/network:messaging:zeromq:git-draft/libzmq

could you please send a fix?

jean-airoldie: sure

bluca: thanks!

PR #3509, opened 2019-05-15
Problem: Potentially unitialized variable in #3508

jean-airoldie:
Solution: Initialized it.


jean-airoldie:
@bluca Seems like waiting [`msleep(SETTLE_TIME)`](https://github.com/zeromq/libzmq/blob/master/tests/test_monitor.cpp#L261) for the endpoint to unbind is not enough some of the time. Any recommendations?

bluca:
use different endpoints

PR #3511, opened 2019-05-16
 Support XPub socket send last value caching to last subscription pipe with ZMQ_XPUB_MANUAL_LAST_VALUE.

imkcy9:
Prevent sub connection recv duplicated message from manual XPub


bluca:
Thanks for the PR, please add:

1) relicense statement as the PR template requires
2) new apis in the draft section of the header and in src/zmq_draft.h
3) documentation about the new api in the socket option manpages
4) unit tests

PR #3514, opened 2019-05-20
Implement thread_t::setThreadName() on windows

moretromain:
Problem: thread_t::setThreadName() is not implemented on windows

Solution: implement setThreadName() by using the exception method (compatible with all win32 versions), and allow a larger thread name on all platforms (from 16 to 128 bytes)

More info here: https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2019

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -425,15 +425,11 @@ void zmq::thread_ctx_t::start_thread (thread_t &thread_,
     thread_.setSchedulingParameters (_thread_priority, _thread_sched_policy,
                                      _thread_affinity_cpus);
 
-    char namebuf[16] = "";
-#ifdef ZMQ_HAVE_WINDOWS
-    LIBZMQ_UNUSED (name_);
-#else
+    char namebuf[128] = "";
     snprintf (namebuf, sizeof (namebuf), "%s%sZMQbg%s%s",
               _thread_name_prefix.empty () ? "" : _thread_name_prefix.c_str (),
               _thread_name_prefix.empty () ? "" : "/", name_ ? "/" : "",
               name_ ? name_ : "");
-#endif
     thread_.start (tfn_, arg_, namebuf);
 }
 
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -47,6 +47,7 @@ static unsigned int __stdcall thread_routine (void *arg_)
 #endif
 {
     zmq::thread_t *self = (zmq::thread_t *) arg_;
+    self->setThreadName (self->_name.c_str ());
     self->_tfn (self->_arg);
     return 0;
 }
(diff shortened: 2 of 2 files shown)


bluca:
Thanks for the PR, but it causes build problems on Windows:

```
1>..\libzmq\src\thread.cpp(102) : error C2864: 'zmq::thread_t::setThreadName::<unnamed-tag>::_type' : only static const integral data members can be initialized within a class
1>..\libzmq\src\thread.cpp(103) : error C2864: 'zmq::thread_t::setThreadName::<unnamed-tag>::_name' : only static const integral data members can be initialized within a class
1>..\libzmq\src\thread.cpp(104) : error C2864: 'zmq::thread_t::setThreadName::<unnamed-tag>::_thread_id' : only static const integral data members can be initialized within a class
1>..\libzmq\src\thread.cpp(105) : error C2864: 'zmq::thread_t::setThreadName::<unnamed-tag>::flags' : only static const integral data members can be initialized within a class
```

Could you please have a look?

src/thread.cpp
```diff
@@ -92,10 +93,32 @@ void zmq::thread_t::setSchedulingParameters (
...
+        return;
+
+    struct {
+        DWORD _type;
+        LPCSTR _name;
+        DWORD _thread_id;
+        DWORD _flags;
+    } thread_info;
+
+	thread_info._type = 0x1000;
```
sigiesec: The indentation seems to be wrong here.

src/ctx.cpp
```diff
@@ -425,15 +425,11 @@ void zmq::thread_ctx_t::start_thread (thread_t &thread_,
     thread_.setSchedulingParameters (_thread_priority, _thread_sched_policy,
                                      _thread_affinity_cpus);
 
-    char namebuf[16] = "";
-#ifdef ZMQ_HAVE_WINDOWS
-    LIBZMQ_UNUSED (name_);
-#else
+    char namebuf[128] = "";
```
sigiesec: I don't think that this really works. If I am not mistaken, the maximum length is 16 chars on Linux and 32 on Windows. Longer names will be truncated, so it should be ensured that the truncated name stays meaningful.

moretromain: Ok, I'll keep the 32 limit on windows and 16 otherwise.

src/thread.cpp
moretromain: Sorry about that, I fixed this on a badly configured editor..I'll open a new PR

moretromain:
I opened a new PR here: https://github.com/zeromq/libzmq/pull/3515, I made too much a mess with this one. Sorry about that

PR #3515, opened 2019-05-20
Implement thread_t::setThreadName() on windows

moretromain:
Problem: thread_t::setThreadName() is not implemented on windows

Solution: implement setThreadName() by using the exception method (compatible with all win32 versions)

More info here: https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2019

### DIFF
--- a/src/thread.hpp
+++ b/src/thread.hpp
@@ -37,7 +37,7 @@
 #include <pthread.h>
 #endif
 #include <set>
-#include <string>
+#include <cstring>
 
 namespace zmq
 {
@@ -56,11 +56,11 @@ class thread_t
     inline thread_t () :
         _tfn (NULL),
         _arg (NULL),
-        _name (""),
         _started (false),
         _thread_priority (ZMQ_THREAD_PRIORITY_DFLT),
         _thread_sched_policy (ZMQ_THREAD_SCHED_POLICY_DFLT)
     {
+        memset (_name, 0, sizeof (_name));
     }
 
 #ifdef ZMQ_HAVE_VXWORKS
@@ -74,6 +74,8 @@ class thread_t
 
     //  Creates OS thread. 'tfn' is main thread function. It'll be passed
     //  'arg' as an argument.
+    //  Name is 16 characters max including terminating NUL. Thread naming is
+    //  implemented only for pthread, and windows when a debugger is attached.
     void start (thread_fn *tfn_, void *arg_, const char *name_);
 
     //  Returns whether the thread was started, i.e. start was called.
@@ -92,16 +94,13 @@ class thread_t
                                   int scheduling_policy_,
                                   const std::set<int> &affinity_cpus_);
 
-    // Sets the thread name, 16 characters max including terminating NUL.
-    // Only implemented for pthread. Has no effect on other platforms.
-    void setThreadName (const char *name_);
-
     //  These are internal members. They should be private, however then
     //  they would not be accessible from the main C routine of the thread.
     void applySchedulingParameters ();
+    void applyThreadName ();
     thread_fn *_tfn;
     void *_arg;
-    std::string _name;
+    char _name[16];
 
   private:
     bool _started;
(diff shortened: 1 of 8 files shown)


moretromain:
Is there a known workaround for the missing snprintf() on VS 2008/2010 ? 
`_snprintf()` is not C99 compliant so it's not a valid candidate here, so maybe I can provide a useful replacement for pre-VS2015 builds, and use it in other places as well (`io_thread_t::start()` for instance)

bluca:
https://github.com/zeromq/libzmq/blob/master/tests/test_security_curve.cpp#L51

moretromain:
Thanks, `_snprintf `doesn't behave the same as `sprintf`, since it writes an empty string if the output string length exceeds the given buffer size.
A better candidate without implementing a custom one would be [`snprintf_s()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-s-snprintf-s-l-snwprintf-s-snwprintf-s-l?view=vs-2019) with the _TRUNCATE argument, that guarantees a null-terminated string with appropriate truncation if the given buffer is too small.
Since those are only used in tests/debugging code, I think we can afford the extra security checks of the win CRT, not to mention that this is only for MSVC 2008/2010 builds.
I'll make that change, run a few test and submit it.

moretromain:
I pushed a change where I replaced the missing `snprintf()` by the best replacement candidate that is `_snprintf_f()`. I'll watch CI builds as I don't have a mingw or VS2010 environment to validate this on.

moretromain:
This turns out to be a platform nightmare...I'll finish fixing and tidying everything up and I'll open a new and clean PR with all this.

bluca:
There's no need to use a new PR, you can just rebase and squash and force push, it's going to be quicker

moretromain:
Or I'll do just that :)

src/thread.hpp
```diff
@@ -74,6 +74,8 @@ class thread_t
 
     //  Creates OS thread. 'tfn' is main thread function. It'll be passed
     //  'arg' as an argument.
+    //  Name is 16 characters max including terminating NUL. Thread naming is
+    //  implemented only for pthread and windows when a debugger is attached.
```
bluca: I suggest a comma here otherwise it's misleading: "only for pthread, and windows when a debugger is attached"

bluca:
Breaks the build on Fedora 30, could you please have a look?

```
[  453s] In file included from /usr/include/string.h:494,
[  453s]                  from /usr/include/c++/9/cstring:42,
[  453s]                  from src/thread.hpp:40,
[  453s]                  from src/thread.cpp:32:
[  453s] In function 'char* strncpy(char*, const char*, size_t)',
[  453s]     inlined from 'void zmq::thread_t::start(void (*)(void*), void*, const char*)' at src/thread.cpp:235:17:
[  453s] /usr/include/bits/string_fortified.h:106:34: error: 'char* __builtin_strncpy(char*, const char*, long unsigned int)' specified bound 16 equals destination size [-Werror=stringop-truncation]
[  453s]   106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
[  453s]       |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

https://build.opensuse.org/package/live_build_log/network:messaging:zeromq:git-draft/libzmq/Fedora_30/x86_64

moretromain:
My bad, strncpy() expects a destination that is at last one byte larger than the given count to ensure null-terminated target string. It's surprising that none of the CI build caught this though..
I fixed it [here](https://github.com/zeromq/libzmq/pull/3516)

bluca:
Thanks - newer versions of gcc are getting wiser with regards to those kind of errors, but the Travis CI runs on old Ubuntu versions

PR #3518, opened 2019-05-21
acinclude.m4, configure.ac: Fix several non-portable uses of `test`

### LINKED ISSUE #3517: configure errors because of syntax errors in the use of `test` shell command

# Issue description

On my system, where `/bin/sh -> /bin/dash`, runnung `./configure` produces errors which I tracked down to `configure.ac` and `acinclude.m4` using the `test` shell command with a `==` operator, syntax which is not valid in a POSIX-like shell like `dash`. See [`man 1p test`](https://www.unix.com/man-page/posix/1P/TEST/) for details.

# Environment

* libzmq version (commit hash if unreleased): cef80655
* OS: Void Linux x86_64

1.  Run `./autogen.sh`
2.  Run `./configure`

```sh
[...output from step 2...]
./configure: 21529: test: x: unexpected operator
./configure: 21534: test: unexpected operator
```
Full output of running the repro steps on my system:
step 1: [output](https://gist.github.com/f7762a406c62228a0012c55d49c6f754)
step 2: [output](https://gist.github.com/65dea53e4a9a48c1131af6d0de7b2069).

# What's the expected result?

Running `configure` should produce no such errors.

ackalker:
Use `test` with `=` operator instead of `==` operator to make POSIX-like shells happy.
This fixes #3517 .

### DIFF
--- a/RELICENSE/ackalker.md
+++ b/RELICENSE/ackalker.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Alain Kalker
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other 
+Open Source Initiative approved license chosen by the current ZeroMQ 
+BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle "ackalker", with
+commit author "Alain Kalker <a.c.kalker@gmail.com>", are copyright of Alain Kalker.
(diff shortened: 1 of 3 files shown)


bluca:
Thanks for the pr, could you please add a relicense statement as the template suggests? https://github.com/zeromq/libzmq/tree/master/RELICENSE

ackalker:
Ah, sure, I will do so. I had been wondering whether my changes would be too trivial to warrant a relicense, but there.

bluca:
It probably is, but it makes our life as maintainers easier to ask for the grant when we see "first-time contributor", so that if a larger contribution is made in the future we don't have to worry about checking

bluca:
thanks!

ackalker:
You're welcome!

Issue #3485, opened 2019-04-28
Router client disconnect() before recv() causing crash (assertion failed)


cenzhao:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description
TLDR;
sequence to reproduce the issue
router(client) <-> router(server) communication.
- server send 3 frames to client: <client-identity-string> <empty-string> <dummy-content>
- client side: poll until data ready to be read.
- client make a zmq_disconnect()
- client do zmq_recv() 3 times, first 2 times got correct data, 3rd call -> "Assertion failed: !_more (src/fq.cpp:112) Aborted (core dumped)"

Detailed Use Cases:
I have an event loop that handles two events:
(1): data ready to read on socket-1
(2): received a command on socket-2 to disconnect a peer. 
event loop keeps looping by calling zmq_poll() over these 2 sockets and handle them accordingly. 

if two events happened at the same time, the handling order is undetermined. If I happen to receive zmq-message form peer-X AND got a command to disconnect() peer-X. 

this following would crash the program:
zmq_poll() indicate two events ready.
zmq_disconnect(peer-X) (process event(2) first)
zmq_recvmsg() (process event(1)) -> assertion failed -> program crash.

based on documentation: http://api.zeromq.org/4-3:zmq-disconnect , if disconnect() called before poll(), poll() will block since message got discarded. 
But if poll() is called BEFORE disconnect(), we should still be able to read the data, shoudn't crash the program.

# Environment

* libzmq version: 4.3.1
* OS: ubuntu 16.04

# Minimal test code / Steps to reproduce the issue

    // create ROUTER client and server socket
    void *server = test_context_socket (ZMQ_ROUTER);
    void *client = test_context_socket (ZMQ_ROUTER);

    // set identity
    zmq_setsockopt(server, ZMQ_IDENTITY, "server", 6);
    zmq_setsockopt(client, ZMQ_IDENTITY, "client", 6);

    // server bind, client connect, let server send 3 frames: <client-id><><dummy-content>
    zmq_bind(server, "inproc://server");
    zmq_connect(client, "inproc://server");
    char msg1[] = "client";
    char msg2[] = "";
    char msg3[] = "test"; 
    zmq_send(server, msg1, 6, ZMQ_SNDMORE);
    zmq_send(server, msg2, 1, ZMQ_SNDMORE);
    zmq_send(server, msg3, 5, 0);

    // poll on client until data ready to read
    std::cout << "poll ..." << std::endl;
[...]

PR #3520, opened 2019-05-23
[WiP] Problem: after disconnecting an inproc router socket, attempting to r…

sigiesec:
…eceive a multipart message causes an assertion in fq.cpp (see #3485)

Solution: allow receiving messages in state term_req_sent1

### DIFF
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -185,7 +185,8 @@ bool zmq::pipe_t::read (msg_t *msg_)
 {
     if (unlikely (!_in_active))
         return false;
-    if (unlikely (_state != active && _state != waiting_for_delimiter))
+    if (unlikely (_state != active && _state != waiting_for_delimiter
+                  && _state != term_req_sent1))
         return false;
 
     for (bool payload_read = false; !payload_read;) {
--- a/tests/test_disconnect_inproc.cpp
+++ b/tests/test_disconnect_inproc.cpp
@@ -126,11 +126,51 @@ void test_disconnect_inproc ()
     test_context_socket_close (sub_socket);
 }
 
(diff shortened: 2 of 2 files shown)


sigiesec:
@cenzhao I was able to fix the issue you describe in #3485, however the change causes the existing test in test_disconnect_inproc to fail. Actually my assumption in #3485 was not correct. While the pipe is terminated asynchronously as I wrote, the assertion failure you saw is caused by the state change that is done synchronously within zmq_disconnect. I wasn't aware of that you are using inproc sockets here, which behave somewhat differently from sockets that use a "real" transport.

@bluca I am not sure if the expectation of that test is valid. After my change, now a message is received that is not expected to be received.

bluca:
I think the test expected the disconnect to be synchronous, at least in terms of stopping receives - it might be fine to change it, but I can't tell off the top of my head of what the full implications would be

cenzhao:
thanks @sigiesec for addressing this issue, I used "inproc" for easily demonstrate the problem. In our Facebook production, we saw the issue over "tcp" socket. Does your fix also address "tcp" case?

after `disconnect()`, I expect either 1. receive the whole message (3 parts) or 2. receives nothing (poll() on that socket will block). I think 2. would be ideal case :)

sigiesec:
@cenzhao ok. I don't think my fix fixes the tcp case. I will give that another look

sigiesec:
I am sorry, but I did not find time yet to give the tcp case a look, and I fear I also won't be able to do this in the near future. I would therefore close this pull request for now, since it doesn't build correctly in its current state.

bluca:
This is getting stale and not progressing, so I'll close it for now.

Issue #2076, opened 2016-08-03
Slow multicast data rate and ZMQ_RATE 


marcovolino:
I have created a simple PUB/SUB application to send small images using epgm multicast. I am hoping to be able to send the images at more than 20fps to many subscribers but so far it is taking about 40 seconds per image. 

I have read that the data rate defaults to 100 kbit/sec so have tried to use the ZMQ_RATE socket option but with no success.This is being run on two separate computers over a local network with the publisher on linux and subscriber on windows.

I have enabled PGM logging by setting the PGM_MIN_LOG_LEVEL=TRACE environment variable.
My publisher regularly puts out to following message
Trace: Recv again on not-full

and the subscriber gives the following output:
![selection_003](https://cloud.githubusercontent.com/assets/6390901/17362896/3776769e-596f-11e6-9692-8ae8c5b75c0d.png)

Does anyone know of any issues with ZMQ_RATE?

Many Thanks,
Marco

RogerFederer03:
I believe increase ZMQ_RATE should help. 100 kbit/sec is really to slow. You may try 1-100Mb/s

iamthebot:
Have you set ZMQ_RATE on BOTH the subscribers and publishers?

iamthebot:
For what it's worth, @marcovolino I've had nothing but trouble with OpenPGM. It buckles at around ~100mb/s, segfaults sometimes for no reason, etc. Doesn't help that openpgm hasn't seen active development in ages. If you need reliable multicast, I'd go with NORM. Otherwise, the UDP support in the new RADIO/DISH sockets is *much* more robust.

RogerFederer03:
@iamthebot  I have the same issue with OpenPGM, my receiver just crashed for no reason with data reached even 10mb/s, I am not sure why is that. 

Also does ZMP support NORM now? According to http://zeromq.org/topics:norm-protocol-transport. In the reference it said that The current NORM code has been tested with the Python pyzmq API binding. How is C/C++ going there? What language did you use for NORM?

iamthebot:
@RogerFederer03 honestly, while we love ZeroMQ for a lot of uses (eg; distributed systems, coordination, etc.). It even works very well for PUB/SUB unicast. That said, it's really no longer the best choice for multicast (aeron is). OpenPGM/NORM are particularly troubling choices because:
1. There's not an active developer community behind either of these protocols. They're hosted on google code (now deprecated) or on government websites. Neither mailing list has seen activity in ages.
[...]

PR #3521, opened 2019-05-24
Add missing information on multicasting with high data rates to PGM docs

jackpimbert:
Problem: PGM doc missing info on high data rates

Solution: Include a new section on configuring both ZMQ and the host
OS tx/rx buffers to facilitate sending large messages at a high data
rate with the PGM protocol.

---

Issue #2076 discusses problems that were encountered when attempting to multicast high data rates with the PGM protocol. There are some settings on both ZMQ and the host OS that must be changed in order to successfully multicast with higher rates. These setting changes have been added to the docs on PGM.

Please let me know if anything should be reformatted/moved to another place in the doc file.


bluca:
Thanks!

RRRaymond:
awesome! this doc helps me a lot!

PR #3532, opened 2019-06-10
Fix issues with SOCKS5 proxy connection

guillon:
Two issues where introduced by commit 12c4b55a:
- the proxy connection was done to the target address instead of
the proxy address
- on error the proxy connection status was not reset to unplugged


guillon:
This pull request fixes the issue #3533

bluca:
Thanks!

PR #3534, opened 2019-06-10
Implement SOCKS5 basic authentication

guillon:
Implement SOCKS5 basic authentication with username/password.

This adds two new setsockopt options ZMQ_SOCKS_USERNAME, and ZMQ_SOCKS_PASSWORD.
Refer to the documentation part for details.

Note that the choice has been done to add username/password through socket options as is done for some other level of authentication (for instance ZMQ_PLAIN_USERNAME/ZMQ_PLAIN_PASSWORD options).
Another choice (or an additional method) could be to augment the ZMQ_SOCKS_PROXY options (currently: "server:port") in order to include authentication information, for instance: "user:password@server:port".

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -656,6 +656,8 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_MULTICAST_LOOP 96
 #define ZMQ_ROUTER_NOTIFY 97
 #define ZMQ_XPUB_MANUAL_LAST_VALUE 98
+#define ZMQ_SOCKS_USERNAME 99
+#define ZMQ_SOCKS_PASSWORD 100
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_ZERO_COPY_RECV 10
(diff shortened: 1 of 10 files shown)


include/zmq.h
```diff
@@ -656,6 +656,8 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_MULTICAST_LOOP 96
 #define ZMQ_ROUTER_NOTIFY 97
 #define ZMQ_XPUB_MANUAL_LAST_VALUE 98
+#define ZMQ_SOCKS_USERNAME 99
```
bluca: these 2 need to be duplicated in src/zmq_draft.hpp otherwise the build will break

guillon: Ah ok. Sorry.
By the way, is it the right way to proceed, i.e. adding new sock opts as draft first?

bluca: yes

guillon: Done.

bluca:
Can you please fix the formatting according to the clang-format tool spec?
https://travis-ci.org/zeromq/libzmq/jobs/543765213

bluca:
I'll fix them separately

bluca:
@guillon thanks for your work on socks - if you had time, it would greately benefit the stability of the socks implementation if there could be at least some basic unit tests, at the moment there's no coverage unfortunately and that's how the regression was introduced

guillon:
Ok, there are some socks server implementation around in python.
Do you think it would be acceptable to have a dependency on some python module for the tests, or is there some scripting language you already use for testing?

bluca:
It would be fine, we can just make it optional and disabled by default, and enable it manually on the CI so that end-users are not affected. Thanks!

guillon:
Will guess about it. It may not be much more work to implement a simple server in C and it may simplify the test setup.
I will look at the existing tests and propose something.

PR #3539, opened 2019-06-11
Add tests for SOCKS proxy support

guillon:
This pull request adds tests/test_socks.cpp for testing socks proxy support.

Please do not merge yet, until I can check if the coverage report is what I should expect.

Also I added the test to Linux only for now as I do not have the possibility to test on other hosts.

### DIFF
--- a/tests/test_socks.cpp
+++ b/tests/test_socks.cpp
@@ -1,0 +1,894 @@
+/*
+    Copyright (c) 2019 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include <fcntl.h>
+#include <errno.h>
+
+#include "testutil.hpp"
+#include "testutil_unity.hpp"
+
+
+SETUP_TEARDOWN_TESTCONTEXT
+
+void recvall (int sock_fd, char *buffer, int len)
+{
+    int res;
+    int total = 0;
+    while (len - total > 0) {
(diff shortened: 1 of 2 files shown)


tests/test_socks.cpp
```diff
@@ -0,0 +1,564 @@
+/*
+    Copyright (c) 2007-2018 Contributors as noted in the AUTHORS file
```
bluca: 2019

bluca:
I've merged the other PR, please rebase to remove the formatting commit when you have a moment. Other than that, this is great, thanks!

tests/test_socks.cpp
guillon: ok, will fix the year.

guillon:
> I've merged the other PR, please rebase to remove the formatting commit when you have a moment. Other than that, this is great, thanks!
Yes, of course, will rebase and check results.

bluca:
The test seems to be failing on Travis:

```
FAIL: tests/test_socks
======================
Push server bound to: tcp://127.0.0.1:45612
Pull client connected to: tcp://127.0.0.1:45612
remote: sending 2 messages
client: receiving 2 messages
socks_server: starting server thread
Push server bound to: tcp://127.0.0.1:32853
Pull client connected to: tcp://127.0.0.1:32853
remote: sending 2 messages
socks_server: listening at address: 127.0.0.1:46387
socks_server: accepted client connection 1/1: client fd 15
socks_server: received greetings
socks_server: answered greetings (method: 0x0)
socks_server: received command (cmd: 1)
socks_server: received address (addr: 0x7f000001)
socks_server: received port (port: 32853)
socks_server: trying to connect to 7f000001:32853
socks_server: replied to request (err: 0x0)
socks_server: waiting for input (client fd: 15, remote fd: 16)
client: receiving 2 messages
socks_server: ready to read from fd 16
socks_server: ready to read from fd 15
socks_server: ready to read from fd 16
socks_server: ready to read from fd 15
socks_server: ready to read from fd 15
socks_server: ready to read from fd 16
socks_server: ready to read from fd 16
socks_server: ready to read from fd 15
socks_server: closed remote connection 1/1: client fd -1
socks_server: closed client connection 1/1: client fd -1
socks_server: closed server
Push server bound to: tcp://127.0.0.1:42580
Pull client connected to: tcp://127.0.0.1:42580
remote: sending 2 messages
socks_server: starting server thread
socks_server: listening at address: 127.0.0.1:42224
```

bluca:
it's a timeout, so it's taking more than 60 seconds

guillon:
> it's a timeout, so it's taking more than 60 seconds

Yes, normally, it's very quick, hence I guess that the subscriber can't connect to the proxy or the proxy to the publisher. May be an issue due to timing, I will check locally.

guillon:
Ok, it's an issue with the startup order (depends upon the host load) of the proxy server versus the zmp pull client.
I will add some semaphore in order to control the startup sequence.

Note that I published pr #3542 as I had to fix the assertion report in order to isolate the right line number where the test failed.

bluca:
The connection should be retried automatically if the endpoint is not reachable, so isn't this highlighting a bug in the implementation?

guillon:
You're right, the problem actually is not the connection hanshake but simply the server bind address which is not yet determined (because we ask for an arbitrary port).
It's a bug in the test. I will fix it, actually I do not need any timing nor semaphore, I just have to bind the sockets first, before launching the proxy thread.

guillon:
Fixed the issue in the test hopefully. The actual threads ordering/timing should not be a problem anymore.

Will wait for the coverage report and see if I can improve it...

bluca:
looks good - seems like the main part missing from coverage is ipv6, which is not easy to get on travis unfortunately

guillon:
I think that I can improve the coverage, please wait before merging.

Note that the appveyor tests do not fail due to this patch.

bluca:
Sure - don't worry about appveyor

guillon:
I completed the socks tests to a reasonable level of coverage for the three related files options.cpp/socks_connecter.cpp/socks.cpp.
I guess that the coverage should be maximized except for the the case of malformed server error (calls to error() in the code).

You can merge it if all is good. Thanks for your help.

PR #3542, opened 2019-06-12
Problem: test assertions not reporting the right line number

guillon:
Solution: pass the line number as part of the testutil functions


sigiesec:
Nice :) This has annoyed me several times, but I never found the time to give it a look and fix it.

PR #3545, opened 2019-06-14
[WIP] Problem: Radio cannot receive subscriptions

### LINKED ISSUE #3474: Detect group join on ZMQ_RADIO
I would like to implement a last-value-cache for the radio-dish pattern. To do so, I would most likely need to be able to detect a dish joining a group of a given radio socket. As this pattern is not well documented since it is still in draft, I would like to know if this is currently supported.

As an aside, is this github repository the right place to ask API related questions?

jean-airoldie:
Add `ZMQ_RADIO_NOTIFY` sockopt that queue join and leave messages
if set to 1. Add `zmq_msg_is_leave` and `zmq_msg_is_join` methods
to check whether a msg is a leave or join notification.

This would close #3474.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -658,6 +658,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_XPUB_MANUAL_LAST_VALUE 98
 #define ZMQ_SOCKS_USERNAME 99
 #define ZMQ_SOCKS_PASSWORD 100
+#define ZMQ_RADIO_NOTIFY 101
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_ZERO_COPY_RECV 10
@@ -671,6 +672,8 @@ ZMQ_EXPORT int zmq_msg_set_routing_id (zmq_msg_t *msg, uint32_t routing_id);
 ZMQ_EXPORT uint32_t zmq_msg_routing_id (zmq_msg_t *msg);
 ZMQ_EXPORT int zmq_msg_set_group (zmq_msg_t *msg, const char *group);
 ZMQ_EXPORT const char *zmq_msg_group (zmq_msg_t *msg);
+ZMQ_EXPORT bool zmq_msg_is_leave (zmq_msg_t *msg);
+ZMQ_EXPORT bool zmq_msg_is_join (zmq_msg_t *msg);
 
 /*  DRAFT Msg property names.                                                 */
 #define ZMQ_MSG_PROPERTY_ROUTING_ID "Routing-Id"
--- a/src/radio.cpp
+++ b/src/radio.cpp
@@ -38,7 +38,9 @@
 
 zmq::radio_t::radio_t (class ctx_t *parent_, uint32_t tid_, int sid_) :
     socket_base_t (parent_, tid_, sid_, true),
-    _lossy (true)
+    _pending_notifs (),
+    _lossy (true),
+    _notify (false)
 {
     options.type = ZMQ_RADIO;
 }
@@ -94,6 +96,19 @@ void zmq::radio_t::xread_activated (pipe_t *pipe_)
                     }
                 }
             }
+
+            //  Notify the user of the subscription.
+            if (unlikely (_notify)) {
+                msg_t notif;
+                int rc = notif.init ();
+                errno_assert (rc == 0);
+
+                rc = notif.move (msg);
+                errno_assert (rc == 0);
+                //  If the user does not `recv` the notifications,
+                //  the queue will grow infinitely.
+                _pending_notifs.push (notif);
+            }
         }
         msg.close ();
     }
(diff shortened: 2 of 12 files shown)


src/socket_base.cpp
```diff
@@ -1630,7 +1630,7 @@ void zmq::socket_base_t::pipe_terminated (pipe_t *pipe_)
 void zmq::socket_base_t::extract_flags (msg_t *msg_)
 {
     //  Test whether routing_id flag is valid for this socket type.
-    if (unlikely (msg_->flags () & msg_t::routing_id))
+    if (unlikely (msg_->flags () & msg_t::routing_id == msg_t::routing_id))
```
bluca: the bitwise and needs to be in brackets

include/zmq.h
```diff
@@ -658,6 +658,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_XPUB_MANUAL_LAST_VALUE 98
 #define ZMQ_SOCKS_USERNAME 99
 #define ZMQ_SOCKS_PASSWORD 100
+#define ZMQ_RADIO_NOTIFY 101
```
bluca: needs to be duplicated in src/zmq_drafts.hpp

src/msg.hpp
```diff
@@ -196,13 +196,13 @@ class msg_t
...
+        type_max = 105,
+
         //  Join message for radio_dish
         type_join = 106,
 
         //  Leave message for radio_dish
-        type_leave = 107,
-
-        type_max = 107
+        type_leave = 107
```
bluca: not sure I follow the reason for this change?

jean-airoldie: Otherwise we get an error when we close the msg because of https://github.com/zeromq/libzmq/blob/master/src/msg.cpp#L50 and https://github.com/zeromq/libzmq/blob/master/src/msg.cpp#L218.

jean-airoldie: I think this was meant to prevent leaking private msg types.

bluca: it's to check that the type is in range and wasn't corrupted - this will make the check fail for a join/leave message

jean-airoldie: Right, I'm blind.

src/socket_base.cpp
```diff
@@ -1630,7 +1630,7 @@ void zmq::socket_base_t::pipe_terminated (pipe_t *pipe_)
 void zmq::socket_base_t::extract_flags (msg_t *msg_)
 {
     //  Test whether routing_id flag is valid for this socket type.
-    if (unlikely (msg_->flags () & msg_t::routing_id))
+    if (unlikely ((msg_->flags () & msg_t::routing_id) == msg_t::routing_id))
```
bluca: also not sure this is required? if the routing_id bit is set, then the result cannot be anything else that routing_id, which already evaluates as true

jean-airoldie: no its not required

src/radio.cpp
```diff
@@ -38,7 +38,9 @@
 
 zmq::radio_t::radio_t (class ctx_t *parent_, uint32_t tid_, int sid_) :
     socket_base_t (parent_, tid_, sid_, true),
-    _lossy (true)
+    _lossy (true),
+    _notify (false),
+    _pending_notifs ()
```
bluca: gcc wants the initialisations in the same order as the declarations (it's very annoying):

```
../src/radio.hpp: In constructor ‘zmq::radio_t::radio_t(zmq::ctx_t*, uint32_t, int)’:
../src/radio.hpp:85:10: error: ‘zmq::radio_t::_notify’ will be initialized after [-Werror=reorder]
     bool _notify;
          ^
../src/radio.hpp:80:23: error:   ‘std::deque<zmq::msg_t> zmq::radio_t::_pending_notifs’ [-Werror=reorder]
     std::deque<msg_t> _pending_notifs;
                       ^
../src/radio.cpp:39:1: error:   when initialized here [-Werror=reorder]
 zmq::radio_t::radio_t (class ctx_t *parent_, uint32_t tid_, int sid_) :
```

jean-airoldie:
Note that for `zmq_msg_is_leave` and `zmq_msg_is_join`, the case where the `zmq_msg_t*` is invalid is not handled.

jean-airoldie:
Also a `recv_hwm` mechanism would be ideal to prevent the notification queue from filling up infinitely if the user does not recv from it.

jean-airoldie:
Also I just realized that `ZMQ_RADIO_NOTIFY` won't work for one way transports such as `PGM`. So I'm reconsidering this PR.

jean-airoldie:
I think that if you need communication in both direction you better use a `SERVER` socket anyway. You could write application code on top of it to provide the same features. I think its an anti-pattern so I'm closing.

PR #3547, opened 2019-06-17
Problem: disconnect & unbind doc is misleading

### LINKED ISSUE #3529: Problem: disconnect & unbind doc is misleading
First of all I did not expect [unbind](https://github.com/zeromq/libzmq/blob/master/src/zmq.cpp#L316) and [disconnect](https://github.com/zeromq/libzmq/blob/master/src/zmq.cpp#L324) to be the same identical function based on their documentation. Moreover, it seems that the behaviour of `term_endpoint` differ depending on the context.

For instance, If is set `linger=0` before disconnecting a `Client` socket from a connected endpoint, then the incoming queue is discarted, which is what I expect based on the disconnect documentation. However, if I set `linger=0` before unbinding a `Server` socket from a bound endpoint, then the incoming queue is not discarded and I can still `recv` messages from it.

So my question is, what is the actual behavior of disconnect / unbind?

jean-airoldie:
This closes #3529.

I also made it explicit that `zmq_disconnect` is the preferred option.

This documentation is inferred from my usage of the `Server` and `Client` socket. However, I am unsure if it can be generalized to all socket types.

### DIFF
--- a/doc/zmq_unbind.txt
+++ b/doc/zmq_unbind.txt
@@ -4,7 +4,7 @@ zmq_unbind(3)
 
 NAME
 ----
-zmq_unbind - Stop accepting connections on a socket
+zmq_unbind - Another name for zmq_disconnect
 
 
 SYNOPSIS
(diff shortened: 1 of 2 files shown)


doc/zmq_unbind.txt
```diff
@@ -14,68 +14,8 @@ int zmq_unbind (void '*socket', const char '*endpoint');
 
 DESCRIPTION
 -----------
-The _zmq_unbind()_ function shall unbind a socket specified
```
bluca: Why the removal of the unbind documentation?

bluca: while the underlying implementation might use common code, the behaviour is different, and most importantly they are semantically different - connect/disconnect and bind/unbind is a simple model to understand, while mixing them I'm afraid would just confuse users

bluca: in other words, the behaviour of a bound socket on unbind and a connected socket on disconnect is quite different, so it makes sense to have differen functions for them

jean-airoldie: > connect/disconnect and bind/unbind is a simple model to understand

Yes it is simpler to understand but thats not how the library works. I agree that a disconnect function whose behaviour change according to the context might confuse users but thats because the function itself is confusing. The underlying implementation is [literally the same](https://github.com/zeromq/libzmq/blob/master/src/zmq.cpp#L316). The current documentation makes it seem like calling disconnect on a bound socket is undefined behaviour and vice verse. As a user I did not expect that at all, and it confused me more than a `disconnect` method whose behaviour depends on the context.

Personally I think that if someone wanted to make disconnect and unbind separate it should first check if the endpoint is indeed connected or bound first and return an error otherwise. However, it is not possible without breaking the API, so new methods should be implemented. I think this would be ideal.

jean-airoldie: In other words, this kinda feels like lying to users to make the API easier to use. I think we should instead make another API with an actual separate behaviour for disconnect vs. unbind, as opposed to pretending like its the case.

PR #3548, opened 2019-06-19
Problem: radix tree's apply function uses incorrect resizing logic

ssbl:
Solution: rewrite the function using a vector

The buffer is extended once by 256 bytes, which will not be enough if
the node holds data larger than this number.

### DIFF
--- a/src/radix_tree.cpp
+++ b/src/radix_tree.cpp
@@ -34,6 +34,7 @@
 
 #include <stdlib.h>
 #include <string.h>
+#include <vector>
 
 node::node (unsigned char *data) : data_ (data)
 {
@@ -517,35 +518,25 @@ bool zmq::radix_tree::check (const unsigned char *key, size_t size)
 
 static void
 visit_keys (node n,
-            unsigned char **buffer,
-            size_t buffer_size,
-            size_t maxbuffer_size,
+            std::vector<unsigned char> &buffer,
             void (*func) (unsigned char *data, size_t size, void *arg),
             void *arg)
 {
-    if (buffer_size >= maxbuffer_size) {
-        maxbuffer_size += 256;
-        *buffer =
-          static_cast<unsigned char *> (realloc (*buffer, maxbuffer_size));
-        zmq_assert (*buffer);
-    }
-
     for (size_t i = 0; i < n.prefix_length (); ++i)
-        (*buffer)[buffer_size++] = n.prefix ()[i];
+        buffer.push_back (n.prefix ()[i]);
     if (n.refcount () > 0)
-        func (*buffer, buffer_size, arg);
+        func (buffer.data (), buffer.size (), arg);
     for (size_t i = 0; i < n.edgecount (); ++i)
-        visit_keys (n.node_at (i), buffer, buffer_size, maxbuffer_size, func,
-                    arg);
-    buffer_size -= n.prefix_length ();
+        visit_keys (n.node_at (i), buffer, func, arg);
+    for (size_t i = 0; i < n.prefix_length (); ++i)
+        buffer.pop_back ();
 }
 
 void zmq::radix_tree::apply (
   void (*func) (unsigned char *data, size_t size, void *arg), void *arg)
 {
-    unsigned char *buffer = NULL;
-    visit_keys (root_, &buffer, 0, 0, func, arg);
-    free (buffer);
+    std::vector<unsigned char> buffer;
+    visit_keys (root_, buffer, func, arg);
 }
 
 size_t zmq::radix_tree::size () const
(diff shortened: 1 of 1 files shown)


src/radix_tree.cpp
```diff
@@ -517,35 +518,25 @@ bool zmq::radix_tree::check (const unsigned char *key, size_t size)
...
-          static_cast<unsigned char *> (realloc (*buffer, maxbuffer_size));
-        zmq_assert (*buffer);
-    }
-
     for (size_t i = 0; i < n.prefix_length (); ++i)
-        (*buffer)[buffer_size++] = n.prefix ()[i];
+        buffer.push_back (n.prefix ()[i]);
     if (n.refcount () > 0)
-        func (*buffer, buffer_size, arg);
+        func (buffer.data (), buffer.size (), arg);
```
bluca: 1>..\libzmq\src\radix_tree.cpp(528) : error C2039: 'data' : is not a member of 'std::vector<_Ty>'
1>        with
1>        [
1>            _Ty=unsigned char
1>        ]
1>trie.cpp
1>thread.cpp

This breaks the build on Windows (see Appveyor) could you please check?

ssbl: Sorry, looks like `vector::data` was added in C++11. Also it doesn't necessarily return NULL when the vector is empty, which was also not correct. I've attempted a fix in the latest commit.

ssbl:
thanks!

bluca:
@ssbl a static analysis tool reports the following in the radix tree source file:

1. MEMORY_LEAK
    - At [src/radix_tree.cpp(407,1)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/radix_tree.cpp#L407)
    - memory dynamically allocated by call to `make_node()` at line 340, column 11 is not reachable after line 407, column 1.
1. MEMORY_LEAK
    - At [src/radix_tree.cpp(407,1)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/radix_tree.cpp#L407)
    - memory dynamically allocated to `current_node._data` by call to `resize` at line 356, column 9 is not reachable after line 407, column 1.
1. MEMORY_LEAK
    - At [src/radix_tree.cpp(527,1)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/radix_tree.cpp#L527)
    - memory dynamically allocated to `current_node._data` by call to `resize` at line 448, column 9 is not reachable after line 527, column 1.
1. MEMORY_LEAK
    - At [src/radix_tree.cpp(527,1)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/radix_tree.cpp#L527)
    - memory dynamically allocated to `parent_node._data` by call to `resize` at line 477, column 9 is not reachable after line 527, column 1.

Could you please have a look?

ssbl:
Thanks for pointing these out, I did take a look into this a couple of weeks ago in [this pull request](https://github.com/TomMD/libzmq/pull/44#issuecomment-495755831), but I couldn't figure out how these allocations might leak. I've installed the tool now and I'll submit a PR when I find something.

bluca:
Thanks!

ssbl:
Just to be sure, did you install the tool on your fork and submit a PR to that fork to run this tool? I've done this [here](https://github.com/ssbl/libzmq/pull/1) to test this but I don't know if that will work.

bluca:
No, I've enabled it on zeromq/libzmq

ssbl:
I couldn't get the tool to run on my fork, the check has been queued for around 20 hours now. However, inspecting the code again I doubt that those allocations can leak. Each of these allocations has a parent object to which we pass ownership through `set_node_at` and in each case the parent is distinct from the allocation, so I don't think we're creating any cycles either. I'll try to verify this with some tests.

@bluca, @sigiesec - any thoughts on why these could be flagged as leaks?

bluca:
Sounds like a false positive perhaps? @TomMD any idea?

TomMD:
Yes, I believe this is a false positive caused by, I'm guessing, the memcpy of the pointer of an element.  I'll make a minimal test case to aid our internal investigation.

@sigiesec @bluca We will be disabling both the preexisiting issues command and the summary comment later today.  Thank you for your feedback and tolerance of the e-mail flood.  Any new issues will still be posted as line comments and appear in the PR comment stream.  The pre-existing issues will be unavailable until we incorporate the github check API, connect the console to the database, and adjust settings to allow display of issues publicly for public repositories.

bluca:
great, thank you!

PR #3550, opened 2019-06-22
Problem: missing heartbeat test for draft sockets

jean-airoldie:
Solution: Add heatbeat tests for RADIO, DISH, CLIENT, SERVER, GATHER and SCATTER
socket types.

Note that currently there are two failures, which are consistent with my experience. I'll open an issue concerning this.
```
test_heartbeat_notimeout_dish_radio:FAIL
test_heartbeat_notimeout_dish_radio_with_curve:FAIL
```


bluca:
We shouldn't commit tests that are known not to work as it will break all the builds, please leave only client server. Also it needs to be ifdefed for non draft builds.

jean-airoldie:
Alright, I did just that. But what is the point of testing if we remove valid tests that fail? In this case it indicates that heartbeating with radio & dish is simply not working.

bluca:
Tests are for bugs and regressions, this is a feature that is not implemented

jean-airoldie:
In this case should I document that this heart-beating is currently not supported for RADIO / DISH?

PR #3566, opened 2019-07-09
Free memory when curve_client sends invalid ready message.

TomMD:
Problem: Malicious clients can cause arbitrary amounts of memory to leak (well, fixed amout an arbitrary number of times).

Solution: Free memory in the error case.

### DIFF
--- a/src/curve_client.cpp
+++ b/src/curve_client.cpp
@@ -240,6 +240,7 @@ int zmq::curve_client_t::process_ready (const uint8_t *msg_data_,
         session->get_socket ()->event_handshake_failed_protocol (
           session->get_endpoint (), ZMQ_PROTOCOL_ERROR_ZMTP_CRYPTOGRAPHIC);
         errno = EPROTO;
+        free (ready_plaintext);
         return -1;
     }
 


sigiesec:
Great, thanks for the fix!

We should really use a limited smart pointer for such cases (where no moving is required). What do you think, @bluca?

bluca:
Is that supported by C++ 98?

sigiesec:
There is no predefined template for that, but it can be implemented. The problem with C++98 and smart pointers in general is that you don't have move semantics, but for cases where the ownership is never passed, a very simple template suffices, along the lines of
```
template<typename T>
class array_ptr_t
{
public:
   array_ptr_t(size_t size) { _ptr = static_cast<T*>(malloc(size * sizeof(T)); alloc_assert(_ptr); }
   T *get() { return _ptr; }
   ~array_ptr_t() { free(_ptr); }
private:
  T* const _ptr;
};
```
However, maybe for such cases it would even be better to use a std::vector instead?

bluca:
yes it seems like a vector would do just fine - do you have time to give it a shot?

sigiesec:
I hope so, in the next few days :)

bluca:
thanks!

PR #3569, opened 2019-07-09
acinclude.m4, configure.ac: use default cacheline size in case of cross-compiling

mdionisio:
when cross-compiling with autotools it is not possible to use machine variable for cacheline size

### DIFF
--- a/RELICENSE/mdionisio.md
+++ b/RELICENSE/mdionisio.md
@@ -1,0 +1,13 @@
+# Permission to Relicense under MPLv2
+
+This is a statement by Thomas M. DuBuisson
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2).
+
+A portion of the commits made by the Github handle "mdionisio", with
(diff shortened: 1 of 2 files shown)


bluca:
@mdionisio thanks for the PR, please push a commit with a relicense statement as the PR template suggests - https://github.com/zeromq/libzmq/tree/master/RELICENSE

bluca:
I had hoped getconf would do the right thing when cross building, should have checked :-/

mdionisio:
getconfig has to work but surelly not 
`cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size`

do you suggest to add the workaround for cross-compiling only in that case?

bluca:
Yeah if getconf is supposed to work (can't check right now) I think it's better to check for cross only for the /sys fallback - arm7 devices can have 32 bytes cacheline sizes for example, would be nice if they got the value right when cross compiling for them

mdionisio:
the possible failure of `getconfig` or `cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size 2>/dev/null` are already managed with `|| echo 64`

PR #3570, opened 2019-07-09
Static analysis warnings

bluca:
@TomMD it seems the tool doesn't see when variables are used for array sizes, these are false positives:

1. DEAD_STORE
    - At [src/tcp_address.cpp(121,5)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/tcp_address.cpp#L121)
    - The value written to &max_port_str_length (type unsigned long const ) is never used.
1. DEAD_STORE
    - At [src/tcp_address.cpp(288,5)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/tcp_address.cpp#L288)
    - The value written to &max_mask_len (type unsigned long const ) is never used.

These are also false positives, the struct is initialized by the systemcall:

1. UNINITIALIZED_VALUE
    - At [src/stream_engine.cpp(79,13)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/stream_engine.cpp#L79)
    - The value read from cred.gid was never initialized.
1. UNINITIALIZED_VALUE
    - At [src/stream_engine.cpp(79,13)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/stream_engine.cpp#L79)
    - The value read from cred.pid was never initialized.
1. UNINITIALIZED_VALUE
    - At [src/stream_engine.cpp(79,13)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/stream_engine.cpp#L79)
    - The value read from cred.uid was never initialized.

These are also false positives:

1. UNINITIALIZED_VALUE
    - At [src/tweetnacl.c(138,15)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/tweetnacl.c#L138)
    - The value read from w[_] was never initialized.
1. UNINITIALIZED_VALUE
    - At [src/tweetnacl.c(149,7)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/tweetnacl.c#L149)
    - The value read from x[_] was never initialized.
1. UNINITIALIZED_VALUE
    - At [src/tweetnacl.c(575,18)](https://github.com/zeromq/libzmq/blob/90ff30c086bc299d619346a04d9eb5a02e96636e/src/tweetnacl.c#L575)
    - The value read from a[_] was never initialized.

### DIFF
--- a/src/udp_engine.cpp
+++ b/src/udp_engine.cpp
@@ -396,8 +396,8 @@ void zmq::udp_engine_t::out_event ()
     if (rc == 0) {
         msg_t body_msg;
         rc = _session->pull_msg (&body_msg);
-        //  TODO rc is not checked here. We seem to assume rc == 0. An
-        //  assertion should be added.
+        //  If there's a group, there should also be a body
+        errno_assert (rc == 0);
 
         const size_t group_size = group_msg.size ();
         const size_t body_size = body_msg.size ();
@@ -412,7 +412,7 @@ void zmq::udp_engine_t::out_event ()
                 rc = group_msg.close ();
                 errno_assert (rc == 0);
 
-                body_msg.close ();
+                rc = body_msg.close ();
                 errno_assert (rc == 0);
 
                 return;
--- a/builds/abi-compliance-checker/ci_build.sh
+++ b/builds/abi-compliance-checker/ci_build.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -x
+set -e
 
 cd ../../
 
--- a/tests/testutil_monitoring.cpp
+++ b/tests/testutil_monitoring.cpp
@@ -339,9 +339,9 @@ void expect_monitor_event_v2 (void *monitor_,
     }
     if (expected_remote_address_
         && 0 != strcmp (remote_address, expected_remote_address_)) {
-        pos += snprintf (pos, sizeof buf - (pos - buf),
-                         "Expected remote address %s, but received %s\n",
-                         expected_remote_address_, remote_address);
+        snprintf (pos, sizeof buf - (pos - buf),
+                  "Expected remote address %s, but received %s\n",
+                  expected_remote_address_, remote_address);
     }
     free (local_address);
     free (remote_address);


bluca:
Can a file be ignored? We don't want to modify external sources (unity source code)

TomMD:
> @TomMD it seems the tool doesn't see when variables are used for array sizes, these are false positives:

Yes I see.  The information is lost by clang, making the dead store that isn't there in source appear in the bitcode.  I am considering solutions.

> These are also false positives, the struct is initialized by the systemcall:
> ...
> These are also false positives:
> ...

While there will always be some false positives they needn't be displayed in every pull request.  If the bug is not related to the pull request then it appears in the summary.  We are considering moving the summary to console.muse.dev via a hyperlinking the text `Muse Analysis Complete`.  Other solutions are possible and I'm trying to understand what is most preferable.

> Can a file be ignored? We don't want to modify external sources (unity source code)

I can add that capability.  The tool options are placed in repo under `.muse/config` for example:

```
ignore = [ "DEAD_STORE", "UNINITIALIZED_VALUE"]
```

It should become a `config.toml` file, I believe, and have another option for ignored files (vs bugs) making the configuration for libzmq something like:

```
ignoreBugs = [ "DEAD_STORE", "UNINITIALIZED_VALUE" ]
ignoreFiles = [ "*tweetnacl*", "*/unity.c" ] # Design Q for team: regex vs bash globbing vs full paths
```

We can have this ready for you guys soon.  I'll close with the team on a design.

bluca:
Thanks - something else that should really be done is moving the output in the "checks" section with the other CI tools, rather than having comments, as they are quite annoying (they spam notifications)

sigiesec:
At least they are annoying since they refer to issues that are already present on master. If they were only new issues, the commenting behaviour might be acceptable.

bluca:
@sigiesec can this be merged in the meanwhile? These fixes are for valid warnings

src/udp_engine.cpp
```diff
@@ -396,8 +396,8 @@ void zmq::udp_engine_t::out_event ()
     if (rc == 0) {
         msg_t body_msg;
         rc = _session->pull_msg (&body_msg);
-        //  TODO rc is not checked here. We seem to assume rc == 0. An
-        //  assertion should be added.
+        //  If there's a group, there should also be a body
+        errno_assert (rc == 0);
```
bluca: @somdoron FYI I think this is the right thing here, but correct me if I'm wrong

PR #3573, opened 2019-07-10
Problem: use of malloc is error-prone

sigiesec:
Solution: use std::vector instead, simplify control-flow

### DIFF
--- a/src/curve_server.cpp
+++ b/src/curve_server.cpp
@@ -327,33 +327,29 @@ int zmq::curve_server_t::process_initiate (msg_t *msg_)
     const size_t clen = (size - 113) + crypto_box_BOXZEROBYTES;
 
     uint8_t initiate_nonce[crypto_box_NONCEBYTES];
-    uint8_t *initiate_plaintext =
-      static_cast<uint8_t *> (malloc (crypto_box_ZEROBYTES + clen));
-    alloc_assert (initiate_plaintext);
-    uint8_t *initiate_box =
-      static_cast<uint8_t *> (malloc (crypto_box_BOXZEROBYTES + clen));
-    alloc_assert (initiate_box);
+    std::vector<uint8_t> initiate_plaintext (crypto_box_ZEROBYTES + clen);
+    std::vector<uint8_t> initiate_box (crypto_box_BOXZEROBYTES + clen);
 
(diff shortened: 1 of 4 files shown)


src/curve_server.cpp
```diff
@@ -410,62 +404,52 @@ int zmq::curve_server_t::process_initiate (msg_t *msg_)
...
     //  Create Box [metadata](S'->C')
-    memset (ready_plaintext, 0, crypto_box_ZEROBYTES);
-    uint8_t *ptr = ready_plaintext + crypto_box_ZEROBYTES;
+    std::fill (ready_plaintext.begin (),
+               ready_plaintext.begin () + crypto_box_ZEROBYTES, 0);
+    uint8_t *ptr = &ready_plaintext[crypto_box_ZEROBYTES];
 
     ptr += add_basic_properties (ptr, metadata_length);
-    const size_t mlen = ptr - ready_plaintext;
+    const size_t mlen = ptr - &ready_plaintext[0];
```
bluca: is pointer arithmetic OK to use with std::vectors?

sigiesec: Yes, unfortunately the syntax is somewhat awkward since std:: vector::data only exists from C++11. However, replacing memcpy by std::copy would be preferable. I planned to do this in another step.

bluca: great, thanks

bluca: We should actually use libsodium's secure memory zones for secrets: https://libsodium.gitbook.io/doc/memory_management

sigiesec: It should be possible to implement a custom allocator that uses sodium_malloc/sodium_free, that can be used with std::vector.

sigiesec: tweetnacl doesn't have those functions though, but I can implement a fallback for the allocator.

bluca: yeah tweetnacl shouldn't be used in production, so it's fine if it falls back to standard malloc

sigiesec: Opened #3576 for this. I already started implementing this in https://github.com/sigiesec/libzmq/tree/use-libsodium-based-allocator

bluca: thanks!

PR #3580, opened 2019-07-14
Use libsodium based allocator

### LINKED ISSUE #3576: CURVE plaintext should be placed in secure memory
For additional security, unencrypted plaintext should be placed in secure memory when using CURVE security.

Points to consider:
- Is this effective for the message payload? Can we ensure this is for all copies that are under control of libzmq, and enable the user to ensure this as well?
- We should care to reduce the number of secure memory allocations to a minimum since this is a very costly operation.

sigiesec:
Fixes #3576

### DIFF
--- a/src/secure_allocator.hpp
+++ b/src/secure_allocator.hpp
@@ -1,0 +1,103 @@
+/*
+    Copyright (c) 2007-2019 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef __ZMQ_SECURE_ALLOCATOR_HPP_INCLUDED__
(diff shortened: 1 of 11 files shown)


bluca:
Nice stuff! Unfortunately there's a couple of build failures, any idea?

```
In file included from ../tests/test_security_curve.cpp:52:0:
../tests/../src/curve_client_tools.hpp:44:2: error: #error "CURVE library not built properly"
 #error "CURVE library not built properly"
```

sigiesec:
Mh, I was struggling to fix the Windows builds but probably broke the other
builds by that.

Will give that another look tomorrow.

Luca Boccassi <notifications@github.com> schrieb am So., 14. Juli 2019,
19:00:

sigiesec:
Strange, I cannot reproduce this problem. This only seems to happen with the builds using a very old gcc...

bluca:
Clang seems to have the same problem: https://travis-ci.org/zeromq/libzmq/jobs/558584750
Include order mixup?

sigiesec:
I think it's fine now :)

tests/test_security_curve.cpp
```diff
@@ -27,6 +27,21 @@
...
+#include "platform.hpp"
+#else
+#include "../src/platform.hpp"
+#endif
+
+#ifndef ZMQ_USE_TWEETNACL
+#define ZMQ_USE_TWEETNACL
+#endif
+#ifdef ZMQ_USE_LIBSODIUM
+#undef ZMQ_USE_LIBSODIUM
```
bluca: If I follow correctly, this means the unit tests are effectively not using the new memory allocators, right?

bluca:
Nice!

tests/test_security_curve.cpp
sigiesec: Well they are on the side of the tested libzmq functions. Just not for the mock curve client messages

bluca: Ah yes of course - thanks!

src/secure_allocator.hpp
```diff
@@ -0,0 +1,103 @@
...
+    {
+        typedef secure_allocator_t<U> other;
+    };
+
+    void construct (pointer p, const_reference val)
+    {
+        new ((void *) p) value_type (val);
+    }
+    void destroy (pointer p) { p->~value_type (); }
+    size_type max_size () const { return SIZE_MAX; }
```
bluca: @sigiesec SIZE_MAX is not available on some old libcs (centos 6, ubuntu 12):

https://build.opensuse.org/package/show/network:messaging:zeromq:git-stable/libzmq

```
[  194s] tests/../src/secure_allocator.hpp: In member function 'zmq::secure_allocator_t<T>::size_type zmq::secure_allocator_t<T>::max_size() const':
[  194s] tests/../src/secure_allocator.hpp:82:42: error: 'SIZE_MAX' was not declared in this scope
```

Could you please add a compat ifnedf in src/stdint?

bluca: Fixed by https://github.com/zeromq/libzmq/pull/3584

PR #3584, opened 2019-07-16
Problems: build broken on centos6/ubuntu12, missing licenses and wrong copyright dates

bluca:
Solutions: fix them

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 0MQ version 4.3.x stable, released on 20xx/xx/xx
 ================================================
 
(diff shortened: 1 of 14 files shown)


bluca:
@somdoron where did the sha1 files come from? We should add a version.txt file with a reference

somdoron:
czmq...

bluca:
lol! I'll see if I can find the origin from there

somdoron:
We can replace with any implementation... @sigiesec suggested one

PR #3591, opened 2019-07-23
Problem: incorrect ZMQ_MAX_SOCKETS default value in doc

jean-airoldie:
Solution: update the doc with the correct default value

The doc incorrectly reports this constant to be `1024`.
https://github.com/zeromq/libzmq/blob/a84ffa12b2eb3569ced199660bac5ad128bff1f0/include/zmq.h#L228


HsuanTingLu:
It seems that the [API web doc](http://api.zeromq.org/master:zmq-ctx-set) is still outdated,

Screen Shot

where should I send the PR?

bluca:
There's no PR, someone needs to regenerate, at some point it will happen

PR #3598, opened 2019-07-26
Problem: PR template is mostly about contributing guidelines

ssbl:
Solution: rename it to CONTRIBUTING.md so that it is shown when
creating a pull request or issue


bluca:
Hi - there's already a template for issues, will this change it?

ssbl:
AFAIK issue templates are a separate thing, so I don't think this change will affect them.

https://help.github.com/en/articles/about-issue-and-pull-request-templates

Pull and issue templates use separate files according to this page.

PR #3608, opened 2019-08-06
[DragonFlyBSD] fix for IPv4 mapping not supported

hxw:
On DragonFly the zmq::enable_ipv4_mapping in ip.cpp fais with a not supported error
So handle this case the same as for OpenBSD

Also update configure files to set the correct definitions for DragonFly OS

This includes all patches from the dports/net/libzmq/dragonfly patches directory

### DIFF
--- a/builds/cmake/platform.hpp.in
+++ b/builds/cmake/platform.hpp.in
@@ -75,8 +75,13 @@
   #define ZMQ_HAVE_MINGW32
 #endif
 
-#if defined(__FreeBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__)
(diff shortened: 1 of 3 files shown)


sigiesec:
Thanks a lot for the fix!

Please consider also adding a known user entry for DragonFly to https://github.com/zeromq/libzmq/blob/master/README.md

Thanks!

bluca:
@hxw could you please send another PR with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE

PR #3613, opened 2019-08-10
no void pointer to store function pointer

TomzBench:
I test with VS16 and MinGW64


bluca:
Thanks for the pr! Could you please add a commit with a relicense statement? As explained in https://github.com/zeromq/libzmq/blob/master/.github/CONTRIBUTING.md and https://github.com/zeromq/libzmq/tree/master/RELICENSE

TomzBench:
Sure,

But I just tested on my build that was failing. It still fails. However it passes when I change line as 

`EXCEPTION_DISPOSITION` **NTAPI** (*Handler) (EXCEPTION_RECORD*, void*, CONTEXT*, void*);

But then this fails on Visual Studio... So I'm not sure entirely what is best to fix yet.

bluca:
Sorry, no idea - @sigiesec @somdoron any suggestion?

sigiesec:
I will give it a look next week, AFK right now!

sigiesec:
I fixed this according to the example at https://docs.microsoft.com/en-us/cpp/cpp/stdcall?view=vs-2019#example, hope this works on MinGW as well.

bluca:
Thanks!

PR #3631, opened 2019-08-13
WIP: memory pool for ZMQ messages to entirely avoid malloc/free

f18m:
This PR is a work in progress and is a simple modification to show the effects of avoiding malloc/free when creating zmq_msg_t with the standard benchmark utils of ZMQ.

In particular the very fast, zero-lock, single-producer/single-consumer queue from:
 https://github.com/cameron314/readerwriterqueue
is used to maintain between the "remote_thr" main thread and its ZMQ background IO thread a list of free buffers that can be used.

Using the same HW for benchmark the 2 sets of CSV files generated are:

no memory pool:
```
# message_size,message_count,PPS[msg/s],throughput[Mb/s]
8,100000000,3237023,207.169
16,100000000,3537600,452.813
32,100000000,3207026,820.999
64,100000000,2369104,1212.982
128,100000000,1519774,1556.249
210,100000000,1389526,2334.404
```

with memory pool:
```
# message_size,message_count,PPS[msg/s],throughput[Mb/s]
8,100000000,3151628,201.704
16,100000000,3345384,428.209
32,100000000,3159686,808.880
64,100000000,2742139,1403.975
128,100000000,2448842,2507.615
210,100000000,2353077,3953.171
```

the memory pooled approach shows a

mostly the same performances for messages <= 32B
 +15% pps/throughput increase @ 64B, 
 +60% pps/throughput increase @ 128B, 
 +70% pps/throughput increase @ 210B

### DIFF
--- a/RELICENSE/skaller.md
+++ b/RELICENSE/skaller.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by John Skallwe
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any
+other Open Source Initiative approved license chosen by the current ZeroMQ
+BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle “skaller", with
+commit author “skaller", are copyright of John Skaller .
+This document hereby grants the libzmq project team to relicense libzmq,
+including all past, present and future contributions of the author listed above.
+
+John Skaller
+2019/09/01


somdoron:
I think we might need mutli-consumer-multi-producer pool, don't you think?
As both the background zeromq thread and multiple user threads are going to use it.

f18m:
> I think we might need mutli-consumer-multi-producer pool, don't you think?
> As both the background zeromq thread and multiple user threads are going to use it.

Well, I know that in practice applications have multiple user threads and multiple ZMQ background threads.
In such a scenario I guess that we might either:
 **a)** ask to the user to create a different memory pool for each of its user thread doing zmq_msg_send();
    this reduces lock-contention (if there's any lock) and help the user to keep track of how many "active"  messages (=msgs available for use inside the mempool) are available on a per-thread basis;

 or 

 **b)** allow the user to use a single global memory pool for packets;
    this approach will generally result in better memory utilization (specially if the load is not well balanced among user threads and some of them are doing a lot of zmq_msg_send() while others are doing very little traffic).

Scenario a) would be absolutely acceptable in my opinion however scenario b) is probably safer to use.

I think scenario a) allows the use of a single-producer/single-consumer (SPSC) queue since the same memory pool will have its "allocate_msg()" always called from the same user app thread and just 1 ZMQ background thread AFAIK.

I think scenario b) instead requires a SPMC queue since the free packets may be consumed from multiple user app threads but they will still be released by a just 1 ZMQ background thread AFAIK.

Am I missing something up to this point?


Scenario b) also lends itself to have some sort of API in ZMQ like:
 - zmq_msg_pool_init(zmq_msg_pool_t*)
 - zmq_msg_pool_allocate(zmq_msg_pool_t*, zmq_msg_t*): 
    this would initialize a zms_msg_t pointing it to the right deallocator of the memory pool, transparently to the user

What do you think about such an API?

sigiesec:
This is a really great extension!

I think it would be great to allow per-user-thread message pools like you suggest in scenario a). Since a socket cannot be migrated between user threads by the specification, this should be usable for most scenarios. However, you are right that this might not lead to an optimal memory usage, if memory usage is distributed unevenly across threads over time. So having the ability to use a global memory pool in addition would be good as well. Maybe the API can be designed in a way that the user can supply an own memory pool? (This might also better integrate with an application-wide memory allocation strategy). libzmq could then provide one or two implementations of a memory pool, but allow the user to extend that in a custom way. The user would need to supply a callback function (or two separate callback functions for allocating and deallocating). This could be associated with the context, and it could be required that this can only be set before any socket is created.

somdoron:
Forwarding msg between threads over one c breaks the single-consumer/producer

f18m:
Hi @somdoron , Hi @sigiesec ,

thanks for the feedback.

> Forwarding msg between threads over one inproc breaks the single-consumer/producer

Well I wonder: in such a case it will be the first user app thread that does pull out a ZMQ message from
the pool and it will be another user app thread doing the zmq_msg_close(), isn't it?
Is any of the ZMQ background threads going to call zmq_msg_close() ?
If not then the single-consumer/producer (SPSC) assumption still holds...


> Maybe the API can be designed in a way that the user can supply an own memory pool?

On the mailing list there were some further discussions about having an "application-wide memory allocation strategy". 
In particular, taking inspiration from the Jens Auer proposal, I would suggest an API like the following one:

In zmq.h:

```
typedef struct zmq_allocator_t {
    void* hint_;
    void* (&allocate_fn)(size_t nbytes, void* hint_);
    void (&release_fn)(void* ptr, void* hint_); 
} zmq_allocator_t;

#define ZMQ_MSG_ALLOCATOR 10


// additional API to access the ZMQ-provided message-pool implementations:

#define ZMQ_MSG_ALLOCATOR_DEFAULT 0               // using malloc/free
#define ZMQ_MSG_ALLOCATOR_PER_THREAD_POOL 1       // using internally a SPSC queue (cannot be used with inproc maybe?) or perhaps an MPMC queue anyway
#define ZMQ_MSG_ALLOCATOR_GLOBAL_POOL 2           // using internally a MPMC queue

void zmq_msg_allocator_init(zmq_allocator_t* alloc_, int type_);   // initializes the given allocator struct with a ZMQ builtin allocator

```

which can then be used as:

```
void main()
{
  ...

  zmq_allocator internalZmqProvidedAllocator;
  zmq_msg_allocator_init(&internalZmqProvidedAllocator, ZMQ_MSG_ALLOCATOR_PER_THREAD_POOL);
  zmq_ctx_set(myCtx, ZMQ_MSG_ALLOCATOR, &internalZmqProvidedAllocator);
  
  ...
  
  zmq_msg_t myMsg;
  zmq_msg_init(&myMsg);   // this message is automatically allocated from ZMQ memory pool instead of regular malloc
}
```

or with a user-defined allocator like:

```
MyOwnMemoryPool g_pool;

void* allocatePool(size_t n, void* pool) {return static_cast<MyOwnMemoryPool*>(g_pool)->allocate_fn(n);}
void releasePool(void* ptr, void* pool) {static_cast<MyOwnMemoryPool*>(g_pool)->release_fn(ptr);}

void main()
{
  ...

  zmq_allocator myPooledAllocator{&pool, allocatePool, releasePool};
  zmq_ctx_set(myCtx, ZMQ_MSG_ALLOCATOR, &myPooledAllocator);
  
  ...
  
  zmq_msg_t myMsg;
  zmq_msg_init(&myMsg);   // this message is automatically allocated from MyOwnMemoryPool
}
```

What do you think?

f18m:
Another question: do you see any issue integrating the projects

https://github.com/cameron314/concurrentqueue
https://github.com/cameron314/readerwriterqueue

inside ZMQ? They are both licensed under "Simplified BSD License" so I think they should be fine with GPL, right?

sigiesec:
Sorry I didn't follow the discussion on the mailing list.

The approach looks good to me basically. However, I would suggest not to define zmq_allocator in zmq.h, but only expose this as an opaque void* like other zmq types. This means that a second initialization/creation function is needed for the user-defined case, and a destruction function as well (cf. the zmq_poller_* API, e.g.). Exposing the data structure in zmq.h makes later changes virtually impossible, and I don't think that is necessary in this case.
Also, as a minor point, the name of the type should be zmq_allocator_t in line with zmq's naming conventions.

sigiesec:
@bluca Can you comment on the licensing question? I am not an expert on that.

bluca:
Yes bsd-3 is compatible - we already have queues implemented here, any reason am external implementation is needed? Every external dependency adds complexity and work for maintenance

f18m:
Hi @bluca ,

> we already have queues implemented here, any reason am external implementation is needed? 

Yeah well I was proposing to integrate those queues because I have used them already and I know how they work. What would be the equivalent in the libzmq world?
I see that there is a yqueue_t object but I see this comment at the top of the file:

```
//  yqueue allows one thread to use push/back function and another one
//  to use pop/front functions. 
```

So my question is: is that a multiple-producer/multiple-consumer queue? or multiple producers/enqueuers calling push() asynchronously will corrupt the queue data structures?
Also the comment says:

```
// However, user must ensure that [...] both threads don't access the same
//  element in unsynchronised manner.
```

This comment means that the object T itself is not magically transformed to a thread-safe object by the queue right (which is very clear and expected from my point of view)? Or it means something else? :)

Thanks

bluca:
they are single producer/consumer - but so are sockets, why is it necessary to have multi producer/consumer?

f18m:
> they are single producer/consumer - but so are sockets, why is it necessary to have multi producer/consumer?

in case we want to provide a global memory pool which could be accessible by all app user threads (consumers) and any ZMQ background thread may return them (as a producer)... 

the single producer/consumer case is useful in case we do a per-user-thread memory pools...

bluca:
I think it makes sense to keep a per socket memory pool. That will have useful side effect like data locality - a common pattern is to have a fast pub-sub and a slow req-rep/router-dealer for management/sync, that way they don't mess with each others caches, pages and TLB entries.

f18m:
Hi @bluca ,

> I think it makes sense to keep a per socket memory pool. That will have useful side effect like data locality 

Well that's yet another possibility... however that means specializing the "allocator pattern" for zmq_msg_t I guess.
I mean: currently (see my comment above https://github.com/zeromq/libzmq/pull/3631#issuecomment-522776749 ) the API proposed so far is not tied in any specific way to zmq_msg_t and, even if as first step I would change only the zmq_msg_t malloc/free pairs, we could in theory change any call to malloc/free in ZMQ code with that approach.

I'm not sure how the API may look like if we want to have a per-socket memory pool... 

>a common pattern is to have a fast pub-sub and a slow req-rep/router-dealer for management/sync, that way they don't mess with each others caches, pages and TLB entries.

From a performance point of view if the req-rep is slow its impact on the cache locality of the fast pub-sub is probably small any way, isn't it?

All in all, I think having per-thread or global memory pool is a simpler approach and is what e.g. DPDK does (well DPDK only provides global memory pools).

f18m:
@bluca , @somdoron , @sigiesec ,

to simplify this feature (memory pooling) and to get an higher chance of reaching the end of it (unfortunately I don't have much free time recently) I would like to start by a very basic pull request first: extending the API of 

`
int zmq_ctx_set (void *context, int option_name, int option_value);
`

that will not allow to pass a structure complex like the zmq_allocator_t, into something like:

`
int zmq_ctx_set(void *context, int option_name, const void *option_value, size_t option_len);
`

is this change ok in your opinion?
Should we just replace the old API breaking compatibility or rather just add a new method instead?

sigiesec:
If you add that, a new method must be added, we must not change existing API (except DRAFT).

f18m:
Hi @bluca ,
thanks for merging #3642 . I will start working on the approach sketched out in 
   https://github.com/zeromq/libzmq/pull/3631#issuecomment-522776749

(I will submit a new MR perhaps so I start working from a clean branch)

f18m:
Hi all,
I just realized that this API model:

> ```
> void main()
> {
>   ...
> 
>   zmq_allocator internalZmqProvidedAllocator;
>   zmq_msg_allocator_init(&internalZmqProvidedAllocator, ZMQ_MSG_ALLOCATOR_PER_THREAD_POOL);
>   zmq_ctx_set(myCtx, ZMQ_MSG_ALLOCATOR, &internalZmqProvidedAllocator);
>   
>   ...
>   
>   zmq_msg_t myMsg;
>   zmq_msg_init(&myMsg);   // this message is automatically allocated from ZMQ memory pool instead of regular malloc
> }
> ```

has a problem:

- the zmq_msg_init() has no pointer to the ZMQ context and indeed the documentation states (something I discovered only now) that in theory multiple ZMQ contexts may exist inside the same application... so how do I get from inside the zmq::msg_t class a pointer to the ZMQ context?

If this is going to be a trouble, I suggest to go back to a slightly different API model (taking from several suggestions in previous comments / emails):

```
// myalloc is of course a long-lived object, which must be destroyed after all the messages it has provided have been consumed:
void* myalloc = zmq_msg_allocator_new(ZMQ_MSG_ALLOCATOR_PER_THREAD_POOL);

zmq_msg_t myMsg;
zmq_msg_init_allocator(&myMsg, myalloc);  
```

What do you think?

somdoron:
Sorry for being late, I still think the single-consumer/single-producer is very problematic and it is actually breaking zeromq.

One of the most common usages of zeromq is having a proxy that distributes messages to multiple workers over inproc. That will be broken, with no easy way to solve that, as when using DEALER proxy cannot know which socket/thread is going to accept the msg, in that use-case you have to go back to heap and do a copy, losing all the benefit of the memory pool (+ the copy).

Moreover, existing users of zeromq that will think that they can just use the built-in memory pool will find out that their system is now broken, this is very dangerous.

I think we should either not have built-in memory pool and only allow for a user to provide its own, or have a safe one. 

We can still have one memory pool for each socket/thread, however, the producer (closing) side must allow multiple access.

Issue #2862, opened 2017-12-13
Problem: UDP engine aborts on networking-related errors from socket syscalls


nayabbashasayed:
# Description
I have a dish socket bind to a multicast address. And dish socket receiving messages in a while loop.
It works fine when there is an Ethernet interface is up.
I did `sudo ifconfig enp7s0 0.0.0.0` and observed the following:
```
No such device (src/udp_engine.cpp:142)
Aborted (core dumped)
```
I am using this dish socket in an application in which the ip address of the interface often becomes `0.0.0.0`.
Is there any way I could exit the while loop safely without core dumping the entire application?

# Environment

* Version: Zeromq 4.2.1
* OS: Ubuntu 16.04

bluca:
I guess the UDP engine needs the same hardening the TCP one got a few months ago, against network errors. That's a setsockopt failing. PRs welcome.

nayabbashasayed:
I am not much into C++. :(
And I see that core dump is happening even for the interface which has static IP and but not connected to the network.

Any condition I could check and break the loop?

nayabbashasayed:
Hi,

I had to add
```
route add -net 224.0.0.0 netmask 240.0.0.0 brtrunk
```
so that radio socket finds a way for multicast traffic. Now core dump is not happening. :)

Thanks..

bluca:
Great, happy you found a workaround and thanks for sharing it.

I'll reopen and retitle the issue, as the UDP implementation should be hardened anyway before it can be declared stable.

nayabbashasayed:
I am able to reproduce the crash with v4.2.3 also. But the scenario is different.
This time GDB prints:
```
(gdb) bt
#0  0xb6bf8424 in __GI_raise (sig=sig@entry=6) at libpthread/nptl/sysdeps/unix/sysv/linux/raise.c:67
#1  0xb6bf27f0 in __GI_abort () at libc/stdlib/abort.c:89
#2  0xb6ed0e14 in zmq::zmq_abort (errmsg_=errmsg_@entry=0xb6c07210 <mylock> "") at src/err.cpp:87
#3  0xb6f07744 in zmq::udp_engine_t::out_event (this=<optimized out>) at src/udp_engine.cpp:285
#4  0xb6f06ca4 in zmq::udp_engine_t::restart_output (this=0x2061d0) at src/udp_engine.cpp:307
[...]

PR #3638, opened 2019-08-22
UDP engine aborts on networking-related errors from socket syscalls #2862

atomashpolskiy:
This refactoring is supposed to prevent ZeroMQ from terminating upon UDP engine tuning errors (in `zmq::udp_engine_t::plug`). As far as I was able to figure out, the session will terminate silently on `protocol_error`, which isn't good, probably. Not really sure, where to go from here...

### DIFF
--- a/src/tcp.cpp
+++ b/src/tcp.cpp
@@ -62,7 +62,7 @@ int zmq::tune_tcp_socket (fd_t s_)
     int nodelay = 1;
     int rc = setsockopt (s_, IPPROTO_TCP, TCP_NODELAY,
                          reinterpret_cast<char *> (&nodelay), sizeof (int));
-    tcp_assert_tuning_error (s_, rc);
+    assert_socket_tuning_error (s_, rc);
     if (rc != 0)
         return rc;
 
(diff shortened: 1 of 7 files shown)


src/tcp.cpp
```diff
@@ -318,46 +318,7 @@ int zmq::tcp_read (fd_t s_, void *data_, size_t size_)
 
 void zmq::tcp_assert_tuning_error (zmq::fd_t s_, int rc_)
```
bluca: why keeping this function, if the usage changes to the new one from ip.cpp?

bluca:
please add a relicense statement, and fix the formatting error that the CI is showing

src/tcp.cpp
atomashpolskiy: It's public, so I was not sure, if this function might have been used somewhere outside libzqm. Removed it completely.

sigiesec: Only the API in zmq.h is exposed to outside of libzmq

atomashpolskiy:
> please add a relicense statement, and fix the formatting error that the CI is showing

I had fixed the formatting of udp_engine.cpp in fe2287d (by `clang-format -i -style=file src/udp_engine.cpp`), so I'm not sure, what's up with Travis... Let's wait for the next build to finish and see, if the issue pertains

bluca:
```make clang-format-diff``` will run it with the template we checked in

PR #3640, opened 2019-08-24
UDP engine aborts on networking-related errors from socket syscalls (2) #2862

atomashpolskiy:
Follow-up for UDP send/receive. Also updated `plug` function to terminate session only when `setsockopt` failed, and request reconnection otherwise. Still haven't looked into how socket monitoring events work, so error handling happens silently.

Note that I added EAGAIN/EWOULDBLOCK and EPIPE in the list of "recoverable" errors.

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -234,20 +234,10 @@ int zmq::bind_to_device (fd_t s_, const std::string &bound_device_)
 #ifdef ZMQ_HAVE_SO_BINDTODEVICE
     int rc = setsockopt (s_, SOL_SOCKET, SO_BINDTODEVICE,
                          bound_device_.c_str (), bound_device_.length ());
-
-#ifdef ZMQ_HAVE_WINDOWS
-    if (rc != SOCKET_ERROR)
-        return 0;
-    const int lastError = WSAGetLastError ();
-    errno = wsa_error_to_errno (lastError);
-    wsa_assert (lastError != WSAENOTSOCK);
-    return -1;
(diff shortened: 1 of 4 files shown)


src/ip.cpp
```diff
@@ -717,11 +716,12 @@ void zmq::assert_socket_tuning_error (zmq::fd_t s_, int rc_)
...
-                      || errno == ECONNABORTED || errno == EINTR
-                      || errno == ETIMEDOUT || errno == EHOSTUNREACH
-                      || errno == ENETUNREACH || errno == ENETDOWN
-                      || errno == ENETRESET || errno == EINVAL);
+        errno_assert (
+          errno == ECONNREFUSED || errno == ECONNRESET || errno == ECONNABORTED
+          || errno == EINTR || errno == ETIMEDOUT || errno == EHOSTUNREACH
+          || errno == ENETUNREACH || errno == ENETDOWN || errno == ENETRESET
+          || errno == EINVAL || errno == EADDRINUSE || errno == EACCES
+          || errno == EWOULDBLOCK || errno == EAGAIN || errno == EPIPE);
```
bluca: Please do not add new errors mixed in with refactoring. It needs to be carefully weighted and tested to avoid introducing deadlocks.

atomashpolskiy: Ok, sure! So did unittest_poller keep running because I made some errors non-critical?

bluca: Not sure - should be investigated

PR #3642, opened 2019-08-25
Introduce extended set/get methods for ZMQ contexts

f18m:
Problem: today we can only set/get numeric options on the ZMQ contexts. This is no good for an existing option (ZMQ_THREAD_NAME_PREFIX) which naturally takes on string values. Moreover it does not allow future developments e.g. the ones discussed in #3631 where a complex data structure should be set on the ZMQ context.

Solution: Introduce DRAFT zmq_ctx_set_ext() and zmq_ctx_get_ext() methods. Change ZMQ_THREAD_NAME_PREFIX to allow for non-numeric thread name prefixes.

NOTE: this PR is missing docs. If it looks fine I can write them.

### DIFF
--- a/doc/zmq_ctx_get_ext.txt
+++ b/doc/zmq_ctx_get_ext.txt
@@ -1,0 +1,82 @@
+zmq_ctx_get_ext(3)
+==================
+
+
+NAME
+----
+
+zmq_ctx_get_ext - get extended context options
+
+
+SYNOPSIS
+--------
+*int zmq_ctx_get_ext (void '*context', int 'option_name', void '*option_value', size_t '*option_len');*
+
+
+DESCRIPTION
+-----------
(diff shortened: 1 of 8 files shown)


sigiesec:
LGTM overall, thanks! Docs can be added later on (but will be required before going out of DRAFT). 

You need to add the new DRAFT functions also to zmq_draft.h (e.g. here: https://travis-ci.org/zeromq/libzmq/jobs/576574156#L1193)

f18m:
Hi @sigiesec ,
I think I fixed all build errors (except for "BUILD_TYPE=valgrind CURVE=tweetnacl DRAFT=enabled" which is failing also on master though)... I also added docs for the new functions.
Let me know if this PR needs further work or not, thanks.

PR #3647, opened 2019-08-31
RELICENSE: add relicense agreement from msune

msune:
Add RELICENSE document for msune contributions, both for:

* marcdevel at gmail dot com
* marc at voltanet io


somdoron:
Thanks

Issue #3646, opened 2019-08-30
Make check failed(undefined reference) with configure --enable-static --disable-shared


forif:
# Issue description
after configure with --enable-static and --disable-shared option, make command succeed but 
make check compilation failed with bunch of "undefined reference" in tests/testutil.cpp

# Environment

* libzmq version (commit hash if unreleased): 4.3.2
* OS: Ubuntu-16.04
* gcc/g++ version: (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609

# Minimal test code / Steps to reproduce the issue
```bash
./configure --enable-static --disable-shared
make
make check
```

# What's the actual result? (include assertion message & call stack if applicable)
```bash
Making check in doc
make[1]: Entering directory '/home/nemo/workspace/git_codes/zeromq-4.3.2/doc'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/home/nemo/workspace/git_codes/zeromq-4.3.2/doc'
make[1]: Entering directory '/home/nemo/workspace/git_codes/zeromq-4.3.2'
[...]

PR #3661, opened 2019-09-02
Problem: static-only test run fails

bluca:
Solution: add libtestutil.a before libzmq.a in LDADD listings, as the
linker will complain about undefined symbols otherwise

Fixes https://github.com/zeromq/libzmq/issues/3646


sigiesec:
Thanks for the fix! Sorry my prior changes were flawed, unfortunately we don't have CI job that catches this. In principle, I am aware of that the library order matters, still I don't really understand why this happens only when doing a static-only build.

bluca:
No worries - I'm not sure either

Issue #3658, opened 2019-09-02
Idea: Configurable event forwarding through zmq_msg


jean-airoldie:
Any opposition to adding ZMQ_EVENT_SUBSCRIBED & ZMQ_EVENT_UNSUBSCRIBED to the `zmq_socket_monitor_versioned` 2? These events would provide the specific topic of that was subscribed / unsubscribed to. This events would be generated by `PUB/SUB` and `RADIO/DISH` sockets.

As a side effect, this would effectively remove for the need for a `XRADIO/XDISH` pattern.

sigiesec:
I am not sure. 

This is an alternative design providing the same functionality as an XPUB socket (and a new XRADIO socket), which provides the events via the socket. It does not, however replace the XSUB (and a new XDISH socket), as these process subscription messages sent via the socket, which is somewhat symmetrical to the XPUB socket behaviour. But maybe there are different use cases for which monitoring events are more appropriate (e.g. when the only action done is to forward the events to an overall application logging or monitoring service). Is this what you have in mind?

jean-airoldie:
I had two things in mind:
* Triggering an application level function when a new subscription is received (like a welcome message but better)
* Proxying a subscription at the application level with RADIO/DISH. e.g. Radio receives a subscription, proxies the subscription to another Dish socket, etc.

More generally, I am finding the stateless nature of zmq sockets to be problematic. I want to allow sockets send/recv operations to be interrupted by events chosen by the user. Essentially, every socket would be paired with a monitor socket so that, for instance a `send` event can be interrupted by a disconnection event for instance. This is really usefull when you need application level code that is linked to socket state (even thought zmq wants to hide its internal state).

jean-airoldie:
Conceptually, all the problems I have with zmq currently can be solved by exposing so state via the `zmq_socket_monitor` API. However I understand that this is opposite to the original ideology of the lib.

somdoron:
I would recommend taking a look at the malamute project source code, the
state issue is solved nicely without any monitoring.

jean-airoldie:
@somdoron Could you implement a application session linked to the lifetime of a connection using malamute? For instance I need a deadman switch that triggers a soon as a client socket disconnect (even if it later reconnects). How can this be achieved without monitoring?

somdoron:
Applicative ping-pong, this is what malamute is doing. Also if I remember correctly, router has a notification through msg.

We can do the same for SERVER and maybe even for RADIO/DISH.

[...]

PR #3665, opened 2019-09-04
[WIP, RFC] Problem: Relevant events on sockets are discarded

jean-airoldie:
Solution: Introduce notification message using `msg_t` to forward these relevant events to the user.

See #3658.

# RFC
The following is a informal rfc so that I can better understand what is involved in such changes.

# Definitions used in this doc
* `server` = a socket which acts as a bound endpoint
* `client` = a socket which connects to a `server`

# Frontend changes

## Add `zmq_connect_peer` function.
This `zmq_connect_peer` function is functionally the same as `zmq_connect`, except
it returns a `routing_id` used to refer to the connection with the peer. This `routing_id` will be
used by notifications to refer to this peer. A `routing_id` of `0` means an
error occured.

## Add `ZMQ_FILTER_NOTIF` sockopt
The `ZMQ_FILTER_NOTIF` sockopt is used to configure what event notifications
should be received by the user. This socket takes a `uint64_t` bitmask,
which defaults to 0 (no notifications). See `NotifType` for
the list of possible events.

## Add `zmq_msg_notif` function.
This `zmq_msg_notif` function can be used to retreive the `NotifType` of
a message. A returned value of `-1` means that the message is not a notification.

## NotifType
These are the possible notifications events. Note that each of these is a bitflag.
* `ZMQ_NOTIF_CONNECTED` means a connection of a `client` was established with
    a remote `server`. The `routing_id` of the message is used to refer to
    the specific `server`.
* `ZMQ_NOTIF_DISCONNECTED` means a connection of a `client` with a remote
    `server` was closed. The `routing_id` of the message is used to refer to
    the specific `server`.
* `ZMQ_NOTIF_CONNECTION_ACCEPTED` means a `server` accepted a connection from a
    remote `client`. The `routing_id` of the message is used to refer to the specific `client`.
* `ZMQ_NOTIF_CONNECTION_CLOSED` means a connection accepted by the `server` from
    a remote `client` was closed. The `routing_id` of the message is used to
    refer to the specific `client`.
* `ZMQ_NOTIF_JOINED` means a remote `DISH` connected to a `RADIO` joined a `group`.
    The `group` of the message is used to refer the specific grouped that was joined.
* `ZMQ_NOTIF_LEAVED` means a remote `DISH` connected to a `RADIO` leaved a `group`.
    The `group` of the message is used to refer the specific grouped that was leaved.
* `ZMQ_NOTIF_HANDSHAKE_FAILED` means a ZAP handshake of a `client` with a
    remote `server` failed. The `routing_id` of the message is used to refer
    to the specific `server`. The `group` of the message is a `HandshakeFailureDetails`.
    See `HandshakeFailureDetails` for details.

## HandshakeFailureDetails
A handshake failure details is used to give details on the type of handshake failure.

```c
enum HandshakeFailureType {
    no_details = 0,
    protocol = 1,
    auth = 2,
};

struct HandshakeFailureDetails {
    HanshakeFailureType type;
    // if HandshakeFailureType::no_details, then code is an errno.
    // if HandshakeFailureType::protocol, then code is a ZMQ_PROTOCOL_ERROR_*
    // if HandshakeFailureType::auth, then code is a ZAP status code (i.e. 300, 400, 500).
    int code;
}
```
# Implementation

## zmq_connect_peer
???. Basically a socket would need to assigne a `routing_id` to each connection, but I don't know how that would be implemented.

## zmq_msg_notif
Check if the msg_type is `type_notif` then return the `NotifType` stored in the union. Else return -1 with EINVAL.

## ZMQ_NOTIF_FILTER sockopt
Add a notif_filter field to the socket options that will be used as a mask, similarly to how `zmq_socket_monitor` is implemented.

## notification msg format
Create a new msg_type::type_notif as well as a new message union. The message union has to contain a `NotifType`, a `group` and a `routing_id`.

## notification queue
??? Unsure how that would be implemented.

## notification mechanism
This would be very similar to how zmq_monitor is currently implemented. In each location where a relevant event occurs:
* Check the `notif_filter` bitflag so see if the notification is activated
* If it is, create a notification msg_t and set the appropriate field depending on the `NotifType`
* Queue the message to the notification queue.

## notification retrieval mechanism
When calling `recv` on a socket, check the notification queue first. If it contains a notification,
return it, otherwise operate as normal.

## Unresolved questions
* Should there be a notification for when a `server` denies a connection (handshake failure)?
* How should zmq_connect_peer be implemented?
* How should the notification queue be implemented?
* Should this mechanism be limited to DRAFT socket types?

### DIFF
--- a/src/msg.hpp
+++ b/src/msg.hpp
@@ -202,7 +202,10 @@ class msg_t
         //  Leave message for radio_dish
         type_leave = 107,
 
-        type_max = 107
+        // Notification message.
+        type_notif = 108,
+
+        type_max = 108
     };
 
     //  Note that fields shared between different message types are not


somdoron:
My idea is that all notification will be generated at the user thread, at least initially, that will limit the number of events that can be generated.

For the first phase I would try to support SERVER only with ATTACHED and TERMINATED events.
At the user thread we can distinguish between connect and accepted, and I would avoid that for now.

So ATTACHED/TERMINATED will replace the CONNECTED,DISCONNECTED,ACCEPTED and CLOSED.

At the moment we cannot support the HANDSHAKE_FAILED at the user thread side as well.

somdoron:
Let's continue by finishing the Msg changes.
Those would be:
* Add another union case to the msg union: https://github.com/zeromq/libzmq/blob/master/src/msg.hpp#L212, for the notif type we need the base fields plus the notif_type. (it should be after metadata, and the following unused should set correctly).
* Add init_notif function to the class, which accept the notification type and initialize the msg.
* Add the notif method (msg class) and zmq_msg_notif (zmq.h & zmq.cpp, should be a draft).

jean-airoldie:
What's the advantage in generating the events in the user thread as opposed to the event loop thread like how the socket_monitor API is implemented?
https://github.com/zeromq/libzmq/blob/69e451c1faa97c08ae7ac34bb4f948f7dfbc3472/src/stream_engine_base.cpp#L584

somdoron:
> What's the advantage in generating the events in the user thread as opposed to the event loop thread like how the socket_monitor API is implemented?

The monitoring socket is a hack (I wish I could remove it), it is not thread-safe and if you would have more than one background thread it would fail, moreover, some of the events are generated at the user thread and some at the io thread (or at least it used to be like that), which is a violation of socket usage.

If we would enqueue the events into the msg pipe it would count for the watermark, and we don't want that.

The correct way to do it is to generate all the events at the user thread, in order to support events that we currently cannot generate at the user thread we should use commands.
Howevet, that is a little more advanced and we use it eventually.

Come for the zeromq hackathon, we could work on the commands for delivring other types of events.

somdoron:
Regarding generating the events for SERVER socket:
1. We first need a queue for the events, so add std::queue notifications.
2. On xattachpipe and terminated insert notifications to the notifications queue.
3. On xrecv, before recv from a fq.recv first try to dequeue a notification.

jean-airoldie:
> Come for the zeromq hackathon, we could work on the commands for delivring other types of events.

I'll consider it.

bluca:
No activity for 6 months on this. Is work going to resume soon, or should I close it for now?

jean-airoldie:
I won't work on it.

PR #3668, opened 2019-09-04
Safer regex replace, keeping options like "/Zc:twoPhase-" intact

straubar:
In our project, we include libzmq via the FetchContent mechanism and add it using add_subdirectory. In the already set CMAKE_CXX_FLAGS_DEBUG and CMAKE_CXX_FLAGS_RELWITHDEBINFO, we have multiple flags set, such as /Zc:twoPhase-. By applying the regex replace, this was converted to /Z7:twoPhase-, which caused the compilation to fail. The now proposed regex replace is a bit safer, as it does not touch flags containing a ":".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -671,8 +671,8 @@ if(MSVC)
 
   # Compile the static lib with debug information included
   # note: we assume here that the default flags contain some /Z flag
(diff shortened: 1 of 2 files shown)


bluca:
Could you please add a commit with a relicense statement? See https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

PR #3670, opened 2019-09-05
problem: ws_engine doesn't send correct host and path

somdoron:
Solution: extract path and host from the address


somdoron:
Appveyor is failing because of the xpub_no_drop test, which is not related. 
This is ready to be merged

PR #3677, opened 2019-09-10
problem: sha external file conflict with czmq

somdoron:
Solution: use external library for sha1.

Part of this commit is making websocket transport optionable

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -52,6 +52,10 @@
 #include <vmci_sockets.h>
 #endif
 
+#ifdef ZMQ_USE_NSS
+#include <nss.h>
+#endif
+
 #define ZMQ_CTX_TAG_VALUE_GOOD 0xabadcafe
 #define ZMQ_CTX_TAG_VALUE_BAD 0xdeadbeef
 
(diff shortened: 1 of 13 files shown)


somdoron:
I still need to install nss as part of the CIs

somdoron:
it seems that building nss on windows is not going to be easy, it requires MozillaBuild...
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Building

src/ctx.cpp
```diff
@@ -87,6 +91,10 @@ zmq::ctx_t::ctx_t () :
 
     //  Initialise crypto library, if needed.
     zmq::random_open ();
+
+#ifdef ZMQ_HAVE_WS
+    NSS_NoDB_Init("/tmp");
```
bluca: doc says parameter should be NULL here

bluca:
> it seems that building nss on windows is not going to be easy, it requires MozillaBuild...
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Building

It should be fine to keep the dependency optional, and fallback to the external sources on windows - given it's now optional, if people need to static link with czmq they can either figure it out or disable this protocol

src/ctx.cpp
somdoron: fixed

somdoron:
@bluca done, now you can use nss instead of built-in or disable ws transport completely

bluca:
Nice!

Issue #3323, opened 2018-12-07
Behaviour of ZMQ_CONFLATE in PUB/SUB pattern


francesco-romano:
# Issue description

**Note: this might be the intended behaviour.**

I am implementing a PUB/SUB pattern where there is one publisher (quite fast) and multiple subscribers (they have different rates).

I am interested in **only the most** recent message. So I thought to enable the `ZMQ_CONFLATE` options.

My understanding (just from the doc, I did not check the code) is that only the newest message is retained -- a queue of size 1 and with the most recent message. Something that we cannot do with the high water mark that discards the new messages.
I am hypothesising that the "user" code sends a message (`zmq_send`), the thread pool in the `zmq_context` will handle the message to be sent on the socket. If a newer message comes before the old one is sent, then (with `ZMQ_CONFLATE` = true), the old message is discarded and the new one is sent. The same on the receiver size. If the user code is slow in calling `zmq_recv`, only the last message is read.

Now, this works perfectly as long as I have 1 publisher and 1 subscriber. 
If I add more subscribers, only the first one that connects get the messages (using `tcp`) or the last one (using `ipc`).
Note that is the publisher side that control the behaviour. Adding `ZMQ_CONFLATE` in the subscriber has no effect (in the sense that the subscribers are able to connect)

So, my question is: is this the intended behaviour?
Does `ZMQ_CONFLATE` set on the publisher side discards the message after it is sent to the first subscriber (and not because a new `zmq_send` has been called)?

# Environment
I tested this on macOS 13.6, lib zmq (installed by homebrew) 4.2.5.
I also tested between two Debian (4.2.3-1 and 4.2.1-4)

The code I run, is the following:

# Code

Publisher:
```cpp
#include <zmq.h>

#include <cstdlib>
#include <chrono>
#include <iostream>
#include <thread>
#include <atomic>
#include <csignal>

static std::atomic<bool> s_shutdown(false);

void signal_handler(int) {
    s_shutdown = true;
}

int main(int argc, char** argv)
{
    signal(SIGINT, signal_handler);
    signal(SIGTERM, signal_handler);

    void* context = zmq_ctx_new();
    void* output = zmq_socket(context, ZMQ_PUB);


    // We set the internal queue to have only one (the most recent) message and throw away
    // everything else.
    int trueValue = 1;
    zmq_setsockopt(output, ZMQ_CONFLATE, &trueValue, sizeof(int));

    // We are ready. Create the endpoints.
[...]

PR #3687, opened 2019-09-23
Fix behavior of ZMQ_CONFLATE on PUB sockets

trya:
This commit actually solves two problems that originated in `dbuffer::write`:
- only one subscriber could receive messages: since the pending message was moved to the back buffer, it was closed and further writes of the same message to other conflate pipes were writes of an empty message.
- leaking messages when one subscriber is not responsive: the front buffer being not closed because of this, swapping the front buffer with the back buffer does not release its enclosed pending message. `msg_t::move` is more indicated in that case.

As such, provides a fix for #3323.

### DIFF
--- a/src/dbuffer.hpp
+++ b/src/dbuffer.hpp
@@ -78,12 +78,12 @@ template <> class dbuffer_t<msg_t>
         msg_t &xvalue = const_cast<msg_t &> (value_);
 
         zmq_assert (xvalue.check ());
-        _back->move (xvalue); // cannot just overwrite, might leak
+        *_back = value_;
 
         zmq_assert (_back->check ());
 
         if (_sync.try_lock ()) {
-            std::swap (_back, _front);
+            _front->move (*_back);
             _has_msg = true;
 
             _sync.unlock ();


src/dbuffer.hpp
```diff
@@ -78,12 +78,12 @@ template <> class dbuffer_t<msg_t>
...
 
         zmq_assert (xvalue.check ());
-        _back->move (xvalue); // cannot just overwrite, might leak
+        *_back = value_;
 
         zmq_assert (_back->check ());
 
         if (_sync.try_lock ()) {
-            std::swap (_back, _front);
+            _front->move (*_back);
```
bluca: The commit message says:

> swap front and back buffer using msg_t::move instead of std::swap

But this does not swap them anymore, instead _back is de initialized. Is that intended?

trya: >But this does not swap them anymore, instead _back is de initialized. Is that intended?
Yes, the idea is to release the message in the front buffer first before moving the message in the back buffer into it. I just changed the commit message to reflect that intent better.

bluca:
Thanks!

Issue #3581, opened 2019-07-15
WebSocket transport progress


somdoron:
- [ ]  draft ZWS 2.0 rfc
- [ ]  document the transport
- [ ]  use NULL mechanism
- [ ]  support http multiline header
- [ ]  support websocket ping-pong
- [ ]  support websocket close command and closing handshake
- [ ]  encoder should encode other message types, not only binary (ping/pong/close)
- [ ]  use a timer for the handshake
- [ ]  support zap and plain mechanism
- [ ]  use Sec-WebSocket-Protocol to negotiate protocol version and mechanism
- [ ] Improve masking, as now it is creating a new buffer and copy it over

smaillns:
What about req/rep sockets ? can we use ZMQ socket to carry http request ?

jean-airoldie:
I have question regarding the websocket transport design.

Some websocket servers require a client to send a binary payload to subscribe to a channel or create an authenticated session. This payload is application specific and must be provided by the user. Moreover, since this payload might contain a nonce, it might change every time it is needed (so it can't be simply cached).

If restarting the websocket is done behind the curtain by zeromq (like for the other transports), there should be a way for the socket to get this user provided payload (or payloads) every time the socket is restarted. Maybe a function pointer would do the trick?

somdoron:
My plan is to use the mechanism framework for that, initially I will support the plain mechanism which include a user and a password. Later I plan to add bearer mechanism. With mechanism each message include metadata of the sender. And you authenticate each new client using the ZAP ptotocol.

jean-airoldie:
I'm not following. Authentication on websockets is kind of the wild west. Each API as its own way of authenticating. How could ZAP be adapted for these abitrary mechanism?

somdoron:
The websocket transport is not general. It is specific to zeromq, for example the more flag is encoded as the first byte of every message. I'm working on the javascript side of it, it will be ready in few days. So only a client that will implement the ZWS protocol (which is not documented yet) will be able to communicate with zeromq on top of the websocket transport.

As part of the protocol NULL & Plain mechanism are going to be used and later bearer mechanism.

jean-airoldie:
Gotcha, I though it was more a general purpose transport. If you also control the remote websocket it makes sense to use ZAP.

wgaylord:
Any news?

somdoron:
I will push to github the pure javascript library.
At the point, you will be able to use it if you compile from source.
[...]

PR #3695, opened 2019-09-29
problem: unsecured websocket is rarely used in production

somdoron:
Solution: support websocket with tls (wss)

- [ ] Support CMake (added to #3581)
- [x] Support self-signed certificates
- [x] Move credentials to connecter and listener
- [ ] Validate certificates when set with setsockopt (added to #3581)
- [x] Load certificate
- [x] Change the option to enable-tls

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -670,6 +670,12 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_SOCKS_PASSWORD 100
 #define ZMQ_IN_BATCH_SIZE 101
 #define ZMQ_OUT_BATCH_SIZE 102
+#define ZMQ_WSS_KEY_PEM 103
+#define ZMQ_WSS_CERT_PEM 104
+#define ZMQ_WSS_TRUST_PEM 105
+#define ZMQ_WSS_HOSTNAME 106
+#define ZMQ_WSS_TRUST_SYSTEM 107
+
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_ZERO_COPY_RECV 10
--- a/src/session_base.cpp
+++ b/src/session_base.cpp
@@ -114,8 +114,14 @@ zmq::session_base_t::session_base_t (class io_thread_t *io_thread_,
     _socket (socket_),
     _io_thread (io_thread_),
     _has_linger_timer (false),
-    _addr (addr_)
+    _addr (addr_),
+    _wss_hostname (NULL)
 {
+    if (options_.wss_hostname.length () > 0) {
+        _wss_hostname = (char *) malloc (options_.wss_hostname.length () + 1);
+        assert (_wss_hostname);
+        strcpy (_wss_hostname, options_.wss_hostname.c_str ());
+    }
 }
 
 const zmq::endpoint_uri_pair_t &zmq::session_base_t::get_endpoint () const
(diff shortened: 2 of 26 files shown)


bluca:
Nice stuff - any chance NSS or GnuTLS could be used as alternative to the vendorized library? For all the usual reasons :-)

somdoron:
Is the license of gnuTLS compatible?

bluca:
Yes, it's LGPL2. Does NSS have the required functionality? Given we already have integration for it

somdoron:
I prefer gnuTLS as the docs are much better and no need for configuration files or database, as much as I can understand from the NSS docs and the nss_init method.

somdoron:
Switched to gnutls, gnutls also support sha1, so we can remove the dependency on NSS.

bluca:
Nice! There's a merge conflict

include/zmq.h
```diff
@@ -670,6 +670,12 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_SOCKS_PASSWORD 100
 #define ZMQ_IN_BATCH_SIZE 101
 #define ZMQ_OUT_BATCH_SIZE 102
+#define ZMQ_TLS_KEY_PEM 103
```
bluca: if these are only for WS, perhaps it's worth naming ZMQ_WS_...? To avoid confusion

somdoron: I will change it (to WSS), actually adding TLS to zmtp will be simple now and would make the name relevant, but if we do we can change them then and obsolete the previous names.

somdoron:
This is ready to be merged, some comments:
* I had to add another build to travis, because a minimum version of gnutls is required, which only available in xenial
* TLS cannot be enabled with CMake, I added to #3581 
* Some minor fixes are needed, I added them to #3581

include/zmq.h
somdoron: fixed

bluca:
aren't we using xenial everywhere by now? anything earlier is deprecated

somdoron:
No... still trusty

bluca:
Looks like Windows builds are all failing, could you please have a look?

src/session_base.cpp
```diff
@@ -114,8 +114,11 @@ zmq::session_base_t::session_base_t (class io_thread_t *io_thread_,
     _socket (socket_),
     _io_thread (io_thread_),
     _has_linger_timer (false),
-    _addr (addr_)
+    _addr (addr_),
+    _wss_hostname (NULL)
 {
+    if (options_.wss_hostname.length () > 0)
+        _wss_hostname = strdup (options_.wss_hostname.c_str ());
```
bluca:
```
1>..\libzmq\src\session_base.cpp(121) : error C2220: warning treated as error - no 'object' file generated
1>..\libzmq\src\session_base.cpp(121) : warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C++ conformant name: _strdup. See online help for details.
1>        C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\include\string.h(207) : see declaration of 'strdup'
1>select.cpp
2>server.cpp
2>..\libzmq\src\session_base.cpp(121) : error C2220: warning treated as error - no 'object' file generated
2>..\libzmq\src\session_base.cpp(121) : warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C++ conformant name: _strdup. See online help for details.
2>        C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\include\string.h(207) : see declaration of 'strdup
```

somdoron: fixed.

PR #3703, opened 2019-10-04
problem: wss transport return incorrect return code for error

somdoron:
Solution: return -1 for errors


bluca:
@somdoron unfortunately it still fails quite often: https://build.opensuse.org/package/show/home:bluca:branches:network:messaging:zeromq:git-draft/libzmq

somdoron:
I will check it again...

PR #3705, opened 2019-10-07
problem: zero-sized messages cause assertion when glibc assertion are on

somdoron:
Solution: when a zero-sized message is encoded or decoded avoid mem copy


somdoron:
@bluca this is what causes the Fedora issue.
Anyway, both wss and ws tests should pass now.

It was hard to find... maybe we should enable the glibcxx assertions when developing

somdoron:
on in one of the builds of travis...

bluca:
Great catch! Yes enabling that sounds like a good idea

PR #3713, opened 2019-10-14
avoid to use std::ceil

mdionisio:
avoid to used "ceil" has the good collateral effect that math library is not used any more and if we build libzmq with statically linked stdc++ we can avoid to add "-lm"


bluca:
Can you remove the -lm given it's no longer necessary? Also please fix the indentation:

```
-    _max_counters (
-      (_max_size + msg_t::max_vsm_size - 1)
-      / msg_t::max_vsm_size)
+    _max_counters ((_max_size + msg_t::max_vsm_size - 1) / msg_t::max_vsm_size)
```

mdionisio:
I will fix the indentation, but I cannot remove the "-lm" because it is an implicit dependency of libstdc++

PR #3717, opened 2019-10-19
Problem: ipc transport not supported under Windows

### LINKED ISSUE #3691: Support unix domain sockets under Win10
# Issue description
Hi there,
thanks for the great work but I have a feature request regarding unix domain sockets on Windows 10. According to the following [blog article](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/), Win10 supports now AF_UNIX domain sockets. If I use `ipc://`, a not supported socket type will be reported. Thus, could please extend the compatible Win10 sockets to AF_UNIX as well. The support of the utilized Win10 version can be checked by `sc query afunix`

Please see also [#1327@pyzmq](https://github.com/zeromq/pyzmq/issues/1327)

# Environment

* libzmq version (commit hash if unreleased): 18.1
* OS: Windows 10

use pyzmq

import zmq
context = zmq.Context()
socket = context.socket(zmq.PUB)
socket.bind('ipc:///tmp/stream')

Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "zmq\backend\cython\socket.pyx", line 550, in zmq.backend.cython.socket.Socket.bind
  File "zmq\backend\cython\checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: Protocol not supported

# What's the expected result?
Bound or connected socket

sigiesec:
Solution: implement support

Fixes #3691 

Open points:
- [x] Fix autotools builds (must define ZMQ_HAVE_IPC)
- [x] Implement temporary directory creation under Windows

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,6 +143,15 @@ if test "x$zmq_militant" = "xyes"; then
     AC_DEFINE(ZMQ_ACT_MILITANT, 1, [Enable militant API assertions])
 fi
 
+# IPC is available on all platforms supported by autotools build at the moment except OpenVMS and VxWorks.
+case "${host_os}" in
+    *vxworks*|*openvms*)
+        ;;
+    *)
+        AC_DEFINE(ZMQ_HAVE_IPC, 1, [Have AF_UNIX sockets for ipc transport])
+        ;;
+esac
+
 # Memory mis-use detection
 AC_MSG_CHECKING([whether to enable ASan])
 AC_ARG_ENABLE(address-sanitizer, [AS_HELP_STRING([--enable-address-sanitizer=yes/no],
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,54 @@ set(API_POLLER "" CACHE STRING "Choose polling system for zmq_poll(er)_*. valid 
 
 set(POLLER "" CACHE STRING "Choose polling system for I/O threads. valid values are
   kqueue, epoll, devpoll, pollset, poll or select [default=autodetect]")
+  
+if (WIN32)
+    # from https://stackoverflow.com/a/40217291/2019765
+    macro(get_WIN32_WINNT version)
+        if (CMAKE_SYSTEM_VERSION)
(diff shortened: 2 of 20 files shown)


somdoron:
Amazing

sigiesec:
Oh, I was too optimistic. The tests worked because they still skipped the ipc stuff :o There is an issue when calling `accept` in the `ipc_listener_t`, which claims `ENOTSOCK`, which is kind of strange. I hope I find some examples on using AF_UNIX on Win10.

But I think I fixed the autotools builds again.

sigiesec:
Ok, I think I got it now. For some reason, our select implementation and AF_UNIX go not together. With epoll, it works. I changed the defaults appropriately. Please give the changes to the autotools build files a particular look when reviewing.

sigiesec:
Note that this only enables the ipc transport for now, it still uses tcp for the signaler. This should probably also be replaceable with AF_UNIX, but since Windows has no socketpair, the pair of sockets needs to be created manually. I am not sure if there are any caveats in that.

configure.ac
```diff
@@ -143,6 +143,9 @@ if test "x$zmq_militant" = "xyes"; then
     AC_DEFINE(ZMQ_ACT_MILITANT, 1, [Enable militant API assertions])
 fi
 
+# IPC is available on all platforms supported by autotools build at the moment.
```
bluca: it used to be disabled on openvms and vxworks though?

sigiesec: Do these work through autotools? I wasn't sure about that. Can you suggest how this should be put here then?

bluca: given it was just checking the OS, you could just add new cases in the hostos switch where it's undefined:

case "${host_os}" in
      vxworks*)
..
     openvms*)

sigiesec: Thanks! Will fix that later on. Need to catch my flight now.

sigiesec: Ok, I added these exceptions now, hope it's fine!

PR #3722, opened 2019-10-26
Problem: condition var methods not explicitly deleted.

evoskuil:


bluca:
breaks some builds:

```
../src/condition_variable.hpp:322:59: error: defaulted and deleted functions only available with -std=c++0x or -std=gnu++0x [-Werror]
../src/condition_variable.hpp:323:76: error: defaulted and deleted functions only available with -std=c++0x or -std=gnu++0x [-Werror]
```

evoskuil:
Ugh, c++11 only.

bluca:
It's fine to have IFDEFS for that, if you want - other places in the code already do

evoskuil:
Prob not worth it.

PR #3735, opened 2019-11-12
Remove the superfluous build requirement of libsodium-devel

viciious:
The following lines in the .spec file ensure that the necessary packages are present on the system in case the rpm is being built with libsodium support enabled:

```
%bcond_with libsodium
%if %{with libsodium}
BuildRequires:  libsodium-devel
%define SODIUM yes
%else
%define SODIUM no
%endif
```

```
%{?_with_libsodium:BuildRequires: libsodium-devel}
%{?_with_libsodium:Requires: libsodium}
```

Unconditionally requiring libsodium-devel makes it impossible to build the spec file with default arguments if the aforementioned package is not installed. This merge request amends that.

### DIFF
--- a/RELICENSE/viciious.md
+++ b/RELICENSE/viciious.md
@@ -1,0 +1,13 @@
+# Permission to Relicense under MPLv2
+
+This is a statement by Victor Luchits
+that grants permission to relicense its copyrights in the libzmq C++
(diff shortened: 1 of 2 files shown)


bluca:

viciious:
> Thanks for the PR - please add a commit with a relicense statement https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

Done

PR #3738, opened 2019-11-13
Allow library output name to be changed externally

eldruin:
Internally we rename the libraries we integrate as to avoid interference with the library versions already available on the target systems by giving them a prefix.
This PR allows users to specify, for example:
```
-DZMQ_OUTPUT_BASENAME:STRING=MYORG_zmq
```
I used `BASENAME` as the flavor and/or version are appended later.

Alternatively, it would be also fine for us to add a prefix variable like `OUTPUT_NAME_PREFIX` or so but I was unable to find a good name that cannot be confused the target `PREFIX` property (e.g. `lib`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,8 @@ if (ENABLE_INTRINSICS)
    add_definitions(-DZMQ_HAVE_ATOMIC_INTRINSICS)
 endif()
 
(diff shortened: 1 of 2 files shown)


bluca:

eldruin:
@bluca done

PR #3741, opened 2019-11-19
Change XSUB -> XPUB multipart message processing.

drolevar:
Now only the first part in a multipart message will be treated as subscribe/unsubscribe.
The rest will be considered user messages regardless of the first byte.

### DIFF
--- a/RELICENSE/drolevar.md
+++ b/RELICENSE/drolevar.md
@@ -1,0 +1,16 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Andrij Abyzov
+that grants permission to relicense his copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other
+Open Source Initiative approved license chosen by the current ZeroMQ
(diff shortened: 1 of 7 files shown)


drolevar:
@somdoron Please, review.

bluca:
Please add a commit with a relicense grant https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

drolevar:
@bluca Done.

sigiesec:
LGTM overall, and great there you added an explicit test for this!

Please just
- [ ] fix two formatting errors in test, see https://travis-ci.org/zeromq/libzmq/jobs/614090448#L744

drolevar:
@sigiesec Done.

somdoron:
I would not change the behavior for frames following subscription/unsubscription as it is might be a breaking change. If you want to keep the new behavior we at least need an option for it, to not break existing behavior.

drolevar:
@somdoron I can implement it in a way that if the first frame is **NOT** a subscription/unsubscription, then all the rest are also treated in the same way. However, the logic will be slightly more complicated.

mrseanmorris:
Agree with @somdoron .  Can this be made a socket option, and probably only enabled in draft mode.

PR #3743, opened 2019-11-23
Fix ZMQ_LAST_ENDPOINT with WS[S] and use it in tests

bluca:


bluca:
@somdoron any idea why the WSS test would hang?

somdoron:
No... I will check out tomorrow

somdoron:
So the ws_address is used both for wss and ws, when you call ZMQ_LAST_ENDPOINT you actually get back a "ws" scheme, although you actually used "wss".
the ws_address_t.to_string is causing the problem, it is always returning the "ws" scheme.

So we either need to make a class for wss and only fix the to string, or let the ws_address constructor knows if it "ws" or "wss" and fix the to_string method.

Let me know what you prefer and I will fix it.

bluca:
I have added a wss_address_t subclass, as adding booleans around was awkward due to the get_socket_name common method

somdoron:
@bluca is it read to be merged?

bluca:
if it looks good to you, then yes

PR #3744, opened 2019-11-27
Problem: static libs are generated without given basename

liedtkeInTUM:
Using static lib generates a wrong output name.  It is just named lib{ZMQ_OUTPUT_BASENAME} and not using the given cmake variable as $ is missing.


bluca:

PR #3746, opened 2019-11-27
Change XSUB -> XPUB message processing.

drolevar:
Now if the first frame in a multipart message is not subscribe/unsubscribe,
the rest of the parts are also considered to be not subscribe/unsubscribe.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -675,6 +675,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_WSS_TRUST_PEM 105
 #define ZMQ_WSS_HOSTNAME 106
 #define ZMQ_WSS_TRUST_SYSTEM 107
+#define ZMQ_ONLY_FIRST_SUBSCRIBE 108
 
 
 /*  DRAFT Context options                                                     */
--- a/tests/test_xpub_manual.cpp
+++ b/tests/test_xpub_manual.cpp
@@ -456,57 +456,66 @@ void test_user_message ()
     test_context_socket_close (sub);
 }
 
+#ifdef ZMQ_ONLY_FIRST_SUBSCRIBE
 void test_user_message_multi ()
 {
+    const int only_first_subscribe = 1;
+
     //  Create a publisher
     void *pub = test_context_socket (ZMQ_XPUB);
     TEST_ASSERT_SUCCESS_ERRNO (zmq_bind (pub, "inproc://soname"));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (pub, ZMQ_ONLY_FIRST_SUBSCRIBE,
+                                               &only_first_subscribe,
+                                               sizeof (only_first_subscribe)));
 
     //  Create a subscriber
     void *sub = test_context_socket (ZMQ_XSUB);
     TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (sub, "inproc://soname"));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (sub, ZMQ_ONLY_FIRST_SUBSCRIBE,
+                                               &only_first_subscribe,
+                                               sizeof (only_first_subscribe)));
 
     //  Send some data that is neither sub nor unsub
     const uint8_t msg_common[] = {'A', 'B', 'C'};
     //  Message starts with 0 but should still treated as user
-    const uint8_t msg_0[] = {0, 'B', 'C'};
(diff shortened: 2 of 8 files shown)


drolevar:
@somdoron @mrseanmorris @bluca Ok, can this be a solution to your concerns?
Should I still look into making it a socket option?

bluca:
If you want to add a new socket option to change the behaviour, that's fine. Also please fix the formatting:

```
--- a/src/xsub.cpp
+++ b/src/xsub.cpp
@@ -104,13 +104,11 @@ int zmq::xsub_t::xsend (msg_t *msg_)
     bool first_part = !_more_send;
     _more_send = (msg_->flags () & msg_t::more) != 0;
 
-    if (first_part)
-    {
+    if (first_part) {
         _first_subscribe = false;
-    }
-    else if (!_first_subscribe)
+    } else if (!_first_subscribe)
         //  User message sent upstream to XPUB socket
-        return _dist.send_to_all(msg_);
+        return _dist.send_to_all (msg_);
 
     if (msg_->is_subscribe () || (size > 0 && *data == 1)) {
         //  Process subscribe message
```

drolevar:
@bluca If the change will be accepted into the master and stay there for subsequent releases, it's fine for me to keep it as it is.
But I'm not the project maintainer. Is it you making strategic decisions for the project?

sigiesec:
@drolevar One small thing: Could you please also update the commit message to follow the "Problem"/"Solution" convention as described in http://wiki.zeromq.org/docs:contributing section "Good commit messages"? Thanks!

The documentation in https://github.com/zeromq/libzmq/blob/master/doc/zmq_socket.txt should be updated to clarify this behaviour.

Regarding the socket option: Strictly, this is a breaking change, since the existing documentation could be interpreted in a way that showed the previous behaviour. However, IMO it seems rather unlikely that someone relied on this, so maybe it's fine without a socket option. In case someone runs into this and deems it a regression, a socket option might be added later on.

somdoron:
I agree with @sigiesec, it is a breaking change, but if it only after a none-sub/unsub I think we are fine without a socket option. @bluca what do you think?

drolevar:
@sigiesec Fixed.
@bluca What do you think?

bluca:
Sorry, I'll have to be the bad cop: there are a gazillion users doing the strangest things. In other words, if there's a chance it might break something, it will - please let's add a socket option

drolevar:
@bluca It's OK to be a bad cop. Someone has to :)

drolevar:
@bluca @sigiesec @somdoron
OK, socket option is implemented. Please, review and comment.

include/zmq.h
```diff
@@ -381,6 +381,7 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
+#define ZMQ_ONLY_FIRST_SUBSCRIBE 93
```
bluca: 93 already exists - needs to be added in the ZMQ_BUILD_DRAFT_API below, and duplicated in src/zmq_drafts.hpp

tests/test_xpub_manual.cpp
```diff
@@ -458,50 +458,57 @@ void test_user_message ()
 
 void test_user_message_multi ()
 {
+    const int only_first_subscribe = 1;
+
     //  Create a publisher
     void *pub = test_context_socket (ZMQ_XPUB);
     TEST_ASSERT_SUCCESS_ERRNO (zmq_bind (pub, "inproc://soname"));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (pub, ZMQ_ONLY_FIRST_SUBSCRIBE,
```
bluca: the test body needs to be wrapper in ```#ifdef ZMQ_ONLY_FIRST_SUBSCRIBE``` since it will be DRAFT for a while

drolevar: @bluca I guess it should the same in the xsub.cpp and xpub.cpp?

include/zmq.h
drolevar: Should be fixed now.

tests/test_xpub_manual.cpp
drolevar: Should be fixed now.

include/zmq.h
bluca: It's still missing from src/zmq_drafts.hpp though? With that, there's no need to ifdef in the implementation, but only for external usage

PR #3747, opened 2019-11-28
Problem: if IPC enabled libzmq does not compile when in uwp build

liedtkeInTUM:
The ipc integration broke the possibility to build on universal windows platform (uwp). My quick fix is to just disable ipc so instead of wepoll select is used. I think that should be the same behavior before ipc was introduced with #3717 

Please let me know if there is a better solution.


sigiesec:
Sorry for breaking this, unfortunately we don't have any UWP build on Appveyor right now, so this remained unnoticed. Should be possible to add one, though. Would you be interested in doing that?

With regard to the patch, I would prefer to do this in the `CMakeLists.txt` rather than by modifying `platform.hpp.in`. Would you mind moving it there?

liedtkeInTUM:
ok I will move it to the cmakeLists.
For my projects I have UWP build configured with gitlab ci. I can have a look on appveyor if that's possible there. You need to install some additional windows libraries and visual studio components.

liedtkeInTUM:
Ok, so after a few iterations you got working uwp build jobs. 
I don't know exactly why ctest fails but there I have until now no further knowledge.  Imo I would disable tests so the jobs succeed as they are still red. I tried to allow failure but that seems not to work for the ctest step. Just let me know what you prefer.

sigiesec:
@liedtkeInTUM Thanks for your work on the UWP CI, this seems to have been somewhat tricky. I will try to reproduce the ctest issue locally.

sigiesec:
Hm, I suspect that the issue is that some dependencies are not available, e.g. `vcruntime140_app.dll`. It's missing on my machine as well, and I have seen other reports that it doesn't seem to suffice to have the VC runtime redistributable installed. Have you got any experience with that?

Apart from that, if this cannot be solved immediately, I think it's fine to disable the tests for now, merge this, and file a bug.

PR #3758, opened 2019-12-13
Fix set cpu affinity

sonoware:


sigiesec:
Thanks for catching this!

PR #3763, opened 2019-12-23
Avoid possible buffers overruns in ws_engine

sigiesec:
... and various other cleanups


somdoron:
travis is failing, do you want me to merge or you want to fix it first?

sigiesec:
@somdoron There seems to be a problem remaining with the autotools build  on MacOS, where `strlcpy` is available from libc. This should be detected by `configure.ac`, but apparently there is something missing. Any idea what I need to fix?

sigiesec:
This should be fixed first before merging IMO. I moved some commits to another PR which should be fine.

bluca:
This rings a bell but can't quite put my finger on it right now - might have some time later to have a look

bluca:
Not sure what's happening on OSX - it looks like both CMake and autotools are finding strlcpy, but then ZMA_HAVE_STRLCPY is not defined. Some issue with the inclusion of platform.hpp maybe?

On Linux on the other hand, strlcpy is not defined in glibc, it comes from an external library, libbsd. So the include has to be <bsd/string.h> and pkg-config has to be used to get the compiler flags to link, and libbsd-dev and libbsd-devel has to be added as a build-dependency on debian and redhat packaging respectively (and to the travis CI)

sigiesec:
Adding support on Linux for using `bsd/string.h` can be done separately, I am not if it's worth it just for this function.

Fox the Mac OS X issue, I will look into it. I thought the failing build was an autotools one, but I was mistaken. There was a typo in the CMake `platform.hpp.in`, which I hopefully fixed now.

bluca:
Ah missed the typo! I'll add support for libbsd-dev in the next few days

bluca:
https://github.com/zeromq/libzmq/pull/3767

PR #3769, opened 2019-12-31
Update array.hpp

xqcool:
Problem:Calling back on an empty container causes undefined behavior.
Solution:Check with the function empty()

### DIFF
--- a/RELICENSE/xqcool.mq
+++ b/RELICENSE/xqcool.mq
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other share-alike OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by xqcool
+that grants permission to relicense its copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other 
+share-alike Open Source Initiative approved license chosen by the current 
+ZeroMQ BDFL (Benevolent Dictator for Life).
+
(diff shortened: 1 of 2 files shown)


bluca:

xqcool:
> Thanks for the PR - please add a commit with a relicense statement https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

OK.

sigiesec:
You are right that calling `back` on an empty container has undefined behaviour, but this is not a legal use of `array::erase` anyway. Rather than changing the condition, the if should be removed entirely, or changed into `zmq_assert (!_items.empty());` or rather `zmq_assert (items.size() > index_);`

xqcool:
OK，Your suggestion is more in line with zmq design，thank for you

发送自 Windows 10 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>应用

发件人: Simon Giesecke<mailto:notifications@github.com>
发送时间: 2020年1月6日 23:12
收件人: zeromq/libzmq<mailto:libzmq@noreply.github.com>
抄送: xqcool<mailto:xuquan316@live.com>; Author<mailto:author@noreply.github.com>
主题: Re: [zeromq/libzmq] Update array.hpp (#3769)


You are right that calling back on an empty container has undefined behaviour, but this is not a legal use of array::erase anyway. Rather than changing the condition, the if should be removed entirely, or changed into zmq_assert (!_items.empty()); or rather zmq_assert (items.size() > index_);

―

PR #3784, opened 2020-01-19
Problem: NEWS out of date with latest changes

bluca:
Solution: update it


bluca:
@somdoron I just added a few lines about websockets, please feel free to add more text/links/etc.
Also, should we make the default of --disable-ws affected by the status of the DRAFT flag?

somdoron:
Looks good to me.

> Also, should we make the default of --disable-ws affected by the status of the DRAFT flag?

I think so.

PR #3790, opened 2020-01-25
EAGAIN: Refer to ZMQ_RCVTIMEO when receiving in blocking-mode

std-any-emplace:


sigiesec:
Thanks for clarifying the documentation!

PR #3793, opened 2020-01-26
Clarify used C++-standard for compilation.

std-any-emplace:


README.md
```diff
@@ -15,6 +15,12 @@ protocols and more.
 
 ## Supported platforms <a name="#platforms"/>
 
+Libzmq is mainly written in C++98 with some optional C++11-fragments. Any platform
+with a conforming C++98-compiler and -library should be able to produce a
```
sigiesec: Actually, the second sentence is not true, for reasons unrelated to the C++ standard version. There are a lot of platform specific library calls, guarded through checks in CMakeLists.txt/autotools and #ifdefs, and it is rather unlikely that it will work on a new platform without some adaptations. The list of platforms supported to various degrees is in the file below. I'd suggest to just remove the second sentence, or specifically refer to the list below.

sigiesec:
@std-any-emplace Thanks for submitting the PR. Please see my comment though, to avoid pairing the clarification in the one aspect with adding confusion to the other aspect.

README.md
std-any-emplace: I removed the second part.

sigiesec:
Great, thanks!

Issue #3776, opened 2020-01-13
How to build for Windows with vs2019?


sun0929:
# I used "open in Visual Studio"  to clone the code to local, but I can't build the project.
# Environment
* libzmq version 4.3.2: 
* OS: win 10

sigiesec:
I am not sure if this is the same what you are doing, but for me, using "File/Open/CMake...", selecting the to top-level CMakeLists.txt file and then running Build All also doesn't work. I get the error:
```C:\src\libzmq-test\src\precompiled.cpp : fatal error C1083: Cannot open compiler intermediate file: 'c:\src\libzmq-test\out\build\x64-debug\precompiled.pch': Permission denied
```

Not sure what is causing this. For now, please run cmake from the command line as described in the INSTALL file, to generate a Visual Studio solution file.

sigiesec:
@sun0929 Can you check if it works with the current master?

PR #3797, opened 2020-01-28
Both libzmq and libzmq-static are trying to work in precompiled.pch. …

std-any-emplace:
…Solution: enforce order.
Fixes https://github.com/zeromq/libzmq/issues/3776.


sigiesec:
Thanks for the fix! IIUC this makes Ninja work and we can consequently switch the Appveyor builds to use Ninja which speeds them up a bit.

PR #3798, opened 2020-01-28
Add an extra space after "/Z7" in CMakeLists.txt.

std-any-emplace:
Else the flag gets joined with the flag following this and clang-cl does not like that.


std-any-emplace:
This is for the clang-cl coming with Msvc2019 BTW. There are more things broken in the CMakeLists.txt when using clang-cl but at least it compiles.

PR #3804, opened 2020-02-03
Submitting permission to relicense under MPLv2 statement.

tberkey:
Submitting permission to relicense under MPLv2 statement.


bluca:
Thank you!

PR #3805, opened 2020-02-03
CURVE: Reduce number of memory allocations and message copies

sigiesec:
I did a few runs on my desktop machine and average throughput via TCP loopback increases from ca. 5800 MB/s to 6100 MB/s for a 10k message size. Not statistically validated, but seems promising.

As a follow-up to this, performance can still be further optimised as discussed by saving one remaining extra memory allocation and copy on the encoding side, and by saving one copy on the decoding side.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,6 +174,7 @@ endif()
 # To disable curve, use --disable-curve
 
 option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" OFF)
+option(WITH_LIBSODIUM_STATIC "Use static libsodium library" OFF)
 option(ENABLE_CURVE "Enable CURVE security" ON)
 
 if(NOT ENABLE_CURVE)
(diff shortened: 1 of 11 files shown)


bluca:
build with libsodium 1.0.0 on debian 8 now fails:

```
[  171s] src/curve_mechanism_base.cpp: In member function 'int zmq::curve_encoding_t::encode(zmq::msg_t*)':
[  171s] src/curve_mechanism_base.cpp:159:61: error: 'crypto_box_easy_afternm' was not declared in this scope
[  171s]        &message_plaintext[0], mlen, message_nonce, _cn_precom);
[  171s]                                                              ^
[  171s] src/curve_mechanism_base.cpp: In member function 'int zmq::curve_encoding_t::decode(zmq::msg_t*, int*)':
[  171s] src/curve_mechanism_base.cpp:213:65: error: 'crypto_box_open_easy_afternm' was not declared in this scope
[  171s]                                         message_nonce, _cn_precom);
```

https://build.opensuse.org/package/live_build_log/network:messaging:zeromq:git-stable/libzmq/Debian_8.0/x86_64

PR #3806, opened 2020-02-03
Problem: intermittent timeouts of test_reconnect_ivl

sigiesec:
Solution: increase timeout to 15 seconds


bluca:
do we need the same in automake?

PR #3807, opened 2020-02-03
Problem: intermittent timeouts on test_heartbeats

sigiesec:
Solution: increase number of retries when reading monitor event


bluca:
Still failed:

../../config/test-driver: line 107: 30502 Alarm clock: 14         "$@" > $log_file 2>&1
FAIL: tests/test_heartbeats

Do we need to increase the test timeout as well?

sigiesec:
I think this could be merged now.

PR #3808, opened 2020-02-03
Fix intermittent test_monitor and test_security_zap timeouts/deadlocks for now

sigiesec:
@bluca @somdoron Adding the additional loop in `test_monitor_versioned_basic` seems to fix the intermittent failures. We might add this to the test, but I think that it shouldn't really be required for a user to always wait for ZMQ_EVENT_MONITOR_STOPPED. I can show you tomorrow what happens in the bad case without this addition.


sigiesec:
So, as a first step we can make the tests safe against races. The underlying problem still needs to fixed, or at least we need to document that somehow.

sigiesec:
Fixed the test failure is `test_monitor_basic`, I mixed up the `get_monitor_event` variants.

PR #3814, opened 2020-02-04
Implement ZMTP 3.1 subscribe/cancel via commands

bluca:
Thanks @somdoron for the suggestion to move the handling of the flag/command to the engine. Backward compatibility is retained. The drawback is that if a single sub/unsub is sent on a sub connected to many pubs, the work will be redone everytime. But it's a one byte or 7/11 bytes fixed size buffer, so it shouldn't make much of a difference, as it's just adding to an existing 2 bytes buffer.

### DIFF
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -212,6 +212,36 @@ int zmq::msg_t::init_leave ()
     return 0;
 }
 
+int zmq::msg_t::init_subscribe (const size_t size_, const unsigned char *topic)
+{
+    int rc = init_size (size_);
+    if (rc == 0) {
+        set_flags (zmq::msg_t::subscribe);
+
+        //  We explicitly allow a NULL subscription with size zero
+        if (size_) {
+            assert (topic);
+            memcpy (data (), topic, size_);
+        }
+    }
+    return rc;
+}
+
+int zmq::msg_t::init_cancel (const size_t size_, const unsigned char *topic)
+{
+    int rc = init_size (size_);
+    if (rc == 0) {
+        set_flags (zmq::msg_t::cancel);
+
+        //  We explicitly allow a NULL subscription with size zero
+        if (size_) {
+            assert (topic);
+            memcpy (data (), topic, size_);
+        }
+    }
+    return rc;
+}
+
 int zmq::msg_t::close ()
 {
     //  Check the validity of the message.
@@ -487,9 +517,12 @@ size_t zmq::msg_t::command_body_size () const
 {
     if (this->is_ping () || this->is_pong ())
         return this->size () - ping_cmd_name_size;
-    if (this->is_subscribe ())
+    else if (!(this->flags () & msg_t::command)
+             && (this->is_subscribe () || this->is_cancel ()))
+        return this->size ();
+    else if (this->is_subscribe ())
         return this->size () - sub_cmd_name_size;
-    if (this->is_cancel ())
+    else if (this->is_cancel ())
         return this->size () - cancel_cmd_name_size;
 
     return 0;
@@ -498,12 +531,17 @@ size_t zmq::msg_t::command_body_size () const
 void *zmq::msg_t::command_body ()
 {
     unsigned char *data = NULL;
+
     if (this->is_ping () || this->is_pong ())
         data =
           static_cast<unsigned char *> (this->data ()) + ping_cmd_name_size;
-    if (this->is_subscribe ())
+    //  With inproc, command flag is not set for sub/cancel
+    else if (!(this->flags () & msg_t::command)
+             && (this->is_subscribe () || this->is_cancel ()))
+        data = static_cast<unsigned char *> (this->data ());
+    else if (this->is_subscribe ())
         data = static_cast<unsigned char *> (this->data ()) + sub_cmd_name_size;
-    if (this->is_cancel ())
+    else if (this->is_cancel ())
         data =
           static_cast<unsigned char *> (this->data ()) + cancel_cmd_name_size;
 
--- a/src/v1_encoder.cpp
+++ b/src/v1_encoder.cpp
@@ -55,11 +55,16 @@ void zmq::v1_encoder_t::size_ready ()
 
 void zmq::v1_encoder_t::message_ready ()
 {
+    size_t header_size = 2; // flags byte + size byte
     //  Get the message size.
     size_t size = in_progress ()->size ();
 
     //  Account for the 'flags' byte.
     size++;
+
+    //  Account for the subscribe/cancel byte.
+    if (in_progress ()->is_subscribe () || in_progress ()->is_cancel ())
+        size++;
 
     //  For messages less than 255 bytes long, write one byte of message size.
     //  For longer messages write 0xff escape character followed by 8-byte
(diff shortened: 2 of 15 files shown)


src/xpub.cpp
```diff
@@ -149,14 +149,20 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
...
             //  breakage since the payload of the message is completely
             //  different. Manually craft an old-style message instead.
-            data = data - 1;
-            size = size + 1;
+            //  Since with inproc we do not store the sub/cancel command
+            //  string anymore in the data buffer, there would be no room
+            //  to hack in the 0/1 byte, and it overwrites the metadata
+            //  pointer in the message struct, so create a new blob_t.
+            //  The pushback makes a copy of the data array anyway, so it
+            //  does not make any difference.
```
sigiesec: I am not sure what this is meant to say and if it is still applicable. There is now the move below, as we discussed yesterday, so there is no copy during the push_back? I guess this also refers to the previous implementation, but this is very confusing in the code comment.

src/xpub.cpp
```diff
@@ -149,14 +149,20 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
             //  we can't give them back to userspace as it would be an API
             //  breakage since the payload of the message is completely
             //  different. Manually craft an old-style message instead.
-            data = data - 1;
-            size = size + 1;
+            //  Since with inproc we do not store the sub/cancel command
+            //  string anymore in the data buffer, there would be no room
+            //  to hack in the 0/1 byte, and it overwrites the metadata
+            //  pointer in the message struct, so create a new blob_t.
```
sigiesec: Today, I still remember what you showed me yesterday, and what "hack in" and "overwrites the metadata pointer" means, but I guess this won't be understandable by anyone else or even by you or me in the not-too-distant future, so I guess it's better to remove this part entirely.

src/xpub.cpp
```diff
@@ -177,14 +183,20 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
...
                 if (subscribe)
-                    *data = 1;
+                    *notification.data () = 1;
                 else
-                    *data = 0;
+                    *notification.data () = 0;
+                memcpy (notification.data () + 1, data, size);
 
-                _pending_data.push_back (blob_t (data, size));
+                _pending_data.push_back (ZMQ_MOVE (notification));
```
sigiesec: See above. 

Also. this whole part is identical to the part modified above. Can this be extracted into a new member function, so that it can be reused at the two call sites?

src/v1_encoder.cpp
```diff
@@ -55,23 +55,37 @@ void zmq::v1_encoder_t::size_ready ()
...
         _tmpbuf[0] = UCHAR_MAX;
         put_uint64 (_tmpbuf + 1, size);
         _tmpbuf[9] = (in_progress ()->flags () & msg_t::more);
-        next_step (_tmpbuf, 10, &v1_encoder_t::size_ready, false);
+        header_size = 10;
     }
+
+    //  Encode the subscribe/cancel byte, as it is no longer stored in
+    //  the data buffer of the message to allow different behaviours in
+    //  the v3.1 or legacy encoders.
```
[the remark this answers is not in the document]
bluca: It refers to the previous internal state - this data was added by the xsub class when the message is created, as opposed to when it is sent. I'll clarify.

src/v3_1_encoder.cpp
```diff
@@ -0,0 +1,101 @@
...
+        || in_progress ()->is_subscribe () || in_progress ()->is_cancel ()) {
+        protocol_flags |= v2_protocol_t::command_flag;
+        if (in_progress ()->is_subscribe ())
+            size += zmq::msg_t::sub_cmd_name_size;
+        else if (in_progress ()->is_cancel ())
+            size += zmq::msg_t::cancel_cmd_name_size;
+    }
+
+    //  Encode the message length. For messages less then 256 bytes,
+    //  the length is encoded as 8-bit unsigned integer. For larger
```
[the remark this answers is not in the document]
bluca: As far as I can see all encoders have minor differences. I guess the separate implementation is clearer than passing flags around. Just following the same pattern.

src/xpub.cpp
bluca: Clarified a bit. It's already in the commit message, but it's a counter-intuitive place to be doing this processing so I think it's better to have inline comments too.

src/xpub.cpp
```diff
@@ -149,14 +149,20 @@ void zmq::xpub_t::xread_activated (pipe_t *pipe_)
             //  we can't give them back to userspace as it would be an API
             //  breakage since the payload of the message is completely
             //  different. Manually craft an old-style message instead.
-            data = data - 1;
-            size = size + 1;
+            //  Since with inproc we do not store the sub/cancel command
+            //  string anymore in the data buffer, there would be no room
```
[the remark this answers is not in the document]
bluca: Clarified a bit. It's already in the commit message, but it's a counter-intuitive place to be doing this processing so I think it's better to have inline comments too.

src/xpub.cpp
bluca: That applies to the inproc case - I left it because for other transport, the data buffer does contain the command string so in theory it is possible to reuse it safely. I want to leave the comment there in case anybody in the future tries to do this optimization, as it's not immediately obvious. I'll clarify.

src/xpub.cpp
bluca: Given it's an if-else, I'll simply move it out of the block

bluca:
> I wonder if we have tests that test for the ZMTP version negotiation and for communicating with older peers?

Unfortunately we don't - it's been on my to-do list forever to add cross-compat tests with the various zmtp implementations, but never had the time to do it

sigiesec:
I'll add it to my to-do list as well ;)

src/v3_1_encoder.cpp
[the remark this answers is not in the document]
sigiesec: Ok. If I find time I will try to merge them and see what the result looks like. I just thought that v2_encoder and v3_1_encoder are probably very similar.

src/xpub.cpp
sigiesec: Thanks for clarifying!

src/xpub.cpp
sigiesec: Nice :)

src/xpub.cpp
sigiesec: Thanks for clarifying!

somdoron:
You can start by porting the zmtp2 test of netmq:
https://github.com/zeromq/netmq/blob/master/src/NetMQ.Tests/ZMTPTests.cs#L27

somdoron:
reviewed it now, very nice implementation. No copy now...

somdoron:
I think it might break the API, so when decoding, where do you convert a subscriber/join command to a msg with 0/1 at the beginning?

If you are not, the existing implementations of XPUB might break. What do you think?

somdoron:
I think we should move the handling of the cancel/subscribe commands from the zmtp_engine to v31_decoder. And those should remove the command_name from the data and leave only the data with the flags of subscribe/join. I think commands should be engine only thing and should not travel up to the sockets.

bluca:
> I think it might break the API, so when decoding, where do you convert a subscriber/join command to a msg with 0/1 at the beginning?
> 
> If you are not, the existing implementations of XPUB might break. What do you think?

Here:

https://github.com/zeromq/libzmq/blob/f17a794d59d5c99b6b9511ae3ec47d2b75872a0a/src/xpub.cpp#L163

I'd prefer to keep the internal representation using flags metadata too, rather than the 0/1 byte. That way in a distant future where zmtp < 3.1 is no longer supported, we just remove the conversion in xpub. It's also cleaner to use flags when reasoning about objects, rather than having to look at the data.

src/msg.cpp
```diff
@@ -212,6 +212,36 @@ int zmq::msg_t::init_leave ()
     return 0;
 }
 
+int zmq::msg_t::init_subscribe (const size_t size_, const unsigned char *topic)
```
sigiesec: `topic` should be `topic_` according to the coding style, same for `init_cancel`

bluca: https://github.com/zeromq/libzmq/pull/3844

sigiesec: Cool, thanks :) I just noticed when I looked into the new violations in the sonarcloud report.

PR #3820, opened 2020-02-06
websocket problems

somdoron:
solution: if msg is not shared or constant, mask the message in place


src/ws_address.cpp
```diff
@@ -67,13 +67,15 @@ zmq::ws_address_t::ws_address_t (const sockaddr *sa_, socklen_t sa_len_)
...
-    _path = std::string ("/");
+    _path = std::string ("");
 
     char hbuf[NI_MAXHOST];
     const int rc = getnameinfo (addr (), addrlen (), hbuf, sizeof (hbuf), NULL,
                                 0, NI_NUMERICHOST);
-    if (rc != 0)
-        _host = std::string ("localhost");
+    if (rc != 0) {
+        _host = std::string("localhost");
```
bluca: Space before bracket - formatting CI is complaining

PR #3823, opened 2020-02-09
problem: zeromq doesn't has a thread-safe peer to peer socket

somdoron:
Solution: a new socket type, called PEER. Very similar to SERVER, but can only connect to other PEERs. Also, a new zmq_connect_peer method, that connects and return a routing-id in a thread-safe and atomic operation.

I created this around two years ago for a crypto project that used NetMQ for the network protocol.

### DIFF
--- a/doc/zmq_connect_peer.txt
+++ b/doc/zmq_connect_peer.txt
@@ -1,0 +1,92 @@
+zmq_connect_peer(3)
+===================
+
+
+NAME
+----
+zmq_connect_peer - create outgoing connection from socket and return the connection routing id in thread-safe and atomic way.
+
+
+SYNOPSIS
+--------
+*uint32_t zmq_connect_peer (void '*socket', const char '*endpoint');*
+
+
+DESCRIPTION
+-----------
+The _zmq_connect_peer()_ function connects a 'ZMQ_PEER' socket to an 'endpoint' and then returns the endpoint 'routing_id'.
+
+The 'endpoint' is a string consisting of a 'transport'`://` followed by an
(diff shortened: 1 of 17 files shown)


doc/zmq_connect_peer.txt
```diff
@@ -0,0 +1,92 @@
...
+*ENOTSUP*::
+The socket is not of type 'ZMQ_PEER'.
+*EFAULT*::
+The 'ZMQ_IMMEDIATE' option is set on the socket.
+
+EXAMPLE
+-------
+.Connecting a peer socket to a TCP transport and sending a message
+----
+/* Create a ZMQ_SUB socket */
```
zedalaye: Shouldn't this be "Create a ZMQ_PEER socket" ?

doc/zmq_connect_peer.txt
```diff
@@ -0,0 +1,92 @@
...
+void *socket = zmq_socket (context, ZMQ_PEER);
+assert (socket);
+/* Connect it to the host server001, port 5555 using a TCP transport */
+uint32_t routing_id = zmq_connect (socket, "tcp://server001:5555");
```
zedalaye: Shouldn't this be "zmq_connect_peer(...)" ?

doc/zmq_connect_peer.txt
somdoron: yes, you are right. Want to make a pull request?

doc/zmq_connect_peer.txt
somdoron: you are right...

PR #3831, opened 2020-02-19
add option to stop trying to reconnect on ECONNREFUSED

WallStProg:
Problem: reconnecting to disconnected endpoint continues "forever"

ZeroMQ's automatic reconnection works reasonably well for processes that bind to "well-known" ports, but it works less well for processes that bind to "wildcard" ports, which are satisfied by the OS choosing an available ephemeral port.

Solution: With @somdoron's help, we've implemented a solution to this problem. This change will cause ZeroMQ to stop reconnecting (and "forget" the endpoint), under the following conditions:

1. If a connection attempt returns ECONNREFUSED, that indicates that the host received the connect, but that there is no code listening at that port.  In that case, we know that the original process has terminated (either normally or abnormally), and stop trying to reconnect.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -126,8 +126,6 @@ src_libzmq_la_SOURCES = \
 	src/own.hpp \
 	src/pair.cpp \
 	src/pair.hpp \
-	src/peer.cpp \
-	src/peer.hpp \
 	src/pgm_receiver.cpp \
 	src/pgm_receiver.hpp \
 	src/pgm_sender.cpp \
@@ -241,8 +239,6 @@ src_libzmq_la_SOURCES = \
 	src/v1_encoder.hpp \
 	src/v2_encoder.cpp \
 	src/v2_encoder.hpp \
-	src/v3_1_encoder.cpp \
-	src/v3_1_encoder.hpp \
 	src/v2_protocol.hpp \
 	src/vmci.cpp \
 	src/vmci.hpp \
@@ -1039,7 +1035,7 @@ test_apps += tests/test_poller \
 	tests/test_app_meta \
 	tests/test_xpub_manual_last_value \
 	tests/test_router_notify \
-	tests/test_peer
+	tests/test_reconnect_options
 
 tests_test_poller_SOURCES = tests/test_poller.cpp
 tests_test_poller_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
@@ -1081,9 +1077,9 @@ tests_test_router_notify_SOURCES = tests/test_router_notify.cpp
 tests_test_router_notify_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
 tests_test_router_notify_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
 
-tests_test_peer_SOURCES = tests/test_peer.cpp
-tests_test_peer_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
-tests_test_peer_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
+tests_test_reconnect_options_SOURCES = tests/test_reconnect_options.cpp
+tests_test_reconnect_options_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
+tests_test_reconnect_options_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
 endif
 
 if ENABLE_STATIC
--- a/Makefile.am
+++ b/Makefile.am
@@ -126,6 +126,8 @@ src_libzmq_la_SOURCES = \
 	src/own.hpp \
 	src/pair.cpp \
 	src/pair.hpp \
+	src/peer.cpp \
+	src/peer.hpp \
 	src/pgm_receiver.cpp \
 	src/pgm_receiver.hpp \
 	src/pgm_sender.cpp \
@@ -239,6 +241,8 @@ src_libzmq_la_SOURCES = \
 	src/v1_encoder.hpp \
 	src/v2_encoder.cpp \
 	src/v2_encoder.hpp \
+	src/v3_1_encoder.cpp \
+	src/v3_1_encoder.hpp \
 	src/v2_protocol.hpp \
 	src/vmci.cpp \
 	src/vmci.hpp \
@@ -1035,6 +1039,7 @@ test_apps += tests/test_poller \
 	tests/test_app_meta \
 	tests/test_xpub_manual_last_value \
 	tests/test_router_notify \
+	tests/test_peer \
 	tests/test_reconnect_options
 
 tests_test_poller_SOURCES = tests/test_poller.cpp
@@ -1076,6 +1081,10 @@ tests_test_app_meta_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
 tests_test_router_notify_SOURCES = tests/test_router_notify.cpp
 tests_test_router_notify_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
 tests_test_router_notify_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
+
+tests_test_peer_SOURCES = tests/test_peer.cpp
+tests_test_peer_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
+tests_test_peer_CPPFLAGS = ${TESTUTIL_CPPFLAGS}
 
 tests_test_reconnect_options_SOURCES = tests/test_reconnect_options.cpp
 tests_test_reconnect_options_LDADD = ${TESTUTIL_LIBS} src/libzmq.la
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -381,7 +381,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
-#define ZMQ_RECONNECT_STOP 109
 
 /*  Message options                                                           */
 #define ZMQ_MORE 1
@@ -399,10 +398,6 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 
 /*  RADIO-DISH protocol                                                       */
 #define ZMQ_GROUP_MAX_LENGTH 255
-
-/*  ZMQ_RECONNECT_STOP options                                                */
-#define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
-#define ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED 0x2
 
 /*  Deprecated options and aliases                                            */
 #define ZMQ_IDENTITY ZMQ_ROUTING_ID
@@ -682,7 +677,10 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_WSS_HOSTNAME 106
 #define ZMQ_WSS_TRUST_SYSTEM 107
 #define ZMQ_ONLY_FIRST_SUBSCRIBE 108
+#define ZMQ_RECONNECT_STOP 109
 
+/*  DRAFT ZMQ_RECONNECT_STOP options                                          */
+#define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_ZERO_COPY_RECV 10
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -381,6 +381,7 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
+#define ZMQ_RECONNECT_STOP 109
 
 /*  Message options                                                           */
 #define ZMQ_MORE 1
@@ -398,6 +399,10 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 
 /*  RADIO-DISH protocol                                                       */
 #define ZMQ_GROUP_MAX_LENGTH 255
+
+/*  ZMQ_RECONNECT_STOP options                                                */
+#define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
+#define ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED 0x2
 
 /*  Deprecated options and aliases                                            */
 #define ZMQ_IDENTITY ZMQ_ROUTING_ID
--- a/src/tcp_connecter.cpp
+++ b/src/tcp_connecter.cpp
@@ -274,6 +274,7 @@ zmq::fd_t zmq::tcp_connecter_t::connect ()
         if (err == WSAEBADF || err == WSAENOPROTOOPT || err == WSAENOTSOCK
             || err == WSAENOBUFS) {
             wsa_assert_no (err);
+            errno = wsa_error_to_errno (err);
         }
         return retired_fd;
     }
--- a/src/tcp_connecter.cpp
+++ b/src/tcp_connecter.cpp
@@ -274,8 +274,8 @@ zmq::fd_t zmq::tcp_connecter_t::connect ()
         if (err == WSAEBADF || err == WSAENOPROTOOPT || err == WSAENOTSOCK
             || err == WSAENOBUFS) {
             wsa_assert_no (err);
-            errno = wsa_error_to_errno (err);
         }
+        errno = wsa_error_to_errno (err);
         return retired_fd;
     }
 #else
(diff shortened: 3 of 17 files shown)


include/zmq.h
```diff
@@ -381,6 +381,7 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
+#define ZMQ_RECONNECT_STOP 109
```
bluca: this needs to be in the DRAFT section below, and it needs to be duplicated in src/zmq_draft.hpp

include/zmq.h
```diff
@@ -399,6 +400,10 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 /*  RADIO-DISH protocol                                                       */
 #define ZMQ_GROUP_MAX_LENGTH 255
 
+/*  ZMQ_RECONNECT_STOP options                                                */
+#define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
+#define ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED 0x2
```
bluca: this is not used anywhere?

bluca:
Please add documentation in docs/zmq_setsockopts.txt and docs/zmq_getsockopts.txt

sigiesec:
Sounds great! Thanks for also adding tests for the new functionality.

There are some build errors that need to be taken care of.

I wonder if it is possible to include the reason for the closure in the ZMQ_EVENT_CLOSED event. But this doesn't need to be done in this PR.

tests/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ set(tests
   test_monitor
   test_socket_null
   test_reconnect_ivl
+  test_reconnect_options
```
bluca: this needs to be moved to the ENABLE_DRAFTS section below

bluca:
Please add the test (in the ENABLE_DRAFTS section) in Makefile.am too, otherwise it will not be ran.

include/zmq.h
WallStProg: If you're referring to `ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED`, that's correct.  The reasons it's included here are:
- To make clear that there are multiple conditions which should stop reconnecting, and that those will be treated as bit-flags in options.
- To support further changes which address failed handshake, but those are a WIP and are not ready yet.

It seemed better to start with something small which is known to work before extending the functionality.  The `ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED` can come out if that would be better.

WallStProg:
Thanks for the comments.  I'll make changes based on your feedback and update the PR -- it may take a day or two.

include/zmq.h
bluca: I think it would be confusing for users if there is an option that is not yet implemented but can be set and does nothing. I'd prefer to leave it out, and add it when the implementation is added.

WallStProg: Makes sense -- just want to make sure that defining options as a bit-wise OR is acceptable, as I didn't see that approach used elsewhere.

WallStProg:
Thanks for the feedback - just updated PR, hope this is more like what it should be.

Note: I also fixed a compile error with `-Werror=reorder` that I noticed in some builds.

WallStProg:
P.S.  I'm getting errors from autotools build that don't seem related to these changes, but am unsure how to resolve:

```
  CXXLD    tools/curve_keygen
src/.libs/libzmq.so: undefined reference to `zmq::peer_t::peer_t(zmq::ctx_t*, unsigned int, int)'
src/.libs/libzmq.so: undefined reference to `zmq::peer_t::connect_peer(char const*)'
src/.libs/libzmq.so: undefined reference to `zmq::v3_1_encoder_t::v3_1_encoder_t(unsigned long)'
collect2: error: ld returned 1 exit status
make[1]: *** [tools/curve_keygen] Error 1
```

Makefile.am
```diff
@@ -126,8 +126,6 @@ src_libzmq_la_SOURCES = \
 	src/own.hpp \
 	src/pair.cpp \
 	src/pair.hpp \
-	src/peer.cpp \
-	src/peer.hpp \
```
bluca: please double check your rebase, these and other changes should not be in the commits - this is why the build is broken

WallStProg: Hmmm... I'm pulling from libzmq/master, and those lines are not in my Makefile.am:

```
	src/own.hpp \
	src/pair.cpp \
	src/pair.hpp \
	src/pgm_receiver.cpp \
	src/pgm_receiver.hpp \
```

Maybe that's why the build fails?

FWIW, peer.cpp does appear in CMakelists.txt.

bluca: That's why I suggested that the rebase was not successful, these lines you are removing need to be left there

WallStProg: You're right -- not sure how that happened, I think it should be OK now.

bluca:
The test is failing on Windows

WallStProg:
- I see lots of [failing tests](https://github.com/zeromq/libzmq/files/4232589/fails.txt) on Windows, with the new test among them.  I'm curious whether these additional failures are expected?

  In both fails of test_reconnect_options, the failure  appears to be caused by the presence of an "extra" monitor event after the expected final CLOSED event.  I'm not sure why there should be another event on the socket at that point, but I'm trying to repro that here (with no success so far).  Given that the assert is more of a "belt and suspenders" test, perhaps the simplest solution is simply to remove it?

```
    expect_monitor_event (sub_mon, ZMQ_EVENT_CLOSED);

    // ZMQ_EVENT_CLOSED should be last event, because of ZMQ_RECONNECT_STOP set above
    int event;
    char *event_address;
    int rc = get_monitor_event_with_timeout (sub_mon, &event, &event_address,
                                             2 * 1000);
    assert (rc == -1);
```

- I did partially fix the clang-format build error, but avoided reformatting code that I did not touch (e.g., msg.hpp).  If you want me to reformat all code, I'm happy to, but didn't want to assume that I should.

bluca:
Please ignore the format tool, there's a bug with the typedef detection

For the test, if more events can happen, then the test should optionally expect them

WallStProg:
> For the test, if more events can happen, then the test should optionally expect them

Well, they're not supposed to -- hence the assert.

In the interest of knowledge I'm going to update the test to print the spurious (?) event.

WallStProg:
The addl info from the failed test (below) seems to show that following the  `ZMQ_EVENT_CLOSED` event we get another event:

- The return from `get_monitor_event_with_timeout` (via `get_monitor_event_internal`) is the event value, which in this case is `ZMQ_EVENT_CONNECT_RETRIED`.
- The event variable has the value 0x86, which is the bit-wise OR of `ZMQ_EVENT_CLOSED`, `ZMQ_EVENT_CONNECT_RETRIED` and `ZMQ_EVENT_CONNECT_DELAYED`.  

Which leads to a couple of questions:

- Can a single event message report multiple events?  I didn't know that was possible -- if it is, my code needs to change.
- It looks like the code in `zmq::tcp_connecter_t::out_event` either closes the connection, or triggers a reconnect -- so the second `zmq::tcp_connecter_t::out_event ` *should* never happen -- right?

Now I know that the socket event *messages* are asynchronous to the actual *events*, but in this case that shouldn't matter -- all the test does is `zmq_connect` (sub => pub) followed by a `zmq_close` (pub), *after* it gets the `ZMQ_EVENT_HANDSHAKE_SUCCEEDED` message.

Is it possible that the socket monitor is returning extra "spurious" events?  Perhaps only under load?  Or only on Windows?  Or is my understanding of the events that should be delivered incorrect?  (If the latter, I'll just remove the offending check, but I don't want to do that if it in fact is triggering a bug).

Thanks for any help here.  

P.S.  There are still several different test failures on Windows that have nothing to do with the PR (esp. `test_many_sockets`) -- is that expected?


```
[00:05:52]       Start 80: test_reconnect_options
[00:05:52] 80: Test command: C:\projects\build_libzmq\bin\Debug\test_reconnect_options.exe
[00:05:52] 80: Test timeout computed to be: 10
[00:05:54] 81: C:\projects\libzmq\unittests\unittest_ypipe.cpp:80:test_create:PASS
[00:05:54] 86: C:\projects\libzmq\unittests\unittest_radix_tree.cpp:280:test_check_prefix:PASS
[00:05:58] 80: Still waiting for monitor event after 250 ms
[00:05:58] 80: Still waiting for monitor event after 500 ms
[00:05:58] 80: Still waiting for monitor event after 750 ms
[00:05:58] 80: Still waiting for monitor event after 1000 ms
[00:05:58] 80: Still waiting for monitor event after 250 ms
[00:05:58] 80: Still waiting for monitor event after 500 ms
[00:05:58] 80: Still waiting for monitor event after 750 ms
[00:05:58] 80: Unexpected event: 0x86, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
[00:05:58] 80: Assertion failed: false, file C:\projects\libzmq\tests\test_reconnect_options.cpp, line 205
[00:05:58] 80: C:\projects\libzmq\tests\test_reconnect_options.cpp:233:reconnect_default:PASS
[00:05:58] 80: C:\projects\libzmq\tests\test_reconnect_options.cpp:234:reconnect_success:PASS
[00:05:58] 85/87 Test #80: test_reconnect_options ...........***Failed    5.32 sec
```

WallStProg:
I've updated the test program to continue to print "unexpected" monitor events, and it shows that for failed tests, the reconnect doesn't stop, and continues to return
ZMQ_EVENT_CONNECT_RETRIED, ZMQ_EVENT_CONNECT_DELAYED, ZMQ_EVENT_CLOSED.

```
test 85
      Start 85: test_reconnect_options
85: Test command: C:\projects\build_libzmq\bin\Release\test_reconnect_options.exe
85: Test timeout computed to be: 10
85: Still waiting for monitor event after 250 ms
85: Still waiting for monitor event after 500 ms
85: Still waiting for monitor event after 750 ms
85: Still waiting for monitor event after 1000 ms
85: Still waiting for monitor event after 250 ms
85: Still waiting for monitor event after 500 ms
85: Still waiting for monitor event after 750 ms
85: Unexpected event: 0x86, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x70, value = 2/0x2 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x1a4, value = 128/0x80 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0xa9, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x70, value = 2/0x2 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x1a4, value = 128/0x80 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0xb2, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x70, value = 2/0x2 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x1e8, value = 128/0x80 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0xa2, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x70, value = 2/0x2 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x1e8, value = 128/0x80 (expected: 0x0, value = -1/0xffffffff)
85: Unexpected event: 0x99, value = 4/0x4 (expected: 0x0, value = -1/0xffffffff)
85/92 Test #85: test_reconnect_options ...........***Timeout  10.02 sec
```

It looks like this may be a Windows-specific issue, at which point I'm out of my depth -- it's been years since I did Windows development, and I don't have access to a Windows development environment.

Perhaps someone with more Windows expertise can take a look?  

- The `zmq::tcp_connecter_t::connect` method has #ifdef'd code specifically for Windows.  It doesn't appear to set `errno` the way the non-Windows code does, but maybe that's not necessary?
- It's not clear that ECONNREFUSED is a valid value for Windows (<https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=vs-2019>), but if not I don't know what is.   
- Given the above, I'm not sure why the test program *ever* succeeded on Windows, but it did pass intermittently in earlier builds (see following).  The latest code appears to fail every single time, but most often with just a plain "Timeout" error, with no addl. information.

```
log.txt:[00:02:15]   test_reconnect_options.cpp
log.txt:[00:02:29]   test_reconnect_options.vcxproj -> C:\projects\build_libzmq\bin\Debug\test_reconnect_options.exe
log.txt:[00:05:52]       Start 80: test_reconnect_options
log.txt:[00:05:52] 80: Test command: C:\projects\build_libzmq\bin\Debug\test_reconnect_options.exe
log.txt:[00:05:58] 80: Assertion failed: false, file C:\projects\libzmq\tests\test_reconnect_options.cpp, line 205
log.txt:[00:05:58] 80: C:\projects\libzmq\tests\test_reconnect_options.cpp:233:reconnect_default:PASS
log.txt:[00:05:58] 80: C:\projects\libzmq\tests\test_reconnect_options.cpp:234:reconnect_success:PASS```

**Comment by bluca** [comment]
@sigiesec any idea?

**Comment by somdoron** [comment]
Regarding the windows issue, the `tcp_connecter::connect` function doesn't set the errno on windows, from the connect function:

```
#ifdef ZMQ_HAVE_WINDOWS
    zmq_assert (rc == 0);
    if (err != 0) {
        if (err == WSAEBADF || err == WSAENOPROTOOPT || err == WSAENOTSOCK
            || err == WSAENOBUFS) {
            wsa_assert_no (err);
        }
        return retired_fd;
    }
#else
    //  Following code should handle both Berkeley-derived socket
    //  implementations and Solaris.
    if (rc == -1)
        err = errno;
    if (err != 0) {
        errno = err;    
        return retired_fd;
    }
#endif
``` 

As you can see the Windows case doesn't set the errno while other platforms do.
The fix is to add `errno = wsa_error_to_errno (err);` before `return retired_fd;` on the Windows case.

By the way, if the connect operation succeeds synchronously the test would pass, as the open function does handle the windows error code correctly.

File: src/tcp_connecter.cpp [hunk 5]
```diff
@@ -265,6 +274,7 @@ zmq::fd_t zmq::tcp_connecter_t::connect ()
         if (err == WSAEBADF || err == WSAENOPROTOOPT || err == WSAENOTSOCK
             || err == WSAENOBUFS) {
             wsa_assert_no (err);
+            errno = wsa_error_to_errno (err);
```
somdoron: it should be after the if block

WallStProg: Yup -- you're right.

WallStProg:
Thanks @somdoron !  That makes sense, and also explains why the test fails only intermittently.  I've pushed a new commit to the PR, so we'll see.

BTW, I've been trying to get a Windows dev env going using one of these: https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/.  So far both VirtualBox and Paralles are completely unusable, at least on my MacBook Pro.  I may try on my CentOS machine, but I'm not optimistic.  If anyone can suggest an alternative for testing/debugging on Windows pls let me know.

somdoron:
```
test 85
      Start 85: test_reconnect_options
85: Test command: C:\projects\build_libzmq\bin\Release\test_reconnect_options.exe
85: Test timeout computed to be: 10
85: Still waiting for monitor event after 250 ms
85: Still waiting for monitor event after 500 ms
85: Still waiting for monitor event after 750 ms
85: Still waiting for monitor event after 1000 ms
85: Still waiting for monitor event after 250 ms
85: Still waiting for monitor event after 500 ms
85: Still waiting for monitor event after 750 ms
85: C:\projects\libzmq\tests\test_reconnect_options.cpp:235:reconnect_default:PASS
85: C:\projects\libzmq\tests\test_reconnect_options.cpp:236:reconnect_success:PASS
85: C:\projects\libzmq\tests\test_reconnect_options.cpp:237:reconnect_stop_on_refused:PASS
85: 
85: -----------------------
85: 3 Tests 0 Failures 0 Ignored 
85: OK
85/92 Test #85: test_reconnect_options ...........   Passed    7.28 sec
```

First build has passed :)

somdoron:
One small formatting issue:
https://travis-ci.org/zeromq/libzmq/jobs/654082266#L788

bluca:
I'll fix the formatting

WallStProg:
> I'll fix the formatting
Your call of course, but I was just going to put things back the way they were, which seems more in line with existing tests (and also fixes formatting).

sigiesec:
Can someone open an issue for the Windows test failures and assign that to me? Will give it a look when I find time. Thanks!

WallStProg:
I added #3835 but I can't assign it

americaboy:
It's only works in tcp protocol, invalid in other protocol like ipc etc. ?

btorpey:
Yes -- the changes as published apply only to tcp, and that is the only scenario that we've tested.  From looking at the code, it *should* be possible to retrofit those changes to ipc and/or tipc, but I'm not familiar enough with those protocols to do so.  (For instance, does a connect return `ECONNREFUSED`?)

For `tcp_connector::out_event`, the relevant bits are:

```
...
    const fd_t fd = connect ();

    if (fd == retired_fd
        && ((options.reconnect_stop & ZMQ_RECONNECT_STOP_CONN_REFUSED)
            && errno == ECONNREFUSED)) {
        send_conn_failed (_session);
        close ();
        terminate ();
        return;
    }
...
```

PR #3832, opened 2020-02-21
Fix unused-variable warning in perf/proxy_thr.cpp

stac47:
The PR fixed the unused-variable warning/error in perf/proxy_thr.cpp:
```
perf/proxy_thr.cpp: In function ‘int main(int, char**)’:
perf/proxy_thr.cpp:335:9: error: unused variable ‘rv’ [-Werror=unused-variable]
  335 |     int rv = zmq_ctx_set (context, ZMQ_IO_THREADS, 4);
      |         ^~
perf/proxy_thr.cpp:403:9: error: unused variable ‘rc’ [-Werror=unused-variable]
  403 |     int rc = zmq_ctx_term (context);
      |         ^~
```

### DIFF
--- a/perf/proxy_thr.cpp
+++ b/perf/proxy_thr.cpp
@@ -74,6 +74,14 @@
 #define TEST_ASSERT_SUCCESS_ERRNO(expr)                                        \
     test_assert_success_message_errno_helper (expr, NULL, #expr)
 
+// This macro is used to avoid-variable warning. If used with an expression,
+// the sizeof is not evaluated to avoid polluting the assembly code.
+#ifdef NDEBUG
+#define ASSERT_EXPR_SAFE(x) do { (void)sizeof(x);} while (0)
+#else
+#define ASSERT_EXPR_SAFE(x) assert(x)
+#endif
+
 
 static uint64_t message_count = 0;
 static size_t message_size = 0;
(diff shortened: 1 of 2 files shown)


perf/proxy_thr.cpp
```diff
@@ -74,6 +74,14 @@
 #define TEST_ASSERT_SUCCESS_ERRNO(expr)                                        \
     test_assert_success_message_errno_helper (expr, NULL, #expr)
 
+// This macro is used to avoid-variable warning. If used with an expression,
+// the sizeof is not evaluated to avoid polluting the assembly code.
+#ifdef NDEBUG
```
bluca: where is this coming from? we never define it, intentionally

stac47: Not sure to understand your question. We build libraries for production with the defined macro:

From `man 3 assert`:

> If  the  macro NDEBUG was defined at the moment <assert.h> was last included, the macro assert() generates no code, and hence does nothing at all.  Otherwise, the macro assert() prints an error message to standard error and terminates the program by calling abort(3) if expression is false (i.e., compares equal to zero).

bluca: You should not do that. You'll get in unrecoverable error states and get stuck there.

stac47: My point of view is different. When you deliver a "release" version of a product, it is of no help to the end-user to know it cored somewhere: the released version should handle smoothly the errors. To me, `assert` macro should be only used in the debug version of your product to fail fast and ease the development.
As written in `man asert`:
> The purpose of this macro is to help programmers find bugs in their programs.  The message "assertion failed in file foo.c, function do_bar(), line 1287" is of no help at all to a user.

bluca: I didn't express a point of view, I merely stated what's going to happen. What you do with that information is entirely up to you.

PR #3847, opened 2020-03-14
[WiP] Refactor protocol literals into constants

JSYoo5B:
Protcol string for PGM, EPGM and NORM is still represented in literal strings.

While changing literals into constants, I also applied conditional compile for them.  
Some code blocks didn't covered with conditional compile, and the others is using both PGM and NORM.  
I tried to change those code blocks as simple as possible, but some code blocks seems messy.

I want someone review and comment especially for following code blocks. (I've marked WiP just for review)
[Code block1](https://github.com/JSYoo5B/libzmq/blob/5a933a0/src/socket_base.cpp#L371-L385)
[Code block2](https://github.com/JSYoo5B/libzmq/blob/5a933a0/src/socket_base.cpp#L556-L563)
[Code block3](https://github.com/JSYoo5B/libzmq/blob/5a933a0/src/socket_base.cpp#L1038-L1046)

### DIFF
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -369,14 +369,14 @@ int zmq::socket_base_t::check_protocol (const std::string &protocol_) const
     //  Specifically, multicast protocols can't be combined with
     //  bi-directional messaging patterns (socket types).
 #if defined ZMQ_HAVE_OPENPGM || defined ZMQ_HAVE_NORM
-    if ((false
-#ifdef ZMQ_HAVE_OPENPGM
-         || protocol_ == protocol_name::pgm || protocol_ == protocol_name::epgm
+#if defined ZMQ_HAVE_OPENPGM && defined ZMQ_HAVE_NORM
+    if ((protocol_ == protocol_name::pgm || protocol_ == protocol_name::epgm
+         || protocol_ == protocol_name::norm)
+#elif defined ZMQ_HAVE_OPENPGM
+    if ((protocol_ == protocol_name::pgm || protocol_ == protocol_name::epgm)
+#else // defined ZMQ_HAVE_NORM
+    if (protocol_ == protocol_name::norm
 #endif
-#ifdef ZMQ_HAVE_NORM
-         || protocol_ == protocol_name::norm
-#endif
-         || false)
         && options.type != ZMQ_PUB && options.type != ZMQ_SUB
         && options.type != ZMQ_XPUB && options.type != ZMQ_XSUB) {
         errno = ENOCOMPATPROTO;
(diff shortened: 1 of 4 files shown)


src/socket_base.cpp
```diff
@@ -546,7 +553,14 @@ int zmq::socket_base_t::bind (const char *endpoint_uri_)
         return rc;
     }
 
-    if (protocol == "pgm" || protocol == "epgm" || protocol == "norm") {
+    if (false
```
bluca: I'm a bit concerned about these kind of ifdefs - when it resolves to false || false won't some compilers or static analyzers flag them as tauthologies?

JSYoo5B: I am considering three other solution for this.

1. Make these protocol checking into inline function. ifdefs are handled in the function.
2. Split these condition checking codes for all ifdef cases
3. Do not cover ifdef for constants (protocol_name::pgm, epgm and norm) and keep previous ifdefs

What seems best solution for this problem?

bluca: I'm fine with making it more verbose and split the condition for ifdefs - it's just 2 or 3

JSYoo5B: Okay, I think second solution is better because it can improve conditional compile.

I'll fix and update.

JSYoo5B:
Added new commit to resolve previous ifdef tricks, rebased from master branch.

bluca:
there's a formatting problem

JSYoo5B:
Fixed formatting issue, ammended in last commit

bluca:
still errors:

```
--- a/src/socket_base.cpp
+++ b/src/socket_base.cpp
@@ -1038,17 +1038,16 @@ int zmq::socket_base_t::connect_internal (const char *endpoint_uri_)
     //  PGM does not support subscription forwarding; ask for all data to be
     //  sent to this pipe. (same for NORM, currently?)
 #if defined ZMQ_HAVE_OPENPGM && defined ZMQ_HAVE_NORM
-    const bool subscribe_to_all = protocol == protocol_name::pgm
-                                  || protocol == protocol_name::epgm
-                                  || protocol == protocol_name::norm
-                                  || protocol == protocol_name::udp;
+    const bool subscribe_to_all =
+      protocol == protocol_name::pgm || protocol == protocol_name::epgm
+      || protocol == protocol_name::norm || protocol == protocol_name::udp;
 #elif defined ZMQ_HAVE_OPENPGM
     const bool subscribe_to_all = protocol == protocol_name::pgm
                                   || protocol == protocol_name::epgm
                                   || protocol == protocol_name::udp;
 #elif defined ZMQ_HAVE_NORM
-    const bool subscribe_to_all = protocol == protocol_name::norm
-                                  || protocol == protocol_name::udp;
+    const bool subscribe_to_all =
+      protocol == protocol_name::norm || protocol == protocol_name::udp;
 #else
     const bool subscribe_to_all = protocol == protocol_name::udp;
 #endif
```

JSYoo5B:
I thought only tab->space problems. I just realized using clang-format.

clang-format patches applied.

PR #3855, opened 2020-03-31
WS and WSS should be optional as it requires additional libraries

Bart-PXL:
Suggestion to make WS and WSS fully optional including the necessary dependencies.


bluca:
This change breaks the build as you can see from the CI results, please fix it. Also please add a commit with a relicense statement https://github.com/zeromq/libzmq/blob/master/RELICENSE/README.md

grmt:
#3857 Probably satisfies this pull request

bluca:
Ok, closing then

PR #3857, opened 2020-04-06
add wss transport and fix tipc tests when building using cmake on linux

grmt:
bringing cmake up-to-par with autogen/configure experience.
triggered by interest in a wss transport running from browser
made GnuTLS default on as I see little reason to only use ws transport

This probably breaks windows build, as I only had V142 build environment and things did not work out


grmt:
hmm, I see this breaks some cmake builds where lib sodium is not installed

Issue #3858, opened 2020-04-06
Support zmq_msg_init_copy for convenience


gummif:
There are 3 ways to initialize `zmq_msg_t`:
1. Empty message
2. With initial size of uninitialized memory
3. Taking ownership/pointer to existing memory

One of the most common things done with ZeroMQ is constructing messages. A very common operation (textbook [example](https://www.oreilly.com/library/view/zeromq/9781449334437/ch01s04.html)) is to construct a message by copying data (especially for other language bindings). Without error checks it looks like so:
```c
zmq_msg_t m;
zmq_msg_init_size(&m, size);
memcpy(zmq_msg_data(&m), data, size);
```
if data can be null, then it becomes
```c
zmq_msg_t m;
zmq_msg_init_size(&m, size);
if (size)
    memcpy(zmq_msg_data(&m), data, size);
```
which are 3 function calls and 1 branch, excluding checks for errors.

May I suggest a 4th option to copy existing data `int zmq_msg_init_copy(zmq_msg_t *msg, const void *data, size_t size);`, used like so
```c
zmq_msg_t m;
zmq_msg_init_copy(&m, data, size);
```

If this is is well received I am willing to see how far I can go implementing it.

bluca:
That would be easily confused with zmq_msg_copy

gummif:
How about `zmq_msg_init_buffer`?

bluca:
sure that's fine, feel free to send a PR

PR #3860, opened 2020-04-08
Problem: boilerplate when init msg from data copy

gummif:
Solution: Add zmq_msg_init_buffer to construct
a message by copying memory from buffer.

See issue https://github.com/zeromq/libzmq/issues/3858

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -450,6 +450,7 @@ test_apps = \
 	tests/test_sub_forward \
 	tests/test_invalid_rep \
 	tests/test_msg_flags \
+	tests/test_msg_ffn \
 	tests/test_msg_init \
 	tests/test_connect_resolve \
 	tests/test_immediate \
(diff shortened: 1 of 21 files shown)


bluca:
THis is a new API, so it must go in the DRAFT section of the header and in src/zmq_drafts.h. Also it's probably better to avoid renaming existing tests - people might have them on SKIP individually

tests/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ set(tests
   test_invalid_rep
   test_msg_flags
   test_msg_ffn
+  test_msg_init
```
bluca, line 21: needs to be in the DRAFT section

Makefile.am
```diff
@@ -451,6 +451,7 @@ test_apps = \
 	tests/test_invalid_rep \
 	tests/test_msg_flags \
 	tests/test_msg_ffn \
+	tests/test_msg_init \
```
bluca, line 454: needs to be in the draft section or it won't build

tests/CMakeLists.txt
gummif: fixed

PR #3866, opened 2020-04-13
Problem: poller item lookup can be simplified

gummif:
Solution: Extract find functinos

### DIFF
--- a/src/socket_poller.cpp
+++ b/src/socket_poller.cpp
@@ -41,6 +41,18 @@ static bool is_thread_safe (const zmq::socket_base_t &socket_)
     return socket_.is_thread_safe ();
 }
 
+// compare elements to value
+template <class It, class T, class Pred>
+static It find_if2 (It b_, It e_, const T &value, Pred pred)
+{
+    for (; b_ != e_; ++b_) {
+        if (pred (*b_, value)) {
+            break;
+        }
+    }
+    return b_;
+}
+
 zmq::socket_poller_t::socket_poller_t () :
     _tag (0xCAFEBABE),
     _signaler (NULL)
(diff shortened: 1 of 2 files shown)


gummif:
One case fails with
```
==31426== Process terminating with default action of signal 14 (SIGALRM)
==31426==    at 0x405E54: zmq::atomic_counter_t::get() const (atomic_counter.hpp:200)
==31426==    by 0x405796: wait_timer_events(test_events_t&) (unittest_poller.cpp:118)
==31426==    by 0x405BA3: test_add_fd_and_remove_by_timer() (unittest_poller.cpp:204)
==31426==    by 0x409D09: UnityDefaultTestRun (unity.c:1339)
==31426==    by 0x405C9F: main (unittest_poller.cpp:269)
```
I don't think this is caused by this change, but maybe something to take a look at.

gummif:
I was unable to reproduce the error on my machine. But with the last commit the valgrind test succeeded (maybe valgrind doesn't like spinning on atomics).

unittests/unittest_poller.cpp
```diff
@@ -103,6 +103,7 @@ void wait_in_events (test_events_t &events_)
 {
     void *watch = zmq_stopwatch_start ();
     while (events_.in_events.get () < 1) {
+        msleep (1);
```
bluca, line 106: If this change introduces a regression, that needs to be fixed by artificial sleeps, then it's not good I'm afraid. If it's just simplifying the internal details it should not have an effect on usage.

gummif: The changes I made are in `socket_poller_t` but the tests that failed is for `poller_t` which I don't any code from the former. So I don't see how the changes I made are correlated to these failure (maybe a random unrelated issue).

PR #3870, opened 2020-04-17
problem: ZMQ_HEARTBEAT is not useful without sending an hello message

somdoron:
When using ZMQ_HEARTBEAT one still needs to implement application-level heartbeat in order to know when to send a hello message.
For example, with the majordomo protocol, the worker needs to send a READY message when connecting to a broker. If the connection to the broker drops, and the heartbeat recognizes it the worker won't know about it and won't send the READY msg.
To solve that, the majordomo worker still has to implement heartbeat. With this new option, whenever the connection drops and reconnects the hello message will be sent, greatly simplify the majordomo protocol, as now READY and HEARTBEAT can be handled by zeromq.

### DIFF
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -88,6 +88,17 @@ void zmq::send_routing_id (pipe_t *pipe_, const options_t &options_)
     pipe_->flush ();
 }
 
+void zmq::send_hello_msg (pipe_t *pipe_, const options_t &options_)
+{
+    zmq::msg_t hello;
+    const int rc =
+      hello.init_buffer (&options_.hello_msg[0], options_.hello_msg.size ());
+    errno_assert (rc == 0);
+    const bool written = pipe_->write (&hello);
+    zmq_assert (written);
+    pipe_->flush ();
+}
+
 zmq::pipe_t::pipe_t (object_t *parent_,
                      upipe_t *inpipe_,
                      upipe_t *outpipe_,
--- a/tests/test_hello_msg.cpp
+++ b/tests/test_hello_msg.cpp
@@ -1,0 +1,106 @@
+/*
+    Copyright (c) 2007-2020 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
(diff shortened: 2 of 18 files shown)


somdoron:
coverage test is failing, I'm not sure why..

src/pipe.cpp
```diff
@@ -88,6 +88,17 @@ void zmq::send_routing_id (pipe_t *pipe_, const options_t &options_)
     pipe_->flush ();
 }
 
+void zmq::send_hello_msg (pipe_t *pipe_, const options_t &options_)
+{
+    zmq::msg_t hello;
+    const int rc = hello.init_size (options_.hello_msg.size ());
+    errno_assert (rc == 0);
+    memcpy (hello.data (), &options_.hello_msg[0], options_.hello_msg.size ());
```
bluca, line 96: memcopy to NULL with 0 size is undefined behavior and will break badly (yes, really). Probably a good place to use the new zmq::msg_t::init_buffer

tests/test_hello_msg.cpp
```diff
@@ -0,0 +1,106 @@
+/*
+    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca, line 2: 2020

bluca:
> coverage test is failing, I'm not sure why..

it's been broken for the past week, some upstream issue

somdoron:
By the way, I'm getting different results when running clang-format locally vs CI, any idea?

bluca:
> By the way, I'm getting different results when running clang-format locally vs CI, any idea?

Different version of clang-format?

somdoron:
I will check it out, I would imagine different versions should yield same output with the same configuration file...

bluca:
> I will check it out, I would imagine different versions should yield same output with the same configuration file...

That would be nice, wouldn't it... unfortunately they don't

src/pipe.cpp
somdoron: fixed

tests/test_hello_msg.cpp
somdoron: fixed

somdoron:
@bluca fixed the issues and travis passes, except the coverage build

sappo:
Great new feature :)

somdoron:
thanks @sappo :)
Can you review my jeromq port of it?
https://github.com/zeromq/jeromq/pull/775

PR #3874, opened 2020-04-18
typo fix root README

shgalym:


shgalym:
do I need to grant RELICENSE details for such little typo fix?

PR #3875, opened 2020-04-18
Fix coverage job and appveyor job

bluca:


somdoron:
Do you want to wait for appveyor or should I merge?

somdoron:
We should consider Github actions, Appveyor is so slow it becomes irrelevant...

bluca:
Let's wait, just to make sure

bluca:
> We should consider Github actions, Appveyor is so slow it becomes irrelevant...

Yes, that would be good

PR #3885, opened 2020-04-27
Problem: no fuzz testing

bluca:
Solution: add some tests that can run as part of the test-suite with trivial input so that they don't break, and that can be built standalone with a fuzzing engine with a specific configure option, that can be used by oss-fuzz.
Partially fixes https://github.com/zeromq/libzmq/issues/3867


bluca:
Next step after this is to generate and store an appropriate corpus for each tests, with variations that mimick some part of ZMTP so that the engine can be tricked in processing inputs. I'll do that as soon as possible.

PR #3879, opened 2020-04-20
problem: appveyor is very slow

somdoron:
Solution: move to github actions

Only visual studio 2019 and 2017 are available. I didn't port all the matrix yet.

Anyway, ctest is failing on win32 build:

```
Test project D:/a/libzmq/libzmq/build_libzmq
      Start  1: test_ancillaries
 1/75 Test  #1: test_ancillaries .................Exit code 0xc000007b
***Exception:   0.16 sec
```

@sigiesec any idea?

PR #3886, opened 2020-04-27
Improve build experience - added new tests in cmake and generate github actions artifacts

grmt:
related recent PRs:
https://github.com/zeromq/libzmq/pull/3879
https://github.com/zeromq/libzmq/pull/3876
https://github.com/zeromq/libzmq/pull/3871

thank you somdoron

### DIFF
--- a/tests/testutil.hpp
+++ b/tests/testutil.hpp
@@ -30,11 +30,7 @@
 #ifndef __TESTUTIL_HPP_INCLUDED__
 #define __TESTUTIL_HPP_INCLUDED__
 
-#if defined ZMQ_CUSTOM_PLATFORM_HPP
 #include "platform.hpp"
-#else
-#include "../src/platform.hpp"
-#endif
 #include "../include/zmq.h"
 #include "../src/stdint.hpp"
 
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -92,7 +92,7 @@ jobs:
     - name: artifact bin
       uses: actions/upload-artifact@v1
       with:
-        name: bin-x64-windows-shared-${{ matrix.MSVCVERSION }
+        name: bin-x64-windows-shared-${{ matrix.MSVCVERSION }}
         path: build_libzmq\bin
     - name: artifact install v141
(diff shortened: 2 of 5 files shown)


tests/testutil.hpp
```diff
@@ -30,11 +30,7 @@
 #ifndef __TESTUTIL_HPP_INCLUDED__
 #define __TESTUTIL_HPP_INCLUDED__
 
-#if defined ZMQ_CUSTOM_PLATFORM_HPP
```
bluca, line 33: why is this being removed?

grmt: I removed it, because visual code intellisense was complaining about the platform header file, as it doesn't exist. Cmake will generate it into the cmake build directory, but not in ../src/platform.hpp. I did not go back in time. Perhaps it was used in the deprecated msvc solutions,  but it is not being used in the current cmake build environment (for windows). 

I indeed did not finish this completely, and I should remove it everywhere.  I'll give it a try.

NB windows x64 test_security_zap has an attitude. It might need some attention.

bluca: Autotools will generate it in src. Please remove this change and fix your edit settings instead.

grmt: Isn't the #ifdef ZMQ_CUSTOM_PLATFORM_HPP logic only used in cmake / windows ?
I didn't find it anywhere else in the repo, The autotools builds should fail miserably if they were missing the header file.

Let's wait on the build results. 
Maybe I should have created a separate PR for this.

bluca: No, this is wrong - autotools always generates in src. Please remove it.

grmt: The builds and test runs seem fine to me.
There is no problem with autotools generating the platform.hpp in src directory.
There is just no need to use the #ifdef  ZMQ_CUSTOM_PLATFORM_HPP  logic

I'm happy to put it back, upon your request, though I don't understand what is going wrong here with the builds.

bluca:
There are merge conflicts, and the broken change has not been reverted after a month. Closing for now. Feel free to reopen after fixing and rebasing.

PR #3887, opened 2020-04-27
Problem: Multiple fd_t definitions

gummif:
Solution: Unify definition

### DIFF
--- a/tests/test_poller.cpp
+++ b/tests/test_poller.cpp
@@ -209,7 +209,7 @@ void test_null_poller_pointers_wait_all_indirect ()
 void test_null_poller_pointer_poller_fd ()
 {
     void *null_poller = NULL;
-    zmq_fd_t fd;
+    fd_t fd;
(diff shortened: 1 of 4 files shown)


tests/testutil.hpp
```diff
@@ -37,6 +37,7 @@
 #endif
 #include "../include/zmq.h"
 #include "../src/stdint.hpp"
+#include "../src/fd.hpp"
```
bluca, line 40: Please don't - I do not want to start including headers from the tests. Yes there's already one or two - they should be removed, not the other way around.

gummif: Fixed

tests/test_poller.cpp
```diff
@@ -41,9 +41,9 @@
 
 SETUP_TEARDOWN_TESTCONTEXT
 
-fd_t get_fd (void *socket_)
+zmq_fd_t get_fd (void *socket_)
```
bluca, line 44: why the renaming churn? it's just tests, there's no need to namespace internal types

gummif: fixed

PR #3896, opened 2020-05-05
Problems: memory sanitizer fails due to uninitialised global, want to run more tests with sanitizer

bluca:
@sigiesec this reverts https://github.com/zeromq/libzmq/commit/c590873ff7a81a3c373884eab37cbe2068fbfe4a - when you have time to have a look and fix the MSAN error, please send a PR to re-enable it


bluca:
With this I think the infrastructure changes are all done and we are in a good position to start tackling the list of bugs found by the fuzzers and add the minimised corpus for regression testing. Will try and get started on the long list soon...

bluca:
> With this I think the infrastructure changes are all done and we are in a good position to start tackling the list of bugs found by the fuzzers and add the minimised corpus for regression testing. Will try and get started on the long list soon...

Actually two things are missing, adding MSAN to the regression tests and support for something like CIFuzz so that we know we don't break the libfuzzer builds

PR #3908, opened 2020-05-11
Problem: release script for OBS build fails often

bluca:
Solution: now that tar_scm supports @PARENT_TAG@ as a revision to
automatically fetch the most recent tag on the default branch, use
it to simplify everything

https://github.com/openSUSE/obs-service-tar_scm/pull/359


bluca:
@somdoron note that I deleted the -win tag, otherwise this doesn't work

PR #3911, opened 2020-05-12
[WIP, RFC] Problem: lot of use of dynamic memory allocation, leading to performance degradation and non-determism

mjvankampen:
See #3644 

Basically the problem we are trying to solve is that the default allocator (malloc/free) is:
- slow
- unpredictable

@f18m showed that using pre-allocated memory that is re-used performance can be increased significantly. My own use-case is somewhat different as I prefer to have as little dynamic memory allocation in my applications as determinism is an element for me. Allocation on the heap is a source of non-determinism.

I took @f18m pull request and tried to make a a bit nicer to allow users to provide their own queues and made the queue grow dynamically.

Open discussion points:
- [ ] Do we need to allow the user to provide an allocator, or is the zero copy feature enough?
- [ ] SPSC queue: remove I guess, @somdoron convinced it will not work, @f18m tends to agree (I don't have the insight to judge this)
- [ ] <C++11 implementation: std::queue + mutex or hide global pool behind ifdefs? I prefer a (slower) <C++11 implementation
- [ ] External integration: ok to integrate concurrentqueue.h like this? I feel like writing a custom queue for zmq seems a too big effort, but it does not align with C4 2.3.5.
- [ ] Naming: ZMQ_MSG_ALLOCATOR_GLOBAL_POOL or ZMQ_ALLOCATOR_GLOBAL_POOL. Is it used anywhere else outside the msg api?
- [ ] What level of testing is sufficient: I would add another test with a larger size
- [ ] Add another message type  for allocator: is this required or is the current setup fine
- [ ] Add allocator to context: so that other msg_init are done using the custom allocator. Is this a feature that is required/possible (at this stage)?
- [ ] Any other suggestions? Such as required features?
- [ ] Is the current dynamic resize ok: it adds a bit of overhead vs the compile time bins.

Todo
- [ ] Add RFC for better PR description
- [ ] Update documentation
- [ ] Add statistics from memory pool
- [ ] Fix CI: broken build on some platforms
- [ ] Add unit tests: coverage below par
- [x] Change to function pointers, virtuals cannot be exposed through zmq.h
- [ ] Fix missing headers (something with pch)
- [ ] Adds free_fn for allocator

# RFC

## Problem

ZMQ uses quite a bit of dynamic memory allocation. This results in 

- some loss in performance as heap allocation takes time, can introduce implicit synchronisation and lead to heap fragmentation, and
- non-determinism as most new/malloc implementations do not have a upper bound time limit. 

The first one is not a problem in itself but as one ZMQ strengths is performance it is always good to improve this. The second one is an issue in some systems that require a bound on latency.

## Solution direction

The most simple solution would be to allow users to provide a custom allocator that can be used instead of the default allocator. This has the advantage that users can customize the application to their needs and solve the above issues. A user can select the trade-off between memory vs. speed vs. determinism themselves. Of course it adds some disadvantages as well:

- some overhead is likely to be introduced (or some templating);
- a new API is required; and
- this allows users to break ZMQ with their custom allocator, resulting in more questions for the ZMQ community.

To mitigate these somewhat a standard implementation for an allocator should be provided.

## Default allocator usage

While there are a quite a few new calls in zmq, Approximately 100 new and 50 malloc calls. Replacing all these is quite a bit of work and most likely not necessary as most of these happen during initialisation. Focus should be on the allocations that happen often and all the time. The best place to look for this is the code that does an allocation for every message sent or received. This results in allocations in:

- msg.hpp/cpp, and
- decoders (decoder_allocators.hpp/cpp).

At the moment (correct me if I'm wrong) the decoder is does not allocate per message but per stream. It seems that the most gain is in the messages, which means the decoder becomes a nice to have (at this point).

## APIs

- message level API: set an allocator per message.

Another option is to set an allocator for all messages (using the context). In my opinion this is a nice to have, and should only be implemented if someone asks for it.

### ```zmq_allocator_new(type)```

Construct an allocator of ```type```. This is a way to provide default allocators to users. If they want to provide their own custom allocator, in that case this function is not used.

### ```zmq_allocator_destroy(allocator)```

Destroy an allocator (either from the user or native) after you are done with it. Can only be called when all messages have been handled. This means after terminating the context.

### ```zmq_msg_init_allocator(message, size, allocator)```

Initialize a message using a custom allocator. Very similar to ```zmq_msg_init_size``` except with an argument to pass an allocator.

### ```zmg_allocator_t```

A struct consisting of function pointers.

```void*(*allocate_fn) (void *allocator, size_t len)```
Allocate function

```void(*deallocate_fn) (void *allocator, void *data_)```
Deallocate function

```bool(*check_tag_fn) (void *allocator)```
Check the tag of an allocator to see if the allocator pointer is valid

```void *allocator```
Pointer to an allocator itself (on which to use the above functions)

optional (not required for the msg interface):
```void(*release_fn) (void *allocator, void *data_)```
Release data, this means the allocator is not the owner anymore.

### Default

A basic allocator using new/delete. Basically the same what we have now but with a slight added cost of indirection through a vtable (might be elimated using an interface and marking things as final).

### Global pool

This is an allocator which keeps queues of memory chunks based on bin sizes. This reduces the amount of dynamic memory allocation (thus gaining speed) at the cost of reserving more memory than strictly needed. To prevent an unreasonable startup size memory gets allocated on demand, but with the power of two. So if more memory chunks are required than available in a bin the number of chunks in a bin is doubled. If the biggest bin is too large, a bin twice the size is created. 

Depending on the workload this way of working can lead to excessive memory use. Scenarios where large burst of same message size, but changing message sizes between bursts, might lead to excessive memory use. If we want to prevent this it might be wise to set a limit to the pre-allocated memory and fall back on new/delete if it is exceeded or return an error. Or we trust the user to not pick this specific allocator.

### Other?

### CODE REVIEW

**Comment by TomMD** [comment]
Interesting.  It seems infer's memory leak detection doesn't realize vector push_back captures the pointer.  Sorry about the noise on your PR, I'll take this learning and see what we can do here.

File: external/mpmcqueue/concurrentqueue.h [hunk 1]
```diff
@@ -0,0 +1,3712 @@
...
+				(p + --i)->~U();
+		}
+		aligned_free<U>(p);
+	}
+
+	template<typename U>
+	static inline U* create()
+	{
+		void* p = aligned_malloc<U>(sizeof(U));
+		return p != nullptr ? new (p) U : nullptr;
```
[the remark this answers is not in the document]
mjvk: Not valid, if aligned malloc does not return a nullptr, new will fill that space and p will be returned for further use and can be free.

external/mpmcqueue/concurrentqueue.h
```diff
@@ -0,0 +1,3712 @@
...
+	}
+
+	template<typename U, typename A1>
+	static inline U* create(A1&& a1)
+	{
+		void* p = aligned_malloc<U>(sizeof(U));
+		return p != nullptr ? new (p) U(std::forward<A1>(a1)) : nullptr;
```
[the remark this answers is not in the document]
mjvk: Same as earlier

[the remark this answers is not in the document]
mjvk: Same as above

src/allocator_global_pool.cpp
```diff
@@ -0,0 +1,137 @@
...
+    _storage_mutex.lock ();
+    size_t oldSize = m_storage.size ();
+    if (oldSize <= bl) {
+        m_storage.resize (bl + 1);
+        m_free_list.resize (bl + 1);
+        for (auto i = oldSize; i <= bl; i++) {
+            size_t msg_size = MsgBlockToBytes (i);
+            m_storage[i].num_msgs =
+              ZMG_GLOBAL_POOL_INITIAL_BLOCK_SIZE / msg_size;
+            m_storage[i].raw_data.push_back (
```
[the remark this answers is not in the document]
mjvk: Reachable through the vector

src/allocator_global_pool.cpp
```diff
@@ -0,0 +1,137 @@
...
+    _storage_mutex.unlock ();
+}
+
+void zmq::allocator_global_pool_t::expand_block (size_t bl)
+{
+    size_t msg_size = MsgBlockToBytes (bl);
+    _storage_mutex.lock ();
+    size_t messagesToAdd = m_storage[bl].num_msgs;
+    m_storage[bl].num_msgs += messagesToAdd;
+    m_storage[bl].raw_data.push_back (
```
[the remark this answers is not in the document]
mjvk: Reachable through the vector

src/allocator_global_pool.hpp
```diff
@@ -0,0 +1,111 @@
...
+        n >>= k;                                                               \
+    }
+        assert (n != 0);
+        int i = 0;
+        S (32);
+        S (16);
+        S (8);
+        S (4);
+        S (2);
+        S (1);
```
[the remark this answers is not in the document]
mjvk: Used by the recursive macro. Will have a look if I can do this another way, but this was the fastest version I could find for log2n of an integer.

mjvk:
@TomMD Cool, thanks. I think all of these are actually false positives. Do you agree?

TomMD:
@mjvk Yes, I agree.  False positives.

tests/test_msg_init.cpp
```diff
@@ -74,6 +74,29 @@ void test_msg_init_buffer ()
...
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_msg_init_allocator (&msg, 6, allocator));
+    TEST_ASSERT_EQUAL_INT (6, zmq_msg_size (&msg));
+    memcpy (zmq_msg_data (&msg), data, 6);
+    TEST_ASSERT_EQUAL_STRING_LEN (data, zmq_msg_data (&msg), 6);
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_msg_close (&msg));
+
+    zmq_msg_t msg2;
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_msg_init_allocator (&msg2, 0, allocator));
+    TEST_ASSERT_EQUAL_INT (0, zmq_msg_size (&msg2));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_msg_close (&msg2));
```
mjvk, line 94: Missing allocator_destroy resulting in leak

mjvk:
I would like to get your comments on a user interfacing API for providing an allocator. How do you feel about this? The zero-copy API might be enough, but a true API might be nicer in my opinion. Especially if we want to use the allocator in more spots.

CMakeLists.txt
```diff
@@ -214,9 +214,9 @@ endif()
 # Select curve encryption library, defaults to tweetnacl To use libsodium instead, use --with-libsodium(must be
 # installed) To disable curve, use --disable-curve
 
-option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" ON)
+option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" OFF)
```
bluca, line 217: why are this and the other options defaults being changed?

include/zmq.h
```diff
@@ -263,10 +263,12 @@ typedef struct zmq_msg_t
 #endif
 } zmq_msg_t;
 
-typedef void(zmq_free_fn) (void *data_, void *hint_);
+typedef void (zmq_free_fn) (void *data_, void *hint_);
 
 ZMQ_EXPORT int zmq_msg_init (zmq_msg_t *msg_);
 ZMQ_EXPORT int zmq_msg_init_size (zmq_msg_t *msg_, size_t size_);
+ZMQ_EXPORT int
+zmq_msg_init_allocator (zmq_msg_t *msg_, size_t size_, void *allocator_);
```
bluca, line 271: this must go in the DRAFT section below

include/zmq.h
```diff
@@ -263,10 +263,12 @@ typedef struct zmq_msg_t
 #endif
 } zmq_msg_t;
 
-typedef void(zmq_free_fn) (void *data_, void *hint_);
+typedef void (zmq_free_fn) (void *data_, void *hint_);
```
bluca, line 266: don't change this, it's a false positive in the formatter tool

bluca: same for other prototypes

src/allocator_default.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca, line 2: 2019-2020

src/allocator_global_pool.hpp
```diff
@@ -0,0 +1,137 @@
...
+    std::vector<basic_concurrent_queue_t<uint8_t *> > _free_list;
+#endif
+    mutex_t _storage_mutex;
+
+    inline size_t MsgBlockToBytes (size_t block)
+    {
+        return ZMQ_GLOBAL_POOL_FIRST_BLOCK_SIZE * 2 ^ block;
+    }
+
+    // by Todd Lehman https://stackoverflow.com/questions/994593/how-to-do-an-integer-log2-in-c
```
bluca, line 107: IIRC all code snippets from SO are licensed under CC-BY-SA v3 which is not compatible with the GPL license

include/zmq.h
mjvk: Thanks I'll have a look. I normally do format on save so this slips in!

src/allocator_global_pool.hpp
mjvk: Removed and replaced by somewhat less efficient (but more readable) implementation. Shouldn't impact performance too much.

CMakeLists.txt
mjvk: No good reason, removed changes

include/zmq.h
mjvk: Fixed

include/zmq.h
mjvk: Fixed

src/allocator_default.cpp
mjvk: Fixed

include/zmq.h
```diff
@@ -698,6 +698,33 @@ ZMQ_EXPORT int zmq_ctx_get_ext (void *context_,
...
+#define ZMQ_MSG_ALLOCATOR_PER_THREAD_POOL 1
+// using internally a MPMC queue, C++11 required
+#define ZMQ_MSG_ALLOCATOR_GLOBAL_POOL 2
+
+ZMQ_EXPORT void *zmq_msg_allocator_new (int type_);
+ZMQ_EXPORT int zmq_msg_allocator_destroy (void **allocator_);
+ZMQ_EXPORT int
+zmq_msg_init_allocator (zmq_msg_t *msg_, size_t size_, void *allocator_);
+
+struct zmq_allocator_t
```
gummif, line 714: Does this need to be exposed? If so, can the type not be used in the API, as in `ZMQ_EXPORT zmq_allocator_t *zmq_msg_allocator_new (int type_);`?

include/zmq.h
```diff
@@ -698,6 +698,33 @@ ZMQ_EXPORT int zmq_ctx_get_ext (void *context_,
...
+{
+    // Allocate a chunk of memory of size len and return the pointer
+    void*(*allocate_fn) (void *allocator, size_t len);
+
+    // Deallocate the memory chunk pointed to by data_
+    void(*deallocate_fn) (void *allocator, void *data_);
+
+    // Return true if this is an allocator and alive, otherwise false
+    bool(*check_tag_fn) (void *allocator);
```
gummif, line 723: Is this required?

mjvk: When the users wants to provide his/her own allocator instead of using one of the built-ins.

src/basic_concurrent_queue.hpp
```diff
@@ -0,0 +1,71 @@
...
+        if (!_queue.empty ()) {
+            item = _queue.front ();
+            _queue.pop ();
+            success = true;
+        }
+        _queue_mutex->unlock ();
+        return success;
+    }
+
+    size_t size_approx () const { return _queue.size (); }
```
gummif, line 64: This is not thread-safe.

include/zmq.h
gummif: What I mean is specifically the `check_tag_fn` function. I don't see why it is necessary at all. Is it used for trying to catch programming errors at runtime?

src/allocator_global_pool.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return _tag == 0xCAFEEBEC;
+}
+
+void zmq::allocator_global_pool_t::allocate_block (size_t bl)
+{
+    _storage_mutex.lock ();
+    size_t oldSize = _storage.size ();
+    if (oldSize <= bl) {
+        _storage.resize (bl + 1);
+        _free_list.resize (bl + 1);
```
gummif, line 65: Is `allocate_block` supposed to be thread-safe? Because here _free_list is being mutated in a thread-unsafe manner.

src/allocator_global_pool.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+    return _tag == 0xCAFEEBEC;
+}
+
+void zmq::allocator_global_pool_t::allocate_block (size_t bl)
+{
+    _storage_mutex.lock ();
+    size_t oldSize = _storage.size ();
+    if (oldSize <= bl) {
+        _storage.resize (bl + 1);
```
gummif, line 64: While this function is being locked with a mutex, these member variables are accessed in other functions without locks, which is not safe.

include/zmq.h
mjvk: It would allow you to check if a void pointer is a pointer to what you think it is. It was in the previous PR so I left it in. Need to check if it is actually used somewhere.

src/allocator_global_pool.cpp
mjvk: Thanks! Changed a few things around. Should be better now. Still need to check thoroughly.

src/basic_concurrent_queue.hpp
mjvk: Hence the size_approx() instead of size. Might be tricky though not sure what size() returns during  a change. Will see if I can find it somewhere in the standard and otherwise will add a mutex.

include/zmq.h
mjvk: If I don't expose this one users cannot provide their own allocator with the current api. This is a design decision. I would allow users to provide their own allocators (maybe they want fixed bins with new, fixed bins with errors, an off the shelves allocator etc.). But of course that depends on how much you want to expose to the user. What do you think about this?

src/allocator_default.hpp
```diff
@@ -0,0 +1,71 @@
...
+    // allocate() typically gets called by the consumer thread: the user app thread(s)
+    void *allocate (size_t len_);
+
+    void deallocate (void *data_);
+
+    bool check_tag () const;
+
+  private:
+    //  Used to check whether the object is a socket.
+    uint32_t _tag;
```
rfarnham, line 67: I see `_tag` is marked as dead in destructor, but the compiler will optimize that out unless `_tag` is volatile-qualified (which could have performance implications elsewhere).

src/basic_concurrent_queue.hpp
rfarnham: Depending on arch, you could get get a torn-read. Although since it is `size_t`, it should be the platform's native word size and therefore safe-ish.
If `size_approx` needs to be fast, alternatively you could use an `std::atomic` size counter with relaxed memory ordering.

axelriet:
I've looked at this as I, too, believe it's worth doing something as in some scenario (inproc and presumably with other very fast transports) the dominant location where time is spent is the heap. However I think the approach taken here is too complex. All that is needed is a couple of strategically placed hooks for malloc/free, a couple of typedefs and one public function to pass-on two function pointers, then plug a faster heap. I did just that and used Intel's TBB scalable allocator instead of the CRT heap and the speed gains are in the 20% range below the vsm threshold (33 bytes message payload or less) and ranging from 30% to 70% depending on message size after that. In fact it works so well I later made using TBB a build option in itself so TBB can be the new default, or you can supply custom functions in case you think you can do better.

f18m:
> I've looked at this as I, too, believe it's worth doing something as in some scenario (inproc and presumably with other very fast transports) the dominant location where time is spent is the heap. However I think the approach taken here is too complex. All that is needed is a couple of strategically placed hooks for malloc/free, a couple of typedefs and one public function to pass-on two function pointers, then plug a faster heap. I did just that and used Intel's TBB scalable allocator instead of the CRT heap and the speed gains are in the 20% range below the vsm threshold (33 bytes message payload or less) and ranging from 30% to 70% depending on message size after that. In fact it works so well I later made using TBB a build option in itself so TBB can be the new default, or you can supply custom functions in case you think you can do better.

hi @axelriet, do you have any "poc quality code" that you can share around your TBB-based approach?
I would be interested in having a look... (I originally raised https://github.com/zeromq/libzmq/pull/3644 but then exhausted the time I had available to complete the feature... as you said maybe it was a too complex approach after all)

axelriet:
Yep you can check my fork of the project. It’s a bit of a battlefield atm, but it builds on Windows, all tests pass, and it’s a drop-in binary replacement so it should be easy to experiment with. The alloc hooks and defaulting to TBB are two separate things that you can combine or not. If you want to try Intel’s allocator just get the latest oneApi from their website. They done supply a static lib (should be possible to build one, though) so you’ll have to add tbbmalloc.dll to your binaries.

axelriet:
Benefits of a custom allocator for inproc workloads. Here I'm using Intel's Scalable Allocator, part of Intel oneAPI / Threading Building Blocks. There is definitely an advantage.

image

axelriet:
I went ahead in my fork and was able to increase the vsm to 40 bytes (up from 33) by repacking the group_t and msg_t structs and reducing the short group names to 7 char.

Therefore, group names from 8 chars and up switch to long group names now (vs 15 chars and up previously) but the fast vsm performance applies to up to 40 bytes messages, which may be important to some users.

I think 40/7 is a better compromise than the original 33/14 choice as users can trivially shrink group names to 7 chars or less if they want max performances for small messages, but they could not increase the vsm threshold beyond 33 if their data didn't fit under it. Now they have seven more bytes to play with.

PR #3918, opened 2020-05-15
Problem: unfinished message can be leaked by client pipe

bluca:


bluca:
@somdoron does this look right?

somdoron:
Yes

PR #3925, opened 2020-05-18
Problem: Usage of invalidated iterator of _timers container in zmq::p…

### LINKED ISSUE #3645: 4.3.2: Crashing use-after-free in poller_base_t::execute_timers
# Issue description

While `poller_base_t::execute_timers` is walking the collection of timers that are expiring / have already expired, a different thread cancels the specific timer whose timer_event method is about to be called.  In my environment, the heap functions including new/delete will fill blocks with 0x5A when they are freed.

The crash happens at line 103 of poller_base.cpp:

```
   it->second.sink->timer_event (it->second.id);
```

when the code attempts to dereference `it->second.sink`.

# Environment

* libzmq version (commit hash if unreleased): 4.3.2
* ***The issue did not occur with 4.2.3.***
* OS: FreeBSD 11.2 / amd64
* Multicore CPU

Calling language is C++ via cppzmq.  I only have one thread, so there are no issues surrounding my code using libzmq objects from more than one thread.  At the moment of the crash, there are three threads:
* A libzmq I/O thread
* My main worker thread, executing my code via boost::asio
* A third thread, also in libzmq, that I'd guess is the reaper.

I don't have a clear picture of what exactly causes this.  I can't tell what kind of object was just removed from the timer multimap because the iterator points to deleted memory (see note above about deleted blocks being filled with 0x5A).  My best guess is a PUB or SUB socket that's being disconnected or unbound from my code at the same moment as the TCP connection to its peer is about to be reconnected, and the poller_base_t is about to dispatch to the timer_event handler.

At certain times, my program will find that its ZeroMQ sockets are disconnected by network events at about the same time that those network events cause a notification to my program to change which endpoints it uses (unbind-then-bind on one socket, disconnect-then-connect on another).

There are also zmq IPC sockets, but they are only rarely changed while the program is running, whereas the network events mentioned above are expected to happen (even if we hope they don't), and they *will* cause the disconnect/connect and unbind/bind sequences.

```
Thread 3 (process 100288):  ***Reaper?***
#0  _kevent () at _kevent.S:3
#1  0x000000080427ccc2 in __thr_kevent (kq=13, changelist=0x0, nchanges=0, eventlist=0x7fffdfffbf30, nevents=256, timeout=0x0)
    at /home/build/snsbsd/lib/libthr/thread/thr_syscalls.c:398
#2  0x00000008025fabc3 in zmq::kqueue_t::loop (this=0x805e18500) at kqueue.cpp:190
#3  0x00000008026283df in thread_routine (arg_=0x805e18540) at thread.cpp:225
#4  0x0000000804279c06 in thread_start (curthread=0x805e10200) at /home/build/snsbsd/lib/libthr/thread/thr_create.c:289
#5  0x0000000000000000 in ?? ()
Current language:  auto; currently asm

Thread 2 (process 100282):  ***My worker thread***
#0  _kevent () at _kevent.S:3
#1  0x000000080427ccc2 in __thr_kevent (kq=3, changelist=0x0, nchanges=0, eventlist=0x7fffffffd370, nevents=128, timeout=0x7fffffffe780)
    at /home/build/snsbsd/lib/libthr/thread/thr_syscalls.c:398
#2  0x0000000800ac0b27 in boost::asio::detail::kqueue_reactor::run () from /usr/Firewall/lib/libnfw++.so
#3  0x0000000800ac056d in boost::asio::detail::scheduler::do_run_one () from /usr/Firewall/lib/libnfw++.so
#4  0x0000000800ac01b2 in boost::asio::detail::scheduler::run () from /usr/Firewall/lib/libnfw++.so
#5  0x0000000800abb913 in sns::DaemonBase::run () from /usr/Firewall/lib/libnfw++.so
#6  0x00000000004097e8 in main ()

Thread 1 (process 100289):  *** Crashed thread***
#0  zmq::poller_base_t::execute_timers (this=0x805e18700) at poller_base.cpp:103
#1  0x00000008025fab58 in zmq::kqueue_t::loop (this=0x805e18700) at kqueue.cpp:177
#2  0x00000008026283df in thread_routine (arg_=0x805e18740) at thread.cpp:225
#3  0x0000000804279c06 in thread_start (curthread=0x805e10e00) at /home/build/snsbsd/lib/libthr/thread/thr_create.c:289
#4  0x0000000000000000 in ?? ()
```

# What's the expected result?

Does not crash.

bjovke:
…oller_base_t::execute_timers ().

Solution: Safe iteration through _timers container, valid only for std::multimap (currently it is).
One minor fix for C++11 features and MSVC added.

Fixes #3645.

### DIFF
--- a/src/poller_base.cpp
+++ b/src/poller_base.cpp
@@ -68,8 +68,14 @@ void zmq::poller_base_t::cancel_timer (i_poll_events *sink_, int id_)
             return;
         }
 
-    //  Timer not found.
-    zmq_assert (false);
+    //  We should generally never get here. Calling 'cancel_timer ()' on
+    //  an already expired or canceled timer (or even worse - on a timer which
+    //  never existed, supplying bad sink_ and/or id_ values) does not make any
+    //  sense.
+    //  But in some edge cases this might happen. As described in issue #3645
+    //  `timer_event ()` call from `execute_timers ()` might call `cancel_timer ()`
+    //  on already canceled (deleted) timer.
+    //  As soon as that is resolved an 'assert (false)' should be put here.
 }
 
 uint64_t zmq::poller_base_t::execute_timers ()
@@ -81,26 +87,31 @@ uint64_t zmq::poller_base_t::execute_timers ()
     //  Get the current time.
     const uint64_t current = _clock.now_ms ();
 
-    //   Execute the timers that are already due.
-    const timers_t::iterator begin = _timers.begin ();
-    const timers_t::iterator end = _timers.end ();
+    //  Execute the timers that are already due.
     uint64_t res = 0;
-    timers_t::iterator it = begin;
-    for (; it != end; ++it) {
-        //  If we have to wait to execute the item, same will be true about
-        //  all the following items (multimap is sorted). Thus we can stop
-        //  checking the subsequent timers.
+    timer_info_t timer_temp;
+    timers_t::iterator it;
+
+    do {
+        it = _timers.begin ();
+
+        //  If we have to wait to execute the item, same will be true for
+        //  all the following items because multimap is sorted. Thus we can
+        //  stop checking the subsequent timers.
         if (it->first > current) {
             res = it->first - current;
             break;
         }
 
-        //  Trigger the timer.
-        it->second.sink->timer_event (it->second.id);
-    }
+        //  Save and remove the timer because timer_event() call might delete
+        //  exactly this timer and then the iterator will be invalid.
+        timer_temp = it->second;
+        _timers.erase (it);
 
-    //  Remove them from the list of active timers.
-    _timers.erase (begin, it);
+        //  Trigger the timer.
+        timer_temp.sink->timer_event (timer_temp.id);
+
+    } while (!_timers.empty ());
 
     //  Return the time to wait for the next timer (at least 1ms), or 0, if
     //  there are no more timers.


src/poller_base.cpp
```diff
@@ -55,7 +55,11 @@ void zmq::poller_base_t::add_timer (int timeout_, i_poll_events *sink_, int id_)
 {
     uint64_t expiration = _clock.now_ms () + timeout_;
     timer_info_t info = {sink_, id_};
+#if __cplusplus >= 201103L || (defined _MSC_VER && _MSC_VER >= 1700)
+    _timers.emplace (expiration, info);
```
sigiesec, line 59: If touching this, it should probably use `ZMQ_MAP_INSERT_OR_EMPLACE`.

IIUC, this is not related to the other changes in this commit, right?

src/poller_base.cpp
```diff
@@ -67,9 +71,6 @@ void zmq::poller_base_t::cancel_timer (i_poll_events *sink_, int id_)
             _timers.erase (it);
             return;
         }
-
-    //  Timer not found.
```
sigiesec, line 71: It's not that nice that this consistency check is removed.

src/poller_base.cpp
```diff
@@ -81,12 +82,13 @@ uint64_t zmq::poller_base_t::execute_timers ()
...
 
-    //   Execute the timers that are already due.
-    const timers_t::iterator begin = _timers.begin ();
-    const timers_t::iterator end = _timers.end ();
+    //  Execute the timers that are already due.
     uint64_t res = 0;
-    timers_t::iterator it = begin;
-    for (; it != end; ++it) {
+    timer_info_t timer_temp;
+    const timers_t::iterator end = _timers.end ();
```
sigiesec, line 88: Strictly speaking, this iterator might also be invalidated when `_timers` is modified. (This might not be the case for the implementations used)

src/poller_base.cpp
```diff
@@ -81,12 +82,13 @@ uint64_t zmq::poller_base_t::execute_timers ()
...
+
+    for (timers_t::iterator it = _timers.begin (); it != end;
```
sigiesec, line 90: The loop looks very weird now. Actually, there is no real loop variable anymore.

If reentrancy is really required, then I would suggest changing to a while loop: `while (!_timers.empty())`, and then just use `_timers.begin()` within the loop.

Another alternative might be to copy or swap `_timers` temporarily, but then handling `cancel_timer` gets more complicated.

sigiesec:
> 
> 
> If reentrancy is required here, then such changes must be made.
> 
> I am curious though where this is actually required. Do you have a stack trace where `cancel_timer` is executed while running `execute_timers`?

Ah, I see you mentioned the situation in https://github.com/zeromq/libzmq/issues/3645#issuecomment-622613741

src/poller_base.cpp
bjovke: Hi. Sorry for the late reply. I saw this construct elsewhere in the code. I wasn't aware of `ZMQ_MAP_INSERT_OR_EMPLACE`. 
I see it's defined only in `blob.hpp`. Including this file might have side effects I'm not aware of.
Do you think it should be included?

Yes, this is not related to the actual issue.

src/poller_base.cpp
bjovke: This code can now be reached because in `execute_timers ()` timer is deleted before `timer_event ()` is executed and `timer_event ()` might call `cancel_timer ()` on already deleted one.

It's very difficult for me to trace through all possible calls to be 100% sure but the issue itself indicates that this might happen.

src/poller_base.cpp
bjovke: `std::multimap` guarantees that only iterators of deleted elements are invalidated and others stay valid - [https://en.cppreference.com/w/cpp/container/multimap/erase](https://en.cppreference.com/w/cpp/container/multimap/erase).
I don't think that underlying container can be changed because the `execute_timers ()` code relies on `std::multimap` properties.
However, the code can be changed to "refresh" the `end ()` iterator on every iteration, there will be some overhead.

src/poller_base.cpp
bjovke: It doesn't stop when all timers are deleted, it stops earlier if there are no more due timers (`break`).

`for` loop is the right construct for this: it sets an initial value, condition to exit the loop and what happens in each iteration, all in one place. It doesn't strictly need to be an increment/decrement operator.

But if you feel that `while` is more suitable in this case I'll change it to `while`.

bjovke:
> 
> 
> If reentrancy is required here, then such changes must be made.
> 
> I am curious though where this is actually required. Do you have a stack trace where `cancel_timer` is executed while running `execute_timers`?

Unfortunately not.
Maybe the people who reported the issue can provide more information?
But it seems that there is a code path for this:
1. `zmq::poller_base_t::execute_timers ()`
2. `zmq::stream_engine_base_t::timer_event ()`
3. `zmq::stream_engine_base_t::error ()`
4. `zmq::session_base_t::engine_error ()`
5. `zmq::session_base_t::reconnect ()`
6. `zmq::io_object_t::cancel_timer ()`
7. `zmq::poller_base_t::cancel_timer ()`

src/poller_base.cpp
bjovke: Actually you're right, it can be `while (!_timers.empty())`. It's just a matter of preference.

sigiesec: I agree that a `for` loop doesn't necessarily need to be restricted to decrement/increment in the iteration expression, but having an iteration expression that looks like the init expression is stretching that somewhat too far.

If you have strong feeling about keeping `for`, keep it, but I just think `while (!_timers.empty())` is much clearer here.

src/poller_base.cpp
sigiesec: Oh, sorry, then I mixed that up with other containers. Thanks for checking it again. Then I also think it should be kept this way.

src/poller_base.cpp
sigiesec: `ZMQ_MAP_INSERT_OR_EMPLACE` should probably be moved to a different header. Maybe macros.hpp?

Can you remove this change from this commit to avoid confusion, given that the other changes are quite critical? Then we can improve this separately.

src/poller_base.cpp
sigiesec: Yes, I understand why it can't be kept given the suggested changes in `execute_timers`. I just wanted to reinforce this side effect. As I said, I can imagine that other alternatives might be even worse, if it's not possible to eliminate this reentrant call altogether. And I understand that the situation leading to this is hard to analyze.

Maybe leave a comment here that "usually" this condition can be asserted except for the reentrant case? If we understand the situation better in the future, this maybe could be changed back again.

src/poller_base.cpp
bjovke: Sorry for the slow replies, I'm pretty busy these days.

I think you're right about this.
That would eliminate the usage of `end ()` iterator and make things simpler.
I'll make changes for this.

bjovke: I've changed it to `while` now.
Actually I've changed it to `do ... while (!_timers.empty())` because at the beginning of `execute_timers ()` we already have this check once:
```
    //  Fast track.
    if (_timers.empty ())
        return 0;
```

src/poller_base.cpp
bjovke: Yes, you're right. I'll add the comment.

bjovke: I've taken another look at the code.
It seems to me now that the assert should stay.
It doesn't make sense to call `cancel_timer ()` from `timer_event ()` handler on the timer that triggered the `timer_event ()`.
Also, the proposed fix from the reported issue did not remove the assert and one of the people who tested it did not run into this assert at any moment.
So I'll put back the assert as it was.

src/poller_base.cpp
bjovke: Yes, I'll remove it. Doesn't functionally change anything.

bjovke: Done.

src/poller_base.cpp
bjovke: Reverted to as it was.

bjovke:
> 
> 
> > If reentrancy is required here, then such changes must be made.
> > I am curious though where this is actually required. Do you have a stack trace where `cancel_timer` is executed while running `execute_timers`?
> 
> Unfortunately not.
> Maybe the people who reported the issue can provide more information?
> But it seems that there is a code path for this:
> 
>     1. `zmq::poller_base_t::execute_timers ()`
> 
>     2. `zmq::stream_engine_base_t::timer_event ()`
> 
>     3. `zmq::stream_engine_base_t::error ()`
> 
>     4. `zmq::session_base_t::engine_error ()`
> 
>     5. `zmq::session_base_t::reconnect ()`
> 
>     6. `zmq::io_object_t::cancel_timer ()`
> 
>     7. `zmq::poller_base_t::cancel_timer ()`

I've taken a more detailed look into the code.
It really makes sense to call `cancel_timer ()` from `timer_event ()` handler. Sometimes two timers are started and if one expires the other one is not needed any more and needs to be canceled.
This means that it makes sense to cancel **only** the timer which is not yet expired, never the already expired one (for which `timer_event ()` was already called) or the one for which `timer_event ()` is being called.

src/poller_base.cpp
bjovke: Hm... This means then that the call to `cancel_timer ()` from `timer_event ()` for the timer on which it is being called will fail this assert.
So in that case we don't need to delete the iterator before the `timer_event ()` call nor make a copy of it's contents.
That makes the code simpler.

bjovke: But then we get to the almost the same code we started with, for which the issue was reported.
I'll put a comment then instead of the assert.

bjovke: OK, I've removed the assert again and put the comment.

sigiesec: Indeed, if the assert wouldn't be hit after the changes to `execute_timers`, then I'd understand the whole problem even less. I still hope to get more insight in the future :)

sigiesec: Thanks for the great comment(s)!

PR #3931, opened 2020-05-24
added Peer Address metadata to UDP messages

wzli:
This is something that I needed for my app. Dunno if it's the best way to do it, but thought I'd share. Please advise if other places need to be changed.

### DIFF
--- a/src/udp_engine.cpp
+++ b/src/udp_engine.cpp
@@ -601,6 +601,10 @@ void zmq::udp_engine_t::in_event ()
     errno_assert (rc == 0);
     memcpy (msg.data (), _in_buffer + body_offset, body_size);
 
+    std::string address (inet_ntoa (reinterpret_cast<sockaddr_in*> (&in_address)->sin_addr));
+    metadata_t::dict_t dict {{ZMQ_MSG_PROPERTY_PEER_ADDRESS, address}};
+    msg.set_metadata(new (std::nothrow) metadata_t (dict));
+
     // Push message body to session
     rc = _session->push_msg (&msg);
     // Message body doesn't fit in the pipe, drop and reset session state


bluca:
This is wrong. We do not expose addresses like this. It leads to broken implementations. Also it breaks IPv6. Read the guide: http://zguide.zeromq.org/page:all

wzli:
> This is wrong. We do not expose addresses like this. It leads to broken implementations. Also it breaks IPv6. Read the guide: http://zguide.zeromq.org/page:all

Can you please explain in more detail? (for my understanding, since I'm not an expert). Why exactly does it break implemetations, why exactly does it break IPV6. What parts of the guide specifically relavent? What's the proper way to retrieve the source IP address from UDP messages?

bluca:
https://linux.die.net/man/3/inet_ntoa

> The inet_ntoa() function converts the Internet host address in, given in network byte order, to a string in **IPv4** dotted-decimal notation.

wzli:
> https://linux.die.net/man/3/inet_ntoa
> 
> > The inet_ntoa() function converts the Internet host address in, given in network byte order, to a string in **IPv4** dotted-decimal notation.

hmm, okay, well it was take for 

> https://linux.die.net/man/3/inet_ntoa
> 
> > The inet_ntoa() function converts the Internet host address in, given in network byte order, to a string in **IPv4** dotted-decimal notation.

hmm, well it was mostly the same code being called in sockaddr_to_msg. But now I see there's an assert for IPV4 before entering, which only occurs in if it's a raw socket. How come raw sockets are IPV4 only but not others?

Is this the only issue? I mean, I'm sure there's something that can handle IPV6 as well. Assuming that was address, what are the other problems?

```
        zmq_assert (in_address.ss_family == AF_INET);
        sockaddr_to_msg (&msg, reinterpret_cast<sockaddr_in *> (&in_address));
```

PR #3932, opened 2020-05-24
Initialize some CMake variables as empty before build

serg06:
This prevents a bug where:
- Parent has a `sources` variable
- Parent does `add_subdirectory("libzmq")`
- LibZMQ does `list(APPEND sources ...)` and proceeds to use that list even though it has some of the parent's value inside of it


bluca:
Please add a commit with a relicense grant

serg06:
> Please add a commit with a relicense grant

Thanks, done

PR #3924, opened 2020-05-17
Small fix for combining flags

Jasper-Bekkers:
Proposed fix for #3923

Issue #3923, opened 2020-05-17
This line doesn't seem correct


Jasper-Bekkers:
I'm pretty sure this should be `|` instead of `||`, though I have no idea what the implications would be. I'm guessing this mostly works by accident because WSA_FLAG_OVERLAPPED is 0x1 but it means WSA_FLAG_NO_HANDLE_INHERIT never gets set.

https://github.com/zeromq/libzmq/blob/e7f802d1ac7d61c89912ab261a19967b7d65b411/src/ip.cpp#L107

bjovke:
Thanks @Jasper-Bekkers , I'll make a pull request.

bjovke:
Resolved by #3943

PR #3943, opened 2020-05-30
Problem: The code is wrong, socket option is never set.

bjovke:
Solution: Incorrect operator used. Correct the operator.


bjovke:
#3923

bjovke:
#3924

PR #3951, opened 2020-06-06
Adds NORM support to cmake

mjvk:
NORM was not yet supported in cmake. This adds that support.


bluca:
You need also to add a cmakedefine to https://github.com/zeromq/libzmq/blob/master/builds/cmake/platform.hpp.in

mjvk:
Will do! Could you explain to me why? The add_definition already outs it everywhere right?

bluca:
I'm not sure, all other defines are there too

mjvk:
I think it helps when recompiling. A define passed impacts all code. On in platform.hpp only the parts that include platform.hpp.

PR #3959, opened 2020-06-16
Problem: mtrie use of non-tail recursion leads to stack overflow 

bluca:


bluca:
@somdoron could you have a look when you got a second, please

somdoron:
Looks good, by the way, it seems the add was tail recursive, was it a problem as well?

bluca:
> Looks good, by the way, it seems the add was tail recursive, was it a problem as well?

It is in case the compiler doesn't optimize it away, so we can't be 100% sure on our end, so I changed it too

PR #3960, opened 2020-06-17
rebase reconnect-redux on master

bill-torpey:
Note: this is an update of https://github.com/nyfix/libzmq/pull/1 

# Connect/reconnect with 0mq

The purpose of this PR is to close the loop on connect/reconnect options that began with <https://github.com/zeromq/libzmq/pull/3831>.

We want to take advantage of 0mq's support for automatic reconnection of sockets that become disconnected, but there are some edge cases that are problematic for us.

Because of the way [OZ](https://github.com/nyfix/OZ/) does [dynamic discovery](https://github.com/nyfix/OZ/blob/master/doc/Naming-Service.md), processes don't bind to known ports -- (almost) all ports are assigned dynamically.  These port assignments are then published to the one-and-only known port, which then forwards them to all the other processes.

With automatic reconnection enabled, if a process terminates and restarts (e.g., because of a crash), 0mq will continue to try to reconnect to its original port.  

The scenarios we care about are:

1. Process terminates, peer attempts to reconnect but gets ECONNREFUSED.  This is already handled by the earlier PR.
2. Process mistakenly attempts to connect to a port that is not a 0mq process:
  1.  The non-0mq process unilaterally closes the connection, presumably because the 0mq process did not adhere to the protocol expected by the non-0mq process.
  2.  The non-0mq process does nothing -- i.e., it does not respond to the 0mq greeting.
  3.  The non-0mq process does respond in some fashion, or at least sends data on the connection that `receive_greeting` can read.

## 1. No process listening at port

This is handled by the ZMQ_RECONNECT_STOP/ZMQ_RECONNECT_STOP_CONN_REFUSED sockopt.

```
#0  zmq::socket_base_t::event_closed (this=0x614590, endpoint_uri_pair_=..., fd_=16) at /home/btorpey/work/libzmq/master/src/src/socket_base.cpp:1866
#1  0x00007ffff7b8f3cc in zmq::stream_connecter_base_t::close (this=0x7fffe80008c0) at /home/btorpey/work/libzmq/master/src/src/stream_connecter_base.cpp:158
#2  0x00007ffff7b68f90 in zmq::tcp_connecter_t::out_event (this=0x7fffe80008c0) at /home/btorpey/work/libzmq/master/src/src/tcp_connecter.cpp:110
#3  0x00007ffff7b8f438 in zmq::stream_connecter_base_t::in_event (this=0x7fffe80008c0) at /home/btorpey/work/libzmq/master/src/src/stream_connecter_base.cpp:168
#4  0x00007ffff7b266a0 in zmq::epoll_t::loop (this=0x60a170) at /home/btorpey/work/libzmq/master/src/src/epoll.cpp:198
#5  0x00007ffff7b42629 in zmq::worker_poller_base_t::worker_routine (arg_=0x60a170) at /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:135
#6  0x00007ffff7b6a3ba in thread_routine (arg_=0x60a1c8) at /home/btorpey/work/libzmq/master/src/src/thread.cpp:250
#7  0x00007ffff7635e65 in start_thread () from /lib64/libpthread.so.0
#8  0x00007ffff6b3f88d in clone () from /lib64/libc.so.6
```

## 2. Non-0mq process listening at port
These all rely on the 0mq handshake failing.  There are a few ways that can happen.

### 2.1 Peer closes connection
In some cases, the non-0mq process will unilaterally close the connection, presumably because the greeting from 0mq process violates the non-0mq process's expected behavior.  This can either be done gracefully (Oracle listener sends a `FIN`) or forcibly (smbd sends a `RST`) -- in either case it is detected in `receive_greeting` with errno == `ECONNRESET`.

### 2.2 Peer does nothing
Some processes don't respond at all to the 0mq greeting.  In this case, the handshake times out (in `stream_engine_base_t::timer_event`)
and triggers the disconnect:

```
#0  zmq::socket_base_t::event_handshake_failed_no_detail (this=0x614590, endpoint_uri_pair_=..., err_=11)
    at /home/btorpey/work/libzmq/master/src/src/socket_base.cpp:1887
#1  0x00007ffff7b66379 in zmq::stream_engine_base_t::error (this=0x7fffe8000f80, reason_=zmq::i_engine::timeout_error)
    at /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:689
#2  0x00007ffff7b66818 in zmq::stream_engine_base_t::timer_event (this=0x7fffe8000f80, id_=64)
    at /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:737
#3  0x00007ffff7b42473 in zmq::poller_base_t::execute_timers (this=0x60a170) at /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:99
#4  0x00007ffff7b2653f in zmq::epoll_t::loop (this=0x60a170) at /home/btorpey/work/libzmq/master/src/src/epoll.cpp:173
#5  0x00007ffff7b42629 in zmq::worker_poller_base_t::worker_routine (arg_=0x60a170) at /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:135
#6  0x00007ffff7b6a3be in thread_routine (arg_=0x60a1c8) at /home/btorpey/work/libzmq/master/src/src/thread.cpp:250
#7  0x00007ffff7635e65 in start_thread () from /lib64/libpthread.so.0
#8  0x00007ffff6b3f88d in clone () from /lib64/libc.so.6
```

This requires the fix described [below](#Possible-bug).

### 2.3 Peer responds to greeting

Some processes do respond to the 0mq greeting. (For instance, sshd responds to the greeting with "SSH-2.0-Open", which just happens to be the same length as the unversioned 0mq greeting).

#### Specifying ZMQ_MIN_ZMTP_VERSION
One way to detect/reject non-0mq processes is to specify a minimum ZMTP version (using the `ZMQ_MIN_ZMTP_VERSION` sockopt).  Since the greeting isn't a proper ZMTP greeting, the code treats it as "unversioned", and as such it doesn't match any ZMTP version.

```
#0  zmq::socket_base_t::event_disconnected (this=0x614590, endpoint_uri_pair_=..., fd_=16) at /home/btorpey/work/libzmq/master/src/src/socket_base.cpp:1880
#1  0x00007ffff7b66394 in zmq::stream_engine_base_t::error (this=0x7fffe8000f80, reason_=zmq::i_engine::protocol_error) at /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:694
#2  0x00007ffff7b8ccb5 in zmq::zmtp_engine_t::handshake_v1_0_unversioned (this=0x7fffe8000f80) at /home/btorpey/work/libzmq/master/src/src/zmtp_engine.cpp:260
#3  0x00007ffff7b8c668 in zmq::zmtp_engine_t::handshake (this=0x7fffe8000f80) at /home/btorpey/work/libzmq/master/src/src/zmtp_engine.cpp:133
#4  0x00007ffff7b64492 in zmq::stream_engine_base_t::in_event_internal (this=0x7fffe8000f80) at /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:252
#5  0x00007ffff7b643f2 in zmq::stream_engine_base_t::in_event (this=0x7fffe8000f80) at /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:242
#6  0x00007ffff7b26745 in zmq::epoll_t::loop (this=0x60a170) at /home/btorpey/work/libzmq/master/src/src/epoll.cpp:206
#7  0x00007ffff7b42629 in zmq::worker_poller_base_t::worker_routine (arg_=0x60a170) at /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:135
#8  0x00007ffff7b6a3ba in thread_routine (arg_=0x60a1c8) at /home/btorpey/work/libzmq/master/src/src/thread.cpp:250
#9  0x00007ffff7635e65 in start_thread () from /lib64/libpthread.so.0
#10 0x00007ffff6b3f88d in clone () from /lib64/libc.so.6
```

#### Without ZMQ_MIN_ZMTP_VERSION
Last but not least is the case where a process responds to the 0mq greeting, but for whatever reason we don't want to use `ZMQ_MIN_ZMTP_VERSION`.  This is similar to the case 2.2 above, and requires the fix mentioned [below](#Possible-bug). 

## Possible bug
In this case, it looks like there's a bug in the code -- the following sequence should move from:

```
bool zmq::zmtp_engine_t::handshake ()
{
    ...
    if (_has_handshake_timer) {
        cancel_timer (handshake_timer_id);
        _has_handshake_timer = false;
    }

    return true;
}
```

to... 

```
void zmq::stream_engine_base_t::mechanism_ready ()
{
    ...
    if (_has_handshake_timer) {
        cancel_timer (handshake_timer_id);
        _has_handshake_timer = false;
    }

    _socket->event_handshake_succeeded (_endpoint_uri_pair, 0);
}
```

Testing appears to show that handshakes will never timeout with the current version of the code.

## Other Stuff

The doc for `ZMQ_EVENT_DISCONNECTED` states:

> The socket was disconnected unexpectedly. The event value is the FD of the underlying network socket. **Warning: this socket will be closed.**

That doesn't seem to happen, and I'm curious why?

## Test Code
Code used to run these tests is at: <https://github.com/WallStProg/zmqtests/tree/master/connect>.


bluca:
Sorry but I seriously dislike the idea of exposing the wire protocol versions as options in the library API. The only reason to refuse handshaking with other versions is if incompatible features are requested - eg, CURVE, or a new socket type.
If all you care about is ensuring only actual peers are talked to, then we are working and have already done some hardening of the handshaking for unrelated reasons.

It really sounds like what you want is ZAP, and are re-implementing it sideways? Have you considered just using that?

somdoron:
@bluca I don't completely agree here, the problem with unversioned version of ZMTP and ZMTP v1 prevents recognizing and disconnecting none zeromq peers. The MIN_ZMTP_VERSION is just a means to block those versions of ZMTP. Maybe we can change the option to ZMQ_BLOCK_OBSOLETE_ZMTP_VERSIONS?

somdoron:
Anyway I think we should deprecate old versions of ZMTP (unversioned and version 1) in next release and remove them in a year?

bluca:
> Anyway I think we should deprecate old versions of ZMTP (unversioned and version 1) in next release and remove them in a year?

You know how it works with backward compatibility - there will be people running libzmq 2.0 for the next 30 years at a minimum :-)
We can change the option to be a strict zmtp enforcement, disabled by default

somdoron:
@bill-torpey what do you think about changing ZMQ_ZMTP_MIN_VERSION to ZMQ_ZMTP_STRICT which will block unversioned and v1?

bill-torpey:
>@bill-torpey what do you think about changing ZMQ_ZMTP_MIN_VERSION to ZMQ_ZMTP_STRICT which will block unversioned and v1?

I think that works (although I'm not sure what the problem is with the approach we came up with -- I'll let you and @bluca debate that, you both know way more than I do about "the 0mq way"). 

As you know, our main concern has to do with using ephemeral ports in [OZ](https://github.com/nyfix/OZ), and what happens when a previously-bound peer terminates, either gracefully or as a result of a crash.

With automatic reconnection enabled (which is pretty much a must-have in our environment -- we have WAN links in parts of the world where unreliable networks are common), 0mq would keep trying *forever* to reconnect to that port.

The change in https://github.com/zeromq/libzmq/pull/3831 mitigates the most common scenario -- where subsequent reconnects fail because there is no longer any code listening at that port.  So far, so good.

A less-likely scenario is where the OS reassigns that ephemeral port to a random process, and our 0mq code keeps trying (and failing) to connect to that port.  This is the scenario that we're trying to address with this PR.  (That, and the possibility that some mis-configuration causes our code to attempt to connect to a "wrong" port at startup).

Presumably, if only unversioned and v1 protocols are rejected, we would want to also implement some kind of security mechanism to handle the case where our code mistakenly attempts to (re-)connect to a port where a non-OZ 0mq process is running.  

Nevertheless, the ability to reject unversioned and v1 protocols should address most of the "pathological" connections.  (For instance, in our tests 0mq connects quite happily to `sshd` ;-)

So, short version: whatever you and @bluca decide about ZMTP versions should be fine -- **the main problem is making sure that non-0mq proccesses fail the handshake**. 

 Thanks for the help!

bill-torpey:
So, if it would help move things along, I'm happy to remove the MIN_ZMTP_VERSION bits.  Those are really more "belts & braces" -- it's the other changes that really accomplish the major part of what we care about.

Please let me know ...

bluca:
ZMQ_ZMTP_STRICT sounds fine

doc/zmq_setsockopt.txt
```diff
@@ -523,7 +523,7 @@ NOTE: in DRAFT state, not yet available in stable releases.
 
 [horizontal]
 Option value type:: char[2]
-Option value unit:: 0, MIN_ZMTP_VERSION_1_0, MIN_ZMTP_VERSION_2_0, MIN_ZMTP_VERSION_3_0, MIN_ZMTP_VERSION_3_1
+Option value unit:: 0, ZMQ_ZMTP_STRICT_V1_0, ZMQ_ZMTP_STRICT_V2_0, ZMQ_ZMTP_STRICT_V3_0, ZMQ_ZMTP_STRICT_V3_1
```
bluca, line 526: What I meant with this was not just to rename the option, but to take away entirely the ability to specify a version - I really don't want to have to start debugging people's setting incompatible versions across nodes who then open bugs about it.
Just make this a boolean please, and if set to true, reject v1 which doesn't have a signature.

bill-torpey: I'm just going to remove the whole thing -- it's not critical for what I need to do, and is just complicating things.  This should probably have been its own PR in the first place, so that's on me.

bill-torpey:
Removed the ZMTP checks -- if desired, that can be resubmitted as a stand-alone patch.

Hopefully the remaining changes are acceptable as-is, but please let me know if addl. changes are needed.  Thanks.

bluca:
> Removed the ZMTP checks -- if desired, that can be resubmitted as a stand-alone patch.
> 
> Hopefully the remaining changes are acceptable as-is, but please let me know if addl. changes are needed. Thanks.

thanks - the changes are fine, but the builds are either failing to compile or the reconnect test is failing

src/stream_engine_base.cpp
```diff
@@ -689,6 +694,15 @@ void zmq::stream_engine_base_t::error (error_reason_t reason_)
             || _mechanism->status () == mechanism_t::handshaking)) {
         const int err = errno;
         _socket->event_handshake_failed_no_detail (_endpoint_uri_pair, err);
+#ifdef ZMQ_BUILD_DRAFT_API
```
bluca, line 697: It's not the ifdef that's missing - the new option needs to be double-defined in src/zmq_drafts.h

bill-torpey: Thanks!

I  *think* I get it:  zmq_draft.h is there to #define constants that would otherwise be undefined in a non-draft build.  Doing that helps to minimize the #ifdef's scattered through the code.  Curious if that's documented anywhere -- I wasn't able to find it .

bluca: Yes that's it - I don't it we've written it anywhere

bluca:
Uh I see a bunch of source files being marked as executable?

bill-torpey:
> Uh I see a bunch of source files being marked as executable?

Should be OK now -- my samba was messing with permissions, think I finally have it sorted.

somdoron:
Nice

somdoron:
I think ZMQ_ZMTP_STRICT is important so I would like to see it as a separate patch.

bill-torpey:
> @somdoron I think ZMQ_ZMTP_STRICT is important so I would like to see it as a separate patch.

I'm willing to take crack at this if you like, but want to be clear what is being discussed:

- iff ZMQ_ZMTP_STRICT sockopt == 1, fail handshake iff peer is unversioned, like so:

```
bool zmq::zmtp_engine_t::handshake_v1_0_unversioned ()
{
    if (_options.zmtp_strict == 1) {
       // reject unversioned connections if ZMQ_ZMTP_STRICT specified
       error (protocol_error);
       return false;
    }
   ...
```

Is that what you're suggesting?

bluca:
That looks fine to me

somdoron:
Looks good to me as well.

Issue #3970, opened 2020-06-28
Invalid argument (../libzmq/src/condition_variable.hpp:308)


wzli:
# Issue description
Rare and sporadic crashing with
Invalid argument (../libzmq/src/condition_variable.hpp:308)

I've encountered this problem twice, once on d0d23446f5797364c885c7fdc982e5a1efe616fa, and once on v4.3.2 (hoping it would be more stable).

# Environment
Ubuntu 18.04 docker container

* libzmq version: 
d0d23446f5797364c885c7fdc982e5a1efe616fa
and v4.3.2

# Minimal test code / Steps to reproduce the issue
It happens rarely, haven't found out how to reliably reproduce. Maybe correlated high processing load, but here's my setup.
A P2P mesh with 32 nodes running inside different docker containers on the same host, talking over the virtual bridge network. Each node binds a dish socket and transmits over radio socket using UDP. On average each node transmits ~8kB messages to 4-6 neighbors at around 30Hz. I encountered the problem on some node after letting the system run for a while.

# What's the actual result? (include assertion message & call stack if applicable)

Invalid argument (../libzmq/src/condition_variable.hpp:308)
Aborted (core dumped)
exited with code 134

Unfortunately, I couldn't retrieve core dump since the docker container shutdown right after the crash. Will will keep the container running next time if I catch it again.

# What's the expected result?
No crash and continued operation.

wzli:
Unexpected return value from a wait condition. Might be some sort of system out of resource error. But still, I wouldn't want the application to crash in deployment. Is it safe to compile without asserts?

bluca:
> Is it safe to compile without asserts?

No, it is not.

The assert message shows your build is using the pthread implementation, might be worth rebuilding with a c++11 compiler so that the STL implementation is used? You can also force it with ```--with-cv-impl=stl11```

@somdoron this comes from the mailbox_safe:

https://github.com/zeromq/libzmq/blob/master/src/mailbox_safe.cpp#L109

> EINVAL
> Different mutexes were supplied for concurrent pthread_cond_timedwait() or pthread_cond_wait() operations on the same condition variable.

https://linux.die.net/man/3/pthread_cond_wait

As far as I can see, there's only one mutex (from the socket_base) being used with a safe mailbox. Any idea?

somdoron:
Only thing that comes to mind is trying closing thread safe socket while another thread tries to send or receive. Which is not supported.

wzli:
[...]

PR #3974, opened 2020-06-28
problem: pthread_cond_timedwait sometimes fail with EINVAL

somdoron:
Solution: fix rounding error and assert if the clock_gettime fails


somdoron:
relates to #3970

PR #3977, opened 2020-06-30
implementation of ZMTP_STRICT

bill-torpey:
As per https://github.com/zeromq/libzmq/pull/3960#issuecomment-650708791 this is an attempt to implement ZMTP_STRICT.

I believe the code is OK, but I'm having trouble coming up with a test program -- the problem appears to be how to "fake" an unversioned greeting.  

The current test results look like this:

```
[btorpey@bt-brix build]$ bin/test_zmtp_strict
Got event: CONNECT_DELAYED
Got event: CONNECTED
Got event: HANDSHAKE_SUCCEEDED
/home/btorpey/work/libzmq/master/src/tests/test_zmtp_strict.cpp:234:connect_success_versioned:PASS
Got event: CONNECT_DELAYED
Got event: CONNECTED
Got event: DISCONNECTED
/home/btorpey/work/libzmq/master/src/tests/test_zmtp_strict.cpp:235:connect_success_unversioned:PASS
Got event: CONNECT_DELAYED
Got event: CONNECTED
Got event: DISCONNECTED
/home/btorpey/work/libzmq/master/src/tests/test_zmtp_strict.cpp:236:connect_failed_unversioned:PASS

-----------------------
3 Tests 0 Failures 0 Ignored 
```

I clearly need to better understand how "unversioned" works -- suggestions welcome.

### DIFF
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -435,7 +435,7 @@ Applicable socket types:: all, when using TCP or IPC transports
 
 
 ZMQ_ZMTP_STRICT: Require versioned ZMTP protocol
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The 'ZMQ_ZMTP_STRICT' option shall retrieve the current setting for the socket.
 
 If set, the handshake will fail if the peer socket's ZMTP version is unspecified.
@@ -446,7 +446,8 @@ NOTE: in DRAFT state, not yet available in stable releases.
 Option value type:: int
 Option value unit:: boolean
 Default value:: 0 (false)
-Applicable socket types:: all but ZMQ_STREAM
+Applicable socket types:: all, only for connection-oriented transports
+(except ZMQ_STREAM)
 
 ZMQ_MULTICAST_HOPS: Maximum network hops for multicast packets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/zmq_getsockopt.txt
+++ b/doc/zmq_getsockopt.txt
@@ -434,6 +434,20 @@ Default value:: ZMQ_NULL
 Applicable socket types:: all, when using TCP or IPC transports
 
 
+ZMQ_ZMTP_STRICT: Require versioned ZMTP protocol
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The 'ZMQ_ZMTP_STRICT' option shall retrieve the current setting for the socket.
+
+If set, the handshake will fail if the peer socket's ZMTP version is unspecified.
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: boolean
+Default value:: 0 (false)
+Applicable socket types:: all but ZMQ_STREAM
+
 ZMQ_MULTICAST_HOPS: Maximum network hops for multicast packets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The option shall retrieve time-to-live used for outbound multicast packets.
--- a/src/options.hpp
+++ b/src/options.hpp
@@ -291,7 +291,7 @@ struct options_t
     // Version of monitor events to emit
     int monitor_event_version;
 
-    // minimum supported version of ZMTP
+    // if set, reject connections to unversioned zmtp
     int zmtp_strict;
 
     //  WSS Keys
--- a/src/options.hpp
+++ b/src/options.hpp
@@ -291,6 +291,9 @@ struct options_t
     // Version of monitor events to emit
     int monitor_event_version;
 
+    // minimum supported version of ZMTP
+    int zmtp_strict;
+
     //  WSS Keys
     std::string wss_key_pem;
     std::string wss_cert_pem;
--- a/tests/test_zmtp_strict.cpp
+++ b/tests/test_zmtp_strict.cpp
@@ -34,18 +34,22 @@
 
 #include <unity.h>
 
+#include <memory.h>
+
 // test connecting to unversioned zmtp w/o strict succeeds
-void connect_success ()
+void connect_success_versioned ()
 {
-    char bind_address[MAX_SOCKET_STRING];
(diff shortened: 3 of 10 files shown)


bill-torpey:
@somdoron -- can you take a look at this when you get a chance?

tests/testutil.hpp
```diff
@@ -67,6 +67,9 @@
 #define PORT_6 5561
 
 //  For tests that mock ZMTP
+const uint8_t zmtp_greeting_unversioned[] = {
+  0x01, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
bluca, line 71: when connecting to a sub, something like this should be a pub message that contains "A":

01 00 02 00 42

bluca: https://rfc.zeromq.org/spec/13/

doc/zmq_getsockopt.txt
```diff
@@ -434,6 +434,20 @@ Default value:: ZMQ_NULL
 Applicable socket types:: all, when using TCP or IPC transports
 
 
+ZMQ_ZMTP_STRICT: Require versioned ZMTP protocol
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
bluca, line 438: For markup the ~ should match the characters above

doc/zmq_getsockopt.txt
```diff
@@ -434,6 +434,20 @@ Default value:: ZMQ_NULL
...
+
+If set, the handshake will fail if the peer socket's ZMTP version is unspecified.
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: boolean
+Default value:: 0 (false)
+Applicable socket types:: all but ZMQ_STREAM
```
bluca, line 449: connected transports - eg, it doesn't apply to inproc/udp

src/options.hpp
```diff
@@ -291,6 +291,9 @@ struct options_t
     // Version of monitor events to emit
     int monitor_event_version;
 
+    // minimum supported version of ZMTP
```
bluca, line 294: the comment needs to be updated

tests/test_zmtp_strict.cpp
```diff
@@ -0,0 +1,239 @@
+/*
+    Copyright (c) 2017 Contributors as noted in the AUTHORS file
```
bluca, line 2: 2020

tests/test_zmtp_strict.cpp
```diff
@@ -0,0 +1,239 @@
...
+*/
+#include <assert.h>
+
+#include "testutil.hpp"
+#include "testutil_unity.hpp"
+#include "testutil_monitoring.hpp"
+
+#include <unity.h>
+
+#include <memory.h>
```
bluca, line 37: public headers inclusion before local headers

tests/testutil.hpp
bill-torpey: Yes, I read the RFC, and it is somewhat unclear, at least to me:

> A peer SHALL send a greeting consisting of either an anonymous or an identity message. An anonymous greeting consists of an empty string.

I've tried different combinations for zmtp_greeting_unversioned (including the one you suggested), but none of them connect, and not sure where to go from here.

doc/zmq_getsockopt.txt
bill-torpey: done

src/options.hpp
bill-torpey: done

tests/test_zmtp_strict.cpp
bill-torpey: done

tests/test_zmtp_strict.cpp
bill-torpey: done

tests/testutil.hpp
bluca: interesting - did you have an old zmq 2.x binary you were doing the connection tests with? Would be easier to look at wireshark I think

bill-torpey: That's my point -- I don't have an old binary to test with, and I'm not eager to try to build one ;-)  I thought you or @somdoron might be able to tell me.

In the meantime, I've tried looking at the code and RFC to try something that works, but no joy so far:

```
const uint8_t zmtp_greeting_unversioned[] = {
  0x01, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
const uint8_t zmtp_greeting_unversioned[] = {
  0x01, 00, 02, 00, 42};
const uint8_t zmtp_greeting_unversioned[] = {
  0x01, 00};
```

Can anyone tell me what an unversioned handshake looks like?

bluca: I think nobody is here who knows it off the top of their head - there's a test for the curve protocol that looks like it's attempting to do one: https://github.com/zeromq/libzmq/blob/master/tests/test_security_curve.cpp#L215

somdoron: I will try to make a test for the unversioned protocol

somdoron: Some history: 
https://github.com/zeromq/libzmq/commit/1bca4f6f033ed6bb56fcfd7762bbfea04ecf684b

somdoron: Source code of stream_engine before versioned protocol:
https://github.com/zeromq/libzmq/blob/a622328a9c089e635e8af6b67f6a87251d3e34bf/src/stream_engine.cpp

The bottom line, there is no greeting with the unversioned ZMTP, you just send messages and the first message must be an identity. 

So {0,0} (empty message) is a valid message. Also {1,0,1} (one byte with value 1). 

Because first message must be an identity, to fake an unversioned peer send the following {0,0,1,0,1}, which sends an empty identity following by a single-byte message.

tests/test_zmtp_strict.cpp
```diff
@@ -0,0 +1,238 @@
...
+#include <unity.h>
+
+// test connecting to unversioned zmtp w/o strict succeeds
+void connect_success_versioned ()
+{
+    char my_endpoint[MAX_SOCKET_STRING];
+    fd_t server = bind_socket_resolve_port ("127.0.0.1", "0", my_endpoint);
+
+    // setup sub socket
+    void *sub = test_context_socket (ZMQ_SUB);
```
somdoron, line 45: I would use a dealer, it would be easier to test

tests/testutil.hpp
somdoron: The bottom line, I would use dealer to dealer for testing, and then the expect one message to arrive on the zeromq dealer.
The message should be {0,0,1,0,1} (empty identity and single byte message)

bill-torpey: > I will try to make a test for the unversioned protocol

Hi @somdoron -- I've been off on vacation, so just checking in to see if you are still planning to take a crack at this?

PR #3982, opened 2020-07-06
Ios135

mileschet:
added examples directory with iOS sample and build_ios.sh script.


bluca:
Hi,

Adding a build script is perfectly fine (but please create a new directory and move it under it, builds/ios, following the same pattern used for everything else), but it doesn't look like the example directory fits. An external repository that you can manage seems more appropriate?
There are a lot of what looks like generated files from some IDE, and even binary files - we do not want these committed here.

If you really want the minimal IOS source file committed (without generated IDE files nor binaries nor copies of the public header), then please move it under builds/ios.

Also please add a relicense statement: https://github.com/zeromq/libzmq/tree/master/RELICENSE

mileschet:
@bluca fantastic, thank you very much really appreciate the directions.

PR #3994, opened 2020-07-21
build: compile with -Werror by default on Darwin

fanquake:
Using `-Werror` has been disabled for Darwin since very early on in the
codebase. However at this point, I can't see an obvious reason why it
should still be disabled compared to when building for the other
operating systems. Unless this was specifically for cross-compiling?

I've tested compiling on macOS using Apple Clang 11.0.3, LLVM Clang
10.0.0 and GCC 10.1.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -196,7 +196,7 @@ else
     AC_MSG_RESULT([no])
 fi
 
-# By default compiling with -Werror except OSX and on Solaris when building
+# By default compiling with -Werror except on OSX and Solaris when building
 # with libsodium.
 AC_ARG_ENABLE([Werror],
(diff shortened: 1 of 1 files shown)


bluca:
It is disabled because it doesn't work, see: https://travis-ci.org/github/zeromq/libzmq/jobs/710220276

fanquake:
> It is disabled because it doesn't work

Right. I think I've misread the original comment. i.e `macOS || (solaris && libsodium)` rather than `(macOS || Solaris) && libsodium`. Regardless, it looks like the issue is just with libsodium, same as Solaris, so it should just be disabled in that case, rather than wholesale. I've tested building in both cases, and have pushed new changes to reflect that.

PR #3997, opened 2020-07-29
make Docker image smaller + use `debian:buster-slim` base

kaczmarj:
The Dockerfile for this project has not been updated in several years, so I tried to improve it in this PR. This PR updates the Dockerfile to use multiple stages. In the first stage, compile-time dependencies are installed, and libzmq is compiled and tested. The second stage starts with a fresh, slim Docker image, and the compiled outputs of the first stage are copied over. Runtime dependencies, like libsodium and kerberos, are installed as well. The resulting Docker image is 28% the size of the original image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,32 @@
-FROM ubuntu:14.04
+FROM debian:buster-slim AS builder
+LABEL maintainer="ZeroMQ Project <zeromq@imatix.com>"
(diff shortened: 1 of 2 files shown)


bluca:
Please add a relicense statement: https://github.com/zeromq/libzmq/tree/master/RELICENSE

kaczmarj:
@bluca - done.

PR #4011, opened 2020-08-15
Problem: use-after-free in WS

bluca:
Solution: bail out of the stream engine out_event if ws_engine raises
an error


bluca:
Happens here:

https://github.com/zeromq/libzmq/blob/1412078db61aac7eead02520ce75940c1ad13359/src/ws_engine.cpp#L973-L977

PR #4028, opened 2020-09-04
Some improvements to the test coverage job

sigiesec:


sigiesec:
Note that the `assert` changes were necessary to unbreak the build of the coverage job after enabling TLS there.

PR #4034, opened 2020-09-07
Problem: 4.3.3 is out, need new version

bluca:
Solution: bump to 4.3.4


bluca:
The ci_deploy.sh is not working anymore due to token issues, I had to do make dist manually. I'll look into it.

PR #4038, opened 2020-09-10
Problem: zmq_ctx_get API broken

bluca:
Solution: restore EINVAL as errno on unknown option.
Broken by https://github.com/zeromq/libzmq/pull/3642 which started to
return EFAULT instead

### DIFF
--- a/tests/test_ctx_options.cpp
+++ b/tests/test_ctx_options.cpp
@@ -283,6 +283,14 @@ void test_ctx_option_blocky ()
     test_context_socket_close (router);
 }
 
+void test_ctx_option_invalid ()
+{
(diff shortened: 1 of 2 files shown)


tests/test_ctx_options.cpp
```diff
@@ -283,6 +283,14 @@ void test_ctx_option_blocky ()
     test_context_socket_close (router);
 }
 
+void test_ctx_option_invalid ()
+{
+    TEST_ASSERT_EQUAL_INT (-1, zmq_ctx_set (get_test_context (), -1, 0));
```
sigiesec, line 288: Just for the future: You can simply write:
```
     TEST_ASSERT_FAILURE_ERRNO (EINVAL, zmq_ctx_set (get_test_context (), -1, 0));
```
instead of the two separate assertions.

PR #4041, opened 2020-09-16
Problem: using targets in CMakeLists.txt creates a corrupted xcode project with cmake

stvales:
Solution: use sources instead which leads to the same result with cmake but also generates a valid xcode project from cmake


bluca:
It's not the same. Targets are used to avoid building every object twice, once for static linking and once for dynamic linking. So this would make the build twice as long. I would suggest filing a bug with xcode to fix their broken integration instead.

stvales:
The bug would be with cmake and i'm not sure at all that this is an actual bug. With any IDE, including xcode, a dynamic and a static version of a library are two different projects/targets. Linking them to the same objects may be debatable. 

Xcode does not (which is sane at many levels) reuse objects between different targets. At the same time, the xcode project generated by cmake does exactly what is described inside the CMakeLists.txt file. Thus the problem.

bluca:
Sorry, but I'm afraid keeping build times and CI times reasonable is more important than making an opinionated proprietary IDE happy. Is it possible to detect with a cmake variable whether it is being run under xcode, for a conditional like the windows one?

stvales:
I was going to suggest using sources for macos only (which is pretty easy to do). But would it be OK ?

bluca:
The OSX CI is the slowest already, so unfortunately no. If autodetection of xcode is not possible, then just add an option, disabled by default

stvales:
Thanks !

PR #4053, opened 2020-09-27
stop reconnect after called zmq_disconnect()

americaboy:
When the server does not exist(Never started), the client is disconnected after the request receive timeout,
but the client socket is still reconnecting in the background.

expected result:
the client socket stop reconnect after called zmq_disconnect().

so,I add a disconnect flag for mark disconnect action,
if user set option ZMQ_RECONNECT_STOP:ZMQ_RECONNECT_STOP_AFTER_DISCONNECT for socket,
it will stop reconnet if disconnect flag is true.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -685,6 +685,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 /*  DRAFT ZMQ_RECONNECT_STOP options                                          */
 #define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
 #define ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED 0x2
+#define ZMQ_RECONNECT_STOP_AFTER_DISCONNECT 0x3
 
 /*  DRAFT Context options                                                     */
 #define ZMQ_ZERO_COPY_RECV 10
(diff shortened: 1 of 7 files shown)


include/zmq.h
```diff
@@ -685,6 +685,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 /*  DRAFT ZMQ_RECONNECT_STOP options                                          */
 #define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
 #define ZMQ_RECONNECT_STOP_HANDSHAKE_FAILED 0x2
+#define ZMQ_RECONNECT_STOP_AFTER_DISCONNECT 0x3
```
bluca, line 688: this needs to go in src/zmq_draft.hpp as well

bluca:
documentation and relicense statement are missing https://github.com/zeromq/libzmq/tree/master/RELICENSE

include/zmq.h
americaboy: ok, I will add it later.

bill-torpey:
A couple of questions re: this option:

- It looks like this is only implemented for ipc sockets -- correct?  
- If a socket is connected to multiple endpoints, and caller disconnects from any one of those endpoints, it will stop reconnecting on all endpoints?

Thanks!

bill-torpey:
@americaboy -- can you pls reply to above questions?  Very curious how this patch is expected to behave.  Thanks!

PR #4060, opened 2020-10-09
fix stack overflow on windows x64

martin-ksti:
in our project we have to use VS2010 Win64 compiler and we have fixed the stack overflow by using this suggestion:
https://github.com/zeromq/libzmq/issues/2876#issuecomment-601434808

### DIFF
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -63,11 +63,18 @@ void zmq::thread_t::start (thread_fn *tfn_, void *arg_, const char *name_)
     _arg = arg_;
     if (name_)
         strncpy (_name, name_, sizeof (_name) - 1);
+
+    // set default stack size to 4MB to avoid std::map stack overflow on x64
+    unsigned int stack = 0;
+#if defined _WIN64
+    stack = 0x400000; 
+#endif
+
 #if defined _WIN32_WCE
     _descriptor =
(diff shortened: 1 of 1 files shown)


bluca:
Please add a relicense statement as explained in the readme here: https://github.com/zeromq/libzmq/tree/master/RELICENSE

martin-ksti:
> 
> 
> Please add a relicense statement as explained in the readme here: https://github.com/zeromq/libzmq/tree/master/RELICENSE

done

https://github.com/zeromq/libzmq/pull/4061#issue-500690748

src/thread.cpp
```diff
@@ -63,11 +63,18 @@ void zmq::thread_t::start (thread_fn *tfn_, void *arg_, const char *name_)
     _arg = arg_;
     if (name_)
         strncpy (_name, name_, sizeof (_name) - 1);
+
+    // set default stack size to 4MB to avoid std::map stack overflow on x64
+    unsigned int stack = 0;
+#if defined _WIN64
+    stack = 0x400000;
```
sigiesec, line 70: If this fix is only relevant for VS2010, shouldn't this be restricted to that version? Sticking with the default if possible seems to be desirable.

martin-ksti: i think you're right. at the moment I have no way to making the change and I don't know if I will ever have that option.

PR #4075, opened 2020-11-02
The _DEBUG preprocessor definition is typically only set on Windows. …

bill-torpey:
… (e.g., see https://stackoverflow.com/questions/2290509/debug-vs-ndebug/29253284#29253284)

### DIFF
--- a/src/poller_base.cpp
+++ b/src/poller_base.cpp
@@ -136,7 +136,7 @@ void zmq::worker_poller_base_t::start (const char *name_)
 
 void zmq::worker_poller_base_t::check_thread () const
 {
-#ifdef _DEBUG
+#ifndef NDEBUG
     zmq_assert (!_worker.get_started () || _worker.is_current_thread ());
 #endif
 }


src/poller_base.cpp
```diff
@@ -136,7 +136,7 @@ void zmq::worker_poller_base_t::start (const char *name_)
 
 void zmq::worker_poller_base_t::check_thread () const
 {
-#ifdef _DEBUG
+#ifndef NDEBUG
```
bluca, line 139: This might have been done intentionally to never use it apart from a few CIs - with this change, it would always be enabled. @sigiesec ?

bill-torpey: I'm not sure -- that's more a question for @sigiesec, who made the original commit.

I was going through our own code looking for "Windows-isms" and spotted this, so I opened the PR for discussion/resolution.

If the original is in fact intended to apply only to Windows, perhaps an `#ifdef _WINDOWS` around it would make that clearer (or at least a comment to that effect).

bluca: @sigiesec ping

sigiesec: Sorry for the delay, giving this a look now.

sigiesec: Yes, it's a Windows-isms owed to the fact I was mainly working on Windows/MSVC at that time. It should be changed.

> with this change, it would always be enabled

... but I don't understand this. Before the change, it was enabled only on MSVC Debug builds, now it will be enabled on all Debug builds, but still not on Release builds, which is intended to avoid these quite heavy-weight sanity checks there.

bluca: The distiction of debug vs release builds is a windows-only thing, it doesn't really make sense anywhere else - ndebug is never defined

sigiesec: For CMake, it's definitely a thing on all platforms.

bluca: Never seen anyone use that on anything but Windows, but let me rephrase then: there is no difference between debug vs release in Linux distributions, ndebug is never defined

sigiesec: I know at least one company that uses a libzmq CMake build on Linux ;)

Also everyone using it through the official conan package (https://conan.io/center/zeromq), which seem to be quite a number of people.

And I am pretty sure Android builds require NDEBUG to be defined (cf. e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=827216)

bluca: Sure, doesn't change the fact that Linux distributions don't use it, so the vast majority of usage of Linux doesn't. The original question was, is this check supposed to be generally enabled in all cases, or not?

sigiesec: It seems this is a highly debated question. I have seen a rather harsh post on a Debian mailing listing stating that compiling with NDEBUG is "evil bad and wrong". I didn't see the arguments behind that. I've worked only on projects that do distinguish between Debug and Release builds, with huge performance differences between them. I think that's a sensible approach in general.

Many assertions in libzmq are about expectations on the environment, and it's sensible to have those in release builds. Since they are usually unrecoverable, it makes sense to make those release assertions. But that doesn't apply to all of them, and not to this one, where the property checked is under total control of the library itself. I think it would make sense to better distinguish between these different types of assertions systematically.

But I don't know if this one is particularly critical.

The proposed patch is fine, but I can also live with removing the condition completely for now.

bill-torpey:
My understanding is that the use of the `NDEBUG` preprocessor symbol to identify code that is included/excluded based on whether doing a debug or release build is not only common in the Linux world, but is also standardized -- e.g., see <https://stackoverflow.com/questions/2290509/debug-vs-ndebug/29253284#29253284>.

The `_DEBUG` preprocessor symbol, on the other hand, is strictly a Windows-ism.

src/poller_base.cpp
bluca: I don't care too much about whether it's right or wrong - what I'm worried about is introducing regressions. Merging this means that for all intents and purposes, bar some exceptions here and there, this assertion goes from universally disabled to universally enabled. Will things break?

sigiesec: I don't expect any regressions. (If there are any, it means we had a data race and thus UB anyway.)

bluca: Great, thank you!

PR #4080, opened 2020-11-09
Fix a typo and wrong order of assert() in the examples

ardrabczyk:


bluca:

PR #4092, opened 2020-11-26
Problem: no ipv6 test coverage

sigiesec:
Solution: add arm64 test coverage job, where ipv6 is available

Also temporarily disable two zmq_socket_monitor_pipes_stats test cases that
permananently fail on arm64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@
 language: c
 
 os:
-- linux
-- osx
+  - linux
+  - osx
 
 dist: bionic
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -28,148 +28,157 @@ env:
...
+    - env: BUILD_TYPE=coverage CURVE=tweetnacl GSSAPI=enabled NORM=enabled TIPC=enabled IPv6=ON TLS=enabled DRAFT=enabled
+      os: linux
+      arch: arm64
+      addons:
+        apt:
+          packages:
+            - libkrb5-dev
+            - libnorm-dev
+            - libgnutls28-dev
+            - lcov
```
bluca, line 69: is lcov included intentionally? it makes the job a lot longer

sigiesec: I think it's required, but I can try to remove it again and see if it still works.

sigiesec:
I wonder why it looks as if all job definitions were changed in .travis.yml... Maybe the indentation changed or so, I will try to get that fixed.

.travis.yml
sigiesec: I wonder: How did you see that lcov makes the job a lot longer?

bluca: it's doing all the coverage things  - gcov files etc

bluca:
> I wonder why it looks as if all job definitions were changed in .travis.yml... Maybe the indentation changed or so, I will try to get that fixed.

yeah, it's whitespace changes

.travis.yml
sigiesec: Well, yes, it's a coverage job.

bluca: ahah, sorry, my bad - I though you meant it in the sense of ensuring ipv6 tests run, rather than actual code coverage

sigiesec:
#4120 now added a non-coverage ARM64 job, so this PR should be changed to just make that into a coverage job.

PR #4094, opened 2020-11-29
resizable_fast_vector_t resize crash fix

boxkey:


bluca:

boxkey:

sorry, i don`n know how to add new file in this pr, so i create a new pr "add relicense.md file #4095" to add the relicense file.
and this fix pr has tow commits.

src/polling_util.hpp
```diff
@@ -78,8 +78,8 @@ template <typename T, size_t S> class resizable_fast_vector_t
     {
         if (_dynamic_buf)
             _dynamic_buf->resize (nitems_);
-        if (nitems_ > S) {
-            _dynamic_buf = new (std::nothrow) std::vector<T>;
+        else {
```
bluca, line 81: @sigiesec could you please have a look? IIRC you added this for some specific use cases - thanks

gummif: Should probably be if else ...  instead to avoid allocations for  few items.

sigiesec: Sorry for the long delay. 

First, this definitely seems wrong. We must ensure the postcondition that the size of the vector is `nitems_` after the call. But as @gummif said, this should not allocate a dynamic buffer if `nitems_` is small.

So having 
```
if (_dynamic_buf) {
            _dynamic_buf->resize (nitems_);
} else if (nitems_ > S) {
            _dynamic_buf = new (std::nothrow) std::vector<T>(nitems_ );
            //  TODO since this function is called by a client, we could return errno == ENOMEM here
            alloc_assert (_dynamic_buf);
}
```
seems to be the right change.

Would be nice to have unit tests for resizable_fast_vector_t.

jhruby: Please guys this is so trivial change. When ZeroMQ crashed in our Debug code I've spotted this instantly. Please merge it. I will use fork until then.

gummif: Since this appears to be a dead PR I made a new one https://github.com/zeromq/libzmq/pull/4247

PR #4109, opened 2020-12-23
Solution: if_nametoindex function is now used when available

### LINKED ISSUE #4108: IPv6 Link-local addresses don`t work on Linux systems
# Issue description

Fix in #3716 introduces new problem in ZeroMq 4.3.3.
It is not possible to connect by using IPv6 link-local address.

Part of ip_resolver.cpp
```c++

unsigned int zmq::ip_resolver_t::do_if_nametoindex (const char *ifname_)
{
#if _WIN32_WINNT > _WIN32_WINNT_WINXP && !defined ZMQ_HAVE_WINDOWS_UWP         \
  && !defined ZMQ_HAVE_VXWORKS
    return if_nametoindex (ifname_);
#else
    LIBZMQ_UNUSED (ifname_);
    // The function 'if_nametoindex' is not supported on Windows XP.
    // If we are targeting XP using a vxxx_xp toolset then fail.
    // This is brutal as this code could be run on later windows clients
    // meaning the IPv6 zone_id cannot have an interface name.
    // This could be fixed with a runtime check.
    return 0;
#endif
}
```

Probably bug is also present in udp_address.cpp
```c++
#if _WIN32_WINNT > _WIN32_WINNT_WINXP && !defined ZMQ_HAVE_WINDOWS_UWP         \
  && !defined ZMQ_HAVE_VXWORKS
            _bind_interface = if_nametoindex (src_name.c_str ());
            if (_bind_interface == 0) {
                //  Error, probably not an interface name.
                _bind_interface = -1;
            }
#endif
```

# Environment

* libzmq version 4.3.3
* OS: Ubuntu 20.04

Use _zmq_connect_ with IPv6 link local address like:
```c++
zmq_connect (requester, "tcp://[fe80::213:b4ff:feff:ff02%br-9bd5d0e6c8ca]:1234")
```
Its not possible to connect on Linux systems.

Function _do_if_nametoindex_ always returns 0.

# What's the expected result?

Function _do_if_nametoindex_ should return network interface ID.

stanpl:
Fixes #4108

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -546,11 +546,13 @@ if(ZMQ_HAVE_WINDOWS)
 
   set(CMAKE_REQUIRED_LIBRARIES "iphlpapi.lib")
   check_cxx_symbol_exists(GetAdaptersAddresses "winsock2.h;iphlpapi.h" HAVE_IPHLAPI)
+  check_cxx_symbol_exists(if_nametoindex "iphlpapi.h" HAVE_IF_NAMETOINDEX)
 
   set(CMAKE_REQUIRED_LIBRARIES "")
   # TODO: This not the symbol we're looking for. What is the symbol?
   check_library_exists(ws2 fopen "" HAVE_WS2)
 else()
+  check_cxx_symbol_exists(if_nametoindex net/if.h HAVE_IF_NAMETOINDEX)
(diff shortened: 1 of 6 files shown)


bluca:
Please add a commit with a relicense statement as indicated in https://github.com/zeromq/libzmq/tree/master/RELICENSE

PR #4110, opened 2020-12-23
Problem: PUB+WS broken, git blame hard to use

bluca:
Solution: fix them


src/ws_encoder.cpp
```diff
@@ -73,6 +73,9 @@ void zmq::ws_encoder_t::message_ready ()
     size_t size = in_progress ()->size ();
     if (_is_binary)
         size++;
+    //  TODO: create an opcode for subscribe/cancel
```
bluca, line 76: @somdoron given WS2 is still draft, I think it would be a good idea to have opcodes for sub/cancel just like ZMTP 3.1

PR #4115, opened 2021-01-04
PROBLEM: missing permission to relicense commits by GH handle skaes

skaes:
SOLUTION: add file containing permission statement


c-rack:
Thanks

PR #4118, opened 2021-01-05
Problem: No direct support for setting socket priority

aheroff1:
Solution: Add ZMQ_PRIORITY socket option, which sets the
SO_PRIORITY socket option on the underlying socket. This
socket option is not supported under Windows. Check option
and set socket option on creation of underlying socket.


include/zmq.h
```diff
@@ -682,6 +682,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_RECONNECT_STOP 109
 #define ZMQ_HELLO_MSG 110
 #define ZMQ_DISCONNECT_MSG 111
+#define ZMQ_PRIORITY 112
```
bluca, line 685: this needs to be duplicated in src/zmq_drafts.hpp

bluca:
Please add the configure check to CMakeLists.txt as well

aheroff1:
Issues above were resolved. Additionally, further internal review found that I did not add documentation to zmq_getsockopt. This is also resolved. New commit to be pushed shortly.

PR #4119, opened 2021-01-07
Avoid crash in wsa_error_to_error when isolated in Microsoft Defender ATP

toonetown:
SOLUTION: The resason for this crash is it appears that an unknown GetLastError is getting returned when an application is "Isolated" within dATP.  It would seem to make the most sense to just return `EFAULT` in `wsa_error_to_error` instead of calling `zmq_abort` if a more accurate error could not be determined or mapped.

### DIFF
--- a/src/err.cpp
+++ b/src/err.cpp
@@ -380,7 +380,7 @@ int zmq::wsa_error_to_errno (int errcode_)
         case WSANO_DATA:
             return EFAULT;
         default:
-            wsa_assert (false);
+            return EFAULT;
     }
     //  Not reachable
     return 0;


toonetown:
A possibly related commit from long ago can be seen at #1072 - that one added a bunch of additional mappings and mapped the majority of them to `EFAULT`.  It was that change that inspired the approach to this fix in order to avoid needing to keep the mapping 100% up-to-date.

bluca:
Sorry but this doesn't look right. We want to allow a very specific set of conditions, and anything outside those parameters requires manual attention. What error exactly are you getting, and what is the situation? What is this defender ATP, some sort of firewall?

toonetown:
Defender ATP is a network isolation product by Microsoft - you can isolate an endpoint, and it prevents (at some kernel level) any network connections from being established.  It appears that in some cases, it is returning some different error code - I can see in the stack trace that it is unable to map the value it gets from `GetLastError()` to an `errno`, but unfortunately, I have been unable to get a crash dump which shows _what_ that error value is (I don't have a license to dATP, and am working via a customer to gather the crashes they are seeing).

The reasoning I had behind this change is that to prevent crashing of the application, it is (IMO) better to have a "generic" error than a raised exception that could not be caught and handled (since it is non-continuable).  I completely recognize that that assumption may not necessarily reflect the views of others.

Making this change has helped in my particular scenario and application - however, it is understandable if it is not desired to be more widely deployed.  I submitted this PR in the event that it might be helpful to the broader community - but I am fine if it is decided to be closed instead, and I can just maintain the change in my own branch that I deliver from.  Thank you for your time.

bluca:
Sorry but blanket-ignore of errors means we'll get bug reports with "messages disappear plz fix", which will make things worse overall. If you can find out exactly what errno is being returned, and document it accordingly, then we can add an exception like the others. Please feel free to open a new PR once that happens, but will close this now.

toonetown:
I accept this decision - thank you for your consideration.

PR #4123, opened 2021-01-16
Problem: norm fails to compile under windows

### LINKED ISSUE #1863: norm_engine.cpp fails to compile (Windows)
On line 229
```
fd_t normDescriptor = NormGetDescriptor(norm_instance);
```
Error:

```
Variable of type 'fd_t' cannot be initialized with value of type NormDescriptor
```
If anyone can suggest workaround or for maintainers.. Specs: VS2015, Win32 build, libzmq+opengpm+Norm.

mjvankampen:
Fixes #1863 along the lines as discussed.

It adds socket pair of which one is used in a thread where NORM events are polled. It sends the events to the other socket which is registered to zmq.

@bebopagogo does this look like what you had in mind? It is not a signalling socket but it passes the events. Is that an issue for NORM? Somewhat easier to implement as polling is level-triggered.

### DIFF
--- a/tests/test_pubsub.cpp
+++ b/tests/test_pubsub.cpp
@@ -115,6 +115,8 @@ int main ()
     RUN_TEST (test_tcp);
     RUN_TEST (test_inproc);
     RUN_TEST (test_inproc_late_bind);
+#if defined ZMQ_HAVE_NORM
     RUN_TEST (test_norm);
+#endif
     return UNITY_END ();
 }
--- a/tests/test_pubsub.cpp
+++ b/tests/test_pubsub.cpp
@@ -1,0 +1,120 @@
+/*
+    Copyright (c) 2007-2020 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    As a special exception, the Contributors give you permission to link
+    this library with independent modules to produce an executable,
+    regardless of the license terms of these independent modules, and to
+    copy and distribute the resulting executable under terms of your choice,
+    provided that you also meet, for each linked independent module, the
+    terms and conditions of the license of that module. An independent
+    module is a module which is not derived from or based on this library.
+    If you modify this library, you must extend this exception to your
+    version of the library.
+
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "testutil.hpp"
+#include "testutil_unity.hpp"
+
+SETUP_TEARDOWN_TESTCONTEXT
+
+void test (const char *address)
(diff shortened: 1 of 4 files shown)


bebopagogo:
This looks like a reasonable approach.  It's similar to what I had in mind as an alternate interim approach but a little different where I would have used  "notification" WinSock within existing the NORM API thread as a compile-time option instead of the Window Event handle that is used now.  Then that WinSock handle would be passed up to the libzmq poller.  It is not if that approach would be any more performant than what you've done here.  

The longer term approach that I had outlined earlier was to modify the libzmq code so a non-socket Event could be used but I would have to relook at that now as think the libzmq poller code has changed a bit.  So the approach you have implemented here is probably worth incorporating to provide Windows support.  Since NORM rides on top of UDP, there is some Windows configuration tuning that needs to be done to achieve very high throughput rates on Windows anyway.

I did not pull and test your code here but I did look through the file changes you made and it looks OK to me for initial Windows NORM support purposes.

bluca:
Also please add a commit with a relicense statement as indicated in https://github.com/zeromq/libzmq/tree/master/RELICENSE

mjvankampen:
@bebopagogo I think the main issue is that zmq uses wepoll (and reactor instead of proactor pattern) as it is a fast approach on Windows. Switching to IO completion ports is a big refactor. 

I think both approaches should be in. One where poll/select/windows equivalents can handle HANDLE. And one where wepoll can use norm as not to impact other code. IOCP is a thing for the distant future I think.

@bluca there is a mjvk in there already. I changed my nick (not sure why anymore :-)). Is this sufficient or should I be in there twice?

bluca:
> @bluca there is a mjvk in there already. I changed my nick (not sure why anymore :-)). Is this sufficient or should I be in there twice?

I see, then please just update the existing grant to list your new Github handle, so that we don't lose track

mjvankampen:
Some tests are failing but I don't think they have much to do with this PR. Is this correct @bluca ?

tests/test_pubsub.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+void test_inproc ()
+{
+    test ("inproc://hello-msg");
+}
+
+void test_inproc_late_bind ()
+{
+#if defined ZMQ_BUILD_DRAFT_API
```
bluca, line 87: I'm not sure I follow, why are you adding these tests? And why only for DRAFTS?

mjvankampen: Fair enough, removed non-norm related tests. As for the draft, for example ZMQ_HELLO_MSG is only available with DRAFT enabled.

mjvankampen:
Cheeky question from my side. It seems I just missed the release. Do you have any indication of when the next one is? I consume packages through dependency managers so typically only use releases.

bluca:
> Cheeky question from my side. It seems I just missed the release. Do you have any indication of when the next one is? I consume packages through dependency managers so typically only use releases.

Given it's new functionality, we wouldn't want to have it in a release immediately, needs to get some testing first

bluca:
We publish packages built from latest HEAD as mentioned in the readme:

http://software.opensuse.org/download.html?project=network%3Amessaging%3Azeromq%3Agit-stable&package=libzmq3-dev

http://software.opensuse.org/download.html?project=network%3Amessaging%3Azeromq%3Agit-stable&package=zeromq-devel

mjvankampen:
Clear! What kind of additional testing, maybe I can add some more unit tests. It might then be nice to add norm to the CI as well?

bluca:
unit tests would be great, although I'm not sure how much works in the CI - is this ready now?

mjvankampen:
Yes this is ready. I could try to add norm to ci in a different PR?

bluca:
the test is failing everywhere with ENODEV, so I've disabled it again

PR #4129, opened 2021-01-26
Problem: exception thrown when debugging cl.exe x64 build with LLDB

s-zhao-3g:
Problem: exception thrown when debugging cl.exe x64 build with LLDB

Solution: Use __try __except with cl.exe anduse pthread_setname_np with MinGW.
Remove usage of pushing exception handler to TIB->ExceptionList.

### DIFF
--- a/RELICENSE/c-zhao-3g.md
+++ b/RELICENSE/c-zhao-3g.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Lingqiao Zhao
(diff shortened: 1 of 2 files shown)


bluca:
Please fix the formatting:

```
--- a/src/thread.cpp
+++ b/src/thread.cpp
@@ -157,13 +157,13 @@ void zmq::thread_t::
 
 #elif defined(__MINGW32__)
 
-    int rc = pthread_setname_np (pthread_self(), _name);
+    int rc = pthread_setname_np (pthread_self (), _name);
     if (rc)
         return;
 
 #else
 
-    // not implemented
+        // not implemented
 
 #endif
 }
```

PR #4144, opened 2021-02-18
Problem: Unused class resizable_fast_vector_t

gummif:
Solution: Remove it


gummif:
Makes https://github.com/zeromq/libzmq/pull/4094 obsolete

gummif:
Nevermind, it is actually used in one place for Windows only.

PR #4162, opened 2021-03-18
fix NORM-related protolib/protokit linking errors

weston-nrl:
NORM build on Ubuntu 18.04 (and likely other platforms) gives linking errors related to protolib/protokit.  This adds -lprotokit and the protolib directory (residing inside of NORM) to fix those errors.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -697,9 +697,9 @@ if test "x$with_norm_ext" != "xno"; then
         if test "x$with_norm_ext" != "xyes"; then
             norm_path="${with_norm_ext}"
             norm_CFLAGS="${norm_CFLAGS} -I${norm_path}/include"
-            norm_LIBS="${norm_LIBS} -L${norm_path}/lib"
+            norm_LIBS="${norm_LIBS} -L${norm_path}/lib -L${norm_path}/protolib/lib"
         fi
-        norm_LIBS="${norm_LIBS} -lnorm"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -697,9 +697,9 @@ if test "x$with_norm_ext" != "xno"; then
         if test "x$with_norm_ext" != "xyes"; then
             norm_path="${with_norm_ext}"
             norm_CFLAGS="${norm_CFLAGS} -I${norm_path}/include"
-            norm_LIBS="${norm_LIBS} -L${norm_path}/lib"
+            norm_LIBS="${norm_LIBS} -L${norm_path}/lib -L${norm_path}/protolib/lib"
         fi
-        norm_LIBS="${norm_LIBS} -lnorm"
+        norm_LIBS="${norm_LIBS} -lnorm -lprotokit"
```
bluca, line 702: This is only for static linked builds though?

weston-nrl: That is what I have tested with, as it's the default NORM/protolib build method.  If there is an alternative method to supply that information via configure to allow compilation to complete, that would work as well, but I didn't see one.

bluca: Ubuntu 18.04 only ships the shared library:

https://packages.ubuntu.com/bionic/amd64/libnorm-dev/filelist

So the build would fail there, as there's no libprotokit.a

Issue #4044, opened 2020-09-16
Cannot build 4.3.3 with mingw


jagerman:
# Issue description

zeromq 4.3.3 fails to build with mingw, where zeromq 4.3.2 was fine:

```
tests/testutil.hpp:105:15: error: conflicting declaration of C function ‘int closesocket(int)’
  105 | #define close closesocket
      |               ^~~~~~~~~~~
In file included from tests/../src/windows.hpp:54,
                 from tests/testutil.hpp:43,
                 from tests/testutil_unity.hpp:34,
                 from tests/testutil_unity.cpp:29:
/usr/share/mingw-w64/include/winsock2.h:1005:34: note: previous declaration ‘int closesocket(SOCKET)’
 1005 |   WINSOCK_API_LINKAGE int WSAAPI closesocket(SOCKET s);
      |                                  ^~~~~~~~~~~
```

Everything builds fine if I delete the:

`#define close closesocket`

referenced above, but I don't whether this is some hack needed for other Windows compilers (e.g. MSVC).  It *does* seem like a pretty dangerous thing to define in a header, though.

Since that define was there in 4.3.2, I am not sure why this started failing in 4.3.3: perhaps some test includes shuffled around?

(Looking at it more closely, it seems testutil.hpp has quite a bit of other Windows-related cruft that isn't needed at all since it has already included ../src/windows.hpp).

# Environment

* libzmq version 4.3.3
* OS: Debian linux
* $ x86_64-w64-mingw32-gcc --version
x86_64-w64-mingw32-gcc (GCC) 10-posix 20200525

sigiesec:
I am pretty sure that this can't simply be removed when building against the Windows SDK (not dependent on the compiler probably).

There are maybe nicer ways to do this than via this preprocessor define. This seemed to be the simplest way to fit in the Windows SDK with the somehow Posix-biased code.

A PR improving this or at least fixing mingw would be definitely welcome!

I think we have a mingw CI job, I wonder if that's broken without notice, or something ia different there.

0x0badc0de:
Got the same issue, just disabled tests by passing `-DBUILD_TESTS=OFF` to cmake.

PR #4172, opened 2021-04-15
MINGW Build Fix.

jlsantiago0:
There is a macro that redefines close to closesocket, but then close is
later defined in unistd.h which is included in testutil_unity.cpp which
then causes the regular close() prototype to be defined with the new
function name closesocket which does not match the signature of the
original closesocket in winsock2.h from the toolchain.

We can work around this issue, by making sure that unistd.h is included
before the macro is defined. HMM> Yes macro substitution is the root of
evil.


jlsantiago0:
NOTE: This should resolve issue #4044

PR #4181, opened 2021-05-02
fix unused variables under windows

neheb:
Fixes compilation as -Werror is passed.

### DIFF
--- a/src/clock.cpp
+++ b/src/clock.cpp
@@ -126,9 +126,11 @@ static f_compatible_get_tick_count64 my_get_tick_count64 =
   init_compatible_get_tick_count64 ();
 #endif
 
+#ifndef ZMQ_HAVE_WINDOWS
 const uint64_t usecs_per_msec = 1000;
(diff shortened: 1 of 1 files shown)


Biswa96:
This commit breaks build in Windows AArch64 system using clang mingw-w64 toolchain. Here is the compiler error.

```
../zeromq-4.3.5/src/clock.cpp:251:48: error: use of undeclared identifier 'nsecs_per_usec'; did you mean 'usecs_per_sec'?
  251 |     return static_cast<uint64_t> (ts.tv_sec) * nsecs_per_usec * usecs_per_sec
      |                                                ^~~~~~~~~~~~~~
      |                                                usecs_per_sec
../zeromq-4.3.5/src/clock.cpp:106:16: note: 'usecs_per_sec' declared here
  106 | const uint64_t usecs_per_sec = 1000000;
      |                ^
```

Full build log can be found here https://github.com/msys2-arm/msys2-autobuild/actions/runs/6616870195/job/17972078012

PR #4190, opened 2021-05-16
Update fd slot type to zmq_fd_t

yitzchak:
The documentation for `zmq_pollitem_t` lists the type of the `fd` field as `int` but it changed to `zmq_fd_t` in v4.3.2

Please note that the documentation for previous versions is also incorrect in that it lists the type as `int` when it was `SOCKET` on Windows and `int` elsewhere. Is the documentation for previous releases corrected for errors?


bluca:

yitzchak:
Done

Issue #4191, opened 2021-05-16
Problem: no downloads available for Windows builds


minrk:
# Issue description

bintray is [shutdown](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/), so all the Windows releases published from appveyor are inaccessible. They are still being built on appveyor, but can no longer be downloaded (and future uploads will presumably fail).

This is *not* a duplicate of https://github.com/zeromq/zeromq.org/issues/113 which was closed as a docs-only issue because the dead links are removed from the docs. This issue should only be closed when:

- a new destination is picked for Windows artifacts ([github packages](https://github.com/features/packages)?)
- Windows build [currently still configured to send to bintray](https://github.com/zeromq/libzmq/blob/04c37982b152945731188f8a149de99a72aef29c/appveyor.yml#L274-L28) is updated to use the new destination
- at least the current stable builds are uploaded to the new destination

pyzmq Windows releases rely on these builds, which means I can't make a pyzmq release until this is fixed or I find another source of libzmq builds.

# Environment

* libzmq version (commit hash if unreleased):  all
* OS: Windows

# Minimal test code / Steps to reproduce the issue

Try to download a Windows release, e.g. `wget https://dl.bintray.com/zeromq/generic/libzmq-v141-x64-4_3_2.zip`

# What's the actual result? (include assertion message & call stack if applicable)

403 forbidden

# What's the expected result?

Windows dll builds of libzmq

bluca:
There's some stuff published on package managers like Conda, are you able to use those for pyzmq?

minrk:
Interesting idea! I can try extracting libraries from the conda packages in the meantime, I'm not sure if that will work. Do you think libzmq should stop publishing releases for Windows? The infrastructure is all already in place, but needs a new host for the artifacts.

bluca:
No opinion on the subject, I'm really not involved in that side - if anyone can find an alternative and make it work, it would be fine too

minrk:
Looked into it for a bit and conda's a no go since they only do 64b builds on Windows, and pyzmq publishes 32b builds for Windows as well.

[cloudsmith](https://cloudsmith.com) seems to be the closest replacement. Doesn't have a built-in appveyor deploy plugin, but it does have a CLI so `cloudsmith-cli push ...` should be all that's needed.


Appveyor supports publishing to GitHub releases, which might work - I can't tell if the multiple artifacts from mutliple builds for one release is a problem, though.

[...]

PR #4192, opened 2021-05-19
[appveyor] upload windows artifacts to GitHub releases

minrk:
now that bintray is gone.

Should address #4191 after the windows files are uploaded. This can be done by either making and pushing a 4.3.4.post0 tag to this repo, or manually copying the files from [my test run of this pr](https://github.com/minrk/libzmq/releases/tag/v4.3.4.post0) to the existing [4.3.4 release](https://github.com/zeromq/libzmq/releases/tag/v4.3.4). The only thing not tested is the upload credentials, which is the only difference between what ran on my fork and this pr.


bluca:
Thanks! I'll do a manual upload, tags are used by automation so we don't want to add new ones

PR #4198, opened 2021-05-25
Remove Windows-specific defines from zmq.h.

TobiSchluter:
These are redundant with the ones in windows.hpp and aren't needed for the public interface.

Since 4.3.3 zmq.h no longer includes windows.h and so these defines have no effect on zmq besides potentially interfering with what the user is doing.

Built on Windows VS 16.9.6, all tests passed.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -57,21 +57,6 @@ extern "C" {
 #endif
 #include <stddef.h>
 #include <stdio.h>
-#if defined _WIN32
-//  Set target version to Windows Server 2008, Windows Vista or higher.
-//  Windows XP (0x0501) is supported but without client & server socket types.
-#ifndef _WIN32_WINNT
-#define _WIN32_WINNT 0x0600
-#endif
-
-#ifdef __MINGW32__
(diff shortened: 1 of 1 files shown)


bluca:
I don't think it's safe to remove things from the public header - you never know what users are relying on. This could break backward compatibility in unexpected ways.

TobiSchluter:
@bluca maybe it helps putting this patch into the context of my previous one which removed #include<winsock2.h> in the same #ifdef block #3681  That definitely was a user-visible change that doesn't seem to have led to an uproar.

This patch basically just cleans up after the previous one because the windows API defines have no purpose any longer besides leading to confusing behavior for people who do their own defines in header files (bad idea!) or to unexpected failures: note that Windows XP i.e. 0x5000 supports less stuff than later versions -- but that is totally independent from the public header. IOW they might think that they are using a version that supports "client & server socket types" but then fail because the library file in the background doesn't actually.

Anyway, it's not an important issue, I just perceived it as an ugly wart.

TobiSchluter:
Thank you @bluca !

PR #4203, opened 2021-06-02
Issue 3146 : Fix strncpy problem with new gcc compilers

awanishgolwara:
GCC-8 and later compilers fail/warn to compile this strncpy as static analysis is detecting a potential string truncation error. The following guidelines are provided by GCC to fix it.

See https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8#forming_truncated_strings_with_snprintf

### DIFF
--- a/src/msg.cpp
+++ b/src/msg.cpp
@@ -701,11 +701,11 @@ int zmq::msg_t::set_group (const char *group_, size_t length_)
         assert (_u.base.group.lgroup.content);
         new (&_u.base.group.lgroup.content->refcnt) zmq::atomic_counter_t ();
         _u.base.group.lgroup.content->refcnt.set (1);
-        strncpy (_u.base.group.lgroup.content->group, group_, length_);
-        _u.base.group.lgroup.content->group[length_] = '\0';
+        strncpy (_u.base.group.lgroup.content->group, group_, sizeof _u.base.group.lgroup.content->group);
+        _u.base.group.lgroup.content->group[sizeof _u.base.group.lgroup.content->group - 1] = '\0';
     } else {
-        strncpy (_u.base.group.sgroup.group, group_, length_);
-        _u.base.group.sgroup.group[length_] = '\0';
+        strncpy (_u.base.group.sgroup.group, group_, sizeof _u.base.group.sgroup.group);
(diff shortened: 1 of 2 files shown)


src/msg.cpp
```diff
@@ -701,11 +701,11 @@ int zmq::msg_t::set_group (const char *group_, size_t length_)
         assert (_u.base.group.lgroup.content);
         new (&_u.base.group.lgroup.content->refcnt) zmq::atomic_counter_t ();
         _u.base.group.lgroup.content->refcnt.set (1);
-        strncpy (_u.base.group.lgroup.content->group, group_, length_);
-        _u.base.group.lgroup.content->group[length_] = '\0';
+        strncpy (_u.base.group.lgroup.content->group, group_, sizeof _u.base.group.lgroup.content->group);
```
bluca, line 704: This doesn't look right to me - it shouldn't blindly to a fixed-size copy, given the length is given. It could go over and copy in garbage. And there's already a check on `length_` at the beginning which guarantees there's enough room. Where exactly are you getting build failures? GCC 8 and newer is used in various places without issues

awanishgolwara: If the length of the source is shorter than length_, the NULL character at the end of the source will also get copied into the destination. So, there should not be any problem with copying some additional garbage after the NULL character (though I agree it should not really be needed). As far as a check on length_ you have mentioned, this is a runtime check whereas the warning/error is during the compilation (static analysis) time and not runtime.

I am getting into this gcc issue just during the compilation of this library using gcc cross compilers https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads

bluca: I actually can't reproduce this warning anywhere, tried with a bunch of compilers - are you sure it's not a temporary false positive from that particular ARM's build?
The existing check is a runtime one, but comparing against a fixed size, which corresponds to the struct size - the compiler should be able to figure out it's safe to do so, and it seems it is even with latest versions of gcc, which is very picky on this stuff

bluca:
no answer in a year, no warning can be seen even on very recent compilers, so closing

PR #4205, opened 2021-06-06
problem: no way to know when connection is temporarly dropped

somdoron:
This is important in order to send the login sequence of a client to the server.

Solution: add ZMQ_HICCUP_MSG to a socket, socket would send that message whenever a connection get temporarly disconnected

### DIFF
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -615,3 +615,15 @@ void zmq::pipe_t::set_disconnect_msg (
       _disconnect_msg.init_buffer (&disconnect_[0], disconnect_.size ());
     errno_assert (rc == 0);
 }
+
+void zmq::pipe_t::send_hiccup_msg (const std::vector<unsigned char> &hiccup_)
+{
+    if (!hiccup_.empty () && _out_pipe) {
+        msg_t msg;
+        const int rc = msg.init_buffer (&hiccup_[0], hiccup_.size ());
+        errno_assert (rc == 0);
+
+        _out_pipe->write (msg, false);
+        flush ();
+    }
+}
--- a/tests/test_hiccup_msg.cpp
+++ b/tests/test_hiccup_msg.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
+    Copyright (c) 2007-2021 Contributors as noted in the AUTHORS file
 
     This file is part of libzmq, the ZeroMQ core engine in C++.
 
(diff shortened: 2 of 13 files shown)


tests/test_hiccup_msg.cpp
```diff
@@ -0,0 +1,69 @@
+/*
+    Copyright (c) 2007-2016 Contributors as noted in the AUTHORS file
```
bluca, line 2: 2021

tests/test_hiccup_msg.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+SETUP_TEARDOWN_TESTCONTEXT
+
+void test ()
+{
+    //  Create a server
+    void *server = test_context_socket (ZMQ_SERVER);
+
+    //  bind server
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_bind (server, "tcp://127.0.0.1:5569"));
```
bluca, line 41: please bind to 127.0.0.1:* and use the get_endpoint, otherwise the test will break on shared build systems

src/pipe.cpp
```diff
@@ -615,3 +615,19 @@ void zmq::pipe_t::set_disconnect_msg (
...
 }
+
+void zmq::pipe_t::send_hiccup_msg (const std::vector<unsigned char> &hiccup_)
+{   
+    if (!hiccup_.empty () && _out_pipe) {
+        msg_t msg;
+        const int rc = msg.init_buffer (&hiccup_[0], hiccup_.size ());
+        errno_assert (rc == 0);
+ 
+        // Rollback any incomplete message in the pipe, and push the disconnect message.
```
bluca, line 626: Should this behaviour be documented in the manpage? It results in lost (half-received) messages, right?

bluca:
Formatter is unhappy:

```
[100%] Formatting with clang-format (using /usr/lib/llvm-8/bin/clang-format) and showing differences with latest commit
diff --git a/src/pipe.cpp b/src/pipe.cpp
index 4c3f7c9..d785c98 100644
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -617,17 +617,16 @@ void zmq::pipe_t::set_disconnect_msg (
 }
 
 void zmq::pipe_t::send_hiccup_msg (const std::vector<unsigned char> &hiccup_)
-{   
+{
     if (!hiccup_.empty () && _out_pipe) {
         msg_t msg;
         const int rc = msg.init_buffer (&hiccup_[0], hiccup_.size ());
         errno_assert (rc == 0);
- 
+
         // Rollback any incomplete message in the pipe, and push the disconnect message.
         rollback ();
 
         _out_pipe->write (msg, false);
-        flush ();        
+        flush ();
     }
 }
-
diff --git a/src/session_base.cpp b/src/session_base.cpp
index 7804c10..f645214 100644
--- a/src/session_base.cpp
+++ b/src/session_base.cpp
@@ -471,7 +471,7 @@ void zmq::session_base_t::engine_error (bool handshaked_,
         if (_active && handshaked_ && options.can_recv_hiccup_msg
             && !options.hiccup_msg.empty ()) {
             _pipe->send_hiccup_msg (options.hiccup_msg);
-        }            
+        }
     }
 
     zmq_assert (reason_ == i_engine::connection_error
diff --git a/tests/test_hiccup_msg.cpp b/tests/test_hiccup_msg.cpp
index 3c8445f..485f95a 100644
--- a/tests/test_hiccup_msg.cpp
+++ b/tests/test_hiccup_msg.cpp
@@ -42,8 +42,10 @@ void test ()
 
     //  Create a client
     void *client = test_context_socket (ZMQ_CLIENT);
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (client, ZMQ_HELLO_MSG, "HELLO", 5));
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_setsockopt (client, ZMQ_HICCUP_MSG, "HICCUP", 6));
+    TEST_ASSERT_SUCCESS_ERRNO (
+      zmq_setsockopt (client, ZMQ_HELLO_MSG, "HELLO", 5));
+    TEST_ASSERT_SUCCESS_ERRNO (
+      zmq_setsockopt (client, ZMQ_HICCUP_MSG, "HICCUP", 6));
     TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (client, "tcp://127.0.0.1:5569"));
 
     // Receive the hello message from client
```

src/pipe.cpp
somdoron: no, because those messages were not flushed yet. Actually, it is not needed, as `clean_pipes` already did that.

PR #4223, opened 2021-06-25
fixes for UBSAN warnings

bill-torpey:
There's recently been a bit of talk around undefined behavior in C/C++ (e.g., in [ACM Queue](https://queue.acm.org/detail.cfm?id=3468263)), which prompted us to update our [test tools](https://github.com/nyfix/memleaktest) to include clang's [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html).

Running UBSAN against our code (and dependencies) revealed a number of reports of UB in libzmq: 

```
~/ubsan$ cuc.sh | awk '/misalign/{getline; print}' | sort -u
	zmq::yqueue_t<zmq::command_t, 16>::back() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:101:25
	zmq::yqueue_t<zmq::command_t, 16>::front() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:97:26
	zmq::yqueue_t<zmq::command_t, 16>::pop() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:161:42
	zmq::yqueue_t<zmq::command_t, 16>::pop() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:162:32
	zmq::yqueue_t<zmq::command_t, 16>::push() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:114:25
	zmq::yqueue_t<zmq::command_t, 16>::push() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:115:22
	zmq::yqueue_t<zmq::command_t, 16>::push() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:119:25
	zmq::yqueue_t<zmq::command_t, 16>::push() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:119:36
	zmq::yqueue_t<zmq::command_t, 16>::push() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:121:34
	zmq::yqueue_t<zmq::command_t, 16>::~yqueue_t() /home/btorpey/work/libzmq/master/src/src/yqueue.hpp:87:42
~/ubsan$ 
```

All of these point back to the `command_t` struct:

```
ID=7a95132b5a9dafc86223265ddb37e29e Count=2 Error="reference binding to misaligned address 0x0000016e15e0 for type 'zmq::command_t', which requires 64 byte alignment"
```

The `command_t` struct is *declared* to be aligned on a 64-byte boundary:

- On Linux, using the `aligned(64)` attribute
- On Windows, using `declspec(align(64))`

Note that either of these decorations affect _only_ static and automatic variables -- they do _not_ affect dynamic (heap) allocations (see below for sample code that illustrates this). 

- On Linux, the code `posix_memalign` function is used (sometimes) to allocate the struct.
  - But only in autotools builds 
- On Windows, no attempt is made to allocate the struct with the proper alignment.
- In both Linux and Windows the struct alignment is hard-coded to 64 bytes -- rather than using `ZMQ_CACHELINE_SIZE`

So, (a) we're not necessarily getting the optimization we think we are (only autotools builds on Linux), and (b) in the other cases we're lying to the compiler about the alignment of `command_t/chunk_t` structs.

Strictly speaking, the incorrect alignment issue is undefined behavior.  That doesn't appear to have any effect at this time, but that can change without warning at any point.  

This PR does the following:

1.  Adds a check to `CMakeLists.txt` to determine if `posix_memalign` is available, and adding `HAVE_POSIX_MEMALIGN` to the `platform.hpp.in` file if so.
2.  A similar approach is used to expose `_aligned_malloc` on Windows.  
3.  The Linux (`aligned`) attribute is made dependent on the value of `HAVE_POSIX_MEMALIGN`.
4.  The Windows (`declspec(align))` attribute is made dependent on the value of `HAVE_ALIGNED_MALLOC`.
5.  Both the Linux (`aligned`) and Windows (`declspec(align))` attributes are changed to use `ZMQ_CACHELINE_SIZE` instead of the hard-coded value 64.
 

## Code Generation
It's not at all clear whether the attributes mentioned influence compiler code generation:

  - [Older versions of the gcc docs](https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Type-Attributes.html) "hint" that the `aligned` attribute can affect code generation:

    > On a SPARC, having all variables of type struct S aligned to 8-byte boundaries allows the compiler to use the ldd and std (doubleword load and store) instructions when copying one variable of type struct S to another, thus improving run-time efficiency. 
    
  - [More recent versions of the gcc docs](https://gcc.gnu.org/onlinedocs/gcc-9.4.0/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes) no longer mention code generation in conjunction with the `aligned` attribute.

## Signed => unsigned conversions
The following example is not strictly UB, but it is flagged by UBSAN as "suspicious" (we build with `-fsanitize=implicit-conversion` and `-fsanitize=integer`, which both activate `-fsanitize=implicit-integer-sign-change`).

In our tests, UBSAN issued the following warnings (note that we post-process UBSAN output [using the scripts found here](https://github.com/nyfix/memleaktest)):

```
ID=611cf205308f0e242dc476c8f3bb25fa Error=implicit conversion from type 'int' of value -5 (32-bit, signed) to type 'unsigned int' changed the value to 4294967291 (32-bit, unsigned)
#   zmq::epoll_t::reset_pollout(void*) /home/btorpey/work/libzmq/master/src/src/epoll.cpp:152:22
#   zmq::io_object_t::reset_pollout(void*) /home/btorpey/work/libzmq/master/src/src/io_object.cpp:90:14
#   zmq::stream_engine_base_t::reset_pollout() /home/btorpey/work/libzmq/master/src/src/stream_engine_base.hpp:125:42
#   zmq::stream_engine_base_t::out_event() /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:395:13
#   zmq::epoll_t::loop() /home/btorpey/work/libzmq/master/src/src/epoll.cpp:202:29
#   zmq::worker_poller_base_t::worker_routine(void*) /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:146:51
#   thread_routine(void*) /home/btorpey/work/libzmq/master/src/src/thread.cpp:257:5
#   start_thread /build/glibc-eX1tMB/glibc-2.31/nptl/pthread_create.c:477:8
#   clone /build/glibc-eX1tMB/glibc-2.31/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

The offending code is the `static_cast` to `short` below:

```
void zmq::epoll_t::reset_pollout (handle_t handle_)
{
    check_thread ();
    poll_entry_t *pe = static_cast<poll_entry_t *> (handle_);
    pe->ev.events &= ~(static_cast<short> (EPOLLOUT));
    const int rc = epoll_ctl (_epoll_fd, EPOLL_CTL_MOD, pe->fd, &pe->ev);
    errno_assert (rc != -1);
}
```

The fix is to `static_cast` to an unsigned type to avoid the overflow:

```
void zmq::epoll_t::reset_pollout (handle_t handle_)
{
    check_thread ();
    poll_entry_t *pe = static_cast<poll_entry_t *> (handle_);
    pe->ev.events &= ~(static_cast<uint32_t> (EPOLLOUT));
    const int rc = epoll_ctl (_epoll_fd, EPOLL_CTL_MOD, pe->fd, &pe->ev);
    errno_assert (rc != -1);
}
```

A similar situation exists here:

```
ID=eb9df2d868b7169c621d28afebbb8d2e Error=implicit conversion from type 'int' of value -2 (32-bit, signed) to type 'unsigned int' changed the value to 4294967294 (32-bit, unsigned)
#   zmq::epoll_t::reset_pollin(void*) /home/btorpey/work/libzmq/master/src/src/epoll.cpp:134:22
#   zmq::io_object_t::reset_pollin(void*) /home/btorpey/work/libzmq/master/src/src/io_object.cpp:80:14
#   zmq::stream_engine_base_t::in_event_internal() /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:322:9
#   zmq::stream_engine_base_t::in_event() /home/btorpey/work/libzmq/master/src/src/stream_engine_base.cpp:243:22
#   zmq::epoll_t::loop() /home/btorpey/work/libzmq/master/src/src/epoll.cpp:206:29
#   zmq::worker_poller_base_t::worker_routine(void*) /home/btorpey/work/libzmq/master/src/src/poller_base.cpp:146:51
#   thread_routine(void*) /home/btorpey/work/libzmq/master/src/src/thread.cpp:257:5
#   start_thread /build/glibc-eX1tMB/glibc-2.31/nptl/pthread_create.c:477:8
#   clone /build/glibc-eX1tMB/glibc-2.31/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Again, while this is *not* technically UB, it is potentially confusing.  (Not to mention, fixing the problem means we don't have to suppress it).

## Caveats

Some caveats for this PR:

- Only addresses cmake builds.
- The Windows changes *should* work, but I have no way to test these, so I'm hoping that others can please verify. 
- I also have no way of testing the scenario where `posix_memalign` is _not_ available -- the oldest system I have access to is CentOS 6, and that includes the function.

## Appendix:  Sample code re: dynamic allocation 

This demonstrates that the aligned attribute is only effective with static and automatic variables:

```
#include <stdio.h>

class X {
   int member;
} __attribute__ ((aligned (1024)));


int main()
{
   X x;
   printf("&x = %p\n", &x);

   X* x1 = new X();
   printf("&x1 = %p\n", &x1);

   X* x2 = new X();
   printf("&x2 = %p\n", &x2);
}
```
 
When compiled (with gcc 9.3.0 on Ubuntu 20.04), this gives:

```
btorpey@bt-brixu:~/work/test$ ./a.out
&x = 0x7fff68b17400
&x1 = 0x7fff68b173f0
&x2 = 0x7fff68b173f8
btorpey@bt-brixu:~/work/test$
```

What's happening here is that the local `x` is being aligned on a 1K-byte boundary, but *not* the instances allocated on the heap.  In order for the heap allocations to be aligned, the code would need to use `posix_memalign` (or equivalent) to allocate the memory.


bluca:
It looks like the Windows builds/tests are unhappy.

Also, several of the source files are getting the +x attribute, could you please fix that?

bill-torpey:
> It looks like the Windows builds/tests are unhappy.

Unfortunately I have no access to Windows.  AFAICT the changes in the PR should not cause any issues, but as mentioned in the PR I'll need some help from a Windows wiz for that side of things.

> Also, several of the source files are getting the +x attribute, could you please fix that?

Sorry -- done.

bluca:
> > It looks like the Windows builds/tests are unhappy.
> 
> Unfortunately I have no access to Windows. AFAICT the changes in the PR should not cause any issues, but as mentioned in the PR I'll need some help from a Windows wiz for that side of things.

Any chance you could remove the Windows changes then, so that we can merge it? I can't break the build - you can then open another Windows-specific follow-up, until someone helps with it

bill-torpey:
> Any chance you could remove the Windows changes then, so that we can merge it? I can't break the build - you can then open another Windows-specific follow-up, until someone helps with it

Would converting PR to a draft turn off the CI?  That might be a better approach

bluca:
> > Any chance you could remove the Windows changes then, so that we can merge it? I can't break the build - you can then open another Windows-specific follow-up, until someone helps with it
> 
> Would converting PR to a draft turn off the CI? That might be a better approach

It wouldn't - but it wouldn't solve the problem that I can't merge it, as it would introduce a regression for Windows users

bill-torpey:
I'm going to close the PR for now -- hopefully that will stop the bots.  

This really should be applied atomically -- but for better or worse I can't do the Windows bits.

I'll try asking @sigiesec what he thinks -- anyone else on Windows you can suggest?

bill-torpey:
@sigiesec -- When you have a chance can you please take a look at this?  For some reason it's driving the Windows CI crazy -- the odd thing is that the build succeeds, but the tests fail.  Thanks!

bluca:
I think it would be fine to fix the Linux side first by itself

bluca:
It's also fine to keep the PR open

bill-torpey:
"When all else fails, read the instructions" 

I missed this bit in docs for `_aligned_malloc`:

> Use _aligned_free to deallocate memory obtained by both _aligned_malloc and _aligned_offset_malloc. Don't use free, which doesn't reclaim the aligned memory correctly and can lead to hard-to-diagnose bugs.

Removing the test for `_aligned_malloc` allows the CI tests to run succssfully.

I'll prepare a new PR to address the Linux issue, and remove the `declspec(align))` from Windows -- at this point, I see no way to accomplish the desired alignment under Windows without a fair amount of work.

I'll try to post this early next week.

bluca:
sounds good, thanks

bill-torpey:
Re-opening this as that seems simpler than starting over.  Removed the offending Windows changes, which fixes CI on source repo, so should be good to go (I hope).

src/command.hpp
```diff
@@ -218,7 +214,11 @@ __declspec(align (64))
 };
 #pragma warning(pop)
```
bluca, line 215: this needs to be removed too, as it matches the push you removed above

```
C:\projects\libzmq\src\command.hpp(215,17): warning C4193: #pragma warning(pop): no matching '#pragma warning(push)' (compiling source file C:\projects\libzmq\src\channel.cpp) [C:\projects\build_libzmq\libzmq.vcxproj]
```

PR #4224, opened 2021-06-27
Remove extra license file with GPL

kaplanlior:
ZeroMQ is licensed under LGPL (http://wiki.zeromq.org/area:licensing and
source files). Yet, we distribute COPYING file with GPL and referencing
it in the README. Move COPYING.LESSER to COPYING.

Some automated scanner might get confused and organizations disqualify
ZeroMQ due to this.


bluca:
AFAIK it is common to have both files for LGPL licensed projects?

bluca:
https://www.gnu.org/licenses/gpl-howto.en.html

> If you are releasing your program under the Lesser GPL, you should also include the text version of the LGPL, usually in a file called COPYING.LESSER. Please note that, since the LGPL is a set of additional permissions on top of the GPL, it's crucial to include both licenses so users have all the materials they need to understand their rights.

From COPYING.LESSER second paragraph:

>   This version of the GNU Lesser General Public License incorporates
the terms and conditions of version 3 of the GNU General Public
License, supplemented by the additional permissions listed below.

So this is actually required to be valid. If automated scanners are getting confused, they have a bug - please file a report with the developers of such tools.

bluca:
To be clear: it is not enough to declare a license, the software also must ship it together with the sources integrally.

kaplanlior:
Fair point about adding both licenses (will be fixed), but the files names aren't mandatory.

COPYING is being detected as the default / main license on GitHub. If you click "view license" in the repository, you'll get GPL from COPYING instead of LGPL from COPYING.LESSER.

Also, the readme file should reference LGLP only to avoid this mislead in putting GPL file first.
I would use COPYING

bluca:
The readme needs to reference both, because again they _both_ together form the license terms. The naming is the standard convention, most projects I have seen use this format. If you want to add _additional_ words to the README to clarify this, feel free to open a new PR as such.

kaplanlior:
Without the additional exception to the LGPL, the auto detection would work (e.g. [go-ethereum](https://github.com/ethereum/go-ethereum), [glances](https://github.com/nicolargo/glances) and [radare2](https://github.com/radareorg/radare2)), with the exception we should use other names to make it clear and to make sure we don't confuse users (e.g. [wkhtmltopdf](https://github.com/wkhtmltopdf/wkhtmltopdf), [qt](https://github.com/therecipe/qt) and [coursera-dl](https://github.com/coursera-dl/coursera-dl)).

Even if LGPL "extends" GPL, we should reference LGPL  first and make sure the user can access GPL. Otherwise, it creates confusion to most.

PR #4227, opened 2021-06-30
Use standard C and C++

ChrisThrasher:
I removed references to GNU language extensions since it appears that language extensions are not required to build the codebase. It seemed reasonable to better enforce the use of standard C and C++ since this codebase ought to be as portable as possible.


ChrisThrasher:
Because I can only test a small subset of all possible configurations, I'd really appreciate approval to run the entire workflow. Thank you for taking a look at my pull request!

configure.ac
```diff
@@ -90,7 +90,7 @@ AC_ARG_ENABLE([force-CXX98-compat],
     [AS_HELP_STRING([--enable-force-CXX98-compat], [force C++98 build [default=disabled]])])
 if test "x$enable_force_CXX98_compat" = "xyes"; then
     AC_LANG_PUSH([C++])
-    AX_CHECK_COMPILE_FLAG([-std=gnu++98], [CXXFLAGS+=" -std=gnu++98"], [])
+    AX_CHECK_COMPILE_FLAG([-std=c++98], [CXXFLAGS+=" -std=c++98"], [])
```
bluca, line 93: I'm not entirely convinced about the c11, but maybe that's close enough to gnu11, and it should be well tested as it's the default. But please do not change gnu++98/gnu99, as that's fallback for very old platforms that are quite fragile. It might be that on modern compilers they are the same, but that's not the case on ancient gcc versions and so on, which are very difficult to test.

bluca:

configure.ac
ChrisThrasher: Should I retain gnu++98 in the CMake build as well?

bluca: yes, please

ChrisThrasher: Done and pushed. I rephrased the commit to better articulate how we're only using standard C11 and C++11 but older compilation modes retain GNU extensions to ensure we don't break with old compilers.

PR #4242, opened 2021-08-13
Problem: calling randombytes_close with libsodium can crash Contexts in other threads

### LINKED ISSUE #4241: threadafety issue with curve_keypair, libsodium, randombytes_close

# Issue description

#3001 reverted the refcounting of libsodium init/close, meaning that `randombytes_close()` can be called many times while libsodium is still in use, e.g. if `zmq_curve_keypair()` is called after `zmq_context()`.

When running with libsodium, calling curve_keypair triggers `zmq::random_close()` which in turn unconditionally triggers `randombytes_close()`. Calls to e.g. `crypto_box_keypair()` for the internal keys in `curve_server` and `curve_client` do not call `zmq::random_open()` which means the following sequence occurs:

- zmq_context(): `sodium_init()`
- zmq_curve_keypair(): `sodium_init(); randombytes_close()`
- zmq_connect(): `crypto_box_keypair()` with randombytes closed!

The same threadsafety issue can occur if one context is closed while another is in use.

This is not *always* a problem, as `randombytes_close()` is a no-op when `getrandom` is available, but this isn't always the case (e.g. macOS). This `crypto_box_keypair()` call can crash due to thread unsafety reinitializing the random data source.

From the [libsodium docs](https://doc.libsodium.org/generating_random_data#usage):

> Explicitly calling \[[randombytes_close][]\] is almost never required.

[randombytes_close]: https://doc.libsodium.org/generating_random_data#usage

It seems like the 'right' fix is to just remove the call to `randombytes_close()` when using libsodium. Doing so could leave at most one FD open on `/dev/urandom`, but not more than one. Otherwise, we need to restore the refcounting in #2636 for libsodium, which was removed in #3001, and/or call `random::open/close` on every curve call, not just context construct/destruct.

# Environment

* libzmq version (commit hash if unreleased): 4.3.4
* libsodium version: 1.0.18
* OS: macOS 11.1, linux (ubuntu-20.04 on GitHub Actions)

Create a CURVE server and keep connecting clients until it crashes. Make sure to call `zmq_curve_keypair()` after starting a context to trigger the inappropraite use-after-close. In Python:

```python
import itertools
import time
from threading import Thread

import zmq

server_public, server_private = zmq.curve_keypair()

url = 'tcp://127.0.0.1:5555'


def server():
    with zmq.Context() as ctx:
        with ctx.socket(zmq.PULL) as s:
            s.curve_server = True
            s.curve_publickey = server_public
            s.curve_secretkey = server_private
            s.bind(url)
            count = 0
            while True:
                try:
                    print(count, s.recv().decode("ascii"))
                except KeyboardInterrupt:
                    print("interrupted...")
                else:
                    count += 1


def client(ctx, i):
    client_public, client_private = zmq.curve_keypair() # calls randombytes_close
    with ctx.socket(zmq.PUSH) as s:
        s.curve_serverkey = server_public
        s.curve_publickey = client_public
        s.curve_secretkey = client_private
        s.connect(url) # uses closed randombytes, can crash
        s.send(str(i).encode('ascii'))


def main():
    Thread(target=server, daemon=True).start()
    time.sleep(1)
    with zmq.Context() as ctx:
        for i in itertools.count():
            # at most 100 sockets/sec
            # to avoid FD exhaustion waiting for message delivery
            time.sleep(1e-2)
            client(ctx, i)


if __name__ == "__main__":
    main()
```

After a number of sockets (varies from 4 to 1000), crash with:

```
  * frame #0: 0x00007fff2037392e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff203a25bd libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff202f7406 libsystem_c.dylib`abort + 125
    frame #3: 0x0000000101e5c531 libsodium.23.dylib`sodium_misuse + 63
    frame #4: 0x0000000101e666fc libsodium.23.dylib`randombytes_sysrandom_stir + 167
    frame #5: 0x0000000101e66764 libsodium.23.dylib`randombytes_sysrandom_buf + 34
    frame #6: 0x0000000101e44f1a libsodium.23.dylib`crypto_box_curve25519xsalsa20poly1305_keypair + 26
    frame #7: 0x0000000101d66922 libzmq.5.dylib`zmq::curve_server_t::curve_server_t(this=0x0000000101045a00, session_=0x000000010103ec00, peer_address_=<unavailable>, options_=0x0000000101030618, downgrade_sub_=false) at curve_server.cpp:62:10 [opt]
    frame #8: 0x0000000101db540c libzmq.5.dylib`zmq::zmtp_engine_t::handshake_v3_x(this=0x0000000101030600, downgrade_sub_=false) at zmtp_engine.cpp:372:45 [opt]
    frame #9: 0x0000000101db52da libzmq.5.dylib`zmq::zmtp_engine_t::handshake_v3_1(this=<unavailable>) at zmtp_engine.cpp:428:32 [opt] [artificial]
    frame #10: 0x0000000101db46ca libzmq.5.dylib`zmq::zmtp_engine_t::handshake(this=0x0000000101030600) at zmtp_engine.cpp:132:10 [opt]
    frame #11: 0x0000000101d9bd0d libzmq.5.dylib`zmq::stream_engine_base_t::in_event_internal(this=0x0000000101030600) at stream_engine_base.cpp:253:13 [opt]
    frame #12: 0x0000000101d70a3b libzmq.5.dylib`zmq::kqueue_t::loop(this=0x00000001019274f0) at kqueue.cpp:218:30 [opt]
    frame #13: 0x0000000101da052d libzmq.5.dylib`thread_routine(arg_=0x0000000101927530) at thread.cpp:256:5 [opt]
    frame #14: 0x00007fff203a28fc libsystem_pthread.dylib`_pthread_start + 224
    frame #15: 0x00007fff2039e443 libsystem_pthread.dylib`thread_start + 15
```

# What's the expected result?

No crash.

minrk:
**Problem:** When using libsodium, calling `zmq::random_close()` closes a shared resources that may still be in use.

See #4241 for details and repro, but this is called unconditionally on every `curve_keypair()` and every context destructor, since #3001. Closing any context and/or issuing a new curve keypair can crash another context if it is working with curve-enabled sockets at the time, as it is still a consumer of a now-closed resource. libsodium will try to reinitialize the source, so it doesn't crash every time; instead, certain race conditions must be met to result in a crash.

**Solution:** Do not close this shared resource while it is still in use, as suggested by the libsodium [documentation](https://doc.libsodium.org/generating_random_data#usage).

For implementations using /dev/[u]random, this can leave up to one leftover FD per process (restores #2632, but should be bounded to one).

If we want to call `randombytes_close()`, we must restore the use of refcounting in #2636 when using libsodium to avoid crashes.

closes #4241

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,7 @@ endif()
 
 option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" ON)
 option(WITH_LIBSODIUM_STATIC "Use static libsodium library" OFF)
+option(ENABLE_LIBSODIUM_RANDOMBYTES_CLOSE "Automatically close libsodium randombytes. Not threadsafe without getrandom()" ON)
 option(ENABLE_CURVE "Enable CURVE security" ON)
 
 if(ENABLE_CURVE)
@@ -271,6 +272,9 @@ if(ENABLE_CURVE)
       endif()
       set(ZMQ_USE_LIBSODIUM 1)
       set(ZMQ_HAVE_CURVE 1)
+      if (ENABLE_LIBSODIUM_RANDOMBYTES_CLOSE)
+        set(ZMQ_LIBSODIUM_RANDOMBYTES_CLOSE 1)
+      endif()
     else()
       message(
         WARNING
--- a/configure.ac
+++ b/configure.ac
@@ -550,6 +550,20 @@ AS_IF([test "x$with_libsodium" = "xyes"], [
 AC_ARG_ENABLE([curve],
     [AS_HELP_STRING([--disable-curve], [disable CURVE security [default=no]])])
 
+AC_ARG_ENABLE(
+    [libsodium_randombytes_close],
+    [AS_HELP_STRING(
+        [--disable-libsodium_randombytes_close],
+        [Do not call libsodium randombytes_close() when terminating contexts.
+        If disabled, may leave one FD open on /dev/urandom
+        until randombytes_close() is called explicitly,
+        but fixes a crash when multiple contexts are used with CURVE.
+        Has no effect when getrandom() is available. [default=enabled]]
+    )],
+    [],
+    [enable_libsodium_randombytes_close=yes]
+)
+
 if test "x$enable_curve" = "xno"; then
     curve_library=""
     AC_MSG_NOTICE([CURVE security is disabled])
@@ -558,6 +572,12 @@ elif test "x$with_libsodium" = "xyes"; then
     AC_MSG_NOTICE([Using libsodium for CURVE security])
     AC_DEFINE(ZMQ_HAVE_CURVE, [1], [Using curve encryption])
     AC_DEFINE(ZMQ_USE_LIBSODIUM, [1], [Using libsodium for curve encryption])
+    if test "x$enable_libsodium_randombytes_close" = "xyes"; then
+        AC_DEFINE(ZMQ_LIBSODIUM_RANDOMBYTES_CLOSE, [1], [Automatically close libsodium randombytes. Not threadsafe without getrandom()])
+    else
+        AC_MSG_NOTICE([Disabling libsodium randombytes_close(). randombytes_close() may need to be called in applcation code.])
+    fi
+
     curve_library="libsodium"
     enable_curve="yes"
 
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -151,8 +151,6 @@ static void manage_random (bool init_)
     if (init_) {
         int rc = sodium_init ();
         zmq_assert (rc != -1);
-    } else {
-        randombytes_close ();
     }
 #else
     LIBZMQ_UNUSED (init_);
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -151,6 +151,13 @@ static void manage_random (bool init_)
     if (init_) {
         int rc = sodium_init ();
         zmq_assert (rc != -1);
+#if defined(ZMQ_LIBSODIUM_RANDOMBYTES_CLOSE)
+    } else {
+        // randombytes_close either a no-op or not threadsafe
+        // doing this without refcounting can cause crashes
+        // if called while a context is active
+        randombytes_close ();
+#endif
     }
 #else
     LIBZMQ_UNUSED (init_);


bluca:
Unfortunately this is not a solution, because then we go back to leaking file descriptors, which trigger a cascade of other problems (selinux errors, etc etc)

minrk:
OK, then the fix is to revert #3001

bluca:
Eh I wish - that breaks a whole other host of cases due to C++ broken global initialization mechanisms...

minrk:
I don't fully understand how the refcounting broke things, but to be correct, libsodium must have a refcounter to protect the call to `randombytes_close()`, just like tweetnacl. I don't know how to make it properly global to avoid #2991, but this means refcounting is clearly required with libzmq, unfortunately.

As it is now, one must ensure that once a context is active dealing with curve messages, no other contexts can be terminated and no new keypairs issued.

bluca:
It's explained in the comments - it's not possible to do refcounting via globals in C++. The only solution is to get a kernel with getrandom(), everything else fixes one things and breaks another, which is not acceptable, as we really need to keep things stable.

minrk:
>  The only solution is to get a kernel with getrandom()

That's not something folks usually have control over, so not really a solution.

> everything else fixes one things and breaks another, which is not acceptable, as we really need to keep things stable.

Yeah, I get that. Even fixing wrong behavior can break things.

WDYT about a runtime and/or compile-time opt-out of the random_close? When `getrandom` is available, removing it changes nothing. When it's not, you trade a crash for a (bounded to one process-wide) leak. If the leak were opt-in, it seems reasonable to expect the calling code to manage closing the shared resource, in the rare cases where that's useful.

The 'wrong' behavior of the leak is easy to workaround (close the FD) and of little consequence (leak up to one FD in the lifetime of the process), whereas the current behavior is not possible to workaround except with "get a different kernel".

bluca:
> > The only solution is to get a kernel with getrandom()
> 
> That's not something folks usually have control over, so not really a solution.
> 
> > everything else fixes one things and breaks another, which is not acceptable, as we really need to keep things stable.
> 
> Yeah, I get that. Even fixing wrong behavior can break things.
> 
> WDYT about a runtime and/or compile-time opt-out of the random_close? When `getrandom` is available, removing it changes nothing. When it's not, you trade a crash for a (bounded to one process-wide) leak. If the leak were opt-in, it seems reasonable to expect the calling code to manage closing the shared resource, in the rare cases where that's useful.
> 
> The 'wrong' behavior of the leak is easy to workaround (close the FD) and of little consequence (leak up to one FD in the lifetime of the process), whereas the current behavior is not possible to workaround except with "get a different kernel".

A build option would be fine. Note that the leak is not just cosmetic, things like selinux will raise an audit and so on.

minrk:
I added an opt-in to cmake and autotools (my m4 is rusty, so I hope it's right). As you might guess, I would prefer the non-crashing behavior be the default, but I assume no change by default is the right choice for consistency.

I wonder if a runtime `zmq_random_open()` / `zmq_random_close()` that allowed applications to call `sodium_init` / `randombytes_close()` and take it out of the hands of `zmq::random_open/close` that have refcounting troubles would be best?

> things like selinux will raise an audit and so on.

Do you have a link to more information on that? If I have application code that does:

```c
zmq_term() // doesn't call randombytes_close() anymore
randombytes_close()
```

would that avoid the audit issue, or no? I don't have experience with selinux, so don't know exactly what's being audited there.

CMakeLists.txt
```diff
@@ -258,6 +258,7 @@ endif()
 
 option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" ON)
 option(WITH_LIBSODIUM_STATIC "Use static libsodium library" OFF)
+option(ENABLE_LIBSODIUM_RANDOMBYTES_CLOSE "Automatically closing libsodium randombytes (not threadsafe)" ON)
```
bluca, line 261: If you want to list the explanation, please be precise "not threadsafe without getrandom()"

minrk:
Do you think it's better to restore the refcounting instead o disabling entirely (i.e. exactly reverting #3001 if the new flag is set)?

bluca:
> Do you think it's better to restore the refcounting instead o disabling entirely (i.e. exactly reverting #3001 if the new flag is set)?

Current patch is best - the workaround to call randombytes_close manually is trivial, while safely dealing with the global init shenanigans is extremely convoluted, if at all possible

PR #4255, opened 2021-09-16
add zmq_ppoll

### LINKED ISSUE #4220: discussion: add ppoll
For the [ROOT project](https://github.com/root-project/root), I have built a `ppoll` (or rather pselect for macOS support) based alternative to `poll`. As you probably know, in POSIX, the `p*` variants of `poll`, `select`, and even possibly `recv` and `send`, allow for passing in a signal mask. This allows for easier interaction with POSIX signals, as I (somewhat incoherently, I must admit) [tried to motivate in this blog post](https://blog.esciencecenter.nl/combining-zeromq-posix-signals-b754f6f29cd6). I came by this idea from [a discussion on the blog of Martin Sustrik](https://250bpm.com/blog:12/); unfortunately the crucial comment with the original suggestion is missing from the blog since it was moved recently, but at the very bottom of the blogpost the idea is still mentioned in the second EDIT paragraph.

To recap, the basic idea is that by using `ppoll` (or any `p*` version of a blocking call), you can effectively use POSIX signals as an additional channel to poll. POSIX `ppoll` or `pselect` *atomically* unblocks the signals given by the mask, calls the poll and then blocks the signals again (i.e. these three operations are done atomically together). If you block the given signals outside of the `ppoll` call, this means that these signals are only handled during the blocking call. This means you are guaranteed to not be "bothered" by them in some hard to handle way outside of the blocking call.

The usecase for me is that I want to use a POSIX signal (SIGTERM) to kill child processes that also use ZeroMQ sockets, but don't want it to just cause random EINTRs that could also be caused by other things. However, this is just one particular usecase. I suspect that in general, people that do any kind of signal handling (probably for graceful shutdowns after CTRL-C mostly) may want to also switch to `p*` calls, because it eliminates the need for complicated time-out based handling to circumvent the race condition explained in [Sustrik's blog post](https://250bpm.com/blog:12/).

Long story short, I would love to submit this code to libzmq, but I would need some help/pointers getting it in full shape, because the current implementation is rudimentary. That is: I just got it to work in one particular way on Linux and macOS, but probably have missed many corner cases.

The `ppoll` code can be found in a couple of places (dev branches), but perhaps it's easiest to start from [the code I'm submitting in this PR right now](https://github.com/root-project/root/pull/8385). [The `ppoll` implementation file is here](https://github.com/root-project/root/blob/0c0dc78fbe78e198ace3814fd623802e79d9ea4c/roofit/roofitZMQ/src/ppoll.cpp) (please, don't mind the copyright/license statements, they can/will be removed/changed).

As you can see, it is more or less simply a copy-paste of the libzmq `poll` implementation, except:

1. I force it to build with `ZMQ_POLL_BASED_ON_SELECT`
2. I added the extra `sigmask_` parameter
3. [The signal mask is passed on to `pselect` on this line](https://github.com/root-project/root/blob/0c0dc78fbe78e198ace3814fd623802e79d9ea4c/roofit/roofitZMQ/src/ppoll.cpp#L238).

Since I myself do most of my development on a Mac, I did not try to implement an actual `ppoll` based version (unlike Linux, macOS doesn't have `ppoll`, only `pselect`), but I expect this would be desirable for a complete implementation. The alternative would be to rip out all the parts in the `ZMQ_POLL_BASED_ON_POLL` sections. Apart from this, I haven't tested at all on Windows, nor do I know anything about signals in Windows or whether any of this is relevant there, so I'd need some input there as well.

So, through this issue I hope I can gather your input on this, before I start doing a PR that may not even have a chance of success. Let me know what you think and of course I'd be happy to answer any questions.

egpbos:
zmq_ppoll mostly mimics zmq_poll behavior, except for the added feature of being able to specify a signal mask. Signals in this mask will be blocked during execution of zmq_ppoll. Switching of the process' active signal mask happens atomically with the actual poll call, so that no race conditions can occur. This behavior is useful when one wants to gracefully handle POSIX signals without race conditions. See e.g. the discussion below https://250bpm.com/blog:12/ for an explanation.

Also includes two new tests:
1. test_zmq_ppoll_fd does the same thing as test_zmq_poll_fd, demonstrating backwards compatibility with zmq_poll when used with a default signal mask.
2. test_zmq_ppoll_signals demonstrates the use of zmq_ppoll with a signal mask, blocking out SIGTERM everywhere except in zmq_ppoll, allowing to handle the signal in one place without having to worry about race conditions.

Fixes #4220.

Edit: note for reviewers: to prevent code duplication, I factored out some common parts of `zmq_poll` so that they could be reused in `zmq_ppoll`.


include/zmq.h
```diff
@@ -513,6 +516,7 @@ typedef struct zmq_pollitem_t
 #define ZMQ_POLLITEMS_DFLT 16
 
 ZMQ_EXPORT int zmq_poll (zmq_pollitem_t *items_, int nitems_, long timeout_);
+ZMQ_EXPORT int zmq_ppoll (zmq_pollitem_t *items_, int nitems_, long timeout_, const sigset_t *sigmask_ = nullptr);
```
bluca, line 519: move to draft section, and add ifdefs as it will break windows builds

tests/test_zmq_ppoll_fd.cpp
```diff
@@ -0,0 +1,90 @@
+/*
+    Copyright (c) 2016 Contributors as noted in the AUTHORS file
```
bluca, line 2: 2021

tests/test_zmq_ppoll_signals.cpp
```diff
@@ -0,0 +1,241 @@
...
+        _Exit(0);
+    }
+}
+*/
+
+int main ()
+{
+    UNITY_BEGIN ();
+    RUN_TEST (test_ppoll_signals);
+//    RUN_TEST (test_poll_signals);
```
bluca, line 239: don't add commented out stuff

src/polling_util.hpp
```diff
@@ -108,8 +108,8 @@ typedef int timeout_t;
 
 timeout_t
 compute_timeout (bool first_pass_, long timeout_, uint64_t now_, uint64_t end_);
-
-#elif defined ZMQ_POLL_BASED_ON_SELECT
+#endif
```
bluca, line 111: this does not look right

src/zmq.cpp
```diff
@@ -846,25 +851,10 @@ static int zmq_poller_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
 }
 #endif // ZMQ_HAVE_POLLER
 
-int zmq_poll (zmq_pollitem_t *items_, int nitems_, long timeout_)
```
bluca, line 849: This is a new function, why are you changing all the existing other implementations? Leave them as they are

src/polling_util.hpp
egpbos: I think there are checks to exclude the possibility of ZMQ_POLL_BASED_ON_POLL and ZMQ_POLL_BASED_ON_SELECT being defined at the same time, so I think it makes no difference for those cases. I changed it like this, because ZMQ_POLL_BASED_ON_POLL can be true at the same time as ZMQ_HAVE_PPOLL, because ppoll is currently based purely on pselect, so it is independent of the poller chosen for existing poll functions.

bluca:
autoconf checks are missing

egpbos:
Thanks for the super quick review! Do you prefer rebase/fixup to keep commit history clean? Or should I add a new "fix review comments" commit?

bluca:
rebase

src/zmq.cpp
egpbos: Most of `zmq_poll` and `zmq_ppoll` are identical, the only big difference is the `pselect` call which needs an extra `sigmask` parameter. Factoring out the identical parts prevents code duplication. I can revert if you prefer, of course, but I had imagined it would create an unnecessary maintenance burden.

bluca: avoiding to break the stable api is more important

tests/test_zmq_ppoll_signals.cpp
egpbos: It was meant as an illustration for devs of why it is necessary. I will remove it.

egpbos:
Can you give me a pointer on where I should add autoconf checks? I'm not so familiar with autoconf.

bluca:
grep for the same definitions and you'll see them in acinclude configure.ac et al

include/zmq.h
```diff
@@ -763,6 +766,10 @@ ZMQ_EXPORT int zmq_socket_monitor_versioned (
   void *s_, const char *addr_, uint64_t events_, int event_version_, int type_);
 ZMQ_EXPORT int zmq_socket_monitor_pipes_stats (void *s);
 
+#ifdef ZMQ_HAVE_PPOLL
```
bluca, line 769: Don't ifdef in the public header, otherwise the ABI is different depending on the build. Instead return an error OPNOTSUPP or something like that from the implementation

src/polling_util.hpp
egpbos: I made sure that the `elif` behavior is still obeyed by replacing the following `if` clause with `#if (!defined ZMQ_POLL_BASED_ON_POLL && defined ZMQ_POLL_BASED_ON_SELECT) || defined ZMQ_HAVE_PPOLL`.

Note that the elif is not possible, because then the block is always skipped when `ZMQ_POLL_BASED_ON_POLL`, but the block is also necessary in case of `ZMQ_HAVE_PPOLL`.

src/zmq.cpp
egpbos: Reverted `zmq_poll`.

egpbos:
Addressed some comments, still have to do the autoconf checks.

egpbos:
I added everything to the autoconf build system as well now. Compiles and passes tests on my system, but I'm sure there will be failures on others, as I have little experience with autoconf. Please review :)

include/zmq.h
```diff
@@ -763,6 +766,8 @@ ZMQ_EXPORT int zmq_socket_monitor_versioned (
   void *s_, const char *addr_, uint64_t events_, int event_version_, int type_);
 ZMQ_EXPORT int zmq_socket_monitor_pipes_stats (void *s);
 
+ZMQ_EXPORT int zmq_ppoll (zmq_pollitem_t *items_, int nitems_, long timeout_, const sigset_t *sigmask_ = nullptr);
```
bluca, line 769: needs to be duplicated in src/zmq_drafts.cpp

bluca:
documentation is missing

bluca:
many CI builds are failing

egpbos:
Added documentation and added duplicate declaration in zmq_draft.h. Also removed default nullptr sigmask parameter, I think it's more consistent this way, since no other zmq functions seem to have default parameters.

include/zmq.h
```diff
@@ -763,6 +766,8 @@ ZMQ_EXPORT int zmq_socket_monitor_versioned (
   void *s_, const char *addr_, uint64_t events_, int event_version_, int type_);
 ZMQ_EXPORT int zmq_socket_monitor_pipes_stats (void *s);
 
+ZMQ_EXPORT int zmq_ppoll (zmq_pollitem_t *items_, int nitems_, long timeout_, const sigset_t *sigmask_);
```
bluca, line 769: It doesn't look like this (sigset_t) is compatible with windows, so I'd suggest to just ifdef both this and the signal.h include

egpbos: Ah right, makes sense, most builds were passing before, so this should fix most of them again!

egpbos: What if, to maintain full API in at least some way, we do this?

```suggestion
#if !defined _WIN32
ZMQ_EXPORT int zmq_ppoll (zmq_pollitem_t *items_, int nitems_, long timeout_, const sigset_t *sigmask_);
#else
ZMQ_EXPORT int zmq_ppoll (zmq_pollitem_t *items_, int nitems_, long timeout_, const void *sigmask_);
#endif
```

This way, it's still callable, but it will just return with errno ENOTSUP anyway on Windows, so the type doesn't really matter.

egpbos: I implemented this in the latest force push.

include/zmq.h
```diff
@@ -101,6 +101,9 @@ typedef unsigned __int8 uint8_t;
 #include <stdint.h>
 #endif
 
+// needed for sigset_t definition
+#include <signal.h>
```
bluca, line 105: This should be ifdefd for windows too just in case

bluca:
Builds are still failing

include/zmq.h
egpbos: Will do.

egpbos:
Yes, I'm going to look into failing builds now.

egpbos:
- One of the failures seems to be from some C++11 features I used. Converted relevant parts to C++98.
- Another was a missing include in the `test_zmq_ppoll_signals.cpp`.

Remaining [Windows build failures](https://ci.appveyor.com/project/zeromq/libzmq/builds/40863013) I think are unrelated to this PR:
- The first one (on x64) fails `test_monitor`, which I don't think I touched any parts of.
- The cygwin64 build seems to fail on some sort of external server timeout.
- The mingw64 failure fails `test_srcfd` and `test_unbind_wildcard`, which I also didn't touch.

I don't understand the fuzzer builds, do you see whether those are working as expected?

bluca:
Yes this is somehow breaking the fuzzers build, and it needs to be fixed

bluca:
never mind, I've seen it fail in another PR too now

egpbos:
Latest force push fixes yet another C++11 feature that was failing CentOS_6 builds. Other than this, it seems everything is passing now (or explained by something other than this PR).

bluca:
Can you rebase on latest master? Half of the CI was down due to Travis

egpbos:
Rebased on master.

bluca:
There are new build failures to fix

egpbos:
I wasn't able to fix everything yet, but at least some failures were due to building the ppoll tests with draft mode disabled, fixed this in the CMake and autotools files.

Two other failures are puzzling, configurations:

-  build (ubuntu-latest, cmake, libsodium, enabled, cmake libsodium-dev, enabled) 
-  build (ubuntu-latest, cmake, clang-tidy clang-tools, tweetnacl, enabled, clang++) 

These builds have drafts enabled, but still complain that `zmq_ppoll` is missing:

```
/home/runner/work/libzmq/libzmq/libzmq/zeromq-4.3.5/tests/test_zmq_ppoll_fd.cpp:72:27: error: use of undeclared identifier 'zmq_ppoll'
```

The common factor that differs from the passing ubuntu builds seems to be cmake. Could it be that cmake doesn't correctly pass on the draft enabling?

I will fix the two Windows regressions, I know what's happening there. Also the clang-format build.

egpbos:
@bluca is it possible that something went wrong with the added CI jobs on GitHub Actions? I see in CI.yaml a lot of jobs that have a DRAFT variable, but nothing is done with that variable in the cmake run step. Perhaps this should be ENABLE_DRAFTS? Or am I missing something?

egpbos:
Looks like the CI configuration may indeed be an issue. The fix I applied, which should have fixed non-draft builds only, also worked for the failing draft builds (only failing build on GH actions is now the clang-format one).

egpbos:
Last push should fix clang-format build.

egpbos:
Two remaining failures in three builds:

1. [Debian_9.0/i586](https://build.opensuse.org/package/live_build_log/home:bluca:network:messaging:zeromq:git-draft:PR-4255/libzmq/Debian_9.0/i586) and [Fedora_34/s390x](https://build.opensuse.org/package/live_build_log/home:bluca:network:messaging:zeromq:git-draft:PR-4255/libzmq/Fedora_34/s390x) fail on unrelated `test_ws_transport`
2. [CentOS_7/x86_64](https://build.opensuse.org/package/live_build_log/home:bluca:network:messaging:zeromq:git-draft:PR-4255/libzmq/CentOS_7/x86_64) seems to have hung. It's possible this happens when the 1 second sleep on the child process side isn't enough for the parent side to complete its recv. I will make the test more robust to this failure, since it doesn't matter anyway for the test of ppoll. The ppoll calls have already completed succesfully by this point.

egpbos:
Added fix for hanging CentOS_7/x86_64 build, all builds should now pass, except the `test_ws_transport` ones, which I'm pretty sure should not be caused by this PR.

egpbos:
The previously failing builds now succeed.

Fuzzers failing now, but this seems spurious and happens on other recent PRs as well. They did pass on the previous run.

Ready to merge?

PR #4258, opened 2021-09-21
add missing override statements

timblechmann:
silences inconsistent-missing-override warnings for me


bluca:

PR #4263, opened 2021-09-23
Fixing cmake_minimum_required() and project() calls order.

risa2000:
According to the CMake doc for [`cmake_minimum_required()`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html?highlight=project) and [`project()`](https://cmake.org/cmake/help/latest/command/project.html#command:project), the commands should be called in this order:

1) `cmake_minimum_required()`
2) `project()`

in order to correctly set up the policies for the project. Otherwise it can break the policy settings as described here: https://gitlab.kitware.com/cmake/cmake/-/issues/22671

Moving the conditional expression before `project()` however does not work because `CMAKE_SYSTEM_NAME` is not defined at this point. So the minimum version is bumped to the higher one for all platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # CMake build script for ZeroMQ
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 2.8.12)
 
 project(ZeroMQ)
 


CMakeLists.txt
```diff
@@ -1,11 +1,7 @@
 # CMake build script for ZeroMQ
-project(ZeroMQ)
+cmake_minimum_required(VERSION 3.0.2)
```
bluca, line 2: 2.8.12

risa2000: I have pushed the new commit into the patch with the lower version required (2.8.12).

CMakeLists.txt
```diff
@@ -1,11 +1,7 @@
 # CMake build script for ZeroMQ
-project(ZeroMQ)
+cmake_minimum_required(VERSION 2.8.12)
```
bluca, line 2: the conditional also is needed otherwise osx breaks

bluca: that doesn't work, as it breaks osx, and breaking backward compatibility doesn't work either

risa2000: According to the CMake maintainer (https://gitlab.kitware.com/cmake/cmake/-/issues/22671#note_1020156) the current state is broken as well.

bluca: That's self-evidently not true, given it does work just fine

bluca: So you are saying that it's not possible to support all 3 of these:

- keep backward compatibility
- keep OSX working
- add support for setting policies on the command line

That still sounds like a cmake bug/limitation to me. But anyway, the first two are really not negotiable, so you'll need to find a different fix if you want the 3rd one to work I'm afraid.

risa2000: In a way yes, if upping the minimum CMake version to the higher common value (3.0.8) is not acceptable. I am not however that intimate with CMake to claim what exactly does not work when the things are called in the wrong order. Setting the policies on the command line not working could be just one of the consequences. Anyway, your call.

PR #4266, opened 2021-10-06
Manually disable symbol export

egpbos:
Problem: when linking to libzmq in my project, I want zmq symbols to remain invisible to users of my library. There is no way to control this, since visibility is set automatically through `ZMQ_EXPORT` based on OS and compiler.

Solution: add a preprocessor variable `ZMQ_NO_EXPORT` that, when set, bypasses the automatic `ZMQ_EXPORT` determination block and just sets `ZMQ_EXPORT` to empty.

By combining this solution at configuration time with manually passing `-fvisibility=hidden` to `CXXFLAGS`, I solved my visibility problem. Just passing `-fvisibility=hidden` is not enough, because `__attribute__ ((visibility ("default")))` has higher priority.


configure.ac
```diff
@@ -1150,6 +1150,14 @@ AC_ARG_WITH([pkgconfigdir],
     [pkgconfigdir='${libdir}/pkgconfig'])
 AC_SUBST([pkgconfigdir])
 
+AC_ARG_ENABLE([no-export],
+    [AS_HELP_STRING([--enable-no-export],
+        [Build libzmq with empty ZMQ_EXPORT macro, bypassing platform-based automated detection [default=no]])])
+if test "x$enable_no_export" = "xyes"; then
+    CPPFLAGS="-DZMQ_NO_EXPORT $CPPFLAGS"
+    CXXFLAGS="-DZMQ_NO_EXPORT $CXXFLAGS"
```
bluca, line 1158: CPPFLAGS is enough, CXXFLAGS are compiler flags for g++, but this is just preprocessor stuff

egpbos: Thanks, I wasn't sure.

egpbos:
Ah, force pushing a non-changed amended commit apparently doesn't retrigger CI, still stalling :) Could you restart it?

Issue #1478, opened 2015-07-20
Not receiving unsubscription messages in XPUB socket with ZMQ_XPUB_VERBOSE and using a XSUB-XPUB proxy in front


jimenezrick:
It seems there is a bug when using ZMQ_XPUB_VERBOSE with XPUB sockets
and it's specifically triggered when using a proxy in front of the
producers, which is our use case.

We're using ZMQ_XPUB_VERBOSE flag with XPUB sockets because we want to
be notified whenever there is new consumers on certain topic. And we
expect to receive a unsubscribe message when there aren't consumers of
that topic any more.

- Our set up like this:
```
        :                               /--> SUB :
App     : XPUB ---> XSUB/proxy/XPUB --> +--> SUB :     App
        :                               \--> SUB :
```
The bug is manifested in the next way: with the ZMQ_XPUB_VERBOSE you
are expected to receive N subscription messages for N consumers
listening certain topic, and to receive a single unsubscription message
when the last of those consumers disconnects. But what really happens is
that when using the proxy (without the proxy it works fine) is that that
unsubscription message never arrives to the producer application.

The problem lies in that we filter (un)subscription messages in both,
xpub_t and xsub_t and this doesn't play well with the verbose flag and a
proxy.

The next diagram explains what happens with the verbose flag but without
the proxy:

- Here we have a direct connection between one publisher and three
consumers. Each consumer will send a subscription message when they join
the same topic and a unsubscription message when they stop following
that topic. The application using the XPUB socket in verbose mode will
see three subscription messages (1st frame \0x01) and only one
unsubscription message (1st frame \0x00) for the last disconnect. This
works as intended:
```
        :                           /--> SUB :
App     : XPUB -------------------> +--> SUB :     App
        :                           \--> SUB :

<-sub   : <---------------------------sub    :
<-sub   : <---------------------------sub    :
<-sub   : <---------------------------sub    :

        : <---------------------------unsub  :
        : <---------------------------unsub  :
<-unsub : <---------------------------unsub  :
```
(xpub_t filters all unsubcription msgs except the last one to the app)

- The problematic scenario is when we put a proxy in the middle where we
we also enable the verbose flag in its XPUB socket.
```
        :                              /--> SUB :
App     : XPUB ---> XSUB/proxy/XPUB -->+--> SUB :    App
        :                              \--> SUB :

<-sub   :     <-sub     <---sub    <-----sub    :
<-sub   :     <-sub     <---sub    <-----sub    :
[...]

PR #4294, opened 2021-11-04
Problem: if ZMQ_XPUB_VERBOSER is used with proxies unsubscribe messages are lost

thielepaul:
consider the following case:
all XPUB sockets have `ZMQ_XPUB_VERBOSER = true`
```
        :                                                   /--> SUB :
App     : XPUB ---> XSUB/proxy/XPUB ---> XSUB/proxy/XPUB -->+--> SUB :    App
        :                                                   \--> SUB :

<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 

        :                                    <-unsub     <-unsub     : 
        :                                    <-unsub     <-unsub     : 
        :               <-unsub   <-unsub    <-unsub     <-unsub     : 
```
Problem: the app never receives a unsubscribe message and thus keeps sending although all subscribers have unsubscribed.

Fix: introduce new option `ZMQ_XSUB_VERBOSE_UNSUBSCRIBE` and set it to true for the proxy on the right
```
        :                                                   /--> SUB :
App     : XPUB ---> XSUB/proxy/XPUB ---> XSUB/proxy/XPUB -->+--> SUB :    App
        :                                                   \--> SUB :

<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 

        :               <-unsub   <-unsub    <-unsub     <-unsub     : 
        :               <-unsub   <-unsub    <-unsub     <-unsub     : 
<-unsub :   <-unsub     <-unsub   <-unsub    <-unsub     <-unsub     : 
```

This is similar to the issue described [here](https://github.com/zeromq/libzmq/issues/1478), the difference here is only that there are two proxies and not one.

### DIFF
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -1252,6 +1252,21 @@ Default value:: NULL
 Applicable socket types:: ZMQ_XPUB
 
 
+ZMQ_XSUB_VERBOSE_UNSUBSCRIBE: pass duplicate unsubscribe messages on XSUB socket
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Sets the 'XSUB' socket behaviour on duplicated unsubscriptions. If enabled, the socket
+passes all unsubscribe messages to the caller. If disabled, only the last unsubscription
+from each filter will be passed. The default is 0 (disabled).
+
+NOTE: in DRAFT state, not yet available in stable releases.
+
+[horizontal]
+Option value type:: int
+Option value unit:: 0, 1
+Default value:: 0
+Applicable socket types:: ZMQ_XSUB
+
+
 ZMQ_ONLY_FIRST_SUBSCRIBE: Process only first subscribe/unsubscribe in a multipart message
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 If set, only the first part of the multipart message is processed as
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -679,6 +679,7 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_PRIORITY 112
 #define ZMQ_BUSY_POLL 113
 #define ZMQ_HICCUP_MSG 114
+#define ZMQ_XSUB_VERBOSE_UNSUBSCRIBE 115
 
 /*  DRAFT ZMQ_RECONNECT_STOP options                                          */
 #define ZMQ_RECONNECT_STOP_CONN_REFUSED 0x1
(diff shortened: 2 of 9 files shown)


include/zmq.h
```diff
@@ -376,6 +376,7 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_GSSAPI_PRINCIPAL_NAMETYPE 90
 #define ZMQ_GSSAPI_SERVICE_PRINCIPAL_NAMETYPE 91
 #define ZMQ_BINDTODEVICE 92
+#define ZMQ_XSUB_VERBOSE_UNSUBSCRIBE 93
```
bluca, line 379: move to the draft section below, it clobbers an existing draft option like this

doc/zmq_setsockopt.txt
```diff
@@ -1252,6 +1252,19 @@ Default value:: NULL
 Applicable socket types:: ZMQ_XPUB
 
 
+ZMQ_XSUB_VERBOSE_UNSUBSCRIBE: pass duplicate unsubscribe messages on XSUB socket
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
bluca, line 1256: for the formatting to work properly, this line's length must match the above's

include/zmq.h
bluca: also needs to be duplicated in src/zmq_drafts.hpp

src/xsub.cpp
```diff
@@ -109,6 +110,8 @@ int zmq::xsub_t::xsetsockopt (int option_,
         }
         _only_first_subscribe = (*static_cast<const int *> (optval_) != 0);
         return 0;
+    } else if (option_ == ZMQ_XSUB_VERBOSE_UNSUBSCRIBE) {
```
bluca, line 113: this should be enclosed in ifdef so that it can only be enabled in draft builds

bluca:
please add coverage of the new option in the unit tests, there should be existing xpub/verbose/verboser tests to leverage

thielepaul:
Thank you for your comments, @bluca ! Please let me know if you have further suggestions.

thielepaul:
I just realized today that the usecase as described above does not work

```
        :                                                   /--> SUB :
App     : XPUB ---> XSUB/proxy/XPUB ---> XSUB/proxy/XPUB -->+--> SUB :    App
        :                                                   \--> SUB :

<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 
<-sub   :     <-sub      <-sub     <-sub      <-sub       <-sub      : 

        :               <-unsub   <-unsub    <-unsub     <-unsub     : 
        :                                                            : 
        :                                                            : 
```
after the first unsubscribe the other subscribers stop receiving messages, probably because the xpub socket (in the middle) stops sending messages after it has received the first unsibscribe.

Issue #3313, opened 2018-11-21
Assertion failed: (src/mailbox.cpp:99)


paololucente:
# Issue description
I'm from a project called pmacct ( https://github.com/pmacct/pmacct ) which can optionally make use of ZeroMQ for passing messages internally. I got three reports in the last three weeks of users compiling pmacct against ZeroMQ 4.2.5 and getting a failed assertion message back:

shell> nfacctd -f ./nfacctd.detailed.conf
Assertion failed: ok (src/mailbox.cpp:99)
Aborted (core dumped)

Only an hint: one of the users made some testing and apparently all works well with up to gcc6. But compiling with gcc7 or gcc8 shows the above issue.

# Environment
* libzmq version: 4.2.5 
* OS: Debian 10, Ubuntu 18.04

# Minimal test code / Steps to reproduce the issue

1. apt-get install libpcap-dev pkg-config libtool autoconf automake bash
2. compile and install ZeroMQ 4.2.5 
3. git clone https://github.com/pmacct/pmacct.git 
4. cd pmacct ; ./autogen ; ./configure --enable-zmq; make ; make install
5. Edit a nfacctd.conf file, put inside the following line:
    plugin_pipe_zmq: true
6. Start pmacct: nfacctd -f /path/to/nfacctd.conf
7. It should pretty much immediately show the failed assertion message

# What's the actual result? (include assertion message & call stack if applicable)
gdb /usr/local/sbin/nfacctd core
GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
< .. cut .. >
Reading symbols from /usr/local/sbin/nfacctd...done.
[New LWP 6120]
[New LWP 6118]
[New LWP 6117]
[New LWP 6119]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `nfacctd: Core Process [detailed] '.
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51    ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f209000d700 (LWP 6120))]
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
[...]

PR #4309, opened 2021-11-23
Problem: Fix fork detection on gcc 7

dgloe-hpe:
Solution: When compiling with gcc 7 and newer, the program produced by
AC_CHECK_FUNCS(fork) produces a warning, which results in configure
incorrectly disabling fork support. Fix the issue by using an
AC_COMPILE_IFELSE which correctly detects fork availability.
Tested by running configure and make check on a system with gcc 7
installed, and verifying that HAVE_FORK was defined correctly.

See issue #3313.


bluca:

PR #4310, opened 2021-11-24
Problem: In rare cases, afunix.h doesn't contain a definition for struct sockaddr_un

### LINKED ISSUE #3949: Build fails with Visual C++ 2017 due to undefined sockaddr_un
Visual C++ 2017 doesn't know the `sockaddr_un` type, which is used by `ipc_address.hpp`/`ipc_address.cpp`. That seems to be the only thing preventing libzmq from building. Adding a definition like this helped me compile the library at least (doesn't seem to work properly, though):

```c++
typedef unsigned int sa_family_t;

struct sockaddr_un
{
    sa_family_t sun_family; /* AF_UNIX */
    char sun_path[108];     /* Pathname */
};
```

There were no such problems with the latest release, 4.3.2.

myd7349:
According to https://github.com/microsoft/vcpkg/issues/21623,
struct sockaddr_un might be unavailable on some machines even afunix.h exists.

For example, on some machines, the content of afunix.h looks like this:
```c
typedef struct _SOCKADDR_UN
{
     ADDRESS_FAMILY Family;
     wchar_t Path[63];
} SOCKADDR_UN, *PSOCKADDR_UN;
```
but on other machines, it may looks like this:
```c
#define UNIX_PATH_MAX 108

typedef struct sockaddr_un
{
    ADDRESS_FAMILY sun_family;
    char sun_path[UNIX_PATH_MAX];
} SOCKADDR_UN, *PSOCKADDR_UN;
```

Fixes #3949

References:
- [Enable Unix-domain sockets support on Windows](https://git.furworks.de/opensourcemirror/postgresql/commit/8f3ec75de4060d86176ad4ac998eeb87a39748c2)
- [AF_UNIX equivalent for Windows](https://stackoverflow.com/questions/9029174/af-unix-equivalent-for-windows)
- https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/


myd7349:
Another approach to fix this issue is:
In CMakeLists.txt:
```cmake
if(WIN32)
    check_include_files("winsock2.h;afunix.h" ZMQ_HAVE_IPC)
    if(ZMQ_HAVE_IPC)
        check_struct_has_member("struct sockaddr_un" sun_path "winsock2.h;afunix.h" ZMQ_HAVE_STRUCT_SOCKADDR_UN)
    endif()
endif()
```
and in `src/windows.hpp`:
```cpp
#if defined(ZMQ_HAVE_IPC) && !defined(ZMQ_HAVE_STRUCT_SOCKADDR_UN)
struct sockaddr_un
{
    sa_family_t sun_family; /* AF_UNIX */
    char sun_path[108];     /* Pathname */
};
#endif
```
Which one do you prefer?

bluca:
The second one would seem to allow the feature to work in more cases, so probably it's better?

myd7349:
Oops! It seems I made a mistake.
The definition of sockaddr_un should be protected by the include guard:
https://github.com/myd7349/libzmq/tree/fix-sockaddr_un-redefinition

I will create a new PR soon.

PR #4311, opened 2021-11-25
Problem: struct sockaddr_un redefinition

myd7349:
The definition of sockaddr_un should be protected by the include guard to avoid redefinition.
Reported here: https://github.com/microsoft/vcpkg/issues/21623#issuecomment-978144984


danieljbarber:
Sorry for not responding right away holidays and a conference came about and I dropped the ball. I don't have access to the same PC I was having the issue on, but I tried master on a different machine with VS 2019 community and it built without issue.

PR #4314, opened 2021-12-03
Issue 4312 : Fix epoll invalid event data pointer

thomasderive:
Problem inside the event array, epoll return an event with invalid data pointer which create a segmentation fault.
Tests to prevent segmentation fault due to null pointer are added.


bluca:

PR #4317, opened 2021-12-16
Problem: ROUTER loses first frame metadata on message prefetch

tarmo:
Solution: copy metadata on prefetch to the first frame.

### DIFF
--- a/tests/test_metadata.cpp
+++ b/tests/test_metadata.cpp
@@ -127,10 +127,61 @@ void test_metadata ()
     zmq_threadclose (zap_thread);
 }
 
+void test_router_prefetch_metadata ()
+{
+    char my_endpoint[MAX_SOCKET_STRING];
+    setup_test_context ();
+
+    //  Spawn ZAP handler
+    //  We create and bind ZAP socket in main thread to avoid case
+    //  where child thread does not start up fast enough.
+    void *handler = zmq_socket (get_test_context (), ZMQ_REP);
+    TEST_ASSERT_NOT_NULL (handler);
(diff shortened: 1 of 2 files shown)


tests/test_metadata.cpp
```diff
@@ -127,10 +127,61 @@ void test_metadata ()
...
+{
+    char my_endpoint[MAX_SOCKET_STRING];
+    setup_test_context ();
+
+    //  Spawn ZAP handler
+    //  We create and bind ZAP socket in main thread to avoid case
+    //  where child thread does not start up fast enough.
+    void *handler = zmq_socket (get_test_context (), ZMQ_REP);
+    TEST_ASSERT_NOT_NULL (handler);
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_bind (handler, "inproc://zeromq.zap.01"));
```
bluca, line 140: This is not safe to do, and will likely cause the test to be flaky, please rework it so that the socket is used from the thread where it's created, there's other cases of using zap in tests and you should be able to just reuse them

tarmo: I just used the existing pattern already used in the same file and apparently also a few others. Just to be clear, is https://github.com/zeromq/libzmq/blob/ab9c7add58c5f1cb46e26ddcff3158199e29b380/tests/testutil_security.cpp#L312-L315 the approach you're suggesting to reuse?

tarmo: Do you think it would be sensible to handle the ZAP request synchronously in the main thread immediately after the `zmq_connect` call?

bluca: if it's already done in the same file you don't need to fix it yourself, it's fine

PR #4323, opened 2021-12-28
Problem: ZMQ_BUILD_DRAFT_API define is not propagated to dependent CMake projects

### LINKED ISSUE #4194: CMake build on Linux does not export ZMQ_BUILD_DRAFT_API as a compile_definition with the target
libzmq version = 4.3.4

* OS: 
Ubuntu 20.04 with cmake 3.16.3

Building with the following command after extracting the tarball and mkdir build; cd build.

cmake -D BUILD_SHARED=OFF -D ENABLE_CURVE=OFF -D ENABLE_DRAFTS=ON =WITH_LIBBSD=OFF -D WITH_LIBSODIUM=OFF ..
sudo make install

The FindPackage config module in /usr/local/lib/cmake/ZeroMQ has no mention of ZMQ_BUILD_DRAFT_API even though you now need it to use the cmake target libzmq-static.

I add the compile definition in my own CMakeLists.txt but the general idea of CMake is that I shouldn't need to.  This is further exacerbated when using cppzmq which also requires the definition.

boscosiu:
Solution: Add the definition to the relevant CMake targets in public scope.  Fixes #4194

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,6 @@ endif()
 # Enable WebSocket transport and RadixTree
 if(ENABLE_DRAFTS)
   message(STATUS "Building draft classes and methods")
-  set(ZMQ_BUILD_DRAFT_API 1)
   option(ENABLE_WS "Enable WebSocket transport" ON)
   option(ENABLE_RADIX_TREE "Use radix tree implementation to manage subscriptions" ON)
 else()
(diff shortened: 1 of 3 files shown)


bluca:
Build fails:

```
C:\projects\build_libzmq\platform.hpp(74,1): error C2220: the following warning is treated as an error [C:\projects\build_libzmq\libzmq.vcxproj]
C:\projects\build_libzmq\platform.hpp(74,1): warning C4005: 'ZMQ_BUILD_DRAFT_API': macro redefinition [C:\projects\build_libzmq\libzmq.vcxproj]
C:\projects\libzmq\src\precompiled.cpp : message : see previous definition of 'ZMQ_BUILD_DRAFT_API' 
```

boscosiu:
Thanks, sorry about that.  I refactored this to have the build targets receive the definition directly from CMake instead of through the CMake-generated platform.hpp.

PR #4332, opened 2022-01-13
improve check for tipc for Android 7

keith-dev:
Android 7 defines sockaddr_tipc, but not tipc_addr.  The build fails because tipc_addr is used, but not defined.


bluca:

PR #4346, opened 2022-02-06
Fix minor casing issue in include files

Jasper-Bekkers:
This fails when compiling x86_64-pc-windows-msvc targets on a case sensitive file system.


bluca:

Jasper-Bekkers:
Done.

PR #4360, opened 2022-03-15
Fix pkgconfig

darkdragon-001:
Use CMake install variables instead of hardcoded values. See [cmake documentation](https://cmake.org/cmake/help/latest/command/install.html#installing-targets).

Otherwise on systems where it is set differently (e.g. Centos 7 as `lib64`) pkgconfig fails.


darkdragon-001:

done

Issue #4369, opened 2022-04-14
ypipe_conflate_t::reader_awake has data race


alokpr:
# Issue description

TSAN reports data race on ypipe_conflate_t::reader_awake.

# Environment

* libzmq version (commit hash if unreleased): 4.3.4
* OS: Ubuntu 18.04

# Minimal test code / Steps to reproduce the issue
```
zmq::context_t publisher_context(1);
zmq::socket_t publisher(publisher_context, zmq::socket_type::pub);
publisher.set(zmq::sockopt::conflate, 1);
publisher.connect("ipc:///tmp/foo");

zmq::context_t subscriber_context(1);
zmq::socket_t subscriber(subscriber_context, zmq::socket_type::sub);
subscriber.set(zmq::sockopt::conflate, 1);
subscriber.bind("ipc:///tmp/foo");

while (true) {
  publisher.send();
  subscriber.recv();
}
```
# What's the actual result? (include assertion message & call stack if applicable)
```
WARNING: ThreadSanitizer: data race
  Read of size 1 at 0x7b38000102f0 by thread T23:
    #0 zmq::ypipe_conflate_t<zmq::msg_t>::flush() <null>
    #1 zmq::pipe_t::flush() src/pipe.cpp:282
    #2 zmq::session_base_t::flush() src/session_base.cpp:236
    #3 zmq::stream_engine_base_t::in_event_internal() src/stream_engine_base.cpp:325
    #4 zmq::stream_engine_base_t::in_event() src/stream_engine_base.cpp:243
    #5 zmq::epoll_t::loop() src/epoll.cpp:206
    #6 zmq::worker_poller_base_t::worker_routine(void*) src/poller_base.cpp:146 
    #7 thread_routine src/thread.cpp:257
    #8 <null> <null> (libtsan.so.0+0x2e3bf)

  Previous write of size 1 at 0x7b38000102f0 by thread T21:
    #0 zmq::ypipe_conflate_t<zmq::msg_t>::check_read() <null>
    #1 zmq::ypipe_conflate_t<zmq::msg_t>::read(zmq::msg_t*) <null>
    #2 zmq::pipe_t::read(zmq::msg_t*) src/pipe.cpp:205
    #3 zmq::xpub_t::xread_activated(zmq::pipe_t*) src/xpub.cpp:103
    #4 zmq::socket_base_t::read_activated(zmq::pipe_t*) src/socket_base.cpp:1722
[...]

PR #4371, opened 2022-04-15
Remove unnecessary variable ypipe_conflate_t::reader_awake

alokpr:
This variable is not necessary because the only possible value is "false".
This variable also had data race reported by TSAN.

Fixes zeromq/libzmq#4369


alokpr:
@bluca can you please review the PR?

mitchellwrosen:
This change looks good to me, however the comment above `flush` probably needs tweaking as well:

```
    //  Flush is no-op for conflate ypipe. Reader asleep behaviour
    //  is as of the usual ypipe.
    //  Returns false if the reader thread is sleeping. In that case,
    //  caller is obliged to wake the reader up before using the pipe again.
```

bluca:
> This variable also had data race reported by TSAN.

TSAN does not understand the complicated pipes sharing that is implemented here, I suggest to simply avoid running it

Issue #4375, opened 2022-05-02
skipped tests register as failed with earlier versions of cmake


bill-torpey:
# Environment

* libzmq version (master): 
* OS: CentOS 7

# Minimal test code / Steps to reproduce the issue

1.  Run unit tests under CentOS 7

# What's the actual result? (include assertion message & call stack if applicable)

Skipped tests show up as failed

# What's the expected result?

Skipped tests should show up as passed

bill-torpey:
Under CentOS 7 with default cmake (2.8.12), unit tests fail like so:

```
        Start 100: test_shutdown_stress_tipc
100/117 Test #100: test_shutdown_stress_tipc ........***Skipped   0.00 sec
...
91% tests passed, 11 tests failed out of 117

Total Test time (real) = 163.90 sec

The following tests FAILED:
	 73 - test_connect_null_fuzzer (Not Run)
	 74 - test_bind_null_fuzzer (Not Run)
	 75 - test_connect_fuzzer (Not Run)
	 76 - test_bind_fuzzer (Not Run)
	 93 - test_address_tipc (Not Run)
	 94 - test_pair_tipc (Not Run)
	 95 - test_reqrep_device_tipc (Not Run)
	 96 - test_reqrep_tipc (Not Run)
	 98 - test_sub_forward_tipc (Not Run)
	 99 - test_connect_delay_tipc (Not Run)
	100 - test_shutdown_stress_tipc (Not Run)
Errors while running CTest
make: *** [test] Error 8
[btorpey@bt-brix7 master]$ cmake -version
cmake version 3.7.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
[btorpey@bt-brix7 master]$ 
```

Skipped tests do not trigger failure with a more recent version of cmake (3.12.3) on CentOS 7, nor with the default (3.16.3) version of cmake on Ubuntu 20.04:

```
        Start 100: test_shutdown_stress_tipc
100/117 Test #100: test_shutdown_stress_tipc ........***Skipped   0.01 sec
...
100% tests passed, 0 tests failed out of 117

Total Test time (real) = 163.54 sec

The following tests did not run:
	 73 - test_connect_null_fuzzer (Skipped)
	 74 - test_bind_null_fuzzer (Skipped)
	 75 - test_connect_fuzzer (Skipped)
	 76 - test_bind_fuzzer (Skipped)
	 93 - test_address_tipc (Skipped)
	 94 - test_pair_tipc (Skipped)
[...]

PR #4376, opened 2022-05-03
default to not building tests if cmake version can't handle skipped tests

bill-torpey:
(see https://github.com/zeromq/libzmq/issues/4375)


bluca:
please update the CI scripts to always run them

bill-torpey:
> please update the CI scripts to always run them
 no idea what you mean -- didn't touch any ci scripts

bluca:
There are CI scripts that run the tests, that now wouldn't, in builds/cmake/build.sh because of the new default - please change the scripts to explicitly enable the option, so that tests always run regardless of cmake version

bill-torpey:
not at all familiar with CI framework, but at first glance it looks like those builds are (still) being done w/autotools?

bill-torpey:
got it

PR #4379, opened 2022-05-06
remove libbsd

guijan:
libbsd is only used once and as part of a larger, incorrect function.
I rewrote the code that used it without the need for it.

I did this because I wanted to switch libzmq to _libbsd-overlay_ instead of _libbsd_ to make it compatible with my libbsd alternative called [libobsd](https://github.com/guijan/libobsd/) (I only want to support the -overlay mode), my library needs to work with libzmq so I can try getting it into OpenWRT at https://github.com/openwrt/openwrt/pull/9714, but it seemed to make more sense to drop the dependency entirely.

I've build tested this PR on Alpine Linux with Autotools and CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,19 +240,6 @@ if(NOT ZMQ_USE_GNUTLS)
   endif()
 endif()
 
-if(NOT MSVC)
-  option(WITH_LIBBSD "Use libbsd instead of builtin strlcpy" ON)
-  if(WITH_LIBBSD)
-    pkg_check_modules(LIBBSD "libbsd")
-    if(LIBBSD_FOUND)
-      message(STATUS "Using libbsd")
-      set(pkg_config_names_private "${pkg_config_names_private} libbsd")
-      set(ZMQ_HAVE_LIBBSD 1)
-    endif()
-  endif()
-  check_cxx_symbol_exists(strlcpy string.h ZMQ_HAVE_STRLCPY)
-endif()
-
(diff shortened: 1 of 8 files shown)


bluca:
What problem are you trying to solve exactly? The library is optional and used only if available and configured

guijan:
The problem I was originally trying to solve was that my libbsd alternative didn't work with libzmq, and the reason for that is that I only implement libbsd's libbsd-overlay mode because I think it leads to simpler programs and fits all use cases, and so I'd rather try to convince others (libzmq in this case) to use libbsd-overlay than to implement the non-overlay mode.

However, I noticed `strlcpy` is the only libbsd function in use and that it's only used once to implement Microsoft's `strcpy_s` outside Windows systems, and that at the same time libzmq's `strcpy_s` differs from Microsoft's: it places the terminator at the last byte of the destination buffer instead of the first, and it doesn't handle the case where `src` is NULL.

Because of the above, I thought it made more sense to write a `strcpy_s` that complies Microsoft's documentation as a small contribution to libzmq. The new implementation happens to remove the need for a libbsd which also satisfies my goal.

bluca:
Sorry but this doesn't make sense. If you want to have alternative implementations of library, simply provide a compatible ABI. We use libbsd where available because it's safer than reimplementing the wheel, so adding to that goes in the exact opposite direction.

guijan:
Would you consider switching to libbsd's _libbsd-overlay_ mode instead, then? I'll send a PR for that later if so.

neheb:
I'd split this up into two commits: fix strcpy_s and then remove libbsd.

The last one could be kept for OpenWrt if upstream rejects.

Issue #4276, opened 2021-10-13
Android build fails with NDK r23


enwi:
# Issue description
Using the latest Android NDK (r23) the build fails due to the removed folder `platforms` inside the NDK, which the build script is trying to access.

# Environment

* libzmq version (commit hash if unreleased): 5d8d8575
* OS: Linux

# Minimal test code / Steps to reproduce the issue
```shell
export NDK_VERSION=android-29
export MIN_SDK_VERSION=29
./build.sh arm
```

# What's the actual result? (include assertion message & call stack if applicable)
```
Android (arm) build failed for the following reasons:
  The ANDROID_BUILD_SYSROOT directory does not exist
    /Android/Sdk/ndk/23.0.7599858/platforms/android-29/arch-arm
```

# What's the expected result?
The build should not fail, thus the build script needs to be adjusted for new releases of the Android NDK.

enwi:
Sadly I can't I don't know what is needed to fix this atm.

bluca:
the ndk version can be overridden in the scripts, so it shouldn't be an issue

rubar-tech:
i'm trying to build it for android in kivy/buildozer, that uses python4android, i have the same problem, libzmq produces lots of error with ndk r23, but ndk 19c works just fine

benjdero:
> Would it be possible to support building for Android using cmake? (this would make it possible to include libzmq in automated build processes for apps)

The bindings for Czmq, Zyre & other projects based on Zproject already build using Cmake, but the lib itself build with Make. I think it would take considerable efforts to port the build scripts from using Make to CMake for someone who isn't used to the NDK,  CMake or Zeromq build scripts for Android.

@sappo Is this something that is worth to look at? As it could bring more consistency to the Android build scripts.

PR #4380, opened 2022-05-09
Problem: Android build scripts does not work with latest NDK versions

benjdero:
Solution: Update the build scripts to work with NDK versions r23 and r24


benjdero:
This fix #4276

PR #4386, opened 2022-05-19
fall thru to tcp/ip for any AF_UNIX/ipc error

pjaggi1:
https://github.com/zeromq/pyzmq/issues/1505
a few pyzmq clients have encountered an issue where the program they're running ends after this assert:
Bad file descriptor (C:\ci\zeromq_1602704446950\work\src\epoll.cpp:100)

We found this happens for users that have a windows 10 build that supports AF_Unix sockets but does not for those who have older windows 10 builds.  this change is to remove the preserved errno (which doesn't appear to be used/checked anywhere-hence the later assert/abort) and instead after cleanup try to use a tcp/ip socket.

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -664,11 +664,10 @@ error_closelistener:
         }
         filename.clear ();
     }
-
+    /*
     errno = saved_errno;
-
     return -1;
-
+    */
 try_tcpip:
     // try to fallback to TCP/IP
     // TODO: maybe remember this decision permanently?


bluca:
https://github.com/zeromq/libzmq/blob/37224c93de2c7c08602c79a2a1b4d7e582f09281/src/ip.cpp#L584

pjaggi1:
> Uh? This doesn't make much sense, there's already a skip-ahead if the socket creation fails

in our case and in some of these other cases its not necessarily the socket creation that fails (for us it was the bind).

songjt1993:
I had the same problem. Will this pr be merged into master? How long will it take to see it in the release version

Godsix:
I had the same problem.And I found conditions:①Windows 10 username contain no-ASCII Characters,such as Chinese Characters;②libzmq was compiled with msvc v142 or v143.
I compile pyzmq with libzmq-v141,it‘s OK.
Before developers solve the problem,user can avoid the problem by replacing to with libzmq-v141.

Godsix:
> I had the same problem.And I found conditions:①Windows 10 username contain no-ASCII Characters,such as Chinese Characters;②libzmq was compiled with msvc v142 or v143. I compile pyzmq with libzmq-v141,it‘s OK. Before developers solve the problem,user can avoid the problem by replacing to with libzmq-v141.

So this problem mostly occurs in non-English speaking countries, because they are more likely to use non-ASCII characters as Windows username.

PR #4392, opened 2022-06-16
Fix compilation on macOS Ventura

rex4539:
Fixes https://github.com/zeromq/libzmq/issues/4391


rex4539:
Looks like it breaks the compilation on all other platforms :)

Perhaps a conditional could bypass this issue.

sphaero:
Did you call snprint correctly?

../../src/tcp_address.cpp:132:27: error: argument to ‘sizeof’ in ‘int snprintf(char*, size_t, const char*, ...)’ call is the same expression as the destination; did you mean to provide an explicit length? [-Werror=sizeof-pointer-memaccess]

PR #4397, opened 2022-06-23
problem: using obselete libtool macros

fanquake:
solution: use the non-obseleted macros

on `./autogen.sh` using autoconf 2.71 and automake 1.16.5:
```bash
configure.ac:83: warning: The macro `AC_LIBTOOL_WIN32_DLL' is obsolete.
configure.ac:83: You should run autoupdate.
config/ltoptions.m4:148: AC_LIBTOOL_WIN32_DLL is expanded from...
configure.ac:83: the top level
configure.ac:83: warning: AC_LIBTOOL_WIN32_DLL: Remove this warning and the call to _LT_SET_OPTION when you
configure.ac:83: put the 'win32-dll' option into LT_INIT's first parameter.
./lib/autoconf/general.m4:2434: AC_DIAGNOSE is expanded from...
config/ltoptions.m4:148: AC_LIBTOOL_WIN32_DLL is expanded from...
configure.ac:83: the top level
configure.ac:84: warning: The macro `AC_PROG_LIBTOOL' is obsolete.
configure.ac:84: You should run autoupdate.
config/libtool.m4:100: AC_PROG_LIBTOOL is expanded from...
configure.ac:84: the top level
```

Note that this combines the libtool initialization into a single check, as having per-host invocations of `LT_INIT` does not seem to work correctly. This means that static builds would no-longer be disabled by default for Windows targets. Is that acceptable? Note that static libraries are enabled by default (where supported) by `LT_INIT`, so there is no need to explicitly pass `enable-shared`.

`AC_PROG_LIBTOOL` is the deprecated form of `LT_INIT`.
`AC_LIBTOOL_WIN32_DLL` is the deprecated form of passing `win32-dll` to
LT_INIT.
`AC_DISABLE_STATIC` is the deprecated form of passing `disable-shared`
to LT_INIT.
See https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html for more info.


bluca:
Is this backward-compatible with older autotools? It needs to keep working on 20 years old unixes and other such ancient platforms

PR #4411, opened 2022-08-05
Update ctx.cpp

caicaiking:
add lock for get_opt


bluca:
What problem are you trying to solve? Why is this necessary?

caicaiking:
> 

those class members are used in different thread, so I think those members need protect for read and write. that's why I add lock here.

bluca:
Reading a single integer (word size) on the vast majority of platforms is an atomic operation. But I guess it can't hurt. Then update the commit message following the template as the PR description says:

Problem: short explanation

Solution: longer explanation

Also please add a relicense statement https://github.com/zeromq/libzmq/tree/master/RELICENSE

PR #4412, opened 2022-08-10
Problem: get_opt function read the options value not thread safe

caicaiking:
Solution: use scoped lock to prevent them from being modified during reading


caicaiking:
rewrite the commit message.

bluca:
Also please add a relicense statement https://github.com/zeromq/libzmq/tree/master/RELICENSE

No need for a new PR, you can simply push a new commit to the same branch

PR #4422, opened 2022-09-08
Problem: bind can fail after open_socket succeeds on Windows without AF_UNIX

minrk:
closes #4386

Following discussion in #4386, this is a more specific check [requested](https://github.com/zeromq/libzmq/pull/4386#issuecomment-1132078805) by @bluca rather than a catch-all. Only failures up to the initial bind are considered for fallback to tcpip.

I can't reproduce the error, so I don't know exactly what errno to check for to make the condition even more specific.

Also adds the missing check of the returned rc from make_fd_pair, which should have raised the actual error instead of the cryptic `src\epoll.cpp:100` reported in https://github.com/zeromq/pyzmq/issues/1505

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -577,6 +577,11 @@ int zmq::make_fdpair (fd_t *r_, fd_t *w_)
     int rc = 0;
     int saved_errno = 0;
 
+    // It appears that a lack of runtime AF_UNIX support
+    // can fail in more than one way.
+    // At least: open_socket can fail or later in bind
+    bool ipc_fallback_on_tcpip = true;
+
     //  Create a listening socket.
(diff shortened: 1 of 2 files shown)


src/ip.cpp
```diff
@@ -596,9 +601,11 @@ int zmq::make_fdpair (fd_t *r_, fd_t *w_)
     rc = bind (listener, const_cast<sockaddr *> (address.addr ()),
                address.addrlen ());
     if (rc != 0) {
-        errno = wsa_error_to_errno (WSAGetLastError ());
```
bluca, line 599: why is this being removed?

minrk: It's unused, so I figured it wasn't useful to keep it. I put it back, though.

bluca (approved):

PR #4459, opened 2022-11-17
Add ZMQ_SUBSCRIPTION_COUNT socket option

f18m:


f18m:
Hi @bluca ,
can you review this PR ? I think it's ready to go, EXCEPT for the documentation.
But before adding that I would like to have some feedback if the way this (simple) feature is implemented is correc.t.. thanks

bluca:
This is going to be racy - subs are managed in the i/o thread(s), but the option is read from the application thread. At least make it use atomics.

f18m:
> This is going to be racy - subs are managed in the i/o thread(s), but the option is read from the application thread. At least make it use atomics.

good point. I'll change it to use atomics, thanks for the review

f18m:
@bluca ,
can you go over this PR again? 
I have added use of atomics, improve the test_pubsub_num_subscriptions by testing also "nested" prefixes (subscriptions sharing the same initial bytes), and docs.

open points:
* should we implement ZMQ_SUBSCRIPTION_COUNT also for SUB sockets?
* documentation of zmq_getsocktopt() seems to be in alphabetical order in the past but now it's not anymore... I can raise another PR later on to reorder the entries if that's useful

doc/zmq_getsockopt.txt
```diff
@@ -983,6 +983,19 @@ Default value:: 8192
 Applicable socket types:: All, when using TCP, IPC, PGM or NORM transport.
 
 
+ZMQ_SUBSCRIPTION_COUNT: Number of subscriptions received
```
bluca, line 986: "subscription count" is ambiguous as it might be thought to refer to the number of subscribers, but this is just counting the number of different topics

f18m: however the way to subscribe is to use zmq_setsockopt(... ZMQ_SUBSCRIBE ... )... so using "subscription count" as term to refer to the number of different ZMQ_SUBSCRIBE operations (*) was intuitive at least for me :)

(*) I'm assuming it's not a common habit to subscribe multiple times to the same topic prefix... IOW I'm assuming that most likely the number of times zmq_setsockopt(... ZMQ_SUBSCRIBE ... ) is invoked on the SUB socket will match the result of zmq_getsockopt(.. ZMQ_SUBSCRIPTION_COUNT.. ) on the PUB socket (if network works fine)

Anyway I'm open to any other suggestion for the name... ZMQ_PUBSUB_TOPIC_PREFIXES ? ZMQ_TOPIC_COUNT?

bluca: ZMQ_PUB_TOPICS_COUNT?

f18m: fine for me, will update this PR accordingly!

f18m: I guess that if we call this socket option ZMQ_PUB_* we don't want to support this socket option also on SUB sockets right?

f18m: hi @bluca ,
can you review again?
 
latest changes include:
 * renamed ZMQ_SUBSCRIPTION_COUNT -> ZMQ_TOPICS_COUNT
 * implemented ZMQ_TOPICS_COUNT for (X)SUB sockets by making a) atomic the size() method of the radix_tree_t and b) creating a num_prefixes() atomic method for trie_t, obtained by wrapping the generic trie-node trie_t inside a trie_with_size_t class
  * inside new test, invoke ZMQ_TOPICS_COUNT on the subscriber, not just on the publisher

I just wonder why the XSUB socket is using the trie_t when there is also mtrie_t,,,

bluca:
```
CMake Error at tests/CMakeLists.txt:252 (add_executable):
  Cannot find source file:

    test_pubsub_num_subscriptions.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm .h
  .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .hip .ispc


CMake Error at tests/CMakeLists.txt:252 (add_executable):
  No SOURCES given to target: test_pubsub_num_subscriptions
```

f18m:
> ```
> CMake Error at tests/CMakeLists.txt:252 (add_executable):
>   Cannot find source file:
> 
>     test_pubsub_num_subscriptions.cpp
> 
>   Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm .h
>   .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .hip .ispc
> 
> 
> CMake Error at tests/CMakeLists.txt:252 (add_executable):
>   No SOURCES given to target: test_pubsub_num_subscriptions
> ```

thanks - should be fixed now (just a bit hard to understand the CI/CD results, and gitlab does not help marking every single commit with "CI failed")

PR #4488, opened 2023-01-10
Problem: message can't be received due to signal

ilkondr:
Issue caught in Golang runtime, which widely uses signal SIGURG for scheduling. Sometimes messages cannot be received. Technically socket_base_t::process_commands() returns failure even if some commands were processed, but next message from mailbox could not be received during interrupt.

Solution: retry receiving from mailbox with zero timeout after EINTR.

Signed-off-by: Ilya Kondrashkin <ikondrashkin@nfware.com>


bluca:

ilkondr:

added

PR #4481, opened 2022-12-29
Issue 4478: Fix sprintf deprecation error failing build on macOS

nighthawk24:
Fixes https://github.com/zeromq/libzmq/issues/4478
Fixes https://github.com/zcash/zcash/issues/6006 & https://github.com/zcash/zcash/issues/6296

Issue #4478, opened 2022-12-20
sprintf deprecation error failing build on macOS


ghmanoj:
# Issue description
Build is failing in macOS 13.1

# Environment
OS: macOS 13.1
Build Environment:
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin22.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

* libzmq version (commit hash if unreleased): dfbfc59c3fe4e782d1640f32af1056a2e74f9aaa
* OS: macOS 13.1

# Minimal test code / Steps to reproduce the issue
Run the following command after git clone:
1.  ./autogen.sh
2. ./configure
3.  make

# What's the actual result? (include assertion message & call stack if applicable)

make[1]: Nothing to be done for `all'.
  CXX      src/libzmq_la-tcp_address.lo
src/tcp_address.cpp:132:12: error: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
    pos += sprintf (pos, "%d", ntohs (port_));
           ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
1 error generated.
make[1]: *** [src/libzmq_la-tcp_address.lo] Error 1
make: *** [all-recursive] Error 1

# What's the expected result?
Build should pass

ghmanoj:
The patch below fixes the build. It replaces the `sprintf(...)` with `snprintf(...)`  but not sure this solution is optimal or not? 

diff --git a/src/tcp_address.cpp b/src/tcp_address.cpp
[...]

PR #4494, opened 2023-01-24
deprecate sprint and replace with snprintf

sphaero:
Problem: Latest OSX deprecates using sprint resulting is failure to build
Fix by switching to snprint.

As tried by #4481 and fixing #4478

### DIFF
--- a/src/tcp_address.cpp
+++ b/src/tcp_address.cpp
@@ -129,7 +129,8 @@ static std::string make_address_string (const char *hbuf_,
     pos += hbuf_len;
     memcpy (pos, ipv6_suffix_, sizeof ipv6_suffix_ - 1);
     pos += sizeof ipv6_suffix_ - 1;
-    pos += sprintf (pos, "%d", ntohs (port_));
+    pos += snprintf (pos, max_port_str_length + 1 * sizeof (char), "%d",
+                     ntohs (port_));
     return std::string (buf, pos - buf);
 }
 
--- a/src/udp_engine.cpp
+++ b/src/udp_engine.cpp
@@ -367,8 +367,8 @@ void zmq::udp_engine_t::sockaddr_to_msg (zmq::msg_t *msg_,
     const char *const name = inet_ntoa (addr_->sin_addr);
 
     char port[6];
-    const int port_len =
-      sprintf (port, "%d", static_cast<int> (ntohs (addr_->sin_port)));
+    const int port_len = snprintf (port, 6 * sizeof (char), "%d",
+                                   static_cast<int> (ntohs (addr_->sin_port)));
     zmq_assert (port_len > 0);
 
     const size_t name_len = strlen (name);
--- a/tests/test_inproc_connect.cpp
+++ b/tests/test_inproc_connect.cpp
@@ -155,7 +155,7 @@ void test_connect_before_bind_ctx_term ()
         void *connect_socket = test_context_socket (ZMQ_ROUTER);
 
         char ep[32];
-        sprintf (ep, "inproc://cbbrr%d", i);
+        snprintf (ep, 32 * sizeof (char), "inproc://cbbrr%d", i);
         TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (connect_socket, ep));
 
         // Cleanup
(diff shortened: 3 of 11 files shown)


bluca:
can you fix the formatting?

```
--- a/src/tcp_address.cpp
+++ b/src/tcp_address.cpp
@@ -129,7 +129,8 @@ static std::string make_address_string (const char *hbuf_,
     pos += hbuf_len;
     memcpy (pos, ipv6_suffix_, sizeof ipv6_suffix_ - 1);
     pos += sizeof ipv6_suffix_ - 1;
-    pos += snprintf (pos, max_port_str_length * sizeof(char), "%d", ntohs (port_));
+    pos +=
+      snprintf (pos, max_port_str_length * sizeof (char), "%d", ntohs (port_));
     return std::string (buf, pos - buf);
 }
 
diff --git a/src/udp_engine.cpp b/src/udp_engine.cpp
index 4d471a8..47f1359 100644
--- a/src/udp_engine.cpp
+++ b/src/udp_engine.cpp
@@ -367,8 +367,8 @@ void zmq::udp_engine_t::sockaddr_to_msg (zmq::msg_t *msg_,
     const char *const name = inet_ntoa (addr_->sin_addr);
 
     char port[6];
-    const int port_len =
-      snprintf (port, 6 * sizeof(char), "%d", static_cast<int> (ntohs (addr_->sin_port)));
+    const int port_len = snprintf (port, 6 * sizeof (char), "%d",
+                                   static_cast<int> (ntohs (addr_->sin_port)));
     zmq_assert (port_len > 0);
 
     const size_t name_len = strlen (name);
diff --git a/tests/testutil.cpp b/tests/testutil.cpp
index 74d139f..a521d57 100644
--- a/tests/testutil.cpp
+++ b/tests/testutil.cpp
@@ -518,16 +518,15 @@ fd_t bind_socket_resolve_port (const char *address_,
         addr_len = sizeof (struct sockaddr_storage);
         TEST_ASSERT_SUCCESS_RAW_ERRNO (
           getsockname (s_pre, (struct sockaddr *) &addr, &addr_len));
-        snprintf (my_endpoint_, 6 + strlen(address_) + 6 * sizeof(char),
-                  "%s://%s:%u",
-                 protocol_ == IPPROTO_TCP   ? "tcp"
-                 : protocol_ == IPPROTO_UDP ? "udp"
-                 : protocol_ == IPPROTO_WSS ? "wss"
-                                            : "ws",
-                 address_,
-                 af_ == AF_INET
-                   ? ntohs ((*(struct sockaddr_in *) &addr).sin_port)
-                   : ntohs ((*(struct sockaddr_in6 *) &addr).sin6_port));
+        snprintf (
+          my_endpoint_, 6 + strlen (address_) + 6 * sizeof (char), "%s://%s:%u",
+          protocol_ == IPPROTO_TCP   ? "tcp"
+          : protocol_ == IPPROTO_UDP ? "udp"
+          : protocol_ == IPPROTO_WSS ? "wss"
+                                     : "ws",
+          address_,
+          af_ == AF_INET ? ntohs ((*(struct sockaddr_in *) &addr).sin_port)
+                         : ntohs ((*(struct sockaddr_in6 *) &addr).sin6_port));
     }
 
     return s_pre;
```

sphaero:
I'll fix if I can get succesful tests. I'm also still missing one sprintf. (Disclaimer; I don't have any OSX machines so I'm really dependent on CI)

Not sure what's happening with CI tests now as I'm hardly getting any output. I might still be having the max size wrong though

sphaero:
Not sure why it says it's not successful. All tests seem fine?

bluca:
> Not sure why it says it's not successful. All tests seem fine?

there are some failures, but it's just the usual flaky tests

sphaero:
Ok, but where do you see those failures. I only see green checkmarks

bluca:
You don't see these in the list?

Screenshot from 2023-01-25 09-58-13

sphaero:
Aargh, my browser is hiding the scrollbar so I never noticed I can scroll down.

src/tcp_address.cpp
```diff
@@ -129,7 +129,8 @@ static std::string make_address_string (const char *hbuf_,
     pos += hbuf_len;
     memcpy (pos, ipv6_suffix_, sizeof ipv6_suffix_ - 1);
     pos += sizeof ipv6_suffix_ - 1;
-    pos += sprintf (pos, "%d", ntohs (port_));
+    pos += snprintf (pos, max_port_str_length + 1 * sizeof (char), "%d",
```
daira, line 132: The [return value of `snprintf`](https://en.cppreference.com/w/c/io/fprintf) is the "number of characters (not including the terminating null character) which would have been written to buffer if `bufsz` was ignored, or a negative value if an encoding error (for string and character conversion specifiers) occurred". (Arguably this was a misdesign of `snprintf`.) So if there were a buffer overflow, `pos` would be calculated incorrectly and the call to the `std::string` constructor below would still be undefined behaviour.

A better way is to either use `sprintf_s` (requires `__STDC_WANT_LIB_EXT1__` to be defined before `#include <cstdio>`), or add an explicit assertion. I will file a PR shortly.

PR #4502, opened 2023-01-31
Fixes for memory allocation and zero size

casaroli:
Hello,

I am building ZeroMQ with NuttX and this is part of a probably series of minor PRs.

I see a scenario where size is `0`, so we are getting a spurious out of memory error.

BR

### DIFF
--- a/RELICENSE/casaroli.md
+++ b/RELICENSE/casaroli.md
@@ -1,0 +1,16 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Marco Casaroli that grants permission to
+relicense its copyrights in the libzmq C++ library (ZeroMQ) under the
+Mozilla Public License v2 (MPLv2) or any other Open Source Initiative
+approved license chosen by the current ZeroMQ BDFL (Benevolent
+Dictator for Life).
+
+A portion of the commits made by the Github handle "casaroli", with
+commit author "Marco Casaroli <marco.casaroli@gmail.com>", are
+copyright of Marco Casaroli.  This document hereby grants the libzmq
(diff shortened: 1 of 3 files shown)


bluca:
memcpy with null src is undefined behaviour

bluca:

bluca:
there are other cases in this class where there will zero-size and undefined pointer memcpys. It's probably easier to just fix wherever you see that it's doing zero-sized blobs?

casaroli:

done

casaroli:
> there are other cases in this class where there will zero-size and undefined pointer memcpys. It's probably easier to just fix wherever you see that it's doing zero-sized blobs?

I am in a parallel task trying to run the tests (on NuttX) so I will propose changes as I encounter them.

If you prefer I can have a look at this class and just add guards similar to this to all relevant places, but maybe it is cleaner to do this as I encounter problems.

WDYT?

bluca:
Please fix them all at once, we cannot risk adding undefined behaviour

casaroli:
> Please fix them all at once, we cannot risk adding undefined behaviour

done

bluca:
Please fix the formatting as the CI suggests

```
--- a/src/blob.hpp
+++ b/src/blob.hpp
@@ -92,8 +92,7 @@ struct blob_t
         _owned (true)
     {
         alloc_assert (!size_ || _data);
-        if (size_ && _data)
-        {
+        if (size_ && _data) {
             memcpy (_data, data_, size_);
         }
     }
@@ -132,8 +131,7 @@ struct blob_t
         alloc_assert (!other_._size || _data);
         _size = other_._size;
         _owned = true;
-        if (_size && _data)
-        {
+        if (_size && _data) {
             memcpy (_data, other_._data, _size);
         }
     }
```

PR #4505, opened 2023-01-31
tests: use NOT APPLE instead of NOT MATCHES Clang

barracuda156:
GCC can be used on macOS, therefore correct condition is on Apple and not on Clang.
(Sure enough, otherwise building with GCC you get a lot of warnings about non-existent folders after -L.)


bluca:

barracuda156:

@bluca Thank you, done.

PR #4507, opened 2023-02-01
#4494 added calls to snprintf, but did not take into account that snprintf can truncate

daira:
If snprintf truncates, it returns the number of characters that would have been written without truncation. Add assertions to cover this case.

See also https://github.com/zeromq/libzmq/pull/4494/files#r1092829210

### DIFF
--- a/RELICENSE/daira.md
+++ b/RELICENSE/daira.md
@@ -1,0 +1,15 @@
+# Permission to Relicense under MPLv2 or any other OSI approved license chosen by the current ZeroMQ BDFL
+
+This is a statement by Daira Hopwood
+that grants permission to relicense hir copyrights in the libzmq C++
+library (ZeroMQ) under the Mozilla Public License v2 (MPLv2) or any other
+Open Source Initiative approved license chosen by the current ZeroMQ
+BDFL (Benevolent Dictator for Life).
+
+A portion of the commits made by the Github handle "daira", with
+commit author "Daira Hopwood <daira@jacaranda.org>", are copyright of Daira Hopwood.
+This document hereby grants the libzmq project team to relicense libzmq,
+including all past, present and future contributions of the author listed above.
+
+Daira Hopwood
+2023/02/01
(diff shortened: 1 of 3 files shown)


include/zmq.h
```diff
@@ -637,6 +637,8 @@ ZMQ_EXPORT void *zmq_threadstart (zmq_thread_fn *func_, void *arg_);
 /* Wait for thread to complete then free up resources.                        */
 ZMQ_EXPORT void zmq_threadclose (void *thread_);
 
+/* Equivalent of snprintf that asserts on error or truncation.                */
+ZMQ_EXPORT int zmq_snprintf ( char *buffer, size_t bufsz, const char *format, ... );
```
bluca, line 641: no new public APIs for this please, we do not want to be in the business of replacing glibc. There's just 2 cases using the return value, simply fix those instead

bluca:

include/zmq.h
daira: Adding this function isn't equivalent to checking the return value only in cases where that value was used (because it always asserts on errors and truncation, as it should).

It wasn't clear to me what the convention was for adding an internal helper function; what header and source file should that use?

bluca: There's no need for anything special for the tests, it really doesn't matter there. Outside of the tests there's only 2 cases, so please just do the required checks directly there

daira: Actually I'll take a different approach that uses `sprintf_s`; that doesn't require any helper function.

src/precompiled.hpp
```diff
@@ -44,6 +44,9 @@
 
 #define __STDC_LIMIT_MACROS
 
+// This must be defined before including <stdio.h> to get snprintf_s.
+#define __STDC_WANT_LIB_EXT1__ 1
```
bluca, line 48: These will not be available everywhere, especially on old compilers and systems, so they cannot be used. Again, please just check the return value in the 2 cases where it's actually needed, and ignore the rest.

PR #4516, opened 2023-02-20
add Zig build system

kassane:
Notably, no bindings are added to the main project, only the introduction an alternative build-system.

Whats is `zon` file?
See: https://github.com/ziglang/zig/issues/14290

## Steps
- https://github.com/zeromq/zproject/pull/1329
- https://github.com/zeromq/czmq/pull/2258

**Note:** This contribution is not intended to rewrite or deconstruct the main project. It only seeks to expand the working tooling for better use and build alternative.


bluca:
There are already multiple systems supported, why is another one needed?

kassane:
> There are already multiple systems supported, why is another one needed?

@bluca,

The library will now link to zig binding (czmq) under the same condition wm a single build. But it will not be restricted to the binding. One important difference about zig binding (czmq) is that libzmq is c++ and implies dual C++ABI support.
https://github.com/ziglang/zig/wiki/Troubleshooting-Build-Issues#dual-abi-linking

Zig C++ (stdc++ vs libc++)
e.g.: https://gist.github.com/kassane/7e9a2da137e13eb6e1dbab726693bdb7?permalink_comment_id=4475986#gistcomment-4475986

On linux for example, the other builds will use the system standard (gnu toolchain) or conventional (optional) clang linking to glibc + libstdc++ (GNU).
The zig being pure LLVM toolchain, will replace libstdc++ with libc++. For shared libraries, no problem but static will result in linking conflict. It is intended that libzmq and zig-czmq will be linked statically. Although it is the code citing sharedlib.

bluca:
You can link it with whatever you want, by passing the appropriate parameters, there's no need for yet another build system to do that

kassane:
> It is intended that libzmq and zig-czmq will be linked statically.

Yes it can be inconvenient for C/C++ Developer. Having to install dependency by dependency until reaching completion.

Now if another build system is unnecessary, then what I can try to do is simply call the process with cmake (and detect if it is installed) and build the binding as shared (on Windows it will be difficult to use zig in cmake).
If czmq used git submodule it would just pull the files from libzmq and link statically.

kassane:
Hi again @bluca,

Would there be any problem in adding the zig to the CI? If you want I can do that.

An important question regarding the library license, I am not a lawyer. But would it be a problem to statically link libzmq into libczmq?
Currently in zig PR, the libzmq build by default is static and libczmq is shared.

From libzmq's side I won't need to add anything else, just test and fix any failures continuously.

On your question about: "why a new build-system?"

For most C and C++ developers it will seem useless until it is unknown to many.
Unfortunately zig cannot build other build-systems directly without performing some hacks or external scripting.

It may happen that don't merge this PR, but I will continue to support each change in the mainline project. This does not interfere with czmq-zig binding as it would continue to bind build optionally.
I suggest trying to read czmq-zig's `b.dependency` on `build.zig`. Which decides to use system library or package manager.

bluca:
As I already said, I don't want to have yet another build system, there's already too many. This is all completely pointless anyway, just use the library provided by the distribution.

PR #4519, opened 2023-02-27
Switch to CNG instead of the deprecated WinCrypt

madamantis-leviathan:
Since wincrypt.h is [deprecated](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom), I replaced it with [BCryptGenRandom](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom) from Cryptography Next Generation API (CNG). 

I tried to follow the existing code style so as not to break anything. However, the changes that I made to CMakeLists.txt to include the new library (bcrypt.lib) don't appear to have had any effect. Feedback would be really appreciated

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,17 +555,20 @@ endif()
 
 if(ZMQ_HAVE_WINDOWS)
   # Cannot use check_library_exists because the symbol is always declared as char(*)(void)
-  set(CMAKE_REQUIRED_LIBRARIES "ws2_32.lib")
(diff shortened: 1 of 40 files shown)


bluca:
I'm not a windows user so can't help with this, we'll need to wait for somebody who knows their way around it

bluca:
tweetnacl was dropped in https://github.com/zeromq/libzmq/pull/4554 and is not going to come back, so we won't need this change anymore.

PR #4540, opened 2023-04-18
Support clang-16 on Termux

keith-dev:


bluca:

PR #4541, opened 2023-04-20
Add NORM Transport configuration sockopts

weston-nrl:
NORM transport is currently hard-coded to particular NORM settings and modes.  This pull request adds some NORM configuration parameters as socket options to allow for more flexibility in configuring NORM transport for different use cases.  Default values and modes remain the same.  The pull also makes use of existing socket options ZMQ_RATE (for NORM fixed-rate operation, not used in congestion control modes) and ZMQ_TOS.

### DIFF
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -411,6 +411,30 @@ ZMQ_EXPORT const char *zmq_msg_gets (const zmq_msg_t *msg_,
 #define ZMQ_SRCFD 2
 
 /******************************************************************************/
+/*  NORM transport configuration options                                      */
+/******************************************************************************/
+#ifdef ZMQ_HAVE_NORM
+
+/*  NORM socket options                                                       */
+#define ZMQ_NORM_MODE 150
+#define ZMQ_NORM_UNICAST_NACK 151
+#define ZMQ_NORM_BUFFER_SIZE 152
+#define ZMQ_NORM_SEGMENT_SIZE 153
+#define ZMQ_NORM_BLOCK_SIZE 154
+#define ZMQ_NORM_NUM_PARITY 155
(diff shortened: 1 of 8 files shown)


src/options.cpp
```diff
@@ -257,6 +257,16 @@ zmq::options_t::options_t () :
     can_recv_disconnect_msg (false),
     hiccup_msg (),
     can_recv_hiccup_msg (false),
+#ifdef ZMQ_HAVE_NORM
```
bluca, line 260: no need for an ifdef here and in the header

include/zmq.h
```diff
@@ -682,6 +682,23 @@ ZMQ_EXPORT void zmq_threadclose (void *thread_);
 #define ZMQ_XSUB_VERBOSE_UNSUBSCRIBE 115
 #define ZMQ_TOPICS_COUNT 116
 
+/*  DRAFT NORM socket options                                                 */
+#define ZMQ_NORM_MODE 150
```
bluca, line 686: number these from 117 and up to follow from the rest

weston-nrl: Yeah, I wasn't sure what to do about that.  I thought it might make sense to separate it numerically to leave room for other potential NORM settings, since there are a lot of NORM configuration options that could still be added.  I suppose in the end, it doesn't really matter though.

weston-nrl: Done.

bluca:
see https://github.com/zeromq/libzmq/actions/runs/4756140185/jobs/8451271633?pr=4541 for code formatting changes

bluca:
still unhappy https://github.com/zeromq/libzmq/actions/runs/4756551611/jobs/8452255627?pr=4541

PR #4432, opened 2022-10-01
sodium, not Sodium

Teebonne:
sodium, not Sodium

Issue #70, opened 2010-09-11
setsockopt should return EINVAL given option HWM and a negative value


chuckremes:
Disregard. This option takes an unsigned value so it shouldn't matter.

PR #4542, opened 2023-04-22
Problem: merged pr-4432 broke FindSodium on linux

bjornstromberg:
Solution: rename the FindSodium to Findsodium so it matches what the build script actually looks for. tested on ubuntu-20.04lts with cmake 3.16.


bjornstromberg:
#4432 just to link these

bjornstromberg:
the ci failure on the github actions ci is just a long running test that timed out
`70/101 Test  #70: test_security_curve ..............***Timeout  60.01 sec`

on master
`70/101 Test  #70: test_security_curve ..............   Passed   11.74 sec`

that is a pretty bad change, is this normal on github actions that stuff times out?
thinking load on server could be a big factor since everything else seems to have passed.

PR #4545, opened 2023-05-09
Add missing pkg_config_defines when built with CMake

panicgh:
The pkg_config_defines variable was only set in configure.ac, but not in CMake. This lead to the `-DZMQ_BUILD_DRAFT_API=1` flag to be omitted from the generated pkg-config file.

configure.ac contains this to set the variable accordingly:

https://github.com/zeromq/libzmq/blob/532b61275e93a8f7e0f2dee4ee1a83f075325926/configure.ac#L1056-L1063


bluca:

Issue #4514, opened 2023-02-13
Add an option to select how address names are interpreted in zmq_bind


guill-delacourt:
# Issue description
We have a use-case in which we would like to put a domain name in the address given to zmq_bind. However, currently, domain names are only allowed in addresses of zmq_connect. An address name in zmq_bind is interpreted as a network interface name.

Could there be an option to select how names are interpreted in zmq_bind ?

In our use-case, the IP address to use is not known in advance and using a wildcard is not possible due to security concerns. The domain name is the only element we know in advance that would allow us to bind the socket.

# Environment

* libzmq version (commit hash if unreleased): 4.3.2
* OS: Linux

# Minimal test code / Steps to reproduce the issue

1.  Use a domain name in the endpoint of a binding socket (using TCP or UDP protocols).

# What's the actual result? (include assertion message & call stack if applicable)
The address name is interpreted as a network interface name, and zmq_bind fails as there is no network interface with the name given.

# What's the expected result?
The addess name is correctly interpreted as a domain name and zmq_bind binds to the correct address.

bluca:
https://github.com/zeromq/libzmq/pull/4561

PR #4560, opened 2023-06-09
Add an option to select how address names are interpreted in zmq_bind #4514

guill-delacourt:
This PR partially fixes the issue #4514. It is currently limited to the the TCP sockets.
This PR introduces a new socket option ZMQ_BIND_RESOLVE_NIC, by default set to true.
When this option is set to true, names in addresses for binding sockets are resolved as network interface names, which is the current behaviour.
When this option is set to false, names in addresses for binding sockets are resolved as hostnames, like non-binding sockets.

### DIFF
--- a/doc/zmq_setsockopt.txt
+++ b/doc/zmq_setsockopt.txt
@@ -1668,6 +1668,19 @@ Option value unit:: boolean
 Default value:: 0 (false)
 Applicable socket types:: All, when using NORM transport.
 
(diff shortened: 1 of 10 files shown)


bluca:
rather than a new option, probably better to just fallback to hostname resolving if interface resolving fails

guill-delacourt:
OK, I'm closing this PR. I'll open a new one as changing the code to fallback to hostname resolving leads to simpler changes.

Issue #4367, opened 2022-04-13
Wondering if this is a bug, or just my misunderstanding


carlcc:
Hi,

I happened to see the following code, and suspect this is a bug:

https://github.com/zeromq/libzmq/blob/572eb008f197e9d047b14a4e4cdfca41eb7c5ffd/src/v3_1_encoder.cpp#L73

At line 60, only the body size of which messages are encoded to be `LARGE`,
But at line 73,   all command messages of which the `command name length + message body length > 255`, are encoded with 8 bytes length field.

As a result, e.g., if a `subscribe command` with 255 bytes' prefix will be encoded to be:
1. No `LARGE` flag set
2. With 8 bytes length field

This message will be decode mistakely by decoder at:

https://github.com/zeromq/libzmq/blob/572eb008f197e9d047b14a4e4cdfca41eb7c5ffd/src/v2_decoder.cpp#L72

(I'm new to libzmq internal)

carlcc:
After doing experient, it is indeed a bug, I will create a PR for this later.

```C++

// This cannot subscribe the topic correctly
zmq_setsockopt(socket, aTopicWithLength250, 250);
```

PR #4564, opened 2023-06-20
Problem: long flag isn't set for subscriptions if topic has between 246 and 255 characters

soell:
Solution: fix V3.1 encoder to calculate long flag after evaluating the subscribe and cancel commands

Fixes zeromq/libzmq#4367

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -64,6 +64,7 @@ set(tests
   test_srcfd
   test_stream_timeout
   test_xpub_manual
+  test_xpub_topic
   test_xpub_welcome_msg
   test_xpub_verbose
   test_base85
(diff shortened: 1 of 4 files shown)


tests/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set(tests
   test_srcfd
   test_stream_timeout
   test_xpub_manual
+  test_xpub_topic
```
bluca, line 67: please also add this to `Makefile.am`

bluca:
there are some CI failures, have a look at the logs

tests/CMakeLists.txt
soell: I hope the changes to Makefile.am are OK. I'm not familiar with automake...

soell:
Is there a possibility to retry the failed OBS Fedora_Rawhide/x86_64 build? I don't believe that the failing test is due to my code changes. The test passes for all other build configurations.

PR #4574, opened 2023-07-13
Benchmark 'add' and 'rm' of generic_mtrie

gkhandake:
Benchmark of 'add' and 'rm' operation of generic_mtrie

In my project we need to subscribe to lots of unique topics (prefixes). Following benchmark test helps to get performance measurement on publisher side when we have lots of topics to subscribe.

Following are the result of benchmark.
```sh
$ ./perf/benchmark_generic_mtrie 
prefixes = 1000000
[generic_mtrie::add]
Average operation time = 287.7 ns
[generic_mtrie::rm]
Average operation time = 189.1 ns
```


perf/benchmark_generic_mtrie.cpp
```diff
@@ -0,0 +1,141 @@
+/*
+    Copyright (c) 2023 Contributors as noted in the AUTHORS file
+
+    This file is part of libzmq, the ZeroMQ core engine in C++.
+
+    libzmq is free software; you can redistribute it and/or modify it under
```
bluca, line 6: The license has changed, please check other files and update the header to match

perf/benchmark_generic_mtrie.cpp
```diff
@@ -0,0 +1,141 @@
...
+    libzmq is distributed in the hope that it will be useful, but WITHOUT
+    ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+    FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public
+    License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#if __cplusplus >= 201103L
```
bluca, line 30: why is this necessary?

bluca:
CI fails in several places, no answer since a few months, closing, if you are still interested please fix the builds and reopen it

PR #4605, opened 2023-10-22
Introduce SAL2 annotations

axelriet:
Added new zmq_sal.h header. Modified zmq.h and zmq.cpp to add SAL2 annotations to all public functions. Added SAL2 annotations to some internal classes (msg, ctx, some utils)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1176,7 +1176,7 @@ if(MINGW)
   endif()
 endif()
 
-set(public_headers include/zmq.h include/zmq_utils.h)
+set(public_headers include/zmq.h include/zmq_utils.h include/zmq_sal.h)
 
 set(readme-docs AUTHORS LICENSE NEWS)
 
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -33,7 +33,7 @@ extern "C" {
 
 /* Include Microsoft SAL header, or no-nothing macros depending on platform */
 
-#include <zmq_sal.h>
+#include "zmq_sal.h"
 
 /*  Handle DSO symbol visibility                                             */
 
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -31,7 +31,9 @@ extern "C" {
 #include <stddef.h>
 #include <stdio.h>
 
-#include <zmq_sal.h>
+/* Include Microsoft SAL header, or no-nothing macros depending on platform */
+
+#include "zmq_sal.h"
 
 /*  Handle DSO symbol visibility                                             */
 
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -31,7 +31,7 @@ extern "C" {
 #include <stddef.h>
 #include <stdio.h>
 
-#include "zmq_sal.h"
+#include <zmq_sal.h>
 
 /*  Handle DSO symbol visibility                                             */
 
--- a/include/zmq.h
+++ b/include/zmq.h
@@ -33,7 +33,7 @@ extern "C" {
 
 /* Include Microsoft SAL header, or no-nothing macros depending on platform */
 
-#include "zmq_sal.h"
+#include <zmq_sal.h>
 
 /*  Handle DSO symbol visibility                                             */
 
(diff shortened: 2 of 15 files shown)


bluca:
What is this for?

CMakeLists.txt
```diff
@@ -1176,7 +1176,7 @@ if(MINGW)
   endif()
 endif()
 
-set(public_headers include/zmq.h include/zmq_utils.h)
+set(public_headers include/zmq.h include/zmq_utils.h include/zmq_sal.h)
```
bluca, line 1179: Why is this random thing being made a public API? This makes no sense

include/zmq.h
```diff
@@ -31,29 +31,66 @@ extern "C" {
 #include <stddef.h>
 #include <stdio.h>
 
+#include <zmq_sal.h>
```
bluca, line 34: Yeah no, do not touch public headers, this is an unreadable mess and I have no idea if it's breaking API/ABI or what

CMakeLists.txt
axelriet: https://learn.microsoft.com/en-us/cpp/code-quality/understanding-sal?view=msvc-170

include/zmq.h
axelriet: I'm not sure what you mean by "unreadable mess" - It appears that you are unfamiliar with SAL. Please read the content Ilinked above.

axelriet:
> What is this for?

CMakeLists.txt
axelriet: Nothing observably changed, either at the API or ABI level. Annotations enforce correct usage and help reduce defects.

include/zmq.h
bluca: I am not going to read anything - if you propose a change, it's on you to explain why do you think is needed

axelriet: SAL (the source annotation language) is a big deal. The explanation takes an entire book that does not fit in a PR comment. SAL in a nutshell enables to express intent in a concise manner. The motivation is less defects thanks to far more effective static analysis than what can be inferred from the code itself. Since the code gets annotated directly and not all compilers understands the annotations yet (only MSVC and the Intel compiler so far) some gymnastic is required to #define annotations to nil for other compilers, therefore I have to introduce a new header made mostly of #ifdef X #undef X #define X (Nil) for each of the couple hundreds possible annotations. This round annotates the API layer mostly to the benefit of users of the library. Later I plan to annotate internal classes/functions for the benefit of the library itself. I already found a couple of defects that will be part of a subsequent PR.

bluca: In other words, this mess in the public headers is for the sole benefit of some proprietary tools that nobody else can run and nobody else will ever look at. Thanks, but no thanks. If you care about static analyzers, there are dozens of unsolved warnings from muse/coverity/etc, you can start from there. And you could also go ask the owners of this proprietary tool to make it work without needing to make a mess of the codebase - if coverity and friends can work just fine as-is, there's no reason other tools couldn't do the same.

axelriet: I don't think the MSVC and the Intel C/C++ compiler - which are both free - qualify as "tools that nobody else can run." Your last reply suggests you did not read or misunderstood my previous comment, the relevant bits being "far more effective static analysis than what can be inferred from the code itself." I have a hard time justifying onboarding a project that rejects quality improvements.

bluca: This unnecessarily makes an unreadable mess in order to support some subpar proprietary tool, and in public headers on top of all that - you didn't even namespace all those messy public defines despite the very generic names, which tells me you don't really understand what public libraries are and how they are supposed to be maintained. Hence this is not a 'quality improvement', very much the opposite, it's a significant quality degradation, and for no good reason: as already mentioned, there are plenty of existing static analyzers that do not require vandalizing the code base to be used. Some are even open source, and some are already hooked in, like coverity and the GHA one. If you care about static analysis, why don't you start by working on those dozens of existing warnings, and reduce them to zero?

bluca: Where would the ad-hominem be, precisely?

PR #4606, opened 2023-10-22
Fix a random test failure, other minor test fixes.

axelriet:
- Fix a random test failure in either test_req_relaxed or test_reconnect_options due to a shared endpoint.
- Fix test_heartbeats and test_security_curve so they don't fail with ZMQ_ACT_MILITANT.
- Enable benchmark_radix_tree for Windows, use trie_with_size_t consistently with xsub.hpp.
- Exclude VS artifacts in .gitignore.

### DIFF
--- a/perf/benchmark_radix_tree.cpp
+++ b/perf/benchmark_radix_tree.cpp
@@ -1,8 +1,7 @@
 /* SPDX-License-Identifier: MPL-2.0 */
 
-#include "../include/zmq.h"
-
-#if (defined __cplusplus && __cplusplus >= 201103L) || (defined _MSC_VER && _MSC_VER >= 1900)
+#if (defined __cplusplus && __cplusplus >= 201103L)                            \
+  || (defined _MSC_VER && _MSC_VER >= 1900)
 
 #include "radix_tree.hpp"
 #include "trie.hpp"
--- a/perf/benchmark_radix_tree.cpp
+++ b/perf/benchmark_radix_tree.cpp
@@ -1,6 +1,8 @@
 /* SPDX-License-Identifier: MPL-2.0 */
 
-#if __cplusplus >= 201103L
+#include "../include/zmq.h"
+
+#if (defined __cplusplus && __cplusplus >= 201103L) || (defined _MSC_VER && _MSC_VER >= 1900)
 
 #include "radix_tree.hpp"
 #include "trie.hpp"
@@ -73,8 +75,9 @@ int main ()
     //
     // Keeping initialization out of the benchmarking function helps
     // heaptrack detect peak memory consumption of the radix tree.
-    zmq::trie_t trie;
+    zmq::trie_with_size_t trie;
     zmq::radix_tree_t radix_tree;
+
     for (auto &key : input_set) {
         trie.add (key, key_length);
         radix_tree.add (key, key_length);
--- a/tests/test_heartbeats.cpp
+++ b/tests/test_heartbeats.cpp
@@ -370,6 +370,7 @@ void test_setsockopt_heartbeat_ttl_max ()
 
 void test_setsockopt_heartbeat_ttl_more_than_max_fails ()
 {
+#if !defined(ZMQ_ACT_MILITANT)
     void *const socket = test_context_socket (ZMQ_PAIR);
     const int value = heartbeat_ttl_max + 1;
     TEST_ASSERT_FAILURE_ERRNO (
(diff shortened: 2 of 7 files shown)


perf/benchmark_radix_tree.cpp
```diff
@@ -1,6 +1,8 @@
 /* SPDX-License-Identifier: MPL-2.0 */
 
-#if __cplusplus >= 201103L
+#include "../include/zmq.h"
```
bluca, line 3: This is wrong

tests/test_hwm_pubsub.cpp
```diff
@@ -259,7 +259,7 @@ void test_blocking (const char *bind_endpoint_)
 DEFINE_REGULAR_TEST_CASES (tcp, "tcp://127.0.0.1:*")
 DEFINE_REGULAR_TEST_CASES (inproc, "inproc://a")
 
-#if !defined(ZMQ_HAVE_WINDOWS) && !defined(ZMQ_HAVE_GNU)
+#if !defined(ZMQ_HAVE_GNU)
```
bluca, line 262: ipc is not available everywhere on windows, this is wrong

tests/test_heartbeats.cpp
```diff
@@ -370,13 +370,19 @@ void test_setsockopt_heartbeat_ttl_max ()
...
     void *const socket = test_context_socket (ZMQ_PAIR);
     const int value = heartbeat_ttl_max + 1;
     TEST_ASSERT_FAILURE_ERRNO (
       EINVAL,
       zmq_setsockopt (socket, ZMQ_HEARTBEAT_TTL, &value, sizeof (value)));
 
     test_context_socket_close (socket);
+#else
+    TEST_IGNORE_MESSAGE (
+      "libzmq with ZMQ_ACT_MILITANT, ignoring "
```
bluca, line 383: Why?

tests/test_security_curve.cpp
```diff
@@ -477,6 +477,7 @@ void test_curve_security_invalid_initiate_command_encrypted_content ()
 
 void test_curve_security_invalid_keysize (void *ctx_)
 {
+#if !defined(ZMQ_ACT_MILITANT)
```
bluca, line 480: Why?

tests/test_req_relaxed.cpp
```diff
@@ -169,7 +169,7 @@ void test_case_4 ()
     TEST_ASSERT_SUCCESS_ERRNO (
       zmq_setsockopt (req, ZMQ_REQ_CORRELATE, &enabled, sizeof (int)));
 
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (req, ENDPOINT_0));
+    TEST_ASSERT_SUCCESS_ERRNO (zmq_connect (req, ENDPOINT_7));
```
bluca, line 172: This doesn't exist

tests/test_security_curve.cpp
```diff
@@ -569,7 +574,7 @@ int main (void)
 
     void *ctx = zmq_ctx_new ();
     test_curve_security_invalid_keysize (ctx);
-    TEST_ASSERT_SUCCESS_ERRNO (zmq_ctx_term (ctx));
+    TEST_ASSERT_SUCCESS_ERRNO(zmq_ctx_term(ctx));
```
bluca, line 577: this breaks formatting rules

perf/benchmark_radix_tree.cpp
axelriet: Correct (leftover from some change I undid)

tests/test_hwm_pubsub.cpp
axelriet: It's better to have the test fail where it's not supported than not testing where it is.

tests/test_heartbeats.cpp
axelriet: ZMQ_ACT_MILITANT causes the test program to voluntarily crash on invalid options, which causes the test to fail erroneously.

tests/test_req_relaxed.cpp
axelriet: How does it compile then? In fact, those tests were spuriously failing as ENDPOINT_0 was shared between two unrelated tests, sometimes causing one or the other to randomly fail with an "address in use" error. This is a fix.

axelriet: Oh I see the change to testutil.hpp did not make it. Fixed.

tests/test_heartbeats.cpp
bluca: then don't enable it

axelriet: I think the “militant” mode has great value. Maybe it should be renamed “strict” or something and have its use much expanded in the codebase so the library bugchecks on every programmer error (like passing invalid values to a function) instead of returning error codes in those cases. 

What is not OK is having tests fail (not in the usual sense of an expected failure) with the strict mode enabled. 

Unless there is a way to express that crashing is a form of success (and then those tests should be standalone as nothing runs after them) the invalid parameter tests must be disabled in strict mode.

bluca: No, this is fine as it is. Drop this change and the rest of the PR can be merged.

PR #4609, opened 2023-10-24
Fixed the CMake warnings

da115115:
With the current version of the CMake files, there are a few warning messages. This patch fixes those warning (the project is building without warnings).
Note that there will no longer be support for versions of CMake strictly less than 3.5. All the current Linux distributions, including long term support (LTS) ones, have CMake with a version greater than 3.5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,6 @@
 # CMake build script for ZeroMQ
+cmake_minimum_required(VERSION 3.5..3.27)
 project(ZeroMQ)
-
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,12 +1,7 @@
 # CMake build script for ZeroMQ
+cmake_minimum_required(VERSION 3.5..3.27)
```
bluca, line 2: No, this breaks builds in other OSes, please leave this as-is

bluca:
We do not want to break the build on older OSes, closing

PR #4615, opened 2023-10-31
Feature/index page

f18m:
Make the index page of Asciidocs more similar to api.zeromq.org landing page, with a list of available documentation pages.


f18m:
With this PR I hope to:
* make it clear in the landing/index page the version of libzmq referred by the documentation 
* make sure we link each and every .adoc page (some user might forget to add a link in zmq.adoc or even in the Makefile.am at some point)

f18m:
@bluca ,
I noticed that one CI build is failing with:

```
make[2]: Entering directory '/home/runner/work/libzmq/libzmq/libzmq/zeromq-4.3.6/_build/sub/doc'
echo >../../../doc/__pagelist
/bin/bash: line 1: ../../../doc/__pagelist: Permission denied
```

seems like only in 1 case the source directory is read-only... would you know why?

f18m:
> @bluca , I noticed that one CI build is failing with:
> 
> ```
> make[2]: Entering directory '/home/runner/work/libzmq/libzmq/libzmq/zeromq-4.3.6/_build/sub/doc'
> echo >../../../doc/__pagelist
> /bin/bash: line 1: ../../../doc/__pagelist: Permission denied
> ```
> 
> seems like only in 1 case the source directory is read-only... would you know why?

I probably have fixed this issue by generating the __pagelist file inside the build tree instead of the source tree.
However I'm not 100% confident about CMake build system: how can I obtain the same result as Automake? 
Should I move the __pagelist generation into an ancillary script that is then invoked from both CMake and Automake?

bluca:
As long as it doesn't break the build, I don't really care if cmake doesn't generate the list, given it's just for the website

PR #4620, opened 2023-11-03
Feature: improve docs

f18m:
* In all "See also" sections: switch from space-separated list to unordered list representation

Rendering before the change: https://github.com/f18m/libzmq/blob/master/doc/zmq_bind.adoc
Rendering after the change: https://github.com/f18m/libzmq/blob/feature/improve-see-also/doc/zmq_bind.adoc


bluca:
I have added a redirect from wikidot

f18m:
Thanks!

PR #4635, opened 2023-11-29
Build with -D_DEFAULT_SOURCE along with -std=c11

fweimer-rh:
On its own, `-std=c11` hides POSIX and other extensions from C headers such as `<stdlib.h>` when building against glibc.  This causes the `posix_memalign` probe to fail incorrectly with compilers that do not accept implicit function declarations.  `_DEFAULT_SOURCE` is ignored by most non-GNU/Linux systems or not relevant in this context, so there is no separate check for adding it.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


configure.ac
```diff
@@ -66,7 +66,7 @@ ZMQ_ORIG_CXXFLAGS="${CXXFLAGS:-none}"
 
 # Checks for programs.
 AC_PROG_CC
-AX_CHECK_COMPILE_FLAG([-std=c11], [CFLAGS+=" -std=c11"], [AC_PROG_CC_C99])
+AX_CHECK_COMPILE_FLAG([-std=c11], [CFLAGS+=" -std=c11 -D_DEFAULT_SOURCE"], [AC_PROG_CC_C99])
```
bluca, line 69: it's a preprocessor flags, so it should be appended to CPPFLAGS rather than CFLAGS

bluca: yes - would that cause problems?

bluca: got it, let's keep it as-is then

PR #4642, opened 2023-12-08
Non-ASCII characters support in ipc:// socket paths

### LINKED ISSUE #4641: Using non-ASCII characters in Unix domain socket paths doesn't work on Windows platform
When using socket paths containing non-ASCII characters (i.e. Cyrillic) in `ipc://` socket paths, the socket opening may fail.

This seems to happen because Windows expects UDS paths encoded in `utf-8`, however 'narrow' versions of file manipulation functions (like `rmdir()`, `unlink()`, etc) expect the local system code page.

# Environment

* libzmq version (commit hash if unreleased): 4.3.5
* OS: Windows 10 17063

dz-fsd:
Problem: Using non-ascii characters in unix domain socket path doesn't work on Windows.

Solution: Convert utf-8 socket paths to utf-16 file names when using filesystem calls to delete files and directories as Windows doesn't have any filesystem calls that take utf-8 path.

rmdir_utf8() and unlink_utf8() static functions were created which substitute rmdir() and unlink() when building on Windows.

Fixes #4641

### DIFF
--- a/src/windows.hpp
+++ b/src/windows.hpp
@@ -28,6 +28,8 @@
 #include <windows.h>
 #include <mswsock.h>
 #include <iphlpapi.h>
+#include <string>
+#include <vector>
 
 #if !defined __MINGW32__
 #include <mstcpip.h>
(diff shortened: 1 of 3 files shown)


bluca:
```
1>c:\projects\libzmq\src\windows.hpp(78) : error C2065: 'nullptr' : undeclared identifier
1>c:\projects\libzmq\src\windows.hpp(86) : error C2039: 'data' : is not a member of 'std::vector<_Ty>'
1>        with
1>        [
1>            _Ty=wchar_t
1>        ]
1>c:\projects\libzmq\src\windows.hpp(90) : error C2039: 'data' : is not a member of 'std::vector<_Ty>'
1>        with
1>        [
1>            _Ty=wchar_t
1>        ]
1>Build log was saved at "file://c:\projects\build_libzmq\libzmq.dir\Release\BuildLog.htm"
1>libzmq - 3 error(s), 0 warning(s)

```

src/windows.hpp
```diff
@@ -64,6 +66,40 @@ static inline int poll (struct pollfd *pfd, unsigned long nfds, int timeout)
...
+        return std::wstring ();
+
+    const int utf16_length =
+      ::MultiByteToWideChar (CP_UTF8, MB_ERR_INVALID_CHARS, utf8_string,
+                             -1, // assume the input string is null-terminated
+                             nullptr, 0);
+    if (utf16_length == 0)
+        return std::wstring ();
+
+    std::vector<wchar_t> utf16_string (utf16_length);
```
axelriet, line 82: replace by std::wstring, use .resize(utf16_length), and &result[0] as dest address, then return that string. Example:

static std::wstring WideStringFromString (_In_z_ const char *str)
{
    std::wstring retVal;

    if (str && *str) {
        const int cchSource = static_cast<int> (strlen (str));
        const int cchNeeded =
          MultiByteToWideChar (CP_UTF8, 0, str, cchSource, nullptr, 0);

        if (cchNeeded > 0) {
            retVal.resize (cchNeeded);
            MultiByteToWideChar (CP_UTF8, 0, str, cchSource, &retVal[0],
                                 cchNeeded);
        }
    }

    return retVal;
}

bluca (approved):

Issue #4669, opened 2024-03-25
cmake: ZMQ_WIN32_WINNT autodetection broken when cross-compiling with mingw-w64


fanquake:
When cross-compiling for Windows on Ubuntu, using mingw-w64, the functionality in the CMake build:
>  [Value to set _WIN32_WINNT to for building [default=autodetect from build environment]](https://github.com/zeromq/libzmq/blob/4a5ece54b8ce0d3408e4d302b8a64b2a5aaa19cf/CMakeLists.txt#L373)

does not work. After configuring with the following:
```bash
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release \
-DCMAKE_SYSTEM_NAME=Windows \
-DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc \
-DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++

<snip>
-- CURVE security is disabled
-- Detected _WIN32_WINNT from CMAKE_SYSTEM_VERSION: 
-- Looking for kqueue
<snip>

cmake --build build -j$(nproc) --verbose
....
[  2%] Building CXX object CMakeFiles/libzmq.dir/src/ws_engine.cpp.obj
In file included from /root/libzmq/src/precompiled.hpp:22,
                 from /root/libzmq/src/ws_encoder.cpp:3:
/root/libzmq/src/windows.hpp:21:19: error: operator '>=' has no left operand
   21 | #if (_WIN32_WINNT >= 0x0501)
      |                   ^~
/root/libzmq/src/windows.hpp:23:2: error: #error You need at least Windows XP target
   23 | #error You need at least Windows XP target
      |  ^~~~~
[...]

PR #4672, opened 2024-03-27
build: fix mingw-w64 compilation with CMake

fanquake:
Set `ZMQ_HAVE_IPC OFF` for all Windows builds, rather than just `WindowsStore` builds.

This fixes compiling with mingw-w64, and mirrors the autotools builds system:

https://github.com/zeromq/libzmq/blob/2a75ef07be0ba76f86592e10b999331cd6124d52/configure.ac#L145-L152

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,9 +313,9 @@ set(POLLER
   kqueue, epoll, devpoll, pollset, poll or select [default=autodetect]")
 
 if(WIN32)
+  set(ZMQ_HAVE_IPC OFF)
   if(CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND CMAKE_SYSTEM_VERSION MATCHES "^10.0")
     set(ZMQ_HAVE_WINDOWS_UWP ON)
-    set(ZMQ_HAVE_IPC OFF)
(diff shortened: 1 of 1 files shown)


bluca:
Why? AF_UNIX was added some years ago to Windows

fanquake:
> Why? AF_UNIX was added some years ago to Windows

Is there as reason it's still disabled in the autotools build then?

fanquake:
> I don't know if anybody uses autotools on windows, doubt it

I don't know what you mean. The Autotools build system here is used extensively for cross-compiling Windows libs on Linux, via mingw-w64.

bluca:
That may very well be, doesn't change the fact that AF_UNIX is available on windows and it doesn't make sense to disable it

fanquake:
Sure. So what is your preferred approach for handling mingw-w64 builds via CMake? Note that this is the second of two issues (see #4669). Should we handle this in the build system, in the code? Is mingw-w64 official unsupported via CMake?

stephanlachnit:
The actual issue with mingw is actually wrong usage of `_MSC_VER`, I fixed it in https://github.com/zeromq/libzmq/pull/4678

PR #4678, opened 2024-04-23
Do not use _MSC_VER if windows

stephanlachnit:
This fixes several instances where _MSC_VER was used to determine whether to use afunix.h or not. For example, MinGW requires the use of afunix.h but does not define _MSC_VER. The correct behavior is checking ZMQ_HAVE_WINDOWS.

### DIFF
--- a/src/ipc_address.hpp
+++ b/src/ipc_address.hpp
@@ -7,7 +7,7 @@
 
 #include <string>
 
-#if defined _MSC_VER
+#if defined ZMQ_HAVE_WINDOWS
 #include <afunix.h>
 #else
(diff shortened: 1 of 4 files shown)


stephanlachnit:
@bluca would it be possible to create a 4.3.6 release with this change?

thomasjm:
I'm unable to build with `mingw-w64` without this change. It's been a while since a release, would one be possible soon @bluca ?

yansheng1003:
follow you ，yet dont work it
error: #error On Windows, IPC does not work with POLLER=select, use POLLER=epoll instead, or disable IPC transport

PR #4687, opened 2024-05-10
Do not bundle zmq.hpp in Debian packages anymore

stephanlachnit:
This reverts commit 83c042ccda60cc94a6ad07a38fd8eb0e6dc12375 from https://github.com/zeromq/libzmq/pull/2852.

Starting with Debian 12 and Ubuntu 23.04, zmq.hpp is now included in the cppzmq-dev package. Thus, zmq.hpp should be removed from the OBS packages to prevent a conflict with the system cppzmq pacakge.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -54,12 +54,6 @@ else
 endif
 endif
 
-override_dh_auto_install:
-	dh_auto_install
-ifneq ("$(wildcard debian/zmq.hpp)","")
-	cp $(CURDIR)/debian/zmq.hpp $(CURDIR)/debian/tmp/usr/include/
-endif
-
 override_dh_strip:
 	dh_strip --dbg-package=libzmq5-dbg
 
--- a/packaging/obs/_service
+++ b/packaging/obs/_service
@@ -11,18 +11,6 @@
     <param name="filename">zeromq</param>
   </service>
 
-  <!-- embed zmq.hpp like Debian and Ubuntu do -->
-  <service name="download_url">
-    <param name="protocol">https</param>
-    <param name="host">github.com</param>
-    <param name="path">zeromq/cppzmq/archive/master.zip</param>
-  </service>
-  <service name="extract_file">
-    <param name="archive">*master.zip</param>
-    <param name="files">*/zmq.hpp</param>
-    <param name="outfilename">debian.zmq.hpp</param>
-  </service>
-
   <!-- extract redhat packaging -->
   <service name="extract_file">
     <param name="archive">*.tar</param>


bluca:
It really makes no sense to have a separate package for a header, but whatever. The point of this is to provide latest git content, so instead please just add a conflict+replaces.

stephanlachnit:
> It really makes no sense to have a separate package for a header, but whatever. The point of this is to provide latest git content, so instead please just add a conflict+replaces.

Yes and no. It is not actually about the header, but detecting cppzmq via pkg-config/cmake, which gets annoying when debian/ubuntu just ships it as part of libzmq and the rest doesn't. The conflict+replaces is in the cppzmq-dev package is actually there, we can't conflict+replaces OBS packages.

I disagree about the point of the OBS packages though: cppzmq is developed separately which a separate repo and separate releases, so the only reason to put `zmp.hpp` in the libzmq package is to do so to mimic the (former) Debian package decision. This was also the reasoning when this was introduced in https://github.com/zeromq/libzmq/pull/2852.

bluca:
Nah, it's in the same OBS repo as it's a pain to maintain, so it makes it easier. Just add the conflict+replace here and that will solve the problem.

stephanlachnit:
> Nah, it's in the same OBS repo as it's a pain to maintain, so it makes it easier. Just add the conflict+replace here and that will solve the problem.

No, it won't. Only if the `cppzmq.pc` file is also installed, otherwise it is useless (except if you only target Debian). Besides, I can still use system cppzmq (which hasn't seen changes in a while anyway) with OBS draft libzmq.

stephanlachnit:
@bluca can we please merge this? This just mirror the behavior of the rpm packages on OBS, which also don't have the zmq.hpp file.

bluca:
No - as mentioned, add conflicts and replaces instead

stephanlachnit:
Alright, then I will close this since just adding conflict and replaces will not solve my issue with the deb OBS packages.

stephanlachnit:
For anyone having the same problem, we now repackage libzmq by removing the zmq.hpp header:
```dockerfile
# Add libzmq repo with draft API
RUN apt-get install --no-install-recommends -y gpg curl
RUN echo 'deb http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_24.04/ /' > /etc/apt/sources.list.d/network:messaging:zeromq:git-draft.list
RUN curl -fsSL https://download.opensuse.org/repositories/network:messaging:zeromq:git-draft/xUbuntu_24.04/Release.key | gpg --dearmor > /etc/apt/trusted.gpg.d/network_messaging_zeromq_git-draft.gpg
RUN apt-get update
# Remove zmq.hpp from libzmq3-dev due to https://github.com/zeromq/libzmq/pull/4687
RUN apt-get download libzmq3-dev
RUN dpkg-deb -R libzmq3-dev_*.deb libzmq3-dev
RUN sed -i '/zmq.hpp/d' libzmq3-dev/DEBIAN/md5sums
RUN rm libzmq3-dev/usr/include/zmq.hpp
RUN dpkg-deb -b libzmq3-dev libzmq3-dev.deb
RUN apt-get install --no-install-recommends -y ./libzmq3-dev.deb
# Cleanup
RUN apt-get purge --auto-remove -y gpg curl
RUN rm -r libzmq3-dev libzmq3-dev.deb libzmq3-dev_*.deb
```

PR #4705, opened 2024-07-13
Add Bazel build files based on Cmake.

BYVoid:
libzmq is already on Bazel Central Registry: https://registry.bazel.build/modules/libzmq


bluca:
Why is this needed? We already have 2 build systems, which is already one too many. Isn't this a google proprietary thing?

BYVoid:
Bazel is open source, adoped by many users other than Google.

It does not replace camke or automake. The bazel rule I added is just a wrapper of the existing cmake build.

bluca:
> Bazel is open source, adoped by many users other than Google.

That doesn't mean much - very often google doesn't really maintain projects in the open, but just throws sources over the wall. That means they are proprietary with sources available, not open source.

> It does not replace camke or automake. The bazel rule I added is just a wrapper of the existing cmake build.

That still doesn't answer the question: why add this? I do not see any reason to do so

BYVoid:
Bazel is open source and has lots of contributors from the community. Google does have some unmaintained open source projects, but Bazel is not among them.

libzmq is available on the [Bazel central registry](https://registry.bazel.build/modules/libzmq) (like a package manager). I am trying to upstream the "build patch" to the project. This would make Bazel users easier to build their project that links to libzmq. But keeping the patch on the Bazel central registry is acceptable, too.

I understand that whether or not this change is accepted depends entirely on the preference of the owners of libzmq.

bluca:
Once again, I do not see any explanation of why this needs to be here. It can just continue to be maintained in that project's specific repository.

PR #4706, opened 2024-07-17
Fix CMake-generated `libzmq.pc` file

hebasto:
This PR mirrors the Autotools-based build system behavior for cross-compiling for Windows with static linking:
https://github.com/zeromq/libzmq/blob/ee29bcd64ac257e3e8c20ffa0fe8093f697fe429/configure.ac#L365

and properly populates the `Libs.private` section in the CMake-generated `libzmq.pc` file.

Addresses the unresolved issue in https://github.com/zeromq/libzmq/pull/4667:
> Please note that CMake-generated `libzmq.pc` file is also broken as its "Libs.private" section contains only `-lstdc++` when cross-compiling for Windows. I leave this problem out of this PR scope.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -567,12 +567,18 @@ if(ZMQ_HAVE_WINDOWS)
   # Cannot use check_library_exists because the symbol is always declared as char(*)(void)
   set(CMAKE_REQUIRED_LIBRARIES "ws2_32.lib")
   check_cxx_symbol_exists(WSAStartup "winsock2.h" HAVE_WS2_32)
+  if(HAVE_WS2_32)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -585,7 +585,9 @@ else()
   check_cxx_symbol_exists(SO_BUSY_POLL sys/socket.h ZMQ_HAVE_BUSY_POLL)
 endif()
 
-if(NOT MINGW)
+if(MINGW)
```
fanquake, line 588: Should probably be done in the Windows block above, where the other checks for the same libs are being done?

hebasto: Thanks! Reworked.

hebasto:
Rebased to refresh CI.

PR #4734, opened 2024-08-26
Problem: ipc connect can fail on Windows, even after bind

### LINKED ISSUE #4730: Problem: epoll crashes for some Windows users
# Issue description

Reported [in pyzmq](https://github.com/zeromq/pyzmq/issues/1981), but after updating to libzmq 4.3.5, it appears the patch in #4422 did not fix the problem, but just shifted the error.

# Environment

* libzmq version (commit hash if unreleased):  4.3.5
* OS: Windows 11 and 10
* Python 3.12.3 via Microsoft Store (though others may be relevant)

There seems to be some interaction with VPNs or firewalls or something that has yet to be fully understood, making it very hard to reproduce. It appears to be most of the time for affected users, but I've never been able to see it myself.

Python:

```python
import zmq
ctx = zmq.Context()
with ctx:
    with ctx.socket(zmq.PUSH) as s:
        s.bind("tcp://127.0.0.1:5555")
```

If built with cmake defaults (epoll, ipc enabled), this crashes with:

```
Bad file descriptor (C:\Users\runneradmin\AppData\Local\Temp\tmppy2n81h4\build_deps\bundled_libzmq-src\src\epoll.cpp:73)
```

[epoll.cpp:73](https://github.com/zeromq/libzmq/blob/v4.3.5/src/epoll.cpp#L73)

# What's the expected result?

it doesn't crash.

minrk:
Solution:

- move tcpip fallback to after connect instead of assuming successful ipc bind means ipc works (it's very weird and possibly a Windows bug that this can happen, but I can at least confirm that it does in very specific scenarios - Python from Windows Store + non-ascii username + default `$env:TMP`)
- correct some Windows error code checks
- set errno when connect fails (missing errno resulted in treating failure as success, leading to #4730)
- make sure to clear `_w` back to `retired_fd` after closing it

Also addresses todo item to remember the tcpip fallback after ipc has failed once, so the failing path doesn't get retried every time. I can back that out if folks want.

I believe this actually closes #4730, though after some debugging, I do believe there are still missing checks to `signaler::valid()`, because even when make_fd_pair properly fails, send is still called on the invalid signaler, leading to the EBADFD reported in #4730, when it seems like it should probably be an assert in the signaler constructor (or immediately after calling the constructor) that it failed to create an fd pair.

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -23,6 +23,10 @@
 #include "tcp.hpp"
 #ifdef ZMQ_HAVE_IPC
 #include "ipc_address.hpp"
+// Don't try ipc if it fails once
+namespace zmq {
+    static bool try_ipc_first = true;
+}
 #endif
 
 #include <direct.h>
@@ -555,8 +559,13 @@ int zmq::make_fdpair (fd_t *r_, fd_t *w_)
 
     // It appears that a lack of runtime AF_UNIX support
     // can fail in more than one way.
-    // At least: open_socket can fail or later in bind
+    // At least: open_socket can fail or later in bind or even in connect after bind
     bool ipc_fallback_on_tcpip = true;
+
+    if (!zmq::try_ipc_first) {
+        // a past ipc attempt failed, skip straight to try_tcpip in the future;
+        goto try_tcpip;
+    }
 
     //  Create a listening socket.
     const SOCKET listener = open_socket (AF_UNIX, SOCK_STREAM, 0);
@@ -585,8 +594,7 @@ int zmq::make_fdpair (fd_t *r_, fd_t *w_)
         goto error_closelistener;
     }
     // if we got here, ipc should be working,
-    // so raise any remaining errors
-    ipc_fallback_on_tcpip = false;
+    // but there are at least some cases where connect can still fail
 
     //  Listen for incoming connections.
     rc = listen (listener, 1);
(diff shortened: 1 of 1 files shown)


bluca:
```
Formatting with clang-format (using clang-format) and showing differences with latest commit
diff --git a/src/ip.cpp b/src/ip.cpp
index db10b13..30d70eb 100644
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -24,8 +24,9 @@
 #ifdef ZMQ_HAVE_IPC
 #include "ipc_address.hpp"
 // Don't try ipc if it fails once
-namespace zmq {
-    static bool try_ipc_first = true;
+namespace zmq
+{
+static bool try_ipc_first = true;
 }
 #endif
```

minrk:
The truly bizarre property of this situation is that IPC totally does work, but it doesn't work in the default temp directory for some process configurations and usernames. In the situations where I've reproduced this, setting `$env:TMP` to a location outside $HOME/AppData, e.g. $HOME/myAppData fixes everything, but any subdir of $HOME/AppData does not. I'm not sure if there's a property that can be interrogated to pick a different directory, or if just this catch on connect is enough. It's really weird that bind succeeds and connect fails.

BtbN:
This introduced a compilation error:

```
15.95 /50-libzmq/src/ip.cpp: In function 'int zmq::make_fdpair(fd_t*, fd_t*)':
15.95 /50-libzmq/src/ip.cpp:678:1: error: jump to label 'try_tcpip'
15.95   678 | try_tcpip:
15.95       | ^~~~~~~~~
15.95 /50-libzmq/src/ip.cpp:568:14: note:   from here
15.95   568 |         goto try_tcpip;
15.95       |              ^~~~~~~~~
15.95 /50-libzmq/src/ip.cpp:572:18: note:   crosses initialization of 'const SOCKET listener'
15.95   572 |     const SOCKET listener = open_socket (AF_UNIX, SOCK_STREAM, 0);
15.95       |                  ^~~~~~~~
```

PR #4751, opened 2024-10-22
Add Relicense

keith-dev:
Commit Relicense file


bluca:
relicensing has been completed already, this is no longer necessary

Issue #4729, opened 2024-08-20
Never closed sockets when remote ZYRE node goes offline


stephan57160:
# Issue description

On our ZYRE production server, once in a while, we observe never-closed sockets.
Sometimes, it goes up to 200 sockets to the same remote ZYRE node.

# Environment

* libzmq version (commit hash if unreleased): 3.4
* OS: reproduced on
  - Linux CentOS (32 & 64 bits - x86 and ARM), 
  - Rocky (64 bits) (x86)

# Minimal test code / Steps to reproduce the issue

1. Start ZYRE node A
2. Start ZYRE node B
3. On node A, 2 TCP sockets are seen with Node B:
  * Node A connected to Node B (used to send data to B).
  * Node B connected to Node A (used to receive data from B).
4. Node B goes offline (out of WIFI coverage, Ethernet cable unplugged, Windows hybernation, ...)
5. On node A, after some time, the ZYRE layer detects that node B is no more present and the PEER B is destroyed with the socket to it (node A to B).

# What's the actual result? (include assertion message & call stack if applicable)

Socket from node B to node A is never closed, even if 
* node B application is restarted or 
* node B is rebooted.

Note: 
This is not visible if application on node B is properly stopped (thx to TCP layer for sending TCP RESET).

# What's the expected result?

Sockets from remote nodes should be automatically closed when the remote disappear: 
- Either the ZYRE peer destruction should do,
- Use of TCP KEEPALIVE from the ZYRE application,

I failed to have a working implementation in any of those 2 cases.

# Possible solution

I digged into LIBZMQ and ZYRE for quite some time.
I tried different approaches, but I always failed to get an access to the ACCEPT()ed socket
in this particular scenario.

Finally, I have a 'draft' possible workaroung, that enables TCP KEEPALIVE right after a particular ACCEPT() in `tcp_listener.cpp`.
Basically, the idea is like:
```
  sock = accept(s_);
  ...
  tune_tcp_keepalives(sock, x, y, y);
```

keith-dev:
It sounds reasonable to me.  Have to considered submitting a change that fixes the problem?

stephan57160:
@keith-dev I can provide a `draft` PR this afternoon, with what we currently use for a few months.
It works for us, but might require more attention/suggestions.

stephan57160:
See PR [4761](https://github.com/zeromq/libzmq/pull/4761).

PR #4761, opened 2024-12-20
DRAFT - Use TCP KEEPALIVE to force closing socket with offline remotes.

stephan57160:
Close issue #4729.

### DIFF
--- a/src/tcp_listener.cpp
+++ b/src/tcp_listener.cpp
@@ -246,5 +246,33 @@ zmq::fd_t zmq::tcp_listener_t::accept ()
     if (options.priority != 0)
         set_socket_priority (sock, options.priority);
 
+    // OK. Let's talk about this lazy implementation.
+    //
+    // On Linux, after ::accept(), TCP_KEEPALIVE is not set by default.
+    // Moreover, there is no way for an application to retreive this socket
+    // and then, no way to close() or use setsockopt() on it.
+    // So, the only possibility to avoid socket leakage is TCP_KEEPALIVE.
+    //
+    // Regarding the values, the default TCP ones are far too huge (> 2h).
+    // The values below seem reasonnable for a ZYRE application :
+    //   120 + 10 * 10 = 220s (3mn 40s)
+    //
+    // Note:
+    //   A ZYRE client application will reset its side after about 30s, so
+    //   no worry about such applications.
+    //   What about other applications ?
+    //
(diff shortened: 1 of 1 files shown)


stephan57160:
OK. Let's talk about this lazy implementation.

On Linux, after ::accept(), TCP_KEEPALIVE is not set by default.
Moreover, there is no way for an application to retreive this socket
and then, no way to close() or use setsockopt() on it.
So, the only possibility to avoid socket leakage is TCP_KEEPALIVE.

Regarding the values, the default TCP ones are far too huge (> 2h).
The values below seem reasonnable for a ZYRE application :
> 120 + 10 * 10 = 220s (3mn 40s)

Note:
> A ZYRE client application will reset its side after about 30s, so
> no worry about such applications.
> What about other applications ?

bluca:
There are options for this, pretty sure zyre should just set them accordingly, rather than hardcoding?

stephan57160:
Agreed, and I spent quite some time trying to find something less hard-coded,but I did not find anything.

For instance, when the socket is created internally by a ZYRE node, the application has no possibility to access it.
A least, I did not find any way to do that.

If you have a pointer, I'll see what I can do with it.

bluca:
ZMQ_TCP_KEEPALIVE
ZMQ_TCP_KEEPALIVE_CNT
ZMQ_TCP_KEEPALIVE_IDLE
ZMQ_TCP_KEEPALIVE_INTVL

stephan57160:
Not sure to understand what you have in mind.

I you use `zmq_set_sockopt(sock, ZMQ_TCP_XXX, ...)`, you have to handle a bunch of `if error` cases.
All are handled by `tune_tcp_keepalives()`.

What I can do is to use constants for 10, 10 and 120, right, but I still have to `tune` TCP KA right after the call to `accept()`,
as not possible from the application.

bluca:
I mean that zyre or any other application should just use those options if they want to modify this behaviour

stephan57160:
Agreed.
Unfortunatelly, ZYRE applications have a limited access to the socket layer.

It's being a long time I worked on this but I'm pretty sure that even the ZACTOR has no view on the `accept()`'ing socket.

bluca:
I am pretty sure you can get the zmq socket in all of those cases, and then it's just a matter of setting those options

stephan57160:
If this way exists, I lost on this Hide & Seek :-)

PR #4776, opened 2025-03-20
set upper bound in cmake_minimum_required

minrk:
setting an upper bound improves forward-compatibility as legacy version support is dropped. Without an upper bound, the minimum version is used as the policy version, which means that setting a version lower than 3.5 will result in warnings starting with cmake 3.27 and errors starting in 4.0 (release candidate out now). This change allows zeromq to build with cmake 4.

- 3.5 compat is deprecated in 3.27 (2023), removed in 4.0 (2025)
- 3.10 compat is deprecated in 3.31 (2024)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
 # CMake build script for ZeroMQ
 
 if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL Darwin)
-  cmake_minimum_required(VERSION 3.0.2)
+  cmake_minimum_required(VERSION 3.0.2...3.31)
 else()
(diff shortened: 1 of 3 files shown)


myd7349:
This will fix configuration error with CMake 4.0.

mcloughlan:
getting this issue when using fetchcontent in a cmake file with version 4 as minimum as well

a4z:
found out too late that this pr exists, so I added https://github.com/zeromq/libzmq/pull/4789

which this pr would have already been merged, that would have spared me some time

andoorve:
Hi, is there any blocker to this being merged?

PR #4809, opened 2025-07-21
Problem: zmq_abort() may be called on Windows if polling fails

dz-fsd:
### Summary
Fixes occasional `abort()`s on Windows platform caused by unhandled errors returned by `WSAPoll()`, causing occasional crashes.

### Problem
Error handling for the `poll()` function is left to the caller, that should remember to use `WSAGetLastError()` instead of using `errno` to find out the error reason. This requires `#ifdef ZMQ_HAVE_WINDOWS ... #endif` and Windows-specific code to be done for each `poll()` invocation.

### Solution 
On Windows platform, implement custom WSA error to errno translation in the poll() function itself. The goal is to make Windows-specific poll() to appear more Unixy to the callers.

### Changes
* Windows-specific `poll()` function implementation in `windows.hpp`

### Platforms affected
* Windows

### Checklist

- [x]  Code compiles correctly
- [x]  No new compiler warnings
- [x]  No breaking API changes
- [x]  Comments added where needed
- [x]  Change is scoped to Windows-specific logic

### DIFF
--- a/src/windows.hpp
+++ b/src/windows.hpp
@@ -24,6 +24,7 @@
 #endif
 #endif
 
+#include "err.hpp"
 #include <winsock2.h>
 #include <windows.h>
 #include <mswsock.h>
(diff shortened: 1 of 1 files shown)


src/socket_poller.cpp
```diff
@@ -559,6 +559,7 @@ int zmq::socket_poller_t::wait (zmq::socket_poller_t::event_t *events_,
         if (rc == -1 && errno == EINTR) {
             return -1;
         }
+
```
bluca: spurious newline

dz-fsd: Consolidated fixes into one commit and removed unnecessarily changed files.

Issue #4164, opened 2021-03-22
Stable poller


gummif:
I am wondering when poller (and possibly other draft features) will go from draft to stable? It looks like it had been in draft for at least 5 years and I guess should be pretty robust by now.

aengusjiang:
Me too. It seem to long time no update.

PR #4813, opened 2025-08-27
Mark poller API as stable

stephanlachnit:
Closes https://github.com/zeromq/libzmq/issues/4164


stephanlachnit:
Requires handling of `ZMQ_CURRENT_EVENT_VERSION` and adjusting tests which do not properly check for draft macros when testing the poller.

bluca:
IIRC there are still some missing features that would require API breaks, check old tickets

Issue #4491, opened 2023-01-17
VSOCK support


achimnol:
# Issue description

I'd like to make a feature request to explicitly add support for `AF_VSOCK` sockets, as the form of `vsock://CID:PORT`.

[VSOCK](https://man7.org/linux/man-pages/man7/vsock.7.html) is used to implement IPC between VMs and containers without extra network configurations in Linux systems. For instance, VSOCK can replace the use of UNIX domain socket bind-mounts in containers.

A particular usage scenarios is Docker Desktop for Mac. With its recent 4.15 release (December 2022), it has changed the filesystem sharing driver to **virtiofs** which does not support bind-mounting UNIX sockets crossing the macOS filesystem and the QEMU VM's filesystem.

A good workaround of this situation is to use VSOCK, which is well-supproted by the QEMU's virtio stack and many other Linux hypervisors.

This is a relatively new kernel feature, though this originates from VMCI which zeromq already has some integration with. I think the actual implementation should not be much difficult.

# Environment

* libzmq version: (feature request)
* OS: Linux

bluca:
Can you send a PR to implement it?

PR #4822, opened 2025-10-29
Add the Linux VSOCK transport.

remijouannet:
Hello,
VSOCK transport allows communication between the Host and guests.
It requires linux on the host and Guest.

the code is mostly inspire from the tipc and vmci.

there is a conflict between VMCI and VSOCK header, i don't know how to treat this issue, do we support having both build ? or do we make exclusion between them ?
Fixes https://github.com/zeromq/libzmq/issues/4491

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -27,7 +27,6 @@
 
 #ifdef ZMQ_HAVE_VSOCK
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


bluca:
Thanks for the PR

> i don't know how to treat this issue, do we support having both build ?

Yeah they should both work at the same time

Issue #4767, opened 2025-01-15
Slow subscriber generates an assert when ZMTP heartbeat is enabled  


ottawaguy81:
# Issue description
Slow subscriber application generates an ASSERT when the subscriber RX queue is full and **HB** is enabled. This is the same issue as https://github.com/zeromq/libzmq/issues/3937
My observation shows, right after the ZMTP heartbeat triggers a reset of the socket, the subsequent socket read  triggers an assert. 

# Environment

* libzmq version : 4.3.4
* OS:  Linux 

# Minimal test code / Steps to reproduce the issue

I have two programs, pub.c and sub.c program that.   The SNDHW and RCVHW are set to one.  I send 3 publication and wait(15sec) for the HB to trigger a socket reset.  
Here are the steps to reproduce the problem. 
gcc  -g pub.c -o pub  -lczmq -lzmq
gcc  -g sub.c -o sub  -lczmq  -lzmq

1.  Run pub program. (./pub) 
2.  Run the sub (./sub) 
3. Press any key three times in the pub program console. That will send three publications.
4. **Wait 15 sec.** // The HB will cause the TCP socket reset.
5. Press any key in the ./sub window. This will trigger a zstr_recv() and cause an abort. 
6. If there is no abort, repeat from step 3 .

****** *********************File **************sub.c*******
#include <czmq.h>
#include <stdio.h>
#include <unistd.h>
#include <string.h>
#include <assert.h>

char serverIp[100]="127.0.0.1";

void *sub_thread(void *arg)
{
  char server[100];
  int counter = 0;
  int rcvhwm  = 2;
  int heartbeat_timeout = 4000; 
  int heartbeat_ivl     = 5000;

 
  zsock_t *client = zsock_new (ZMQ_SUB);
  char *message;
  printf (" Set option ZMQ_RCVHWM %d \n", zmq_setsockopt (zsock_resolve(client), ZMQ_RCVHWM, &rcvhwm, sizeof (int)));
  printf (" Set option ZMQ_HEARTBEAT_TIMEOUT %d \n", zmq_setsockopt (zsock_resolve(client), ZMQ_HEARTBEAT_TIMEOUT, &heartbeat_timeout, sizeof (int)));
  printf (" Set option ZMQ_HEARTBEAT_IVL %d \n", zmq_setsockopt (zsock_resolve(client), ZMQ_HEARTBEAT_IVL, &heartbeat_ivl, sizeof (int)));
 
[...]

Issue #3937, opened 2020-05-26
Assertion failed: _input_stopped (stream_engine.cpp:467)


lukaszsamson:
# Issue description

Assertion failed: _input_stopped (stream_engine.cpp:467)

# Environment

* libzmq version (commit hash if unreleased): 4.3.2
* OS: macos 10.15.4

# Minimal test code / Steps to reproduce the issue

Issue happens indeterministically so no direct reproduction steps. I have a process with several threads each running one dedicated DEALER socket. Every socket is in the same zmq context and is connected to the same endpoint (ROUTER) over TCP transport and uses asynchronous request-response pattern. The socket threads each call zmq_poll (on that one socket) and alternate zm_send/zmq_recv with ZMQ_DONTWAIT. The endpoint is generally slow at consuming messages and sending responses and the responses may be produced in a different order or not arrive at all. Sockets are created and used only on 1 thread.

# What's the actual result? (include assertion message & call stack if applicable)
Creshing thread stack trace
* thread #35, name = 'ZMQbg/IO/0', stop reason = signal SIGABRT
  * frame #0: 0x00007fff7331f33a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff733dbe60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff732a6808 libsystem_c.dylib`abort + 120
    frame #3: 0x000000001f9236e9 libzmq.5.dylib`zmq::zmq_abort(char const*) + 9
    frame #4: 0x000000001f953f0b libzmq.5.dylib`zmq::stream_engine_t::restart_input() + 651
    frame #5: 0x000000001f9426b8 libzmq.5.dylib`non-virtual thunk to zmq::session_base_t::write_activated(zmq::pipe_t*) + 136
    frame #6: 0x000000001f924afc libzmq.5.dylib`zmq::io_thread_t::in_event() + 60
    frame #7: 0x000000001f928658 libzmq.5.dylib`zmq::kqueue_t::loop() + 408
    frame #8: 0x000000001f958d9d libzmq.5.dylib`thread_routine(void*) + 61
    frame #9: 0x00007fff733dc109 libsystem_pthread.dylib`_pthread_start + 148
    frame #10: 0x00007fff733d7b8b libsystem_pthread.dylib`thread_start + 15

All threads

```
  thread #1, queue = 'com.apple.main-thread'
[...]

Issue #4364, opened 2022-04-02
【zmq 4.3.4】pub-sub mode crash by _io_error


ABlueLight:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description
when use ZMQ_PUB ZMQ_SUB and set heart_beat_timout maybe crash by _io_error
error message 
Assertion failed: !_io_error (/home/users/jiangkuan.liu/tools/zeromq-4.3.4/src/stream_engine_base.cpp:331)

# Environment

* libzmq version (commit hash if unreleased):  4.3.4
* OS:  ubuntu 16.04 gcc 5.4.0

# Minimal test code / Steps to reproduce the issue

## PUB code
~~~cpp
void* sockets;
std::mutex mtx;
std::atomic_int item_nums;
void *context = nullptr;

int main() {
  std::vector <std::string> zmq_protocol_items;
  std::string ip_port = "tcp://127.0.0.1:4124";
  zmq_protocol_items.push_back(ip_port);

  context = zmq_ctx_new();
  if (context == nullptr) {
    std::cout << "init zmq context failed!\n";
    return -1;
  }
  for (auto &proto : zmq_protocol_items) {
    std::cout << proto << "\n";
    void *socket = zmq_socket(context, ZMQ_PUB);

    int zmq_sndhwm = 100;
    zmq_setsockopt(socket,
                       ZMQ_SNDHWM,
                       &zmq_sndhwm,
                       sizeof(int));
    
    int zmq_heart_beat_ivl = 5;
    zmq_setsockopt(socket,
                     ZMQ_HEARTBEAT_IVL,
                     &zmq_heart_beat_ivl,
                     sizeof(int));

    int zmq_heart_beat_timeout = 5;
    zmq_setsockopt(socket,
                     ZMQ_HEARTBEAT_TIMEOUT,
                     &zmq_heart_beat_timeout,
                     sizeof(int));
    int ret = zmq_bind(socket, proto.c_str());
    if (ret < 0) {
      std::cout << "zmq linke failed!\n";
      continue;
    }
    sockets = socket;
  }
  std::vector<char *> data_vec;
  std::vector<int> data_size_vec;
  data_size_vec.push_back(10);
  data_size_vec.push_back(10);
  data_size_vec.push_back(100);
  char *data_0 = new char[10];
  data_0[0] = 1;
  char *data_1 = new char[10];
   data_1[0] = 2;
  char *data_2 = new char[100];
   data_2[0] = 3;
[...]

Issue #4229, opened 2021-07-02
Assertion failed: input_stopped (src/stream_engine.cpp:443) after setting send/receive high water marks


scottbrogden-iheartmedia:
*Please use this template for reporting suspected bugs or requests for help.*

# Issue description

We have been seeing memory issues lately we think is related to zeromq, we had previously set the high water mark for send and receive to 0.  Thinking setting the high water marks to 0 was contributing to our memory issue we set the high water mark to 1000.  But after setting this we see random crashes with the debug line 

**Assertion failed: input_stopped (src/stream_engine.cpp:443)**

# Environment

* libzmq version (commit hash if unreleased): 
* 4.2.5
* OS: 
* Linux in docker container based off of runtime:3.1-bionic

scottbrogden-iheartmedia:
This looks to be related to https://github.com/zeromq/libzmq/issues/3937.  I have both rcvhwm and heartbeat_ivl set.  I have turned off rcvhwm and left sndhwm set to 1000 and no crashes after 9 hours.

Has that issue been fixed in later versions, I'm using 4.2.5 now installed through apt-get install -y libzmq3-dev

lukaszsamson:
@scottbrogden-iheartmedia if your issue is indeed related to #3937 then no, 4.3 does not fix it. In our case we disabled heartbeat_ivl and worked it around with custom heartbeats via ordinary zmq messages.

ABlueLight:
bug is still exist

PR #4830, opened 2026-01-11
problem: session restarts input on an engine that never stalled

lukaszsamson:
This PR reproduces and fixes the `_input_stopped` assertion crash from https://github.com/zeromq/libzmq/issues/3937

This PR originally made `restart_input ()` ignore the stale call, which was deemed a wrong approach. I investigated the root cause and the PR is now rewritten so that the assert stays in place and the actual bug is addressed

## Root cause

The pipe between session and socket, outlives the engine. `activate_write` is an asynchronous cross-thread command from the reader. It can still be in flight while the stalled engine is replaced. Instrumented trace from the reproducer:

    input stopped    engine=A            (pipe full, A stops reading TCP)
    engine_error     engine=A            (heartbeat timeout)
    process_attach   engine=B            (reconnect)
    write_activated  engine=B            (stale activation from A's stall)
    restart_input    engine=B  -> Assertion failed: _input_stopped

The "input stalled on the full pipe" state lives in the engine and dies with it,
but `session_base_t::write_activated ()` unconditionally called `restart_input ()` on the new engine

## Related observations (not addressed here)

- Heartbeats can kill a healthy connection whenever input is stopped by
  backpressure, because the engine cannot observe incoming PINGs it is not
  reading. That is what triggers the reconnect in this reproducer, but the
  stale-activation race exists for any reconnect-while-stalled.
- `mechanism_ready ()` bails out of routing-id delivery on `EAGAIN` assuming the
  pipe is shutting down, which is also reachable when the pipe is merely still
  full after a reconnect.

Fixes:
https://github.com/zeromq/libzmq/issues/3937

Possibly fixes:
https://github.com/zeromq/libzmq/issues/4767 (Slow subscriber, RX queue full, heartbeat triggers reset)
https://github.com/zeromq/libzmq/issues/4229 (HWM configured (especially rcvhwm), random crashes)
https://github.com/zeromq/libzmq/issues/3596 (CLIENT/SERVER, heartbeat=1s, rcvhwm=1, slow consumer)
https://github.com/zeromq/libzmq/issues/4364 (PUB/SUB, heartbeat with small timeouts)

### DIFF
--- a/src/i_engine.hpp
+++ b/src/i_engine.hpp
@@ -35,8 +35,18 @@ struct i_engine
     //  events are not fired on termination.
     virtual void terminate () = 0;
 
+    //  Indicates whether the engine has stopped reading input because
+    //  it could not push more messages to the session (e.g. the pipe
+    //  is full). Only if this is the case may restart_input () be called.
+    //  The session must check this before calling restart_input (),
+    //  since the pipe and its flow-control state outlive the engine:
+    //  a pipe writability notification may be delivered after the engine
+    //  that stalled on the full pipe has been replaced by a new one.
+    virtual bool input_stopped () const = 0;
+
     //  This method is called by the session to signalise that more
     //  messages can be written to the pipe.
+    //  May only be called if input_stopped () returns true.
     //  Returns false if the engine was deleted due to an error.
     //  TODO it is probably better to change the design such that the engine
     //  does not delete itself
--- a/src/norm_engine.hpp
+++ b/src/norm_engine.hpp
@@ -44,6 +44,8 @@ class norm_engine_t ZMQ_FINAL : public io_object_t, public i_engine
     //  events are not fired on termination.
     void terminate () ZMQ_FINAL;
 
+    bool input_stopped () const ZMQ_FINAL { return !zmq_input_ready; }
+
     //  This method is called by the session to signalise that more
     //  messages can be written to the pipe.
     bool restart_input () ZMQ_FINAL;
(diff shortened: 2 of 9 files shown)


bluca:
If an assert is hit it means there is a programming error somewhere else that results in an invalid/undefined state. Just removing the assert and ignoring the error is wrong and leaves the underlying problem lurking about.

lukaszsamson:
> If an assert is hit it means there is a programming error somewhere else that results in an invalid/undefined state. Just removing the assert and ignoring the error is wrong and leaves the underlying problem lurking about.

I understand your point @bluca but the assert in question happens in production software using libzmq and is 100% reproducible in test from this PR using documented API in supported way. Either the assert is incorrect and should be removed, or the implementation is broken somewhere else and should be changed to avoid this code path. Alternatively the documentation needs to be updated to explicitly forbid using sub socket with hwm options

bluca:
> or the implementation is broken somewhere else and should be changed to avoid this code path.

Yes, that's the point, it means something is going wrong elsewhere, and that's what needs to be investigated and fixed

lukaszsamson:
> Yes, that's the point, it means something is going wrong elsewhere, and that's what needs to be investigated and fixed

@bluca I went ahead and attempted to address the issue by adding `i_engine::input_stopped ()` and making `session_base_t::write_activated ()` forward the event only to an engine that is actually stalled

Investigation and root cause in the updated PR description

PR #4832, opened 2026-01-21
cmake: Don't hardcode include directories for installed targets

hebasto:
This fixes scenarios where user specifies `CMAKE_INSTALL_INCLUDEDIR`.


hebasto:
Rebased to refresh the CI.

PR #4838, opened 2026-01-31
Re-run system calls when EINTR occurs

yurivict:
This prevents "Interrupted system call" failures when signals are delivered to the process.

### DIFF
--- a/src/devpoll.cpp
+++ b/src/devpoll.cpp
@@ -37,7 +37,10 @@ zmq::devpoll_t::~devpoll_t ()
 void zmq::devpoll_t::devpoll_ctl (fd_t fd_, short events_)
 {
     struct pollfd pfd = {fd_, events_, 0};
-    ssize_t rc = write (devpoll_fd, &pfd, sizeof pfd);
+    ssize_t rc;
+    do {
+        rc = write (devpoll_fd, &pfd, sizeof pfd);
+    } while (rc == -1 && errno == EINTR);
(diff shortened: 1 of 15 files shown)


bluca:
Looping forever is not right, all of these need an upper bound of retries to ensure eventual termination

yurivict:
It's not going to be looping forever, only until signals are exhausted which can't last forever.

This code was extensively tested in production on the product that was randomly failing due to EINTR signals and it looks good.

bluca:
There's no guarantee of that, so you need to add a counter to bound it

yurivict:
Oh no, there is a guarantee: the number of such interruptions is limited by the number of child processes which is up to the software package to decide how many. However practically speaking this number is limited.
There can be a lot of such processes or just one such process.

Adding an arbitrary limit would impose a failure condition that doesn't have to be there. Any failure with the EINTR "Interrupted system call" message should never be reported to the user.

bluca:
This has nothing to do with processes or packages. Please apply the requested fix, or this will not be merged.

jfolker:
Yuri is in the right. Putting a counter on EINTR breaks with decades of UNIX precedent and deviates needlessly from expectation.

Do what people naturally expect.

EDIT: As a matter of fact, this precedent began inside UNIX itself. However, if you're still not convinced then here are some modern examples of this precedent being followed: OpenSSH, Apache Server, PostgreSQL, cURL, and rsync. This list is not exhaustive.

PR #4840, opened 2026-02-01
Fix Mach port leak in `alt_clock_gettime` on macOS

qingweizhang054:
**Description:**

This PR fixes a Mach port resource leak in the macOS implementation of `alt_clock_gettime`.

On macOS versions prior to 10.12 (Sierra, released in 2016), `clock_gettime` is not available, libzmq falls back to `alt_clock_gettime`, which uses Mach APIs to retrieve timing information. The current implementation calls `mach_host_self()` to obtain a host port, but does not release the associated send right after use.

Since `mach_host_self()` returns a send right that must be explicitly deallocated, repeated calls to `alt_clock_gettime` in long-running processes can lead to Mach port leaks and eventual resource exhaustion.

This change ensures that the host port returned by `mach_host_self()` is properly released via `mach_port_deallocate()` after use. The fix is limited to the legacy macOS code path and does not affect platforms that use the native `clock_gettime` implementation.

**Impact:**

* Prevents Mach port leaks on older macOS systems
* Improves stability for long-running applications
* No behavior change on modern macOS or other platforms

### DIFF
--- a/src/clock.cpp
+++ b/src/clock.cpp
@@ -37,9 +37,11 @@ int alt_clock_gettime (int clock_id, timespec *ts)
 {
     clock_serv_t cclock;
     mach_timespec_t mts;
+    host_name_t host_self = mach_host_self ();
     host_get_clock_service (mach_host_self (), clock_id, &cclock);
     clock_get_time (cclock, &mts);
     mach_port_deallocate (mach_task_self (), cclock);
+    mach_port_deallocate (mach_task_self (), host_self);
     ts->tv_sec = mts.tv_sec;
     ts->tv_nsec = mts.tv_nsec;
     return 0;


qingweizhang054:
Note: CI reports failures on the GitHub Actions macOS runner in
`tests/test_zmq_poll_fd` and `tests/test_zmq_ppoll_fd`, where `sendto()`
returns `EHOSTUNREACH` (errno 65).

I cannot reproduce this locally on macOS, and this PR does not modify any
networking-related code paths. This appears to be an environment-specific
issue on the macOS runner rather than a regression introduced by this
change.

Relevant CI output:

FAIL: tests/test_zmq_poll_fd
sendto(...) failed, errno = 65 (Host unreachable)

FAIL: tests/test_zmq_ppoll_fd
sendto(...) failed, errno = 65 (Host unreachable)

PR #4852, opened 2026-04-17
fix: replace assert with early return in out_event when _io_error is set

ssam18:
When the I/O thread collects both POLLIN and POLLOUT from the same epoll_wait batch, processing the POLLIN can set _io_error via in_event_internal(), and the already-queued POLLOUT still gets dispatched to out_event() — hitting the assert and crashing. Replacing the assert with an early return handles this gracefully, consistent with the same guard already present in restart_output(). Fixes #4841.


bluca:
If an assert is hit it means there is a programming error somewhere else that results in an invalid/undefined state. Just removing the assert and ignoring the error is wrong and leaves the underlying problem lurking about.

PR #4853, opened 2026-04-17
Fix double free in pipe_t::process_pipe_term_ack

### LINKED ISSUE #4806: Double free happening in ZMQ 4.3.4 version, causing reaper and IO thread crashes when system goes down
Double free happening in ZMQ 4.3.4 version, causing reaper and IO thread crashes when system goes down

Thread 1:

#10 <> in zmq::yqueue_t<zmq::msg_t, 256>::<del>yqueue_t (<>, __in_chrg=<optimized out>) at src/yqueue.hpp:81
#11 zmq::ypipe_t<zmq::msg_t, 256>::</del>ypipe_t (<>, in_chrg=<optimized out>) at src/ypipe.hpp:46
#12 zmq::ypipe_t<zmq::msg_t, 256>::~ypipe_t (<>, in_chrg=<optimized out>) at src/ypipe.hpp:46
#13 <> in zmq::pipe_t::process_pipe_term_ack (<>) at src/pipe.cpp:398
#14 <> in zmq::socket_base_t::process_commands (<>, timeout=timeout@entry=0, throttle=throttle@entry=false) at src/socket_base.cpp:1505
#15 <> in zmq::socket_base_t::in_event() () at src/socket_base.cpp:1687
#16 <> in zmq::epoll_t::loop () at src/epoll.cpp:206
#17 <> in thread_routine () at src/thread.cpp:402

Thread 2:

#5 <> in zmq::session_base_t::pipe_terminated (, pipe=<optimized out>) at src/session_base.cpp:298
#6 <> in zmq::pipe_t::process_pipe_term_ack () at src/pipe.cpp:372
#7 <> in zmq::io_thread_t::in_event () at src/io_thread.cpp:91
#8 <> in zmq::epoll_t::loop () at src/epoll.cpp:206
#9 <> in thread_routine () at src/thread.cpp:402

Thread 2 frees up the memory, and thread 1 doesn't clean up and double free. Kindly help out.

ssam18:
The `_in_pipe` pointer was being used to drain messages and then deleted via `LIBZMQ_DELETE`, but the command dispatch loop in `io_thread_t::in_event` processes all mailbox commands in a single pass. If a second command arrives for the same `pipe_t` after `delete this` is executed, the object's memory has already been freed and `_in_pipe` becomes a dangling pointer, leading to a double free.

This fix saves `_in_pipe` to a local and nulls out the member before draining, so any stale access to the deleted object sees NULL rather than a freed ypipe pointer. Fixes #4806.

### DIFF
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -359,16 +359,18 @@ void zmq::pipe_t::process_pipe_term_ack ()
     //  First, delete all the unread messages in the pipe. We have to do it by
     //  hand because msg_t doesn't have automatic destructor. Then deallocate
     //  the ypipe itself.
+    upipe_t *const in_pipe = _in_pipe;
+    _in_pipe = NULL;
 
-    if (!_conflate) {
+    if (!_conflate && in_pipe) {
         msg_t msg;
-        while (_in_pipe->read (&msg)) {
+        while (in_pipe->read (&msg)) {
             const int rc = msg.close ();
             errno_assert (rc == 0);
         }
     }
 
-    LIBZMQ_DELETE (_in_pipe);
+    delete in_pipe;
 
     //  Deallocate the pipe object
     delete this;
--- a/src/pipe.cpp
+++ b/src/pipe.cpp
@@ -359,16 +359,18 @@ void zmq::pipe_t::process_pipe_term_ack ()
     //  First, delete all the unread messages in the pipe. We have to do it by
     //  hand because msg_t doesn't have automatic destructor. Then deallocate
     //  the ypipe itself.
+    upipe_t *in_pipe = _in_pipe;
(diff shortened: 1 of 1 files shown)


src/pipe.cpp
```diff
@@ -359,16 +359,18 @@ void zmq::pipe_t::process_pipe_term_ack ()
...
         msg_t msg;
-        while (_in_pipe->read (&msg)) {
+        while (in_pipe->read (&msg)) {
             const int rc = msg.close ();
             errno_assert (rc == 0);
         }
     }
 
-    LIBZMQ_DELETE (_in_pipe);
+    delete in_pipe;
```
bluca: keep the LIBZMQ_DELETE macro usage

PR #4857, opened 2026-04-24
Free memory used by gss_wrap

### LINKED ISSUE #4856: Memory leak in GSSAPI codepath
The GSSAPI code leaks memory.  Here's an address-sanitizer output:

```
Direct leak of 378 byte(s) in 14 object(s) allocated from:
    #0 0x773c656fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x773c6559521b in zmq::gssapi_mechanism_base_t::encode_message(zmq::msg_t*) src/gssapi_mechanism_base.cpp:61
    #2 0x773c65599e3a in zmq::gssapi_mechanism_base_t::produce_ready(zmq::msg_t*) src/gssapi_mechanism_base.cpp:302
    #3 0x773c65599e3a in zmq::gssapi_server_t::next_handshake_command(zmq::msg_t*) src/gssapi_server.cpp:55
    #4 0x773c655c8bcd in zmq::stream_engine_base_t::next_handshake_command(zmq::msg_t*) src/stream_engine_base.cpp:459
    #5 0x773c655c44b7 in zmq::stream_engine_base_t::out_event() src/stream_engine_base.cpp:326
    #6 0x773c655961e8 in zmq::io_thread_t::in_event() src/io_thread.cpp:64
    #7 0x773c65595118 in zmq::epoll_t::loop() src/epoll.cpp:183
    #8 0x773c655c1f08 in thread_routine src/thread.cpp:229
    #9 0x773c6565ea41 in asan_thread_start ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234
    #10 0x773c6529caa3 in start_thread nptl/pthread_create.c:447
```

The issue is that the buffer containing the plain-text passed to `gss_wrap` is not freed so every message is leaked.

jake-scott:
Fixes #4856

### DIFF
--- a/src/gssapi_mechanism_base.cpp
+++ b/src/gssapi_mechanism_base.cpp
@@ -66,6 +66,10 @@ int zmq::gssapi_mechanism_base_t::encode_message (msg_t *msg_)
     maj_stat = gss_wrap (&min_stat, context, 1, GSS_C_QOP_DEFAULT, &plaintext,
                          &state, &wrapped);
 
+    free (plaintext_buffer);
+    plaintext.value = NULL;
+    plaintext.length = 0;
+
     zmq_assert (maj_stat == GSS_S_COMPLETE);
     zmq_assert (state);
 


bluca:
This is not right. First of all the buffer is still in use where you are freeing it and you don't null it, and secondly and most importantly, https://salsa.debian.org/debian/krb5/-/blob/master/src/lib/gssapi/generic/rel_buffer.c gss_release_buffer() already frees it later, so this introduces a double-free.

jake-scott:
hey @bluca - what am I missing here?  `plaintext_buffer` is alloced on line 56 and then assigned to `plaintext.value` on line 63.  I don't see that the buffer is ever seen outside of the function that its allocated in nor any `gss_release_buffer` related to it in that function.  (I do see the wrapped - encrypted buffer being freed though, just not the plaintext buffer)

bluca:
Yes I've confused it with the wrapped buffer. Still, better to NUL plaintext.value, just in case

PR #4932, opened 2026-09-16
Problem: no way to fully disable ZMTP 1.0/2.0 fallback

jamesdillonharvey:
A ZMTP engine that receives a greeting not matching a valid ZMTP 3.x signature falls back to unversioned/ZMTP 1.0 or ZMTP 2.0 framing (zmtp_engine.cpp receive_greeting()). That legacy framing has no mechanism negotiation, so any TCP peer that is not a ZMTP client at all (e.g. a **port scanner probing an open port**) can have arbitrary bytes it sends decoded and surfaced as an application message, instead of the connection simply being rejected.

The existing zap_enabled() check in handshake_v1_0_unversioned() / handshake_v1_0() / handshake_v2_0() only closes this gap when ZAP is already configured on the socket (mechanism != NULL, or a non-empty ZAP_DOMAIN) - a NULL-mechanism socket with no ZAP domain set has no protection against this at all.

Solution: add a build-time option, ENABLE_ZMTP_FALLBACK (CMake) / --disable-zmtp-fallback (Autotools), default on (matching current behavior). When disabled, the three legacy handshake paths reject unconditionally, regardless of ZAP/security mechanism configuration.

Adds tests/test_zmtp_fallback_disabled.cpp, built only when the option is disabled, verifying a raw peer sending unversioned ZMTP 1.0 framing (with no ZAP domain configured) is rejected outright rather than having its bytes decoded and delivered.


mrseanmorris:
i've run into this problem myself on different projects i manage. each time we spent a bunch of time in the app layer working around it.

nnog:
Including an opt-in build flag for this seems like a no-brainer to me, and seems like a good hardening option when not using ZAP.

